PR #170, opened 2016-11-21
Pr/trails

alexanderbock:
Pull request for the new RenderableTrails

Scenes checked:
 - `default.scene`
 - `newhorizons.scene`
 - `osirisrex.scene`
 - `rosetta.scene`
 - `globebrowsing.scene`


data/scene/atmosphereearth/atmosphereearth.mod
```diff
@@ -107,15 +107,20 @@ return {
...
+                Type = "SpiceTranslation",
+                Body = "EARTH",
+                Observer = "SUN"
+            },
+            Color = { 0.5, 0.8, 1.0 },
+            -- StartTime = "2016 JUN 01 12:00:00.000",
+            -- EndTime = "2017 JAN 01 12:00:00.000",
+            -- SampleInterval = 3600
+            Period = 365.242,
+            Resolution = 1000
```
emiax: maybe could be good to add a unit as a comment

modules/base/rendering/renderabletrail.cpp
```diff
@@ -23,346 +23,355 @@
...
+    static const char* KeyRendering = "Rendering";
+
+    // The possible values for the _renderingModes property
+    enum RenderingMode {
+        RenderingModeLines = 0,
+        RenderingModePoints,
+        RenderingModeLinesPoints
+    };
+
+    static std::map<std::string, RenderingMode> RenderingModeConversion = {
```
emiax: should this be a static const std::map?

modules/base/rendering/renderabletrail.cpp
```diff
@@ -23,346 +23,355 @@
...
+
+    _renderingModes.addOptions({
+        { RenderingModeLines, "Lines" },
+        { RenderingModePoints, "Points" },
+        { RenderingModeLinesPoints, "Lines+Points" }
+    });
+
+    if (dictionary.hasKeyAndValue<std::string>(KeyRendering)) {
+        _renderingModes = RenderingModeConversion[
+            dictionary.value<std::string>(KeyRendering)
```
emiax: Is there anything that ensures that the string value found in the dictionary is a key in the RenderingModeConversion map? Maybe check for this and throw a readable error

modules/base/rendering/renderabletrail.cpp
```diff
@@ -23,346 +23,355 @@
...
+    // offset
+    int primaryOffset =
+        _primaryRenderInformation._iBufferID == 0 ? 0 : _primaryRenderInformation.first;
+
+    // Render the primary batch of vertices
+    render(_primaryRenderInformation, totalNumber, primaryOffset);
+
+    // The secondary batch is optional,. so we need to check whether we have any data here
+    if (_floatingRenderInformation._vaoID == 0 || _floatingRenderInformation.count == 0) {
+        render(
```
emiax: I'm curious: Why is `render` a lambda, when it is only reffered to once from the same method where it is defined?

modules/base/rendering/renderabletrail.h
```diff
@@ -40,53 +44,117 @@ namespace opengl {
...
+protected:
+    explicit RenderableTrail(const ghoul::Dictionary& dictionary);
+
+    /// Returns the documentation entries that the con
+    static openspace::Documentation Documentation();
+   
+    /// The layout of the VBOs
     struct TrailVBOLayout {
-        float x, y, z, e;
+        float x, y, z;
```
emiax: love it!

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -0,0 +1,450 @@
...
+    glm::vec3 p = _translation->position(data.time);
+    _vertexArray[_primaryRenderInformation.first] = { p.x, p.y, p.z };
+
+    glBindVertexArray(_primaryRenderInformation._vaoID);
+    glBindBuffer(GL_ARRAY_BUFFER, _primaryRenderInformation._vBufferID);
+
+    // 3
+    if (!report.needsUpdate) {
+        // If no other values have been touched, we only need to upload the floating value
+        glBufferSubData(
```
emiax: Possible optimization: Looks like this is happening regardless if the last point moved or not. Would it be nice to not do this when time is paused?

modules/base/rendering/renderabletrailorbit.h
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * Performs a full sweep of the orbit and fills the entire vertex buffer object.
+     * \param time The current time up to which the full sweep should be performed
+     */
+    void fullSweep(double time);
+
+    /// This structure is returned from the #updateTrails method and gives information
+    /// about which parts of the vertex array to update
+    struct UpdateReport {
+        static const int All = 0; ///< The entire array was touched in the update
```
emiax: No way to denote that zero points were updated?

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -0,0 +1,450 @@
...
+        }
+    }
+
+    glEnableVertexAttribArray(0);
+    glVertexAttribPointer(0, 3, GL_FLOAT, GL_FALSE, 0, 0);
+
+    glBindVertexArray(0);
+}
+
+RenderableTrailOrbit::UpdateReport RenderableTrailOrbit::updateTrails(
```
emiax: Is there a reason why this method stores some values in the update report, and some in the instance state (_needsFullSweep etc)?

modules/base/shaders/renderabletrail_fs.glsl
```diff
@@ -22,19 +22,28 @@
...
+uniform vec3 color;
+uniform int renderPhase;
+
 #include "fragment.glsl"
 
 Fragment getFragment() {
-    vec4 c = vec4(color * fade * forceFade, 1.0);
+    if (renderPhase == RenderPhasePoints) {
+        vec2 circCoord = 2.0 * gl_PointCoord - 1.0;
+        if (dot(circCoord, circCoord) > 1.0) {
```
emiax: Nice way to check if length > 1. I assume it's faster since there is no square root involved. Maybe deserves a commend to explain? I have not checked the visual result yet, but it looks like this could cause jagged edges? How about something like: 
`alpha *=  smoothstep(dot(circCoord, circCoord),  1, 1 - 1.0 / pointSize) `

src/properties/optionproperty.cpp
```diff
@@ -57,9 +57,7 @@ const std::vector<OptionProperty::Option>& OptionProperty::options() const {
 }
 
 void OptionProperty::addOption(int value, std::string desc) {
```
emiax: Would addOption(int value, const std::string& desc) be better?

src/properties/optionproperty.cpp
```diff
@@ -57,9 +57,7 @@ const std::vector<OptionProperty::Option>& OptionProperty::options() const {
...
-    Option option;
-    option.value = value;
-    option.description = desc;
+    Option option = { std::move(value), std::move(desc) };
```
emiax: if desc is const string&, then move would not be necessary, I guess?

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -0,0 +1,450 @@
...
+// This class is using a VBO ring buffer + a constantly updated point as follows:
+// Structure of the array with a _resolution of 16:
+//  ---------------------------------------------------------------------------------
+//  | FF |    |    |    |    |    |    |    |    |    |    |    |    |    |    |    |
+//  ---------------------------------------------------------------------------------
+//    0     1    2    3    4    5    6    7    8    9   10   11   12   13   14   15  
+//                    <------ newer in time                                    oldest
+//
+// In the begining the floating value starts at 0; this means that array element 0 is
+// updated and uploaded to the GPU at every frame. The F+1 element is the newest fixed
```
emiax: What is F and what is FF ?

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -0,0 +1,315 @@
...
+                "samples are spaced along the trajectory. The time interval between "
+                "'StartTime' and 'EndTime' is split into 'SampleInterval' * "
+                "'TimeStampSubsampleFactor' segments.",
+                Optional::No
+            },
+            {
+                KeyTimeStampSubsample,
+                new IntVerifier,
+                "The factor that is used to create subsamples along the trajectory. This "
+                "value (together with 'TimeStampSubsampleFactor') determines how far "
```
emiax: This is TimeStampSubsampleFactor.
"together with SampleInterval" ?

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -0,0 +1,450 @@
...
+// the newest fixed point. If the time processes backwards, the floating point moves
+// towards the upper areas of the array instead.
+// In both cases, only the values that have been changed will be uploaded to the GPU.
+//
+// For the rendering, this is achieved by using an index buffer that is twice the size of
+// the vertex buffer containing identical two sequences indexing the vertex array.
+// In this example:
+// -----------------------------------------------------------------
+// |0|0|0|0|0|0|0|0|0|0|1|1|1|1|1|1|0|0|0|0|0|0|0|0|0|0|1|1|1|1|1|1|     << Values
+// |0|1|2|3|4|5|6|7|8|9|0|1|2|3|4|5|0|1|2|3|4|5|6|7|8|9|0|1|2|3|4|5| written top to bottom
```
emiax: Took a while for me to figure out that "top to bottom" was not referring to the data structure, but to the comment itself. Maybe decrease the size of the list from 16 to 8, to make everything fit in one line?

data/scene/atmosphereearth/atmosphereearth.mod
alexanderbock: Changed the documentation to better reflect that

modules/base/rendering/renderabletrail.cpp
alexanderbock: Yes

modules/base/rendering/renderabletrail.cpp
alexanderbock: Yes; the documentation will check for this with the StringInListVerifier

Added comments to make that clearer

modules/base/rendering/renderabletrail.cpp
alexanderbock: It was actually used twice.. in the line you highlight and just a line before then

modules/base/rendering/renderabletrailorbit.cpp
alexanderbock: Added a boolean to the UpdateData structure and test for both 0 delta time and pausing

modules/base/rendering/renderabletrailorbit.cpp
alexanderbock: The update report is only used in the parent classes render method.  The instance variables are used elsewhere, too

modules/base/shaders/renderabletrail_fs.glsl
alexanderbock: Added a comment about it and added the smoothstep function

src/properties/optionproperty.cpp
alexanderbock: We want the move though.. if we don't have a move, we would need a copy as the OptionProperty needs to store the values internally and can't rely on the reference being around forever

modules/base/rendering/renderabletrailorbit.cpp
alexanderbock: Made that comment a bit clearer by explaining FF and adding the missing F to F

modules/base/rendering/renderabletrailtrajectory.cpp
alexanderbock: Fixed

modules/base/rendering/renderabletrailorbit.cpp
alexanderbock: Made it fit into one row instead

PR #172, opened 2016-11-28
Support for Kepler and TLE Translations

alexanderbock:
A few other things have slipped into the PR as well, but well..

### DIFF
--- a/modules/base/rendering/renderableplanet.cpp
+++ b/modules/base/rendering/renderableplanet.cpp
@@ -349,7 +349,7 @@ void RenderablePlanet::render(const RenderData& data) {
     glm::dmat4 rot = glm::rotate(glm::dmat4(1.0), M_PI_2, glm::dvec3(1, 0, 0));
     glm::dmat4 roty = glm::rotate(glm::dmat4(1.0), M_PI_2, glm::dvec3(0, -1, 0));
     //glm::dmat4 rotProp = glm::rotate(glm::dmat4(1.0), glm::radians(static_cast<double>(_rotation)), glm::dvec3(0, 1, 0));
-    modelTransform = modelTransform * rot /** roty*/ /** rotProp*/;
+    modelTransform = modelTransform * rot * roty /** rotProp*/;
 
     glm::dmat4 modelViewTransform = data.camera.combinedViewMatrix() * modelTransform;
 
--- a/modules/base/translation/keplertranslation.cpp
+++ b/modules/base/translation/keplertranslation.cpp
@@ -1,0 +1,378 @@
+/*****************************************************************************************
+ *                                                                                       *
+ * OpenSpace                                                                             *
+ *                                                                                       *
+ * Copyright (c) 2014-2016                                                               *
+ *                                                                                       *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this  *
+ * software and associated documentation files (the "Software"), to deal in the Software *
+ * without restriction, including without limitation the rights to use, copy, modify,    *
+ * merge, publish, distribute, sublicense, and/or sell copies of the Software, and to    *
+ * permit persons to whom the Software is furnished to do so, subject to the following   *
+ * conditions:                                                                           *
+ *                                                                                       *
+ * The above copyright notice and this permission notice shall be included in all copies *
+ * or substantial portions of the Software.                                              *
+ *                                                                                       *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,   *
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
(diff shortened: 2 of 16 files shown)


include/openspace/scene/translation.h
```diff
@@ -46,7 +46,16 @@ class Translation : public properties::PropertyOwner {
 
     glm::dvec3 position(double time);
 
+    // Registers a callback that gets called when a significant change has been made that
+    // invalidates potentially stored points, for example in trails
+    void onParameterChange(std::function<void()> callback);
```
emiax: I know we have a lot of these void onXYZ(function<void()>) methods already, and they are nice and create a simple API. But as soon as we want an observable have more than 1 observer, it becomes a bit more tricky. I don't think this is a problem here at all, but it would be good to have a discussion about these kind of pub-sub issues at some point :)

modules/base/rendering/renderableplanet.cpp
```diff
@@ -349,7 +349,7 @@ void RenderablePlanet::render(const RenderData& data) {
     glm::dmat4 rot = glm::rotate(glm::dmat4(1.0), M_PI_2, glm::dvec3(1, 0, 0));
     glm::dmat4 roty = glm::rotate(glm::dmat4(1.0), M_PI_2, glm::dvec3(0, -1, 0));
     //glm::dmat4 rotProp = glm::rotate(glm::dmat4(1.0), glm::radians(static_cast<double>(_rotation)), glm::dvec3(0, 1, 0));
-    modelTransform = modelTransform * rot /** roty*/ /** rotProp*/;
+    modelTransform = modelTransform * rot * roty /** rotProp*/;
```
emiax: There is a comment right before these lines saying "earth needs to be rotated for that to work." Can these rotations be explained more clearly? Why is there a rotation of pi around x and then around y? That should be the same as one single rotation around z, am I right?

modules/base/translation/tletranslation.cpp
```diff
@@ -0,0 +1,374 @@
...
+#include <chrono>
+#include <fstream>
+#include <vector>
+
+namespace {
+    const char* KeyFile = "File";
+    
+    // The list of leap years only goes until 2056 as we need to touch this file then
+    // again anyway ;)
+    static const std::vector<int> LeapYears = {
```
emiax: There is a simple algorithm for leap years:

if (year is not divisible by 4) then (it is a common year)
else if (year is not divisible by 100) then (it is a leap year)
else if (year is not divisible by 400) then (it is a common year)
else (it is a leap year)

https://en.wikipedia.org/wiki/Leap_year#Algorithm

would it make sense to use this instead, or is this list better for some reason? I suppose something like this could live in something more "central", like the Time class, in order to avoid duplication of code.

modules/base/translation/tletranslation.cpp
```diff
@@ -0,0 +1,374 @@
...
+        1956, 1960, 1964, 1968, 1972, 1976, 1980, 1984, 1988, 1992, 1996,
+        2000, 2004, 2008, 2012, 2016, 2020, 2024, 2028, 2032, 2036, 2040,
+        2044, 2048, 2052, 2056
+    };
+    
+    // Count the number of full days since the beginning of 2000 to the beginning of
+    // the parameter 'year'
+    int countDays(int year) {
```
emiax: Yeah, since there is a lot of time handling here, I think it would be a good idea to move it to the time class, so more classes can use it. Nice and tidy implementation, I like it :)

include/openspace/scene/translation.h
alexanderbock: Yeah, I agree.. and maybe Michael and Sebastian's event system is the way to go... at the moment my thought was that those objects are only owned by a single objects anyway, so there is not going to be an issue about it.. but yeah.. something to keep in mind!

modules/base/rendering/renderableplanet.cpp
alexanderbock: Yeah, I need to look more into it again...  the first rotation moves the earth from rotating on its belly (aka. north pole pointing at the sun) to the correct side up.  The second rotation rotates the meridian into the correct location. I don't know why, but both rotations are necessary to produce the correct locations on Earth for their local time (without the second rotation, every position is 12h off sync, basically)

modules/base/translation/tletranslation.cpp
alexanderbock: Yeah, I know (i implemented it a few times before). The reason i did it this way is to be able to easily compute the number of leap years between two dates, which is less trivial in the other way.

And yes; this should move to an improved (and independent of SPICE) time class later on

modules/base/translation/tletranslation.cpp
alexanderbock: Yeah... I agree;  we should think about a new Time class that can do these kind of conversions (see previous comment)

modules/base/translation/keplertranslation.cpp
```diff
@@ -0,0 +1,378 @@
...
+}
+
+glm::dvec3 KeplerTranslation::position() const {
+    return _position;
+}
+
+double KeplerTranslation::eccentricAnomaly(double meanAnomaly) const {
+    // Compute the eccentric anomaly (the location of the spacecraft taking the
+    // eccentricity of the orbit into account) using different solves for the regimes in
+    // which they are most efficient
```
GPayne: Suggest adding more to this comment to explain the different solver criteria for different values of eccentricity.

PR #177, opened 2016-12-30
Refactor scene graph: Move scene loading to separate class.

emiax:
Since this is a quite big refactor, it would be great to get some help with verifying that nothing important has been broken. I would appreciate if you guys would check out this branch and try it on your own machines, as well as browse through the code.

Key changes:
- Move scene loading to separate class (SceneLoader). Refactor the loading code and break it up into smaller methods.
- RenderEngine does no longer own the scene. OpenSpace engine owns one SceneManager, which may own several scenes. In the future, I hope this will make it easier to load scenes in the background, and start rendering them when they are ready.
- Change ownership of scene graph nodes (Nodes own unique pointers to their children, root node is owned by Scene)
- Improve detection for circular dependencies and make it possible to recover if this kind of errors exist in the scene spec.
- Remove the "common" module.
- Reactivate rudimentary unit testing for scene loading.

I'd be happy to improve the code based on your feedback, @alexanderbock @jccosta @GPayne @EricMyers47 et al. The more eyes on this, the better. Happy new year!


EricMyers47:
I'd like to pull this into a separate branch first, try building on all platforms on Jenkins, and let people perhaps try it out on their own first, then merge into develop.

EricMyers47:
Okay, I pulled your branch into a branch on OpenSpace called feature/scenegraph-refactor so we can all look at it and try it out, and I put builds of it on Jenkins for Linux, Mac, and Windows.
The builds fail, so we should address that first, and then when those are working we'll merge into develop.

Issue #155, opened 2016-10-29
Launcher checkboxes hard to read


EricMyers47:
The checkboxes used by the Launcher (eg. for the synchronize feature) are difficult to see, as they just change color from dark to light.  Which is which?   There needs to be some positive indication that the box is checked, such as a check mark in another (contrasting) color.

alexanderbock:
The CSS for that is located in
apps/Launcher/main.cpp
lines 76 to 97

:)

EricMyers47:
Thanks, I'll take a look at it.  Meanwhile, I've at least documented the meaning in the wiki.   It makes sense that light is selected (in a low light environment), but a check mark will be more definite.

EricMyers47:
Note for later, this seems to be how to change the CSS to do this: http://doc.qt.io/qt-4.8/stylesheet-examples.html#customizing-a-qpushbutton-using-the-box-model

EricMyers47:
I like the idea, but I don't like the red (it looks like something is wrong) or the bright green, so I'll change it grey and a more gray green.

PR #228, opened 2017-02-13
Modification to Launcher checkboxes - clearly defined status-indicators.

noahdasanaike:
This is a simple alteration to how Launcher checkboxes display status; green indicates a check, whereas red does not.

This is an attempt to deal with [Issue #155.](https://github.com/OpenSpace/OpenSpace/issues/155)


EricMyers47:
Okay, let's not merge directly into master anymore.   We want this to go into develop first, so we can try it out, and then it will go into master when everything else goes into master.

PR #232, opened 2017-02-16
Feature/luaextensions

alexanderbock:


src/scripting/systemcapabilitiesbinding.cpp
```diff
@@ -0,0 +1,341 @@
...
+                "cacheSize",
+                &luascripting::general::cacheSize,
+                "",
+                "Returns the cache size."
+            },
+            {
+                "extensions",
+                &luascripting::general::extensions,
+                "",
+                "Returns all supported exteions as comma-separated string."
```
emiax: Is there a reason to join the extensions to one string? How about returning a lua array-of-strings?

src/scripting/scriptengine.cpp
```diff
@@ -41,12 +41,12 @@
 #include "scriptengine_lua.inl"
 
 namespace {
-    const std::string MainTemplateFilename = "${OPENSPACE_DATA}/web/luascripting/main.hbs";
-    const std::string ScriptingTemplateFilename = "${OPENSPACE_DATA}/web/luascripting/scripting.hbs";
-    const std::string HandlebarsFilename = "${OPENSPACE_DATA}/web/common/handlebars-v4.0.5.js";
-    const std::string JsFilename = "${OPENSPACE_DATA}/web/luascripting/script.js";
-    const std::string BootstrapFilename = "${OPENSPACE_DATA}/web/common/bootstrap.min.css";
-    const std::string CssFilename = "${OPENSPACE_DATA}/web/common/style.css";
+    const char* MainTemplateFilename = "${OPENSPACE_DATA}/web/luascripting/main.hbs";
```
emiax: I'm just curious: Why is const char* preferred rather than const std::string?

alexanderbock: const std::string has some initialization that happens at application startup time, whereas const char* can be directly used.  Probably noone will be able to notice, but i think it is a good habit not to use std::string if not necessary

src/scripting/systemcapabilitiesbinding.cpp
alexanderbock: Good idea;  I'll change that

PR #236, opened 2017-02-19
* When Launcher first starts only the default scene is selected.

EricMyers47:
This now works, and builds on Jenkins on all platforms.

### DIFF
--- a/apps/Launcher/syncwidget.cpp
+++ b/apps/Launcher/syncwidget.cpp
@@ -225,8 +225,11 @@ void SyncWidget::setSceneFiles(QMap<QString, QString> sceneFiles) {
         const QString& sceneName = keys[i];
 
         QCheckBox* checkbox = new QCheckBox(sceneName);
-        checkbox->setChecked(true);
-
+        checkbox->setChecked(false);
+        QString defaultName = "default";
+        if(QString::compare(defaultName, sceneName, Qt::CaseInsensitive)==0){
+            checkbox->setChecked(true);
+        }
         _sceneLayout->addWidget(checkbox, i / nColumns, i % nColumns);
     }
 }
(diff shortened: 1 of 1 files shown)


apps/Launcher/syncwidget.cpp
```diff
@@ -225,8 +225,11 @@ void SyncWidget::setSceneFiles(QMap<QString, QString> sceneFiles) {
         const QString& sceneName = keys[i];
 
         QCheckBox* checkbox = new QCheckBox(sceneName);
-        checkbox->setChecked(true);
-
+        checkbox->setChecked(false);
```
alexanderbock: By default the QCheckBox is unchecked, so this line is not needed

apps/Launcher/syncwidget.cpp
```diff
@@ -225,8 +225,11 @@ void SyncWidget::setSceneFiles(QMap<QString, QString> sceneFiles) {
...
+        QString defaultName = "default";
+        if(QString::compare(defaultName, sceneName, Qt::CaseInsensitive)==0){
```
alexanderbock: 1. Instead of `QString::compare`, it's better to use the overloaded `==` operator:  `defaultName == sceneName`.  The cases should all be lowercase anyway.

2. Coding style:
```
if (defaultName == sceneName) {
```

apps/Launcher/syncwidget.cpp
EricMyers47: Okay.  Do we want to keep it to be explicit, or rely on anyone reading it knowing that?

apps/Launcher/syncwidget.cpp
EricMyers47: Ah, very nice.  I didn't know that the operator was overlayed, and that does look much better.

apps/Launcher/syncwidget.cpp
alexanderbock: I'm more in favor of using the default value as it is;  we don't overwrite all the default values elsewhere in the code

EricMyers47:
Okay, cleaned it up as you suggested.

PR #239, opened 2017-02-22
Feature/modular sgct config

alexanderbock:


alexanderbock:
Most functionality is in
scripts/configuration_helper.lua and configurationmanager.cpp

PR #273, opened 2017-03-16
Pr/scenegraph refactor

emiax:
Any feedback is appreciated :)

Changelist:
- Move scene loading into separate class (SceneLoader)
- Camera is now owned by the scene
- Introduce scene manager to handle loading of scenes.
- Remove SceneGraph class. (Scene and SceneGraphNode remains)
- Scene owns the root node
- Nodes own their children (unique pointers)
- Add attachment/detachment API for scene graph nodes
- Remove "common" scene module
- Add unit tests for scene loading
- Gracefully unload invalid scene graph contents after loading (at least some, including circular dependencies, missing connection to root)
- SceneLoader API allows for importing mod files into existing scenes (as opposed to reloading full scene)

Future work: (This pull request will make it easier to implement...)
- The scene should be able to contain/own more than one camera.
- Load scenes asynchronously in the background (Scene manager can help out here)
- Possibly load/unload individual mod files in the background using the improved SceneLoader API.
- Do not rely on setting the global working directory while loading the scene (stateful => hard to parallelize)
- Improved DSG API
- Possibly move topologically sorted scene graph nodes from the Scene to a new UpdateManager class or similar. (Rendering algorithm does not need to traverse in this order, so it's not really a responsibility of Scene)
- ...

@GPayne This is scheduled for building on Jenkins, but the osx node seems to be down at the moment.


EricMyers47:
Still doesn't build on Linux.   See the log on Jenkins to figure out why, but it looks like something to do with the smart pointers:
`18:35:42 /usr/include/c++/5/bits/unique_ptr.h:74:22: error: invalid application of ‘sizeof’ to incomplete type ‘openspace::Renderer’`

I'll see if I can try it on Mac later today (the Jenkins Mac is down again, but there is a build queued if/when it resets).

emiax:
Thank you @EricMyers47 for looking into the issue and @GPayne for bringing the osx node back up (either manually or through cron magic)
I was able to solve the linux/osx issue now, and finally we have a green tick on this branch!

include/openspace/engine/openspaceengine.h
```diff
@@ -193,6 +199,9 @@ class OpenSpaceEngine {
     // Others
     std::unique_ptr<properties::PropertyOwner> _globalPropertyNamespace;
     
+    bool _switchScene;
```
alexanderbock: Bool variables should have some verb in them in order to make the if statements more readable and make them less ambigous For example `_shouldSwitchScene`  or `_scheduleSwitchScene`.

include/openspace/scene/scenegraphnode.h
```diff
@@ -63,21 +65,32 @@ class SceneGraphNode : public properties::PropertyOwner {
     SceneGraphNode();
     ~SceneGraphNode();
 
-    static SceneGraphNode* createFromDictionary(const ghoul::Dictionary& dictionary);
+    static std::unique_ptr<SceneGraphNode> createFromDictionary(const ghoul::Dictionary& dictionary);
 
     bool initialize();
     bool deinitialize();
 
+    void traversePreOrder(std::function<void(SceneGraphNode*)> fn);
```
alexanderbock: Some better names for these functions? `PreOrder` and `PostOrder` don't say pre *what*

include/openspace/scene/sceneloader.h
```diff
@@ -0,0 +1,119 @@
...
+* PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+* HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+* CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+* OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+****************************************************************************************/
+
+#ifndef __OPENSPACE_CORE___SCENELOADER___H__
+#define __OPENSPACE_CORE___SCENELOADER___H__
+
+#include <memory>
```
alexanderbock: Nitpicky:
Different order of includes:

1. `openspace`
2. `ghoul`
3. `others`

include/openspace/rendering/renderengine.h
```diff
@@ -32,6 +32,10 @@
 #include <openspace/properties/triggerproperty.h>
 #include <openspace/properties/scalar/boolproperty.h>
 #include <openspace/properties/scalar/intproperty.h>
+#include <openspace/rendering/screenspacerenderable.h>
```
alexanderbock: Nitpicky:
Sorting includes alphabetically

include/openspace/scene/sceneloader.h
```diff
@@ -0,0 +1,119 @@
...
+
+#include <openspace/scene/scenegraphnode.h>
+#include <openspace/util/camera.h>
+
+namespace openspace {
+
+class Scene;
+
+class SceneLoader {
+public:
```
alexanderbock: Do these two structs need to be public?  If they are only used internally, they can be private as well

modules/newhorizons/rendering/renderableplaneprojection.cpp
```diff
@@ -265,8 +265,9 @@ void RenderablePlaneProjection::updatePlane(const Image& img, double currentTime
     if (!_moving) {
         SceneGraphNode* thisNode = OsEng.renderEngine().scene()->sceneGraphNode(_name);
         SceneGraphNode* newParent = OsEng.renderEngine().scene()->sceneGraphNode(_target.node);
-        if (thisNode != nullptr && newParent != nullptr)
-            thisNode->setParent(newParent);
+        if (thisNode != nullptr && newParent != nullptr) {
```
alexanderbock: Autocasting to `bool` is fine for `ptr` types:

`if (thisNode && newParent) {`

src/engine/openspaceengine.cpp
```diff
@@ -497,22 +499,124 @@ void OpenSpaceEngine::initialize() {
...
+    LINFO("Finished initializing");
+}
+
+void OpenSpaceEngine::scheduleLoadScene(const std::string& scenePath) {
+    _switchScene = true;
+    _scenePath = scenePath;
+}
+
+void OpenSpaceEngine::loadScene(const std::string& scenePath) {
+
```
alexanderbock: One newline too much

src/engine/openspaceengine.cpp
```diff
@@ -497,22 +499,124 @@ void OpenSpaceEngine::initialize() {
...
-    // Initialize the RenderEngine
-    _renderEngine->setSceneGraph(sceneGraph);
     _renderEngine->initialize();
+
+    scheduleLoadScene(scenePath);
+
+    LINFO("Finished initializing");
+}
+
+void OpenSpaceEngine::scheduleLoadScene(const std::string& scenePath) {
```
alexanderbock: Here you can pass `scenePath` as a `std::string` instead and `std::move` it into the `_scenePath` variable.  You'll potentially save one copy

src/engine/openspaceengine.cpp
```diff
@@ -497,22 +499,124 @@ void OpenSpaceEngine::initialize() {
...
+    // Write keyboard documentation.
+    const std::string KeyboardShortcutsType =
+        ConfigurationManager::KeyKeyboardShortcuts + "." +
+        ConfigurationManager::PartType;
+
+    const std::string KeyboardShortcutsFile =
+        ConfigurationManager::KeyKeyboardShortcuts + "." +
+        ConfigurationManager::PartFile;
+
+    std::string type, file;
```
alexanderbock: Reducing scope:
```
const bool hasType = configurationManager().hasKeyAndValue<std::string>(KeyboardShortcutsType);
const bool hasFile = configurationManager().hasKeyAndValue<std::string>(KeyboardShortcutsFile);
if (hasType && hasFile) {
  interactionHandler().writeKeyboardDocumentation(
    configurationManager().value<std::string>(KeyboardShortcutsType),
    configurationManager().value<std::string>(KeyboardShortcutsFile)
);
}
```

src/engine/openspaceengine.cpp
```diff
@@ -497,22 +499,124 @@ void OpenSpaceEngine::initialize() {
...
+
+    const std::string KeyPropertyDocumentationFile =
+        ConfigurationManager::KeyPropertyDocumentation + '.' +
+        ConfigurationManager::PartFile;
+
+    hasType = configurationManager().hasKey(KeyPropertyDocumentationType);
+    hasFile = configurationManager().hasKey(KeyPropertyDocumentationFile);
+
+    if (hasType && hasFile) {
+        std::string propertyDocumentationType;
```
alexanderbock: Can use `configurationManager().value<std::string>` since you already checked that the key exists

src/engine/settingsengine.cpp
```diff
@@ -83,7 +83,7 @@ void SettingsEngine::initialize() {
         std::string sceneFile = _scenes.getDescriptionByValue(_scenes);
         OsEng.configurationManager().setValue(
             ConfigurationManager::KeyConfigScene, sceneFile);
-        OsEng.renderEngine().scene()->scheduleLoadSceneFile(sceneFile);
+        OsEng.scheduleLoadScene("${SCENE}/" + sceneFile);
```
alexanderbock: Here by dragons, if the scenefile is *not* in the `${SCENE}` directory

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
+    if (updateDeps) {
+        updateDependencies();
+    }
+}
 
-            // After loading the scene, the keyboard bindings have been set
-            const std::string KeyboardShortcutsType =
-                ConfigurationManager::KeyKeyboardShortcuts + "." +
-                ConfigurationManager::PartType;
+void Scene::removeNode(SceneGraphNode* node, Scene::UpdateDependencies updateDeps) {
```
alexanderbock: The `Scene` qualifier for `UpdateDependencies` is extra

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
-        try {          
-            loadSceneInternal(_sceneGraphToLoad);
+Camera* Scene::camera() const {
+    return _camera.get();
+}
 
-            // Reset the InteractionManager to Orbital/default mode
-            // TODO: Decide if it belongs in the scene and/or how it gets reloaded
-            //OsEng.interactionHandler().setInteractionMode("Orbital");
+void Scene::addNode(SceneGraphNode* node, Scene::UpdateDependencies updateDeps) {
```
alexanderbock: The `Scene` qualifier for `UpdateDependencies` is extra

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
 
-            const std::string KeyboardShortcutsFile =
-                ConfigurationManager::KeyKeyboardShortcuts + "." +
-                ConfigurationManager::PartFile;
+void Scene::updateDependencies() {
+    sortTopologically();
+}
 
+void Scene::sortTopologically() {
+    std::copy(_circularNodes.begin(), _circularNodes.end(), std::back_inserter(_topologicallySortedNodes));
```
alexanderbock: The `std::back_inserter` is potentially slow. A faster method is:
```
_topologicallySortedNodes.insert(
  _topologicallySortedNodes.end(), 
  make_move_iterator(_circularNodes.begin()),
  make_move_iterator(_circularNodes.end())
);
```

Your version has to call `push_back` many times, the other version does only zero or one reallocation and a single move.

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
+    zeroInDegreeNodes.push(root);
+    
+    std::vector<SceneGraphNode*> nodes;
+    nodes.reserve(_topologicallySortedNodes.size());
+    while (!zeroInDegreeNodes.empty()) {
+        SceneGraphNode* node = zeroInDegreeNodes.top();
+        nodes.push_back(node);
+        zeroInDegreeNodes.pop();
+
+        for (auto& n : node->dependentNodes()) {
```
alexanderbock: Better to not use `auto` whereever reasonable

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
+            auto it = inDegrees.find(n);
```
alexanderbock: Here it is fine.. writing out iterators is pain

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
+            it->second -= 1;
+            if (it->second == 0) {
+                zeroInDegreeNodes.push(n);
+                inDegrees.erase(it);
+            }
         }
-        catch (const ghoul::RuntimeError& e) {
-            LERRORC(e.component, e.what());
+        for (auto& n : node->children()) {
```
alexanderbock: Better to not use `auto` whereever reasonable

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
+            auto it = inDegrees.find(n);
```
alexanderbock: Here it is fine.. writing out iterators is pain

src/scene/scene.cpp
```diff
@@ -79,148 +82,122 @@ namespace {
...
-    OnExit(
-        [](){ OsEng.windowWrapper().setSynchronization(true); }
-    );
-    
-    lua_State* state = ghoul::lua::createNewLuaState();
-    OnExit(
-           // Delete the Lua state at the end of the scope, no matter what
-           [state](){ ghoul::lua::destroyLuaState(state); }
-           );
+    for (auto& it : inDegrees) {
```
alexanderbock: `auto`

src/scene/scene.cpp
```diff
@@ -232,221 +209,70 @@ bool Scene::loadSceneInternal(const std::string& sceneDescriptionFilePath) {
             LERRORC(std::string(_loggerCat) + "(" + e.component + ")", e.what());
         }
     }
+}
 
-    // update the position of all nodes
-    // TODO need to check this; unnecessary? (ab)
-    for (SceneGraphNode* node : _graph.nodes()) {
+void Scene::update(const UpdateData& data) {
+    for (auto& node : _topologicallySortedNodes) {
```
alexanderbock: `auto`

src/scene/scenegraphnode.cpp
```diff
@@ -200,6 +197,20 @@ bool SceneGraphNode::deinitialize() {
     return true;
 }
 
+void SceneGraphNode::traversePreOrder(std::function<void(SceneGraphNode*)> fn) {
+    fn(this);
+    for (auto& child : _children) {
```
alexanderbock: `auto`

src/scene/scenegraphnode.cpp
```diff
@@ -200,6 +197,20 @@ bool SceneGraphNode::deinitialize() {
...
+        child->traversePreOrder(fn);
+    }
+}
+
+void SceneGraphNode::traversePostOrder(std::function<void(SceneGraphNode*)> fn) {
+    for (auto& child : _children) {
```
alexanderbock: `auto`

src/scene/scenegraphnode.cpp
```diff
@@ -356,35 +367,134 @@ void SceneGraphNode::render(const RenderData& data, RendererTasks& tasks) {
...
-    _parent = parent;
+    if (_scene && updateScene) {
+        _scene->updateDependencies();
+    }
 }
 
-void SceneGraphNode::addChild(SceneGraphNode* child) {
-    _children.push_back(child);
+void SceneGraphNode::clearDependencies(UpdateScene updateScene) {
+    for (auto& dependency : _dependencies) {
```
alexanderbock: `auto`

src/scene/scenegraphnode.cpp
```diff
@@ -356,35 +367,134 @@ void SceneGraphNode::render(const RenderData& data, RendererTasks& tasks) {
...
-//
-//    if (it != _children.end()){
-//        _children.erase(it);
-//        return true;
-//    }
-//
-//    return false;
-//}
+    _dependencies = dependencies;
+    for (auto& dependency : dependencies) {
```
alexanderbock: `auto`

src/scene/scenegraphnode.cpp
```diff
@@ -543,15 +665,16 @@ bool SceneGraphNode::sphereInsideFrustum(const psc& s_pos, const PowerScaledScal
...
 }
+*/
 
 SceneGraphNode* SceneGraphNode::childNode(const std::string& name)
 {
     if (this->name() == name)
         return this;
     else
-        for (SceneGraphNode* it : _children) {
+        for (auto& it : _children) {
```
alexanderbock: `auto`

src/scene/scenemanager.cpp
```diff
@@ -0,0 +1,52 @@
...
+ *                                                                                       *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,   *
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include <openspace/scene/scenemanager.h>
```
alexanderbock: Nitpicky:  Ordering

include/openspace/scene/scenegraphnode.h
emiax: These are standard names for traversal orders: https://en.wikipedia.org/wiki/Tree_traversal#Pre-order_2

include/openspace/engine/openspaceengine.h
emiax: Will switch to _scheduledSceneSwitch

include/openspace/scene/sceneloader.h
emiax: :+1:

include/openspace/scene/sceneloader.h
emiax: Good point. I think I originally intended to expose public methods for returning these as intermediate steps in the loading (to allow other classes to control the loading process in more detail). Since this is not exposed in the interface right now, it makes more sense to make them private.

src/engine/openspaceengine.cpp
emiax: How will this potentially save a copy? Is this correct:
Const ref: No copy into the new stack frame. 1 copy from the reference to the member variable. Total 1 copy.
Pass by value and move semantics: One copy into the new stack frame. No copy into the member value (only swap pointers). Total 1 copy.

Semantically though, I guess one could argue that since the method is making a copy of the value anyway, it makes sense to show that explicitly in the signature, instead of hiding that in the implementation. However, you could also argue that the fact that the value is stored in the class is an implementation detail, and if we change the internal structure of the class, the API can be unchanged (while the copy is avoided) if we keep it as const std::string&.

:) I'll change it to std::string and std::move for now, but I think this is an interesting discussion

src/engine/settingsengine.cpp
emiax: I agree, but dragons were here before as well. I'll clarify the dragons by passing in the constant sceneDir instead. Since the scene files are collected in the same method, I think this is fine until future refactors.

src/scene/scenegraphnode.cpp
emiax: _children is a vector of unique pointers: cannot copy those pointers.

src/scene/scenegraphnode.cpp
emiax: _children is a vector of unique pointers: cannot copy those pointers.

src/scene/scenegraphnode.cpp
emiax: _children is a vector of unique pointers: cannot copy those pointers.

src/scene/scenegraphnode.cpp
emiax: _children is a vector of unique pointers: cannot copy those pointers.

src/engine/openspaceengine.cpp
emiax: I managed to reduce scope with putting documentation generation in 2 separate blocks.
I'd rather stick to checking for existence and retrieving in the same step. (Saving one string based lookup. Not that it matters here, but I think it's good practice anyway.)

include/openspace/scene/scenegraphnode.h
alexanderbock: 👍  Didn't know those terms

src/scene/scenegraphnode.cpp
alexanderbock: You can use:  `std::unique_ptr<SceneGraphNode>& child`

src/scene/scenegraphnode.cpp
alexanderbock: You can use:  `std::unique_ptr<SceneGraphNode>& child`

src/scene/scenegraphnode.cpp
alexanderbock: You can use:  `std::unique_ptr<SceneGraphNode>& child`

PR #286, opened 2017-04-10
Feature/virtualproperty

alexanderbock:
Implementing a virtual property feature that is used in combination with the regular rexpression and tag syntax for properties.  Virtual properties do not represent a program state per se, but is used to access a list of properties instead.

An example is provided in the `default.scene`, where a virtual property is added that toggles all trails


modules/onscreengui/src/renderproperties.cpp
```diff
@@ -46,13 +46,22 @@ void renderTooltip(Property* prop) {
...
     OsEng.scriptEngine().queueScript(script, scripting::ScriptEngine::RemoteScripting::Yes);
 }
 
+void executeScriptGroup(const std::string& id, const std::string& value) {
+    std::string script =
+        "openspace.setPropertyValue('" + id + "', " + value + ");";
+    OsEng.scriptEngine().queueScript(script, scripting::ScriptEngine::RemoteScripting::Yes);
+}
+
+#define executeScript executeScriptGroup
```
emiax: Why this?

src/engine/openspaceengine.cpp
```diff
@@ -176,6 +177,13 @@ OpenSpaceEngine::OpenSpaceEngine(
     TransformationManager::initialize();
 }
 
+OpenSpaceEngine::~OpenSpaceEngine() {
+    // We know that we own all of the properties in the virtual namespace
+    for (properties::Property* prop : _virtualPropertyNamespace->propertiesRecursive()) {
+        delete prop;
```
emiax: I guess we should consider creating unique pointers to props later on, and pass ownership to the propery owners. Probably not in this PR though.

modules/onscreengui/src/renderproperties.cpp
```diff
@@ -46,13 +46,22 @@ void renderTooltip(Property* prop) {
     }
 }
 
-void executeScript(const std::string& id, const std::string& value) {
+void executeScriptSingle(const std::string& id, const std::string& value) {
```
emiax: I understand that this is more efficient since it does not use regex, but is it ever used?

src/query/query.cpp
```diff
@@ -108,15 +108,24 @@ std::vector<properties::Property*> allProperties() {
         p.end()
     );
 
+    std::vector<properties::Property*> q =
+        OsEng.virtualPropertyOwner().propertiesRecursive();
+
+    properties.insert(
```
emiax: Since `allProperties` are used from`property_setValueRegex`, it seems a bit strange that virtual properties (which are already regex-formatted) are also included in the search. Is there a reason why virtual properties are added to this list?

modules/onscreengui/src/renderproperties.cpp
alexanderbock: Woops.. It's a leftover from the testing phase.

src/engine/openspaceengine.cpp
alexanderbock: Sounds good

modules/onscreengui/src/renderproperties.cpp
alexanderbock: Yes, if I write keybindings I'll do it with the `setPropertyValueSingle` function. I think it is good to have the option in case we don't want to pay the cost of the regex

emiax: Agreed! I just wanted to check if you meant to use the method somewhere in this pull request. Right now, all calls are made to the `executeScript` (alias for `executeScriptGroup`) (as far as I can see)

PR #288, opened 2017-04-13
Feature/domecasting

emiax:
- Update parallel connection to work better with GlobeBrowsing interaction mode and with dynamic time. Interpolation is now perforned in a coordinate frame relative to the camera focus node. 
- Replace buffer size heuristics with manual settings, yielding a more predictable behavior.
- Add properties to control buffer sizes, keyframe intervals, time tolerance etc.
- Add GUI for parallel connection.


src/engine/openspaceengine.cpp
```diff
@@ -132,7 +133,7 @@ OpenSpaceEngine::OpenSpaceEngine(
     , _settingsEngine(new SettingsEngine)
     , _timeManager(new TimeManager)
     , _downloadManager(nullptr)
-    , _parallelConnection(new ParallelConnection)
+    , _parallelConnection(std::make_unique<ParallelConnection>())
```
alexanderbock: `new ParallelConnection` would be fine as well..  The old version calls the constructor taking a raw pointer and taking ownership, the new version calls the move constructor

src/network/parallelconnection_lua.inl
```diff
@@ -101,6 +101,29 @@ int setPassword(lua_State* L) {
     }
 }
 
+int setHostPassword(lua_State* L) {
+    bool isFunction = (lua_isfunction(L, -1) != 0);
```
alexanderbock: You can check the number of arguments first with:
```
    int nArguments = lua_gettop(L);
    SCRIPT_CHECK_ARGUMENTS("loadScene", L, 1, nArguments);
```

`SCRIPT_CHECK_ARGUMENTS` is defined in `openspace/scripting/script_helper.h`

src/engine/openspaceengine.cpp
emiax: Yeah, I figured, but I think this makes it more explicit what's going on :)

src/network/parallelconnection_lua.inl
emiax: Nice, thanks :) Since all those setters just control properties now, I can actually kill these lua methods.

PR #289, opened 2017-04-13
Feature/planetradius

emiax:
- Removed some power scaled coordinates
- Removed some dead code
- Removed spice dependency from sphere geometry class
- Make scene graph nodes rely on their renderable to compute their bounding sphere

@jonathanbosson, this fixes the issue with the uninitialized bounding boxes for RenderablePlanet.
Not sure if we should pull in this into master before the upcoming Earth Day event, but you can merge it into your touch branch temporarily if you want.

### DIFF
--- a/src/rendering/renderengine.cpp
+++ b/src/rendering/renderengine.cpp
@@ -282,8 +282,6 @@ void RenderEngine::updateScene() {
         Time::ref().timeJumped(),
         _performanceManager != nullptr
     });
-
-    _scene->evaluate(_camera);
     
     LTRACE("RenderEngine::updateSceneGraph(end)");
 }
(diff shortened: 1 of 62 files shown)


src/rendering/renderengine.cpp
```diff
@@ -282,8 +282,6 @@ void RenderEngine::updateScene() {
         Time::ref().timeJumped(),
         _performanceManager != nullptr
     });
-
-    _scene->evaluate(_camera);
```
alexanderbock: This was not necessary?

alexanderbock:
It's fine to merge it back, I created a branch for the earth day

src/rendering/renderengine.cpp
emiax: Nope, it was only used for an old version of frustum culling that was commented out anyways.

PR #309, opened 2017-05-17
Feature/luaconsole

alexanderbock:
Rendering a better version of the Lua console that looks nicer, is easier to read, and shows the history of the console


include/openspace/properties/property.h
```diff
@@ -293,6 +293,18 @@ class Property {
     void setReadOnly(bool state);
 
     /**
+    * Default view options that can be used in the Property::setViewOption method. The
+    * values are: Property::ViewOptions::Color = <code>color</code>,
+    * Property::ViewOptions::LightPosition = <code>lightPosition</code>,
+    * Property::ViewOptions::PowerScaledScalar = <code>powerScaledScalar</code>, and
```
kbladin: Remove comments on PSC

kbladin:
Current issues on OSX:

- `luaconsole.vert` does not compile when defining `const float y`. Works if removing the `const` initialiser. What was the reason for `const`?
- Text does not render
- The size is wrong due to retina screen resolution. The console is half the height it should be.
- On my laptop (not sure for others) I have to press the key `<`to bring up the console instead of the regular `§` key otherwise used. This was a problem before the new console but should probably be adressed.
- Is it possible to make use of the `cmd` key for pasting text instead of `ctrl`? Just a suggestion.

kbladin:
Reopen

kbladin:
Retina issue solved.

alexanderbock:
The problem with the keys is due to the keyboard language settings..  As it says in the `luaconsole.cpp:272`:
```cpp
    if (key == CommandInputButton) {
        // Button left of 1 and above TAB
        // How to deal with different keyboard languages? ---abock
```

I don't have a good idea since I'm not sure if using the Windows API to query the keyboard language will be a good solution either

alexanderbock:
Leaving the language issue open for now (#311)

PR #312, opened 2017-05-19
Feature/launcherconfigs

territo-sci:
Creates a configuration section for the Launcher in openspace.cfg to prevent slow console output, especially on Windows.

Allows you to set the logging levels to `Trace`, `Debug`, `Info`, `Warning`, `Error`, `Fatal`, or `None` in the config file, or on the command line with `-d` or `-debug` 0-7.


apps/Launcher/mainwindow.cpp
```diff
@@ -263,3 +329,20 @@ void MainWindow::newsReadyRead() {
...
+    _optionParser->setApplicationDescription("OpenSpace Launcher");
+    _optionParser->addHelpOption();
+    _optionParser->addVersionOption();
+
+    _optionParser->addOptions({
+        { { "d", "debug" },
+        QCoreApplication::translate("main", "Debug output level"),
+        QCoreApplication::translate("main", "0, 1, 2, 3, 4, 5, 6"),
+        QCoreApplication::translate("main", "1"),
+        }
```
alexanderbock: Shouldn't this be used in the `configureLogging`?

territo-sci: Indeed it should. I think I pulled it out while I was testing things out and never put it back in. Thanks for catching it!

PR #333, opened 2017-06-20
Feature/packaging

eriksunden:
Packing for both Windows (to ZIP) and OSX (to TGZ) is working correctly using CPACK, by just building the "PACKAGE" project generated by CMake.

I would recommend distributing OpenSpace binaries primarily as these archives, as installers are not actual making anything extra for this project, everything is already self-contained in the archive, nothing has to be installed or set in the OS itself anyway, so installers only add cosmetics, and makes it time-consuming for developers.

However, installer creation works as well, on Windows only, by using NSIS generator when running the packaging. The CMake option "OPENSPACE_CREATE_INSTALLER" needs to be checked for this to be happening when the package project is built. It will also create an archive alongside the installer. [NSIS v2.51](https://sourceforge.net/projects/nsis/files/NSIS%202/) is the one I have installed on my PC.
Installer on OSX (meaning, creating a bundle) is a more time-consuming project, so the benefit of that needs to be clear. 

I have also deleted the openspace.h file, as it is now generated by CMake such that all version numbering is specified in CMake instead, such that this can be used in naming the archive/installer.

I now include the "documentation" folder in the archive, which is generated at runtime, meaning that the OpenSpace application needs to be started at least one time in a fresh build before the package project can be built.

On another note, the "data" folder is also include in the packing, as OpenSpace will crash without it. However, if this contains a lot of data, packing will take a long time. I would also argue that the data should NOT be distributed inside the generated archive, if it is of substantial size.


klaseskilson:
> Installer on OSX (meaning, creating a bundle)

Do you mean building a `.app` or a `.dmg`? I have no experience when it comes to `.dmg`, but there is a working `.app` CMake setup in the [`feature/add-cef`](https://github.com/openspace/openspace/tree/feature/add-cef) branch, if that is something that might be interesting here.

PR #334, opened 2017-06-21
Feature/globebrowsing

kbladin:
Layer support for globe browsing:

- Add layers using the function openspace.globebrowsing.addLayer
- Delete layers using openspace.globebrowsing.deleteLayer
- Layer type does not necessarily have to be of tile type. For example solidcolor does not use tiles
- Blend modes for layers are Normal, Add, Subtract, Multiply, Color
- Layer adjustments to affect layers. The current only active one is chroma key to cut out a color from the layer. Transfer functions or clipping masks are examples of layer adjustments for the future.
- Support for adding layer specifications for quickly accessing GIBS layers:
  - openspace.globebrowsing.createGibsGdalXml
  - openspace.globebrowsing.createTemporalGibsGdalXml
  - The arguments for these functions are currently strings. Would it be better to use a lua dictionary?
- No data values for height layers are correctly regarded (can be seen on Earth. No longer bumps on the poles)

Other minor things:
- Worked a bit on point globe to render globes at large distances. Currently not in use and doesn't have anything to do with the other things.
- Concurrent job manager takes a thread pool as argument and not a pointer to one. This is because the concurrent job manager needs to have ownership of the thread pool for correct deinitialization. Will cause breaking change for users of concurrent job manager if merged in to master.

PS: I accidentally reused an old branch so some commits that are already in master shows up in the list..

PR #354, opened 2017-07-10
Globebrowsing tmp

kbladin:
Made a new pull request since Jenkins did not like the other branch.. It had some cached files left so better to start from a new branch branched from master. Same changes as for the previous pull request so comments have already been looked over. I will remove both branch once this is merged in.

From the original pull request:

Layer support for globe browsing:

- Add layers using the function openspace.globebrowsing.addLayer
- Delete layers using openspace.globebrowsing.deleteLayer
- Layer type does not necessarily have to be of tile type. For example solidcolor does not use tiles
- Blend modes for layers are Normal, Add, Subtract, Multiply, Color
- Layer adjustments to affect layers. The current only active one is chroma key to cut out a color from the layer. Transfer functions or clipping masks are examples of layer adjustments for the future.
- Support for adding layer specifications for quickly accessing GIBS layers:
  - openspace.globebrowsing.createGibsGdalXml
  - openspace.globebrowsing.createTemporalGibsGdalXml
  - The arguments for these functions are currently strings. Would it be better to use a lua dictionary?
- No data values for height layers are correctly regarded (can be seen on Earth. No longer bumps on the poles)

Other minor things:
- Worked a bit on point globe to render globes at large distances. Currently not in use and doesn't have anything to do with the other things.
- Concurrent job manager takes a thread pool as argument and not a pointer to one. This is because the concurrent job manager needs to have ownership of the thread pool for correct deinitialization. Will cause breaking change for users of concurrent job manager if merged in to master.


kbladin:
No need to look at the code. Already done at #334. This contains the same code. However this is the one that will be merged once Jenkins accepts it.

PR #358, opened 2017-07-12
Feature/multitouch

jonathanbosson:
MS thesis with multitouch interface implementation is done.

[OpenSpace Multi-Touch Interface Instructions](https://github.com/OpenSpace/OpenSpace/files/1142961/OpenSpace.Multi-Touch.Interface.Instruction.pdf)

**Includes:**
- Client side of TUIO protocol that allows OpenSpace to listen for new touch input from any device using a TUIO server
- Rendered screen-space markers (for now only for touch input) to show where the user interacts
- Two intertwined interaction mode (velocity-inertia and direct-manipulation) to allow for intuitive control through touch devices. Direct-manipulation uses a non-linear least square minimization method called LMA.
- Ability to trace objects in the scene the input points at and project model/world view coordinates to screen-space

**Possible improvements:**
- Refactor new classes and current interaction classes into a more abstract style that handles any type of input.
- Put some functions such as double tap or raytracing into LUA callbacks such that they're not locked to touch input.
- Improve LMA's ability to converge when 3+ fingers are exceedingly close to each other.


emiax:
Great job!
To get this mergable, there are a few things that need to be done/looked into.

- All tabs need to be converted to spaces.
- Is it possible to include TUIO as a submodule? If not, we should fork the project and keep it in a separate repository. This will make it much easier to update and reason about which code that is OpenSpace specific.

jonathanbosson:
We could make [TUIO11](https://github.com/mkalten/TUIO11_CPP) a submodule, but it contains a lot that we don't need. It's only really the `TUIO` and `oscpack` subdirectories we need.

emiax:
I would still prefer having it as a submodule, unless it's ridiculously heavy. What do you think, @alexanderbock ?

jonathanbosson:
It's 416kb out of  5401kb we need. 5 Mb isn't much though so I think you're right that we can go for it.
I remember I had some trouble with the original include path for files in those subdirectories though, to get it to compile I had to add `oscpack` or `TUIO` to the include path. For example `#include "oscpack/ip/PacketListener.h"` over the original `#include "ip/PacketListener.h"`.
Is there a way I can add add include paths in the `CMakeLists.txt` so that we don't have to change the source of the submodule?

.gitignore
```diff
@@ -119,3 +119,6 @@ data/spice/OsirisRexKernels
 data/spice/plu055.bsp
 data/spice/Rosetta
 data/spice/sat375.bsp
+data/scene/lodglobes/saturn/Launcher.config
```
alexanderbock: No need for these

CREDITS.md
```diff
@@ -21,7 +21,7 @@ Rickard Lindtstedt
 Michael Sjöström
 Michael Novén
 Oskar Carlbaum
-Jonathas Bosson
+Jonathan Bosson
```
alexanderbock: Oops.. sorry for that

data/scene/missions/juno/juno/juno.mod
```diff
@@ -126,16 +126,19 @@ return {
...
-            Translation = {
-                Type = "SpiceTranslation",
-                Body = "JUNO",
-                Observer = "JUPITER BARYCENTER"
-            },
-            Color = { 0.70, 0.50, 0.20 },
-            StartTime = "2016 JUN 28",
-            EndTime = "2016 APR 01",
-            SampleInterval = 3600
+            Type = "RenderableTrail",
```
alexanderbock: This is removing the new specification and replacing it with the old ...   it shouldn't be changed at all

data/spice/naif0011.tls
```diff
@@ -0,0 +1,148 @@
+KPL/LSK
```
alexanderbock: The file was replaced with naif0012.tls

This can be deleted

data/web/log/script.js
```diff
@@ -1,4 +1,4 @@
-var levels = ['trace', 'debug', 'info', 'warning', 'error', 'fatal'];
+var levels = ['debug', 'info', 'warning', 'error', 'fatal'];
```
alexanderbock: Not sure why the `trace` part was lost here and in the next one

modules/touch/include/TouchMarker.h
```diff
@@ -0,0 +1,82 @@
...
+
+class TouchMarker : public properties::PropertyOwner
+{
+    public:
+        TouchMarker();
+
+        bool initialize();
+        bool deinitialize();
+
+        void render(const std::vector<TUIO::TuioCursor> list);
```
alexanderbock: `const ...&`?

modules/touch/include/TouchMarker.h
```diff
@@ -0,0 +1,82 @@
...
+
+
+    private:
+        void createVertexList(const std::vector<TUIO::TuioCursor> list);
```
alexanderbock: `const std::vector&`?  If you move the values out of the `vector`, you can use non-`const` instead

modules/touch/src/TouchInteraction.cpp
```diff
@@ -0,0 +1,779 @@
...
+    , _spinSensitivity("Sensitivity of spinning", "Sensitivity of spinning in direct-manipulation", 1.0f, 0, 2)
+    , _inputStillThreshold("Input still", "Threshold for interpreting input as still", 0.0005f, 0, 0.001)
+    , _centroidStillThreshold("Centroid stationary", "Threshold for stationary centroid", 0.0018f, 0, 0.01) // used to void wrongly interpreted roll interactions
+    , _interpretPan("Pan delta distance", "Delta distance between fingers allowed for interpreting pan interaction", 0.015f, 0, 0.1)
+    , _slerpTime("Time to slerp", "Time to slerp in seconds to new orientation with new node picking", 3, 0, 5)
+    , _guiButton("GUI Button", "GUI button size in pixels.", glm::ivec2(32, 64), glm::ivec2(8, 16), glm::ivec2(128, 256))
+    , _friction("Friction", "Friction for different interactions (orbit, zoom, roll, pan)", glm::vec4(0.01, 0.025, 0.02, 0.02), glm::vec4(0.0), glm::vec4(0.2))
+    
+    , _vel{ glm::dvec2(0.0), 0.0, 0.0, glm::dvec2(0.0) }
+    , _sensitivity{glm::dvec2(0.0808181818181818, 0.0454545454545455), 4.0, 2.75, glm::dvec2(0.0808181818181818, 0.0454545454545455) }
```
alexanderbock: Reason for the values?

modules/touch/src/TuioEar.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include <openspace/engine/wrapper/windowwrapper.h>
+#include <openspace/interaction/interactionhandler.h>
+#include <openspace/rendering/renderengine.h>
+#include <openspace/rendering/screenspacerenderable.h>
+
+#include <ghoul/logging/logmanager.h>
+
+using namespace TUIO;
+namespace {
+    const std::string _loggerCat = "TuioEar";
```
alexanderbock: `_loggerCat` not used in this file and cane be removed

modules/touch/src/TuioEar.cpp
```diff
@@ -0,0 +1,162 @@
...
+    _mx.unlock();
+}
+
+// save id to be removed and remove it in clearInput
+void TuioEar::removeTuioCursor(TuioCursor *tcur) {
+    _mx.lock();
+    _removeList.push_back(tcur->getSessionID());
+
+    // Check if the cursor ID could be considered a tap
+    double dist = 0;
```
alexanderbock:
```cpp
double dist = std::accumulate(tcur->getPath().begin(), tcur->getPath().end(), 0.0, [](double a, const TuioPoint& p) { return a + glm::length(...);});
```

modules/touch/src/TuioEar.cpp
```diff
@@ -0,0 +1,162 @@
...
+void TuioEar::addTuioBlob(TuioBlob *tblb) { }
+
+void TuioEar::updateTuioBlob(TuioBlob *tblb) { }
+
+void TuioEar::removeTuioBlob(TuioBlob *tblb) { }
+
+void TuioEar::refresh(TuioTime frameTime) { } // about every 15ms
+
+std::vector<TuioCursor> TuioEar::getInput() {
+    std::lock_guard<std::mutex> lock(_mx);
```
alexanderbock: This lock_guard shouldn't be necessary as the return value will be elided

modules/touch/src/TuioEar.cpp
```diff
@@ -0,0 +1,162 @@
...
+        _tap = false;
+        return !_tap;
+    }
+    else {
+        return _tap;
+    }
+}
+
+TuioCursor TuioEar::getTap() {
+    std::lock_guard<std::mutex> lock(_mx);
```
alexanderbock: This lock_guard shouldn't be necessary as the return value will be elided

modules/touch/touchmodule.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <glm/ext.hpp>
+
+#include <sstream>
+#include <string>
+#include <iostream>
+
+using namespace TUIO;
+
+namespace {
+    const std::string _loggerCat = "TouchModule";
```
alexanderbock: This `_loggerCat` is not used

openspace.cfg
```diff
@@ -10,7 +10,10 @@ return {
     SGCTConfig = sgct.config.single{},
 
     -- A regular 1920x1080 window
-    -- SGCTConfig = sgct.config.single{1920, 1080},   
+    -- SGCTConfig = sgct.config.single{1920, 1080}, 
+    
+    -- A windowed 1920x1080 fullscreen
+    SGCTConfig = sgct.config.single{1920, 1080, border=false, windowPos={0,0}, shared=true, name="WV_OBS_SPOUT1"},
```
alexanderbock: This should be commented out

src/interaction/interactionmode.cpp
```diff
@@ -262,74 +263,77 @@ OrbitalInteractionMode::MouseStates::MouseStates(double sensitivity, double velo
 {}
 
 void OrbitalInteractionMode::MouseStates::updateMouseStatesFromInput(const InputState& inputState, double deltaTime) {
-    glm::dvec2 mousePosition = inputState.getMousePosition();
+    ghoul::any t = property("Global Properties.Touch.TouchInteraction.TouchEvents")->get();
```
alexanderbock: I think it would be necessary here to check first whether the module exists?

modules/touch/src/TouchInteraction.cpp
jonathanbosson: Reason is to nullify sensitivity difference between X-axis and Y-axis on a 16:9 aspect ratio. Not the best solution as aspect ratio easily can change, it's not a super important feature for the user experience anyways. I'll change them to `0.08` and `0.045` for now.

src/interaction/interactionmode.cpp
jonathanbosson: I'm not sure, if the module doesn't exist `ghoul::any t` will return false and should thus work fine without included module as well. I can encapsulate the start and end of the if-statement with module-checks for safety.

modules/touch/src/TuioEar.cpp
jonathanbosson: I wish this worked. It seems the `std::list` that is returned from `tcur->getPath()` is a circle-linked list. We never reach `.end()` when I try it, just repeatedly goes through the entire list.

PR #363, opened 2017-07-13
Feature/globe shading

kbladin:
Enable the ability to calculate accurate normals for globes using extra samples from the height layers without the use of bump / normal maps.

Also: Oren Nayar diffuse shading to get better representation of surface reflections (Moon should have a higher roughness for example).

Useful for showing the terminator of globes as well as solving some of the issues regarding the atmosphere implementation.

Bonus:
An issue with scaled globes solved. Now heights are represented correctly even if the model transform of the globe changed.


modules/globebrowsing/globes/renderableglobe.cpp
```diff
@@ -65,11 +65,12 @@ RenderableGlobe::RenderableGlobe(const ghoul::Dictionary& dictionary)
         BoolProperty("enabled", "Enabled", true),
         BoolProperty("performShading", "perform shading", true),
         BoolProperty("atmosphere", "atmosphere", false),
+        BoolProperty("useAccurateNormals", "useAccurateNormals", false),
         FloatProperty("lodScaleFactor", "lodScaleFactor",10.0f, 1.0f, 50.0f),
-        FloatProperty("cameraMinHeight", "cameraMinHeight", 100.0f, 0.0f, 1000.0f)
+        FloatProperty("cameraMinHeight", "cameraMinHeight", 100.0f, 0.0f, 1000.0f),
+        FloatProperty("orenNayarRoughness", "orenNayarRoughness", 0.0f, 0.0f, 1.0f)
```
alexanderbock: That's a cool name

modules/globebrowsing/globes/renderableglobe.h
```diff
@@ -32,6 +32,9 @@
 
 #include <openspace/properties/scalar/floatproperty.h>
 #include <openspace/properties/scalar/intproperty.h>
+#include <openspace/properties/scalar/boolproperty.h>
+
+#include <string>
```
alexanderbock: Why add `<string>`?

modules/globebrowsing/rendering/chunkrenderer.cpp
```diff
@@ -112,6 +112,76 @@ ghoul::opengl::ProgramObject* ChunkRenderer::getActivatedProgramWithTileData(
...
 }
 
+void ChunkRenderer::setCommonUniforms(ghoul::opengl::ProgramObject& programObject,
+                                      const Chunk& chunk, const RenderData& data)
+{
+    glm::dmat4 modelTransform = chunk.owner().modelTransform();
+    glm::dmat4 viewTransform = data.camera.combinedViewMatrix();
+    glm::dmat4 modelViewTransform = viewTransform * modelTransform;
+
+    if (_layerManager->layerGroup(
```
alexanderbock: If the short-circuit evaluation is not necessary, it might be simpler to define `const bool` before the if and make the if statement itself shorter and thus easier to understand.

Or:  Instead of `.size() > 0` use `!XX.empty()`

modules/globebrowsing/shaders/blending.hglsl
```diff
@@ -27,27 +27,27 @@
 
 vec4 blendNormal(vec4 oldColor, vec4 newColor)
 {
```
alexanderbock: `{` on new line

modules/globebrowsing/shaders/blending.hglsl
```diff
@@ -94,43 +94,43 @@ vec3 HueShift (in vec3 Color, in float Shift)
...
+
+        if ( h < 0.0) {
+            h += 6.0;
+        }
+        h = h / 6.0;
+    }
+    return vec3( h, s, l );
 }
 
 vec3 rgb2hsv(vec3 c)
```
alexanderbock: `{`

modules/globebrowsing/shaders/globalchunkedlodpatch_vs.glsl
```diff
@@ -70,11 +79,24 @@ void main() {
...
 
     // Get the height value
-    float height = getTileVertexHeight(in_uv, levelWeights);
+    float height = getTileHeight(in_uv, levelWeights);
 
     // Apply skirts
     height -= getTileVertexSkirtLength();
-    
+
+#if USE_ACCURATE_NORMALS
```
alexanderbock: Is the `#if` necessary?

modules/globebrowsing/globes/renderableglobe.h
kbladin: Whoops

modules/globebrowsing/globes/renderableglobe.cpp
kbladin: It's a reflectance model

modules/globebrowsing/shaders/globalchunkedlodpatch_vs.glsl
kbladin: I don't want to have to calculate tangents unless they are used. And they are only used if accurate normals are used.

PR #403, opened 2017-10-09
Feature/fieldlinesSequence

Carlbaum:


modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    RenderEngine& renderEngine = OsEng.renderEngine();
+    if (_shaderProgram) {
+        renderEngine.removeRenderProgram(_shaderProgram);
+        _shaderProgram = nullptr;
+    }
+
+    // Stall main thread until thread that's loading states is done!
+    while (_isLoadingStateFromDisk) {
+        LWARNING("TRYING TO DESTROY CLASS WHEN A THREAD USING IT IS STILL ACTIVE");
```
emiax: Looks like this could result in infinite printing?

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -0,0 +1,323 @@
...
+bool RenderableFieldlinesSequence::isReady() const {
+    return _isReady;
+}
+
+void RenderableFieldlinesSequence::render(const RenderData& data, RendererTasks&) {
+    if (_activeTriggerTimeIndex != -1) {
+        _shaderProgram->activate();
+
+        // Calculate Model View MatrixProjection
+        const glm::dmat4 ROT_MAT = glm::dmat4(data.modelTransform.rotation);
```
emiax: Regular camelCase for constants. ALL_CAPS indicates preprocessor macros.

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -0,0 +1,323 @@
...
+    }
+}
+
+void RenderableFieldlinesSequence::update(const UpdateData& data) {
+    if (_shaderProgram->isDirty()) {
+        _shaderProgram->rebuildFromFile();
+    }
+
+    // This node shouldn't do anything if its been disabled from the gui!
+    if (_enabled) {
```
emiax: Consider early return `if (!_enabled)`

modules/fieldlinessequence/rendering/renderablefieldlinessequence.h
```diff
@@ -0,0 +1,175 @@
...
+    bool   extractMandatoryInfoFromDictionary(SourceFileType& sourceFileType);
+    void   extractOptionalInfoFromDictionary(std::string& outputFolderPath);
+    void   extractOsflsInfoFromDictionary();
+    void   extractTriggerTimesFromFileNames();
+    bool   loadJsonStatesIntoRAM(const std::string& OUTPUT_FOLDER);
+    void   loadOsflsStatesIntoRAM(const std::string& OUTPUT_FOLDER);
+    void   setModelDependentConstants();
+    void   setupProperties();
+    bool   prepareForOsflsStreaming();
+#ifdef OPENSPACE_MODULE_KAMELEON_ENABLED
```
emiax: Can these methods be moved out to a separate class? To make this class completely independent from Kameleon?

modules/fieldlinessequence/rendering/renderablefieldlinessequencesetup.cpp
```diff
@@ -0,0 +1,874 @@
...
+
+#include <fstream>
+#include <sstream>
+
+namespace {
+    std::string _loggerCat = "RenderableFieldlinesSequence";
+
+    // ----- KEYS POSSIBLE IN MODFILE. EXPECTED DATA TYPE OF VALUE IN [BRACKETS]  ----- //
+    // ---------------------------- MANDATORY MODFILE KEYS ---------------------------- //
+    const char* KEY_INPUT_FILE_TYPE         = "InputFileType";   // [STRING] "cdf", "json" or "osfls"
```
emiax: Typically we use CamelCaseWithCapitalFirstLetter for these types of constants

modules/fieldlinessequence/rendering/renderablefieldlinessequencesetup.cpp
```diff
@@ -0,0 +1,874 @@
...
+    };
+    static const openspace::properties::Property::PropertyInfo OriginButtonInfo = {
+        "focusCameraOnParent", "Focus Camera", "Focus camera on parent."
+    };
+    static const openspace::properties::Property::PropertyInfo TimeJumpButtonInfo = {
+        "timeJumpToStart", "Jump to Start Of Sequence", "Performs a time jump to the start of the sequence."
+    };
+
+    enum class SourceFileType : int {
+        CDF = 0,
```
emiax: CamelCaseWithCapitalFirstLetter

modules/fieldlinessequence/rendering/renderablefieldlinessequencesetup.cpp
```diff
@@ -0,0 +1,874 @@
...
+        _pMaskingMax = std::to_string(_maskingRanges[0].y);
+    }
+}
+
+void RenderableFieldlinesSequence::definePropertyCallbackFunctions() {
+    // Add Property Callback Functions
+    bool hasExtras = _states[0].nExtraQuantities() > 0;
+    if (hasExtras) {
+        _pColorQuantity.onChange([this] {
+            LDEBUG("CHANGED COLORING QUANTITY");
```
emiax: Are these debug messages necessary?

modules/fieldlinessequence/shaders/fieldlinessequence_fs.glsl
```diff
@@ -0,0 +1,48 @@
...
+
+#include "fragment.glsl"
+
+in vec4 vs_color;
+in float vs_depth;
+
+uniform bool usingAdditiveBlending;
+
+Fragment getFragment() {
+    if (vs_color.a == 0) {
```
emiax: Safer to check if smaller than some threshold

modules/fieldlinessequence/util/fieldlinesstate.h
```diff
@@ -0,0 +1,100 @@
...
+    class Kameleon;
+}
+#endif // OPENSPACE_MODULE_KAMELEON_ENABLED
+
+namespace openspace {
+
+class FieldlinesState {
+public:
+
+#ifdef OPENSPACE_MODULE_KAMELEON_ENABLED
```
emiax: Can parts with Kameleon dependency be moved out to separate class?

emiax:
Great job! Is this everything fieldline related from the thesis work?
Looks like there is no line morphing included in this - I completely understand if that would need more work in order to be stable + useful.
I wrote a few minor thoughts/questions as code review comments. Overall I think the code looks neat! I'll checkout the branch and play around with it as soon as I get time :)

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
Carlbaum: Yeah, it looks that way, but I don't think it will. 
The reason for that `while` loop is that I'm sometimes loading states on a different thread; if so the `_isLoadingStatesFromDisk` is set to `true` and as soon as the thread is done it is set back to `false`. In this other thread it makes use of some member variables, which is  why I must let that thread finish before the destrutor can continue.
Is there another way of achieving this functionality?

BTW: It is the function `RenderableFieldlinesSequence::readNewState(const std::string& FILEPATH)` that is being executed if `_isLoadingStatesFromDisk == true`

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
Carlbaum: Aaah, alright! Will fix it!

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
Carlbaum: Definitely!

modules/fieldlinessequence/rendering/renderablefieldlinessequencesetup.cpp
Carlbaum: Nopes ;)

modules/fieldlinessequence/util/fieldlinesstate.h
Carlbaum: Yeah, it shouldn't be too much of a hassle. Will fix it!

Carlbaum:
No, you're right. I removed the line morphing, the GPU/live traced fieldlines and the 3D lines using the billboard technique (like HC used) as I didn't think it looked well enough to be in the master branch.

I could probably add the functionality for morphing back in if you think it will be used. However, I think that preparing the lines for morphing should be done outside of OpenSpace. I mean, I don't really think that CDFs should be used as input. It would be better that all of that is prepared outside.

modules/fieldlinessequence/rendering/renderablefieldlinessequencesetup.cpp
Carlbaum: I'm on it!

modules/fieldlinessequence/rendering/renderablefieldlinessequencesetup.cpp
Carlbaum: Thumbs up!

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
Carlbaum: Oh, btw: I have never seen that message being triggered, so it's something that is VERY unlikely to happen!

alexanderbock: I assume the `_isLoadingStateFromDisk` is `std::atomic_bool`  if so, there is no other good way of doing that sans using a full mutex, which I think would be overkill here

Carlbaum: Alright! I'll leave it be then.

Carlbaum:
Oh, btw: the part of Kameleon that is used when tracing field lines from ENLIL cdfs is slightly broken; there are two small fixes that has to be done.
Errors:
1. There's a variable that's being used while uninitialized. 
2. The `binary_search` function they've created can go out of vector bounds if the value isn't found.

Those issues are easily fixed:
1.  Comment out or remove the entire `if`-statement where error occurs as it does nothing. (Rows 1294-1301 of `Kameleon/src/ccmc/Tracer.cpp`)
2. E.g. ensure that `middle + 1` isn't out of the vector's bounds at row 51 of `Kameleon/src/ccmc/Utils.h`

I wasn't sure if I should change that, as it is external code.

PR #532, opened 2018-02-23
Added support to G-Buffer.

jccosta:
Added missing support for G-Buffer from RenderableModel.


jccosta:
@emiax No, the normal must be in object space. All pre-calculation is done in object space and when the normal is neeed, it must be in object space.

PR #533, opened 2018-02-24
Add implementation of pure URL synchronization from foreign sources

### LINKED ISSUE #530: Implement a third download method for resources that takes a URL path and downloads it

alexanderbock:
UrlSynchronization class that allows direct download from non Big Bang servers.  Either a single URL or multiple URLs 

Closes #530


emiax:
Looks good!
I’m thinking: would it be good to base the sync directory on the url(s) to avoid directory name collisions, if two users use the same identifier? (This is more likely here since there is no centralized server side collision)
For example, hash the url(s)?

emiax:
Second comment: can the commented out blocks be removed?

PR #540, opened 2018-03-03
Issue/525

alexanderbock:
Introduce ProgramObjectManager static variable to some OpenSpaceModule classes to handle reusing ProgramObjects between multiple instances of the same Renderable, drastically reducing startup time due to reducing shader compilation times (about 200+ shaders less compiled on default_full,  saving about 10-15 seconds startup time on default).

ProgramObjects are indexed by unique name which is a anonymous namespace variable for each Renderable; ProgramObjectManager keeps a reference count of the program objects and deletes them when the last Renderable loses interest in them.

Considerations:
1. Using static ghoul::opengl::Program* wasn't feasible as it would require a second variable to keep the reference count and make the code messy
2. std::shared_ptr wasn't an issue as code was getting messy to make sure that the ProgramObject would be deleted at the correct time when there was still an OpenGL context available

### DIFF
--- a/include/openspace/engine/moduleengine.h
+++ b/include/openspace/engine/moduleengine.h
@@ -59,10 +59,20 @@ public:
     void initialize(const ghoul::Dictionary& moduleConfigurations);
 
     /**
+     * Calls the initializeGL functions of all registered OpenSpaceModule%s.
+     */
+    void initializeGL();
+
+    /**
      * Deinitializes all of the contained OpenSpaceModule%s by calling the
      * OpenSpaceModule::deinitialize methods.
      */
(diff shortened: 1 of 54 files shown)


alexanderbock:
TODO(jccosta);   Could you apply the same pattern to the atmosphere shaders? All of them are compiled multiple times right now as well and I have the feeling that we'll only need each of them once?

TODO(alexanderbock):  Investigate how to apply this to localchunk and globalchunk. Option 1: Encode the Dictionary state into the name and make use of the resuse mechanism that way.  Option 2: Convert the shaders into using uniforms instead of compile time constants and only compile shaders once

PR #549, opened 2018-03-12
Issue/24

alexanderbock:
Adding property interpolation to `setPropertyValue`


src/scene/scene_lua.inl
```diff
@@ -105,13 +96,22 @@ void applyRegularExpression(lua_State* L, std::regex regex,
                     )
                 );
             } else {
-               executePropertySet(prop, L);
+                if (interpolationDuration == 0.0) {
+                    prop->setLuaValue(L);
```
emiax: This should probably also clear the ongoing interpolation (if any) ?

include/openspace/scene/scene.h
```diff
@@ -203,6 +207,17 @@ class Scene
     std::mutex _programUpdateLock;
     std::set<ghoul::opengl::ProgramObject*> _programsToUpdate;
     std::vector<std::unique_ptr<ghoul::opengl::ProgramObject>> _programs;
+
+    struct InterpolationInfo {
+        // @TODO(abock): The property might have been deleted underneath us, which would
```
emiax: Good point: I think this needs to be fixed before it can be merged.
would it be good/feasable to change the interface of property owners to store shared pointers instead of raw ones? We would need to change pretty much all concrete property owners...

include/openspace/properties/numericalproperty.inl
```diff
@@ -395,4 +395,57 @@ std::string NumericalProperty<T>::generateAdditionalDescription() const {
...
+void NumericalProperty<T>::setLuaInterpolationTarget(lua_State* state) {
+    bool success = false;
+    T thisValue = PropertyDelegate<NumericalProperty<T>>::template fromLuaValue<T>(
+        state,
+        success
+    );
+    if (success) {
+        _interpolationStart = TemplateProperty<T>::_value;
+        _interpolationEnd = std::move(thisValue);
+    }
```
emiax: Throw some type of error if !success

src/scene/scene.cpp
```diff
@@ -453,41 +508,35 @@ scripting::LuaLibrary Scene::luaLibrary() {
...
-                "through all property owners to find those that are tagged with this "
-                "group name, and set their property values accordingly."
+                "third is not present or is '0', the value changes instantly, otherwise "
+                "the change will take that many seconds and the value is interpolated at "
+                "each steap in between. The fourth argument must be either empty, "
+                "'regex', or 'single'. If the last argument is empty (the default), the "
+                "URI is interpreted using a wildcard in which '*' is expanded to '(.*)' "
+                "and bracketed components '{ }' are interpreted as group tag names. "
+                "Then, the passed value will be set on all properties that fit the regex "
+                "+ group name combination. If the third argument is 'regex' neither the "
```
emiax: fourth?

src/scene/scene.cpp
```diff
@@ -453,41 +508,35 @@ scripting::LuaLibrary Scene::luaLibrary() {
                 "setPropertyValue",
                 &luascriptfunctions::property_setValue,
                 {},
-                "string, *",
+                "name, value [, duration, optimization]",
```
emiax: "optimization" could be renamed to something that indicates that it has to do with regex

emiax: Would be cool to allow for different easing options: Linear, EaseIn, EaseOut, EaseInOut.

src/scene/scene_lua.inl
```diff
@@ -153,78 +154,22 @@ int setPropertyCall_single(properties::Property* prop, std::string uri, lua_Stat
...
-                if (pathRemainderToMatch.compare(thisPropMatchId) == 0)  {
-                    properties::PropertyOwner* matchingTaggedOwner
-                        = findPropertyOwnerWithMatchingGroupTag(prop, tagToMatch);
-                    if (matchingTaggedOwner) {
-                        setPropertyCall_single(prop, uri, L, type);
-                    }
-                }
-            }
+        if (duration == 0.0) {
+            prop.setLuaValue(L);
```
emiax: Same here: clearing interpolation if duration is set to 0.

src/scene/scene.cpp
```diff
@@ -440,6 +441,60 @@ SceneGraphNode* Scene::loadNode(const ghoul::Dictionary& dict) {
     return rawNodePointer;
 }
 
+void Scene::addInterpolation(properties::Property* prop, float durationSeconds) {
```
emiax: A change for the future: I think properties shouldn't really live in the scene. (Since there are non-scene properties.) I think it would make more sense to have a `PropertyManager` or similar. With the current structure this makes sense though.

include/openspace/properties/numericalproperty.inl
alexanderbock: No need; the "fromLuaValue" already throws an error and since the argument is the same as for the uninterpolated `setPropertyValue`, the error message still makes sense

alexanderbock:
Closes #24

PR #567, opened 2018-03-21
Change Name to Identifier in globe browsing helper code

emiax:
@alexanderbock: I had some issues with .info files after the Name -> Identifier change.
Is this the proper way to solve it? This would also require changing the info files themselves. Not sure if there is a central repo for those, where that should be changed in that case?

(I tried this locally and it works) :)

### DIFF
--- a/modules/globebrowsing/scripts/layer_support.lua
+++ b/modules/globebrowsing/scripts/layer_support.lua
@@ -166,7 +166,7 @@ openspace.globebrowsing.parseInfoFile = function (file)
     local color = nil
     if ColorFile then
         color = {
-            Name = Name,
+            Identifier = Identifier,
             Description = Description or "",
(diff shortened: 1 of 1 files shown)


alexanderbock:
No, we don't have a central repo for those yet since they are distributed together with the GeoTIFFs.  How about leaving the info files as they are but doing:
```
Name = Namem
Identifier = nameToIdentifier(Name)
```

with `nameToIdentifier` exchanging spaces and dots to `_`?

emiax:
That would work, but if we are in control of the GeoTIFFS, I think it might be better to add an `Identidier` to those files. Older versions of OpenSpace will continue to use the `Name`.
Who distributes the GeoTIFFs? Are the info files following some standard?

alexanderbock:
Right now, we are distributing them;  and we define the info files completely as well

PR #570, opened 2018-03-23
Fix crash caused by setting properties using regex

emiax:
The Lua stack was thrown away after setting the first matching property, causing the system to crash when trying to set the other matching properties.
Fixed by introducing a `leaveOnStack` parameter to `setLuaValue` for properties.


alexanderbock:
How about not introducing that parameter, but leaving it up to the caller of setLuaValue and getLuaValue to clear the stack after the call?

emiax:
Yeah, that’s probably better! Let’s not merge this.

PR #596, opened 2018-04-17
Issue/87

alexanderbock:
Time interpolation instead of fixed jump

@emiax   Could you give this a spin and give feedback?  The time interpolation doesn't end at the correct time, and the difference seems to depend both on the total time travelled as well as the framerate.

To test:
```lua
openspace.time.setDeltaTime(0);
openspace.time.setTime("2018 JUL 01");
openspace.time.setTime("2018 JUL 02", 5);
```

### DIFF
--- a/include/openspace/scene/scene.h
+++ b/include/openspace/scene/scene.h
@@ -202,7 +202,7 @@ public:
      */
     void removePropertyInterpolation(properties::Property* prop);
 
-    void addTimeInterpolation(float targetTime, double durationSeconds);
+    void addTimeInterpolation(double targetTime, double durationSeconds);
     void removeTimeInterpolation();
 
     /**
--- a/include/openspace/scene/scene.h
+++ b/include/openspace/scene/scene.h
@@ -188,7 +188,7 @@ public:
      * \pre \p durationSeconds must be positive and not 0
      * \post A new interpolation record exists for \p that is not expired 
      */
-    void addInterpolation(properties::Property* prop, float durationSeconds,
+    void addPropertyInterpolation(properties::Property* prop, float durationSeconds,
         ghoul::EasingFunction easingFunction = ghoul::EasingFunction::Linear);
 
     /**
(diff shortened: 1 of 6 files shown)


emiax:
I started on a keyframe-based solution to this problem: https://github.com/OpenSpace/OpenSpace/compare/master...feature/time-interpolation

I think it may be easier to get this to work with the parallel connection.
Things that I tried interpolating so far:
- Pause/play (slowing down, speeding up)
- changing delta time  (slowing down, speeding up)
- Interpolating time (same as `openspace.time.setTime("2018 JUL 01", 5);` above) 

Some work remains:
- Moving `addTimeInterpolation` to `TimeManager` since all other time methods live there.
- Getting it to work with `ParallelConnection`
- Adjusting `spacecraftinstrument` module to `Time` and `TimeManager` interface changes
- Possibly adding more easing options
- Make it possible to specify interpolation duration in lua, when setting pause and delta time.

One issue to think about: What should happen if an interpolation is started while one is already running?

emiax:
This one (https://github.com/OpenSpace/OpenSpace/compare/master...feature/time-interpolation) is getting close to ready. @alexanderbock, if you want to check this out and see what you think, that would be great :)

```
openspace.time.setDeltaTime(0);
openspace.time.setTime("2018 JUL 01"); -- no interpolation
openspace.time.interpolateTime("2018 JUL 02"); -- default interpolation speed based on global property
openspace.time.interpolateTime("2018 JUL 02", 5); -- custom interpolation speed
openspace.time.interpolateDeltaTime();  -- default interpolation speed
openspace.time.interpolateDeltaTime(60);  -- custom interpolation speed
openspace.time.interpolateTogglePause(); -- default interpolation speed
openspace.time.interpolateTogglePause(2); -- custom interpolation speed
```

...etc

The GUI and keybindings bind to the interpolation methods, so smooth interpolation is the default option.

PR #671, opened 2018-07-12
Fix issue #660 and provide percentage for http syncs

emiax:


modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -219,33 +227,57 @@ bool HttpSynchronization::trySyncFromUrl(std::string listUrl) {
             ghoul::filesystem::FileSystem::PathSeparator +
             filename + TempSuffix;
 
+        if (sizeData.find(line) != sizeData.end()) {
+            LWARNING(
```
alexanderbock: For the `fmt::format` strings, I moved away from our usual styling to:
```
LWARNING(fmt::format(
    "{}: Duplicate entries: {}", _identifier, line
));
```

modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -219,33 +227,57 @@ bool HttpSynchronization::trySyncFromUrl(std::string listUrl) {
...
+                fmt::format("{}: Duplicate entries: {}", _identifier, line)
+            );
+            continue;
+        }
+
         downloads.push_back(std::make_unique<AsyncHttpFileDownload>(
             line, fileDestination, HttpFileDownload::Overwrite::Yes));
 
         auto& fileDownload = downloads.back();
```
alexanderbock: Unnecessary `auto`;  `back` returns the type, not an iterator

modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -219,33 +227,57 @@ bool HttpSynchronization::trySyncFromUrl(std::string listUrl) {
...
             filename + TempSuffix;
 
+        if (sizeData.find(line) != sizeData.end()) {
+            LWARNING(
+                fmt::format("{}: Duplicate entries: {}", _identifier, line)
+            );
+            continue;
+        }
+
         downloads.push_back(std::make_unique<AsyncHttpFileDownload>(
```
alexanderbock: [i know it's not your code, but since you are already in there]

```
downloads.push_back(std::make_unique<AsyncHttpFileDownload>(
    line,
    fileDestination,
    HttpFileDownload::Overwrite::Yes
));
```

modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -219,33 +227,57 @@ bool HttpSynchronization::trySyncFromUrl(std::string listUrl) {
...
+                p.totalBytesKnown,
+                p.totalBytes,
+                p.downloadedBytes,
+            };
+
+            SizeData size = std::accumulate(
+                sizeData.begin(),
+                sizeData.end(),
+                SizeData{ true, 0, 0 },
+                [](const SizeData& a, const auto& b) -> SizeData {
```
alexanderbock: The auto is not necessary as it is the type of `sizeData`, which is `std::pair<const std::string, SizeData>`

modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -270,16 +302,14 @@ bool HttpSynchronization::trySyncFromUrl(std::string listUrl) {
             FileSys.deleteFile(originalName);
             int success = rename(tempName.c_str(), originalName.c_str());
             if (success != 0) {
-                LERRORC(
-                    "HTTPSynchronization",
+                LERROR(
```
alexanderbock: Same as above

PR #690, opened 2018-08-21
Feature/globals

alexanderbock:


alexanderbock:
Fixes in the branch
 - Move global objects out of OpenSpaceEngine
 - General cleanup of main.cpp
 - Add default_joystick asset to all scenes
 - No longer suppress mouse interaction on slave nodes
 - Window delegate uses function pointers rather than subclassing
 - Fix for false overwriting of ImGUI configuration file
 - Change default color and tilt angle of fisheye rendering
 - Restructured performance manager
 - Simplify screenshot handling

alexanderbock:
@emiax  Could you give it a run in the dome and see if it is working?  If it will fail, it will fail at startup and camera + script sychronization and should be obvious immediately

PR #761, opened 2018-11-08
Use bin folder instead of cwd to find configuration file

emiax:
@micahnyc pointed out that when double clicking an application on mac, the current working directory is set to "/". One way to get around this problem would be to set cwd to the application path, using NSGetExecutablePath. However, instead, I suggest that we don't change the current working directory, and use the newly introduced ${BIN} to find the configuration file.

Let me know what your thoughts are. I have only tested this on mac yet.


emiax:
Micah tested successfully on windows. Will go ahead and merge.

Issue #709, opened 2018-09-04
Add ability to modify zooming speed using keybinds


alexanderbock:
For example:
Zooming + Shift key:  1/10th of the zoom speed
Zooming + Ctrl key : 1/100th speed

mayby apply to orbital motion, too?

This would provide the user with the ability to fine-tune navigation, when needed

emiax:
I think this is a good idea, but Shift and Ctrl are already used to control roll and pan, so I think we need to come up with some other modifier/shortcut to change the zoom/orbital speed. There is a `MouseSensitivity` property that perhaps could be divided into several properties - orbital, zoom, pan, roll etc. If we do this, the user can bind any key to modify these parameters in runtime.

PR #792, opened 2019-02-11
Add support for zoom sensitivity adjustment with Ctrl/Shift

mingenuity:
* Holding Ctrl while zooming with the Right mouse button or Alt-Left mouse will now increase the sensitivity, causing the zoom speed (in or out) to increase by 8x
* Likewise, holding Shift while zooming will decrease the sensitivity by 2x (1/2 of the speed)
* These values are different as zooming is already relatively insensitive, and it seems likely (?) that the user won't want to make it much less insensitive when holding down Shift
* These are hardcoded values but ideally should later be made configurable, e.g. via the Orbital Navigator sliders
* This is related to issue #709, although it allows for both speeding up and down rather than just slowing down
* Thanks to Micah for his support during the hackathon!

### DIFF
--- a/src/interaction/mousecamerastates.cpp
+++ b/src/interaction/mousecamerastates.cpp
@@ -26,6 +26,11 @@
 
 #include <openspace/interaction/inputstate.h>
 
+namespace {
+    const double SENSITIVITY_ADJUSTMENT_INCREASE = 8.0;
+    const double SENSITIVITY_ADJUSTMENT_DECREASE = 0.5;
+}
+
 namespace openspace::interaction {
 
 MouseCameraStates::MouseCameraStates(double sensitivity, double velocityScaleFactor)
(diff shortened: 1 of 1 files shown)


emiax:
Thank you for the contribution @mingenuity,
I think this is a great idea, that would be really useful especially when zooming into really small spacecraft from far away. However binding this to the Shift and Ctrl keys will interfere with the roll/pan behavior. Right now, it is possible to zoom towards an object while pressing the shift key to adjust the roll, or use Ctrl to pan. I cannot think of any better way to rebind the keys right now, so I will not merge this right now. I'll keep this in idea in mind for future updates to our camera navigation. Thanks again, and if you have more ideas in this area, let us know!

micahnyc:
I think we can move it to z and x instead of shift and ctrl and then we should be able to use it.

alexanderbock:
Great work!  Thanks a lot!!

PR #797, opened 2019-02-15
Fixed unforms problems when eclipses are activated.

jccosta:
Fixed uniforms not being used when eclipses are activated.


modules/globebrowsing/src/renderableglobe.h
```diff
@@ -68,6 +68,11 @@ struct Chunk {
     std::array<Chunk*, 4> children = { { nullptr, nullptr, nullptr, nullptr } };
 };
 
+enum ShadowCompType {
```
alexanderbock: Should be an `enum class`

modules/globebrowsing/src/renderableglobe.h
```diff
@@ -93,7 +98,7 @@ class RenderableGlobe : public Renderable {
     LayerManager& layerManager();
     const glm::dmat4& modelTransform() const;
 
-private:
+private:
```
alexanderbock: Extra empty spaces

PR #799, opened 2019-02-18
Feature/anchor and aim

emiax:
Introduce the ability to navigate using an anchor and aim, in addition to the previous concept of a focus node. When this mode is active, the camera will stay fixed on a sphere around the anchor node, while moving on the sphere to keep both anchor (foreground) and aim (background) nodes positions in screen space.

Example usage:
- Set spacecraft as anchor and planet as aim to always look down at a planet, when orbiting.
- Set spacecraft as anchor and planet as aim to visualize a flyby without having to reposition the planet in screen space.
- Set moon as anchor and planet as aim or vice versa, to observe the orbit without having to manually compensate for the movement.

Also add ability to reassign a scene graph node without invoking the retargeting interpolation.

This depends on an update to the webgui as well, so once we have a thumbs up on merging both pull requests, I'll push to the webgui and also update the webgui hash.
Here is the pull request to the webgui frontend: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/2


alexanderbock:
Doesn't compile right now when the `touch`module is enabled

PR #808, opened 2019-03-06
Changed touch module to allow build with/without webbrowser module enabled

GPayne:
Tested touch interaction with and without the webbrowser module.


emiax:
Nice!

PR #827, opened 2019-04-05
Cleanup for Spheres and PSC

emiax:
- Improve the way fading is handled in DU objects.
- Restrict static rotation interface to only take euler rotaitons
  (no matrices anymore, since they are not guaranteed to be solely rotations.)
- Remove RenderablePlanet. Use RenderableGlobe instead.
- Remove power scale coordinates from RenderableSphere, including its shader.
- Remove PowerScaledScalar altogether.
- Remove PSC related code from Renderable.
- Fix bug with rendering spheres with translation
- Create a basic scene with only spheres as an example
- Make it possible to load NH scene together with the default scene
- Use ISO standard spherical coordinate system in spheres, so to remove the need for "magic" rotations in client code.

Remaining work: Remove all power scaled coordinates, from model projection, renderableconstellationbounds, stars.


alexanderbock:
:+1:
I would prefer to keep the option to have matrices in the staticrotation as we might want to compute non euler angles as the rotation when creating the `StaticRotation` from Lua.

alexanderbock:
2>c:\development\openspace\modules\galaxy\rendering\galaxyraycaster.cpp(62): error C2039: 'setPscUniforms': is not a member of 'openspace::Renderable'

Issue #831, opened 2019-04-24
Textfields provide no feedback or cursor in CEF


emiax:
To reproduce: Click the search field in the Scene sidebar. No cursor is visible.
This seems to be related to the recent CEF version upgrade.

emiax:
fixed with #833

PR #833, opened 2019-04-25
Feature/upgrade cef

emiax:


emiax:
I skipped upgrading on MacOS. This upgrade fixes https://github.com/OpenSpace/OpenSpace/issues/831 on Windows. The issue is not present on Mac.

PR #840, opened 2019-05-03
Latest changes to session recording with support for new webGUI record/playback controls

GPayne:
These changes were made to work with the session recording GUI controls, which are currently in the 'sessionRecording' branch of the OpenSpace-WebGuiFrontend repo.


GPayne:
Currently doing build & test after resolving merge conflicts.

micahnyc:
@GPayne I see commit b21f4b9 - "Fixed problem with occasional large jump in camera pos/rotation after playback finishes." 

However it still is jumping for me at least 50% of the time.

GPayne:
@micahnyc Thanks for trying this out. I am still seeing a small jump at the end (which I am still looking into), but commit b21f4b9 prevented very large jumps in position and rotation that would sometimes happen.

Referring to the jumps you are still seeing 50% of the time: are these large or small? (I define a "small" jump distance as the focus node moving no more than 10% of the width of the window)

emiax:
I tried it out together with @micahnyc and there were jumps from one side of the planet to another. This seems to be correlated with OpenSpace having to jump in time when the simulation is started. To reproduce: 1. Record a session when zooming in on Earth. Stop recording. 2. Offset time with a few hours. 3. Play back recording. 4. When playback is done, the jump appears.

GPayne:
OK @micahnyc this latest 40a1bb3 change completely fixed it for me. I no longer see any visible change in position/rotation when playback finishes. Please let me know if your experience is different.

GPayne:
@emiax I just tried your steps with 40a1bb3 and I cannot reproduce the problem now.

emiax:
Awesome! Just tried it out on my mac and fixed a minor issue with paths. (For some reason the absPath does not support trailing slashes properly on mac).
I have only browsed through the code briefly yet, and I think it overall looks good. To promote separation of concern, I tried to make the OrbitalNavigator less aware of the session recording, and introduce a more general `resetNodeMovements` method. This avoids additional complexity in the main navigation logic, which is already pretty complex. Let me know if you think this may cause any issues.

With this change, do you think the `_firstFrameAfterPlaybackFinished` can be removed from NavigationHandler, and call the reset methods immediately when playback stops, in the `NavigationHandler::stopPlayback` method? To my experience it's usually a good idea to minimize the amount of (redundant) state and special cases based on that state.

Micah and I will take a look at the GUI asap as well. But this change could be merged first, independently of the GUI update, am I right?

GPayne:
@emiax the new method for resetting the camera looks good. I don't see a problem with making the change you're talking about in NavigationHandler, will do that tomorrow.
Yes, this change is independent of the GUI, since it doesn't affect the interface between them.

modules/server/include/topics/sessionrecordingtopic.h
```diff
@@ -0,0 +1,55 @@
...
+
+private:
+    const int UnsetOnChangeHandle = -1;
+
+    //Provides the idle/recording/playback state int value in json message
+    nlohmann::json state();
+
+    int _stateCallbackHandle = UnsetOnChangeHandle;
+    bool _isDone = false;
+    std::chrono::system_clock::time_point _lastUpdateTime;
```
emiax: This seems to be unused

modules/server/src/topics/getpropertytopic.cpp
```diff
@@ -70,6 +72,11 @@ void GetPropertyTopic::handleJson(const nlohmann::json& json) {
     else if (requestedKey == RootPropertyOwner) {
         response = wrappedPayload(global::rootPropertyOwner);
     }
+    else if (requestedKey == SessionRecordingPlaybackList) {
```
emiax: I think we should strive to minimize the number of these special case keys in `getpropertytopic`. Since this one has nothing to do with the property tree, how about putting this into the session recording topic for now?
In the future we can replace this with a javascript call into the lua api (luascripttopic) and get return values from there.

modules/server/src/topics/sessionrecordingtopic.cpp
```diff
@@ -0,0 +1,96 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include "modules/server/include/topics/sessionrecordingtopic.h"
+
+#include <modules/server/include/connection.h>
+#include <openspace/engine/globals.h>
+#include <openspace/properties/property.h>
```
emiax: Is this used?

src/interaction/keyframenavigator.cpp
```diff
@@ -134,12 +140,11 @@ bool KeyframeNavigator::updateCamera(Camera* camera, const CameraPose prevPose,
 
     // Linear interpolation
     t = std::max(0.0, std::min(1.0, t));
-    camera->setPositionVec3(
-        prevKeyframeCameraPosition * (1 - t) + nextKeyframeCameraPosition * t
-    );
-    camera->setRotation(
-        glm::slerp(prevKeyframeCameraRotation, nextKeyframeCameraRotation, t)
-    );
+    glm::dvec3 nowCameraPosition = prevKeyframeCameraPosition * (1 - t) + nextKeyframeCameraPosition * t;
```
emiax: This does not fit in a line of 90 characters, so better to keep old formatting

src/interaction/sessionrecording.cpp
```diff
@@ -586,6 +599,17 @@ void SessionRecording::preSynchronization() {
     else if (_cleanupNeeded) {
         cleanUpPlayback();
     }
+
+    //Handle callback(s) for change in idle/record/playback state
+    if (_state != _lastState)
+    {
```
emiax: Keep { on same line

src/interaction/sessionrecording.cpp
```diff
@@ -1255,6 +1283,44 @@ void SessionRecording::saveKeyframeToFile(std::string entry) {
...
+    return handle;
+}
+
+void SessionRecording::removeStateChangeCallback(CallbackHandle handle) {
+    const auto it = std::find_if(
+        _stateChangeCallbacks.begin(),
+        _stateChangeCallbacks.end(),
+        [handle](const std::pair<CallbackHandle, std::function<void()>>& cb) {
+        return cb.first == handle;
+    }
```
emiax: Indentation

src/interaction/sessionrecording.cpp
```diff
@@ -1255,6 +1283,44 @@ void SessionRecording::saveKeyframeToFile(std::string entry) {
...
+        it != _stateChangeCallbacks.end(),
+        "handle must be a valid callback handle"
+    );
+
+    _stateChangeCallbacks.erase(it);
+}
+
+std::string SessionRecording::playbackList() {
+    std::string fileList;
+    std::string recordingsPath = absPath("${RECORDINGS}");
```
emiax: can be a const std::string

src/interaction/navigationhandler.cpp
```diff
@@ -124,15 +124,25 @@ void NavigationHandler::updateCamera(double deltaTime) {
...
     }
-    else if ( ! _playbackModeEnabled ) {
-        if (_camera) {
-            if (_useKeyFrameInteraction) {
-                _keyframeNavigator->updateCamera(*_camera, _playbackModeEnabled);
-            }
-            else {
-                _orbitalNavigator->updateStatesFromInput(*_inputState, deltaTime);
-                _orbitalNavigator->updateCameraStateFromStates(deltaTime);
+    else {
```
emiax: From discussion yesterday: let's see if we can simplify this logic by calling 
```
_orbitalNavigator->resetVelocities();
_orbitalNavigator->resetNodeMovements();
```
directly when the playback is stopped.

emiax:
Thanks for the feedback on that change. I have looked through the rest of the code now and made some comments - other than those you have my :+1: for merge

modules/server/src/topics/getpropertytopic.cpp
GPayne: I remember not wanting to put it here vs. session recording topic, but it seemed that I had to do it this way in order for it to work in the web GUI.
In any case, I don't want to risk breaking compatibility with the GUI right now, since the session recording part of it will be changed by @emiax  and/or @micahnyc.

PR #870, opened 2019-05-18
Bring back stars to the Earth's night.

jccosta:
Fixed wrong behavior in the Atmospheres at night.


alexanderbock:
There is a lot of red-based attenuation when on the nightside looking up..    That doesn't seem to be too realistic, no?

PR #876, opened 2019-05-20
Feature/check latest version

emiax:


src/rendering/renderengine.cpp
```diff
@@ -1223,14 +1223,25 @@ void RenderEngine::renderVersionInformation() {
...
+            global::versionChecker.latestVersion();
+
+        VersionChecker::SemanticVersion currentVersion {
             OPENSPACE_VERSION_MAJOR,
             OPENSPACE_VERSION_MINOR,
             OPENSPACE_VERSION_PATCH
         };
         if (currentVersion < latestVersion) {
-            versionString += " [" + latestVersion.format() + " is available]";
+            LINFO(fmt::format(
```
emiax: This will happen each frame

emiax: Will push a fix

PR #905, opened 2019-06-17
Feature/transparency rendering last

emiax:


emiax:
Will not merge this - instead I will sync with @jccosta how to proceed with this. Will probably pick some ideas from here and the most part from feature/hdrOS

PR #931, opened 2019-07-11
Feature/timeline translation

emiax:
Introduce TimelineTranslation that brings basic keyframing cababilities to Translation objects.
This makes it possible to interpolate between different translations depending on the simulation time, for example switch between spice data and static data, or specify movement along a path by interpolating between a series of static translations.

First use case of this is data from the Apollo 11 landing, converted from KML.


modules/base/rotation/timelinerotation.cpp
```diff
@@ -0,0 +1,114 @@
...
+        next = prev;
+    }
+    const double prevTime = prev->timestamp;
+    const double nextTime = next->timestamp;
+
+    double t = 0.0;
+    if (nextTime - prevTime > 0.0) {
+        t = (now - prevTime) / (nextTime - prevTime);
+    }
+    return t * next->data->matrix(data) + (1.0 - t) * prev->data->matrix(data);
```
emiax: I think here we should use a slerp instead of a lerp. Just interpolating the individual components of a matrix can give wierd results, which will probably be visible if the rotation between two keyframes is big enough.

modules/base/rotation/timelinerotation.h
```diff
@@ -0,0 +1,49 @@
...
+****************************************************************************************/
+
+#ifndef __OPENSPACE_MODULE_BASE___TIMELINEROTATION___H__
+#define __OPENSPACE_MODULE_BASE___TIMELINEROTATION___H__
+
+#include <openspace/scene/rotation.h>
+#include <openspace/util/timeline.h>
+
+namespace openspace {
+
```
emiax: Here be tabs. Also, the namespace should not be indented.

alexanderbock:
Merged the navigationhandler branch and now there are conflicts

Issue #934, opened 2019-07-15
Kiosk version of OpenSpace sometimes crashes from a heap corruption when switching stories


GPayne:
Switching stories often enough will cause a heap corruption at the resize call in `SyncBuffer::data()`. If the size of `vector<char> _dataStream` is initialized to a much larger value (defaults to 4096), this error will not occur.
The kiosk version exists in the **feature/gui-for-touch** branch.

GPayne:
Changes in pull request #937 fixed this.

PR #937, opened 2019-07-17
Added a check to prevent buffer sync buffer overflow, and expand buff…

GPayne:
…er size if needed.

This addresses issue #934.

### DIFF
--- a/src/util/syncbuffer.cpp
+++ b/src/util/syncbuffer.cpp
@@ -37,7 +37,13 @@ SyncBuffer::~SyncBuffer() {} // NOLINT
 void SyncBuffer::encode(const std::string& s) {
     ghoul_assert(_encodeOffset + sizeof(char) * s.size() + sizeof(int32_t) < _n, "");
 
-    int32_t length = static_cast<int32_t>(s.length());
+    int32_t anticpatedBufferSize = _encodeOffset + (sizeof(char) * s.size())
+        + sizeof(int32_t);
+    if (anticpatedBufferSize >= _n) {
+        _dataStream.resize(anticpatedBufferSize);
+    }
+
+    int32_t length = static_cast<int32_t>(s.size() * sizeof(char));
     memcpy(
         _dataStream.data() + _encodeOffset,
         reinterpret_cast<const char*>(&length),
--- a/src/util/syncbuffer.cpp
+++ b/src/util/syncbuffer.cpp
@@ -35,8 +35,6 @@ SyncBuffer::SyncBuffer(size_t n)
 SyncBuffer::~SyncBuffer() {} // NOLINT
 
 void SyncBuffer::encode(const std::string& s) {
-    ghoul_assert(_encodeOffset + sizeof(char) * s.size() + sizeof(int32_t) < _n, "");
-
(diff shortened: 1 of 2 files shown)


src/util/syncbuffer.cpp
```diff
@@ -37,7 +37,13 @@ SyncBuffer::~SyncBuffer() {} // NOLINT
 void SyncBuffer::encode(const std::string& s) {
     ghoul_assert(_encodeOffset + sizeof(char) * s.size() + sizeof(int32_t) < _n, "");
```
emiax: This assert should be removed now right? It would trigger and exit the application even though larger buffer size is actually supported below.

src/util/syncbuffer.cpp
```diff
@@ -37,7 +37,13 @@ SyncBuffer::~SyncBuffer() {} // NOLINT
...
 
-    int32_t length = static_cast<int32_t>(s.length());
+    int32_t anticpatedBufferSize = _encodeOffset + (sizeof(char) * s.size())
+        + sizeof(int32_t);
+    if (anticpatedBufferSize >= _n) {
+        _dataStream.resize(anticpatedBufferSize);
+    }
+
+    int32_t length = static_cast<int32_t>(s.size() * sizeof(char));
     memcpy(
```
emiax: If you change the memcpy to a std::insert, or similar (https://en.cppreference.com/w/cpp/container/vector/insert), you don't need to explicitly resize the vector before. I think that would be easier to read and less prone to errors.

GPayne: I had originally considered doing this, but decided not to make this change. This code is using a `std::vector`, but accesses its memory directly without using any of the safe/convenient features of the container (besides resize). `memcpy` is very fast, so I suspect this was written for performance, and so I'll keep it as-is. Otherwise I would want to rewrite everything to use the vector member functions for access.

emiax: Check this for performance comparison between insert and memcpy:
http://thewayofc.blogspot.com/2014/05/copying-memory-from-c-to-c-using.html?m=1

GPayne: When I say that `memcpy` is fast, I'm not advocating using it over C++ standard operators. I'm just saying this is the only reason I can think of why it had been written that way.
The options I see for this fix:
1. Stick with `memcpy` to stay consistent with the "C style" of memory access that this file currently uses (I'm not counting resize because it was already in there).
2. Use a mix of C and C++ styles of memory access by introducing std::insert.
3. Rewrite everything to use C++ standard operators.

If I was writing this from scratch I would use only C++ standard operations on a vector. Even if there was a performance hit, it would be more readable. I have other things to work on right now so option #3 is out, at least for now.

Functionally, either way will work but what is best for code readability? For this fix, I prefer option 1 because the code is more consistent. I think that mixing access types makes it more confusing to the next person. What is your opinion on option 1 or 2? I will make my change based on which you think is best for readability.

src/util/syncbuffer.cpp
GPayne: Correct, this line has been removed.

Issue #885, opened 2019-05-28
Issue with transparency with RenderableTrails in the Apollo scene


alexanderbock:
After the TLI event, the Earth-centric launch orbit are not correctly rendered relative to the atmosphere

alexanderbock:
Additionally, the global trail is not correctly culled against Earth

emiax:
Currently, all renderables decide what render bin they should belong to based on their opacity. However, if they are using additive blending (which trails should, since the fade-over-time aspect makes them translucent generally) they should always render in the trasparent bin.

Here is the change that introduced a general renderbin detection for renderables: https://github.com/OpenSpace/OpenSpace/pull/618/files#diff-164881db16899efa14f4b2d3abc68736R104

This should also be a problem for other renderables that have a blend mode option, like RenderableSphere.

I suggest we remove the automatic selection of render bin, since the bin may depend on other things than the `_opacity` property. If trails render to the transparent bin, and do so with additive blending, the problem will disappear.

emiax:
Will create a PR for this change

emiax:
I thought this would be straight forward, but simply making trails use additive blending makes one problem remain: The atmosphere will pick up the position and normals written by the trail, and get affected by it, so even if the trail is completely transparent, it will affect the look of the atmosphere.
An alternative would be to have transparent geometry NOT write to the position + normal buffers. The result of this is that the atmosphere will always be rendered "on top" of of the trail. There is no longer any trace of the transparent trail, but there is a drastic shift in appearance of the trail depending on if the atmosphere is rendered on the same pixel or not. The atmosphere is rendered on top, changing the color of the trail as if would be located on the planetary surface.

As I currently see it, the proper way to solve this is to draw the trails (potentially along with other transparent geometry) after the atmosphere. This is not trivial at this point, since the atmosphere is so involved in the HDR rendering. @jccosta, I think it would be good to have a meeting soon and talk more about your HDR changes, and catch up in general.

jccosta:
@emiax I think we should talk too. There is another inherent problem with the trails and transparency.  The Sun's glare is also a transparent billboard and it must be rendered after all trails, otherwise you can get the wrong blending sorting. What is the requirement for having transparent trails? The end "tail" being eroding?

emiax:
Here are some suggestions:

![rendering-pipeline](https://user-images.githubusercontent.com/1081118/59104580-39a51100-8932-11e9-8654-be4349024b80.png)

[...]

PR #943, opened 2019-07-29
HDR Rendering Branch

alexanderbock:
currently not fit for merge,  this PR is just to check the changes


modules/atmosphere/shaders/atmosphere_deferred_fs.glsl
```diff
@@ -426,10 +416,16 @@ vec3 inscatterRadiance(inout vec3 x, inout float t, inout float irradianceFactor
     // we can change it on the fly with no precomputations)
     // return radiance * sunRadiance;
     vec3 finalScatteringRadiance = radiance * sunIntensity;
+    
     if (groundHit) {
         return finalScatteringRadiance;
     } else {
-        return ((r-Rg) * invRtMinusRg)*spaceColor.rgb * backgroundConstant + finalScatteringRadiance;
+        //return ((r-Rg) * invRtMinusRg)*spaceColor.rgb + finalScatteringRadiance;
+        //return attenuation * spaceColor.rgb + finalScatteringRadiance;
```
emiax: `attenuation * spaceColor.rgb + finalScatteringRadiance;` is equivalent to the current return value if attenuation is between 0 and 1. Replace current return with this line?

openspace.cfg
```diff
@@ -166,7 +166,7 @@ LoadingScreen = {
     ShowNodeNames = true,
     ShowProgressbar = false
 }
-CheckOpenGLState = false
+CheckOpenGLState = true
```
emiax: This should probably not be changed?

shaders/fragment.glsl
```diff
@@ -32,6 +32,7 @@ struct Fragment {
     vec4 color;
     vec4 gPosition;
     vec4 gNormal;
+    uint filterFlag;
```
emiax: Seems that filtering was taken out of this branch -  I suppose this should go as well?

shaders/framebuffer/renderframebuffer.frag
```diff
@@ -25,14 +25,25 @@
...
+layout(location = 3) out vec4 filterBuffer;
 
 void main() {
-     Fragment f   = getFragment();
-     _out_color_  = f.color;
-     gPosition    = f.gPosition;
-     gNormal      = f.gNormal;
-     gl_FragDepth = normalizeFloat(f.depth);
+    Fragment f  = getFragment();
+    _out_color_ = vec4((log2(vec3(1.0) - (f.color.rgb - vec3(DeltaError)))/(-exposure)), f.color.a);
```
emiax: Is this the inverse of the tone mapping? In that case, I guess it is only the inverse of one of the TMOs - in that case, which one?

I thought the idea was to do this only for non-hdr fragments, but this seems to be happening for all fragments (?)

My interpretation of this is that all fragments are now supposed to be rendered in "LDR" (non-linear), get transformed to "HDR" (where they can be added linearly) and then get mapped back? I think this calls for some discussion. @jccosta when are you available to talk?

What is `DeltaError`?

src/rendering/framebufferrenderer.cpp
```diff
@@ -271,6 +341,67 @@ void FramebufferRenderer::deferredcastersChanged(Deferredcaster&,
     _dirtyDeferredcastData = true;
 }
 
+void FramebufferRenderer::resolveMSAA(float blackoutFactor) {
```
emiax: This method is not used anywhere (?)

src/rendering/framebufferrenderer.cpp
```diff
@@ -1241,6 +1403,7 @@ const std::vector<double>& FramebufferRenderer::mSSAPattern() const {
 void FramebufferRenderer::updateRendererData() {
     ghoul::Dictionary dict;
     dict.setValue("fragmentRendererPath", std::string(RenderFragmentShaderPath));
+    dict.setValue("hdrExposure", std::to_string(_hdrExposure));
```
emiax: I don't see how only the exposure is needed here, but not other parameters for the current TMO, such as which operator that is used, the gamma etc.

include/openspace/rendering/renderengine.h
```diff
@@ -206,15 +228,28 @@ class RenderEngine : public properties::PropertyOwner {
...
+    properties::PropertyOwner _tmoOwner;
     properties::FloatProperty _hdrExposure;
-    properties::FloatProperty _hdrBackground;
+    properties::FloatProperty _maxWhite;
+    properties::OptionProperty _toneMapOperator;
+    
+    properties::PropertyOwner _imageOwner;
     properties::FloatProperty _gamma;
+
+    properties::FloatProperty _hue;
```
emiax: Can you name a use-case for adjusting the hue?

include/openspace/rendering/renderengine.h
```diff
@@ -206,15 +228,28 @@ class RenderEngine : public properties::PropertyOwner {
...
     properties::FloatProperty _globalBlackOutFactor;
     properties::IntProperty _nAaSamples;
+    
+    properties::PropertyOwner _tmoOwner;
     properties::FloatProperty _hdrExposure;
-    properties::FloatProperty _hdrBackground;
+    properties::FloatProperty _maxWhite;
+    properties::OptionProperty _toneMapOperator;
+    
+    properties::PropertyOwner _imageOwner;
```
emiax: Not used?

include/openspace/rendering/renderengine.h
```diff
@@ -206,15 +228,28 @@ class RenderEngine : public properties::PropertyOwner {
 
     properties::FloatProperty _globalBlackOutFactor;
     properties::IntProperty _nAaSamples;
+    
+    properties::PropertyOwner _tmoOwner;
```
emiax: Not used?

src/rendering/renderengine.cpp
```diff
@@ -255,9 +303,17 @@ RenderEngine::RenderEngine()
...
-    , _hdrExposure(HDRExposureInfo, 0.4f, 0.01f, 10.0f)
-    , _hdrBackground(BackgroundExposureInfo, 2.8f, 0.01f, 10.0f)
-    , _gamma(GammaInfo, 2.2f, 0.01f, 10.0f)
+    , _tmoOwner(TMOInfo)
+    , _hdrExposure(HDRExposureInfo, 1.68f, 0.01f, 10.0f)
+    , _maxWhite(MaxWhiteInfo, 4.f, 0.001f, 100.0f)
+    , _toneMapOperator(ToneMapOperatorInfo, properties::OptionProperty::DisplayType::Dropdown)
+    , _imageOwner(ImageInfo)
+    , _gamma(GammaInfo, 0.86f, 0.01f, 5.0f)
+    , _hue(HueInfo, 1.f, 0.0f, 5.0f)
```
emiax: Do we really need hue? If we do, what is the reasoning for a 0-5 range?

emiax:
Good to see that this is coming together!
Here are my two main concerns about this branch:

1) I'm not sure what happened to the idea of supporting both HDR and LDR fragments.

I put in some comments above, in `renderframebuffer.frag` related to this. It seems now that only LDR is supported, and that all fragments are blended into HDR space. I think it is important that we make this part of the code reaonsbly easy to follow, as it will affect the rendering of everything. I'm not sure how the inverse TMO was derived, but it can only be a valid inverse for one of the TMOs, and does not seem to adapt to which of the ones that are in use. Can we remove all other TMOs than the one used here? I also don't understand how it's possible that the inverse does not involve the gamma.

2) The problem with blending trails with the atmosphere persists. 

If trails are rendered before the atmosphere, we can either let them write to the position buffer or not and get different results. 
 1) Trails write to position buffer =>The atmosphere depth will be affected by the translucent trail, and cause artifacts. Not acceptable.
  2) Trails don't write to the position buffer => Trails will be blended in as is they were projected onto the surface of the planet, but only if the fragment is inside the atmosphere bounds. If outside the atmosphere, there will be no blending, causing a drastic change in appearance on a trail as it enters the atmosphere. Not acceptable either.

I posted a suggestion here: https://github.com/OpenSpace/OpenSpace/issues/885#issuecomment-499871327 on how to address the issue, namely to render trails (and potentially other translucent geometry) after the atmosphere. I implemented the idea here: https://github.com/OpenSpace/OpenSpace/pull/905. We could not merge it, since it reqiured changes related to this branch, which was work in progress.

I think both these matters need to be addressed before we can merge this. Thoughts?
If you agree, I can go ahead and implement the change in rendering order on top of this branch. Regarding the TMOs and inverses, I think we need to have a meeting to get everyone on the same page.

data/assets/default.scene
```diff
@@ -10,6 +10,13 @@ asset.onInitialize(function ()
     openspace.globebrowsing.goToGeo("Earth", 58.5877, 16.1924, 20000000)
 
     openspace.markInterestingNodes({ "Earth", "Mars", "Moon", "Sun" })
+
+     -- HDR / Image options:
+    openspace.setPropertyValueSingle('RenderEngine.Gamma', 0.95);
+    openspace.setPropertyValueSingle('RenderEngine.HDRExposure', 3.7);
+    openspace.setPropertyValueSingle('RenderEngine.ToneMapOperator', 8);
+    openspace.setPropertyValueSingle('RenderEngine.Lightness', 1.1);
+    openspace.setPropertyValueSingle('RenderEngine.Saturation', 1.2);
```
emiax: Isn't it strange that the default behavior is to boost up the colors?
I think it would make sense if pretty much all the settings were set to 1.0, and then adjust individual renderables to look good with this default setting.

PR #951, opened 2019-08-15
Pull code from 2019 space debris thesis into Master

GPayne:
This is code space debris code from Elon & Jonathan's 2019 thesis project. The main intent is to pull the satellite rendering in. The volume rendering code is here, but is disabled from the build currently.


modules/space/translation/keplertranslation.cpp
```diff
@@ -304,6 +304,26 @@ glm::dvec3 KeplerTranslation::position(const UpdateData& data) const {
     return _orbitPlaneRotation * p;
 }
 
+glm::dvec3 KeplerTranslation::debrisPos(const double& time) const {
```
emiax: Is there a reason to have a separate method for this, instead of the regular `position` method?

modules/space/rendering/renderablesatellites.h
```diff
@@ -0,0 +1,141 @@
+// /****************************************************************************************
```
emiax: Double comments

openspace.cfg
```diff
@@ -6,13 +6,14 @@
 -- occurs in a single window, a fisheye projection, or a dome cluster system
 
 -- A regular 1280x720 window
-SGCTConfig = sgct.config.single{}
```
emiax: This should be kept

modules/space/translation/keplertranslation.cpp
GPayne: Must have been an intention to do position differently but never happened. I changed this to just having one `position` method to be consistent with other Translations.

modules/space/rendering/renderablesatellites.h
GPayne: Fixed

openspace.cfg
GPayne: Agreed, fixed.

modules/space/translation/keplertranslation.cpp
emiax: Cool, thanks!

PR #967, opened 2019-08-29
Feature/gui for touch

liuloppan:
Feature gui-for-touch


data/assets/gui/images.asset
```diff
@@ -0,0 +1,48 @@
+-- This asset requires OpenSpace to be built with the OPENSPACE_MODULE_SYNC enabled
+
+local assetHelper = asset.require("util/asset_helper")
+
+asset.syncedResource({
+    Type = "UrlSynchronization",
+    Name = "Icons",
+    Identifier = "planet_icons",
+    Url = {
+        "http://data.openspaceproject.com/dist/sci/mastertheses/2018/icons/solarsystem/earth.png",
```
emiax: We may want to move this before merging?

data/assets/util/webgui.asset
```diff
@@ -38,12 +38,12 @@ asset.onInitialize(function ()
     openspace.setPropertyValueSingle("Modules.WebGui.ServerProcessEnabled", enabled)
 
     if guiCustomization.webguiDevelopmentMode then
-        -- Route CEF to the deveopment version of the GUI.
+        -- Route CEF to the development version of the GUI.
```
emiax: Thanks :)

data/assets/util/webgui.asset
```diff
@@ -38,12 +38,12 @@ asset.onInitialize(function ()
...
         -- This must be manually served using `npm start`
         -- in the OpenSpace-WebGuiFrontend repository.
         openspace.setPropertyValueSingle(
             "Modules.CefWebGui.GuiUrl",
-            "http://127.0.0.1:4690/frontend/#/onscreen"
+            "http://127.0.0.1:4690/frontend/#/ontouch"
```
emiax: We could possibly move this string to assets/customization/gui.asset, or perhaps just set a flag there and check it here.

include/openspace/interaction/orbitalnavigator.h
```diff
@@ -142,6 +143,14 @@ class OrbitalNavigator : public properties::PropertyOwner {
     properties::FloatProperty _followAnchorNodeRotationDistance;
     properties::FloatProperty _minimumAllowedDistance;
 
+    properties::FloatProperty _velocitySensitivity;
+    properties::BoolProperty _flyTo;
+    properties::BoolProperty _overview;
+    //If 1st story selected, allow an additional camera update iteration before overview
+    bool _cameraInitializedBeforeFirstStoryOverview = false;
```
emiax: I think we need to deal with this special case before merging it in - orbitalnavigator is such a core part of the system and should not need to care about the concept of a story

include/openspace/scene/scenegraphnode.h
```diff
@@ -187,6 +192,14 @@ class SceneGraphNode : public properties::PropertyOwner {
     glm::dmat4 _modelTransformCached;
     glm::dmat4 _inverseModelTransformCached;
 
+    properties::BoolProperty _computeScreenSpaceValues;
+    properties::IVec2Property _screenSpacePosition;
+    properties::BoolProperty _screenVisibility;
+    properties::DoubleProperty _distFromCamToNode;
+    properties::DoubleProperty _screenSizeRadius;
+    properties::FloatProperty _visibilityDistance;
+    std::chrono::high_resolution_clock::time_point _lastUpdate;
```
emiax: Is this _lastUpdate only used for the new properties regarding screen space data? In that case, its name is too generic.

modules/base/scale/staticscale.cpp
```diff
@@ -58,7 +58,7 @@ double StaticScale::scaleValue(const UpdateData&) const {
     return _scaleValue;
 }
 
-StaticScale::StaticScale() : _scaleValue(ScaleInfo, 1.0, 1.0, 1e6) {
+StaticScale::StaticScale() : _scaleValue(ScaleInfo, 1.0, 0.1, 100) {
```
emiax: This is kind of funny - I keep changing this locally depending on what I want to do for the moment.
At some point it would be nice to have a straight-forward way to set these limits more dynamically based on what kind of object the scale applies to. Now's not the time though :) I think the factor 100 is probably at least as good as 1e6, so I'm fine with using 100 for now. :)

openspace.cfg
```diff
@@ -130,6 +129,7 @@ ModuleConfigurations = {
         WebSocketInterface = "DefaultWebSocketInterface"
     },
     CefWebGui = {
+        GuiUrl = "http://localhost:4680/#/ontouch/",
```
emiax: Apparently we have two ways to set this now... :) We will have to figure this out before merging

src/interaction/navigationhandler.cpp
```diff
@@ -52,6 +52,14 @@ namespace {
     constexpr const char* KeyReferenceFrame = "ReferenceFrame";
     const double Epsilon = 1E-7;
 
+    constexpr const openspace::properties::Property::PropertyInfo OriginInfo = {
```
emiax: I think this accidentally slipped in? The `origin` was split into `anchor` and `aim` and moved into `OrbitalNavigator`. I think all this `origin` related stuff can be removed...  (?)

src/interaction/navigationhandler.cpp
```diff
@@ -142,7 +150,24 @@ NavigationHandler::NavigationState::NavigationState(std::string anchor, std::str
 NavigationHandler::NavigationHandler()
     : properties::PropertyOwner({ "NavigationHandler" })
     , _useKeyFrameInteraction(KeyFrameInfo, false)
+    , _origin(OriginInfo)
 {
+    _origin.onChange([this]() {
```
emiax: This should probably be removed as well. And the property itself in the header file.

src/interaction/orbitalnavigator.cpp
```diff
@@ -136,8 +135,45 @@ namespace {
...
+        "Triggering this property makes the camera move to an overview."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ApplyFlyToInfo = {
+        "ApplyFlyTo",
+        "Apply fly to node",
+        "Triggering this property makes the camera move to the focus node."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo StoryOverviewLimitInfo = {
```
emiax: It would be great if we can figure out how this can be either moved out to a class that cares about stories, or if the concept can be generalized to be useful for non-story use-cases.

src/interaction/orbitalnavigator.cpp
```diff
@@ -403,6 +459,66 @@ void OrbitalNavigator::updateCameraStateFromStates(double deltaTime) {
         _camera->rotationQuaternion()
     };
 
+    if (_flyTo || _overview) {
```
emiax: I think this method is already very long and hard to get an overview of. Could this code be moved out to a separate method? Also there are some assumptions about planets here like `planetRadius` and the constant `4` etc.

src/interaction/orbitalnavigator.cpp
```diff
@@ -1173,6 +1289,32 @@ glm::dvec3 OrbitalNavigator::translateHorizontally(double deltaTime,
...
+    const double focusLimitBuffer = 0.1;
+
+    // Calculate and truncate the factor which determines the velocity of the 
+    // camera movement towards the focus node
+    if (flyTo) {
+        velocityFactor = (1 - ((focusLimit * (1.0 - focusLimitBuffer))
+        / distFromCameraToFocus));
+    }
+    else {
+        velocityFactor = (1 - (distFromCameraToFocus
```
emiax: The line is split in a strange way, right where one of the highest priority operators are.

src/interaction/orbitalnavigator.cpp
```diff
@@ -1173,6 +1289,32 @@ glm::dvec3 OrbitalNavigator::translateHorizontally(double deltaTime,
...
+        / (focusLimit * (1.0 + focusLimitBuffer))));
+    }
+    velocityFactor *= _velocitySensitivity;
+    velocityFactor = floor(velocityFactor * 1000) / 1000;
```
emiax: Why is this necessary? Why 1000?

src/scene/scenegraphnode.cpp
```diff
@@ -595,6 +652,91 @@ void SceneGraphNode::setDependencies(const std::vector<SceneGraphNode*>& depende
     }
 }
 
+void SceneGraphNode::computeScreenSpaceData(RenderData& newData) {
+    // Purposely slow the update rate of screen space position in order to reduce the
+    // effects of jittering in the position of information icon markers in web gui.
```
emiax: Is this really the reason? I thought it would be for performance reasons?

src/scene/scenegraphnode.cpp
```diff
@@ -595,6 +652,91 @@ void SceneGraphNode::setDependencies(const std::vector<SceneGraphNode*>& depende
...
+    if ((now - _lastUpdate) < std::chrono::milliseconds(100)) {
+        return;
+    }
+    _lastUpdate = now;
+
+    // Calculate ndc
+    const Camera& cam = newData.camera;
+    const glm::dvec3& worldPos = _worldPositionCached;
+    const glm::dvec4 clipSpace = glm::dmat4(cam.projectionMatrix()) *
+                                 cam.combinedViewMatrixNoScale() * glm::vec4(worldPos, 1.0);
```
emiax: The method `combinedViewMatrixNoScale` seems to be have introduced only for this use case. Is it really needed? At first glance, it seems that a scaling shouldn't matter here... I'm not sure and would need to think about this a little more.

src/util/camera.cpp
```diff
@@ -139,6 +139,19 @@ float Camera::maxFov() const {
     return _maxFov;
 }
 
+const glm::dmat4& Camera::combinedViewMatrixNoScale() const {
```
emiax: I'm unsure whether this method is necessary. Right now it seems to only be used by `computeScreenSpaceData`, which I'm not sure actually needs it. Maybe it's a good method to have anyways... Thoughts?

modules/base/scale/staticscale.cpp
liuloppan: Ok, you are right. I changed it because I figured it was impossible to use the slider for objects such as spacecrafts, planets etc. but I can imagine stars and other far away objects would need a very large scale when testing.

src/interaction/navigationhandler.cpp
liuloppan: Fixed now

src/interaction/navigationhandler.cpp
liuloppan: Fixed now

src/util/camera.cpp
liuloppan: I will remove this for now, I could not spot any difference with using the regular combinedViewMatrix.

src/scene/scenegraphnode.cpp
liuloppan: Removed combinedViewMatrixNoScale() and replaced it with combinedViewMatrix()

data/assets/gui/images.asset
liuloppan: @alexanderbock Can put it on the server maybe?

include/openspace/interaction/orbitalnavigator.h
alexanderbock: Yea, I agree.  I quick workaround for that would be to set the camera position from the overview screen twice?  A bit hackish, but I think that would work

src/scene/scenegraphnode.cpp
alexanderbock: It is;  sending updates too often slows down the rendering, but also overloads CEF as can't handle that many updates iirc

data/assets/gui/images.asset
alexanderbock: It's moved and the URLs are updated

include/openspace/interaction/orbitalnavigator.h
alexanderbock: @liuloppan  could you check if that works?
Removing this variable and the code that it executes and instead solving it on the GUI part?

alexanderbock:
Latest commit had braking changes:
 - removed endpoint from openspace.cfg and moved into touch.scene
 - renamed StoryOverviewLimit -> OverviewLimit property for orbitalnavigator

src/interaction/orbitalnavigator.cpp
liuloppan: This logic has been updated to only handle a generic linearflight to a set distance from the anchor node. Distinction between overview (flyout) and flyto is now made on the frontend instead.

src/interaction/orbitalnavigator.cpp
liuloppan: This code has been updated now, simplyfying a lot of the math.

PR #999, opened 2019-10-25
Feature/interaction monitor

liuloppan:
Created a minimal interaction monitor. Currently it registers mouse move, mouse click and touch events in the OpenSpace window. 
It can be used for applications (for example web gui etc) that needs to know if OpenSpace is in an idle state.

You can set the state directly to idle/active or set the time it takes from the latest interaction until the state turns idle.

In the future I see how this could be expanded to track more kinds of interactions, to be able to filter what type of interactions that affects "idleness" as well as create logs of user interactions etc.


modules/touch/touchmodule.cpp
```diff
@@ -46,54 +47,59 @@ using namespace TUIO;
 
 namespace openspace {
 
-bool TouchModule::hasNewInput() {
+bool TouchModule::processNewInput() {
```
emiax: :+1:

src/interaction/interactionmonitor.cpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        double currentApplicationTime = global::windowDelegate.applicationTime();
+        double timeDiff = currentApplicationTime - _lastInteractionTime;
+
+        if (timeDiff >= _idleTime.value() && _isInActiveState) {
+            setActivityState(false);
+        }
+    }
+
+    void InteractionMonitor::registerInteraction(std::string interactionType)
```
emiax: Better to use an enum type here instead of string, both for maintainability and performance.
However, since the feature of `interactionType` is not used for anything right now, I'd even suggest removing the parameter altogether until we have a concrete use case for it. Thoughts?

src/interaction/interactionmonitor.cpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        _lastInteractionTime = global::windowDelegate.applicationTime();
+        _lastInteractionType = interactionType;
+        setActivityState(true);
+        // LDEBUG(fmt::format("Interaction Type {} at time {}", _lastInteractionType, _lastInteractionTime));
```
emiax: Comment can be removed

src/interaction/navigationhandler.cpp
```diff
@@ -642,7 +642,7 @@ scripting::LuaLibrary NavigationHandler::luaLibrary() {
                 "locally or remotely. The latter being the default."
             },
             {
-                "clearJoystickButotn",
+                "clearJoystickButton",
```
emiax: Well spotted :)

include/openspace/interaction/interactionmonitor.h
```diff
@@ -0,0 +1,73 @@
...
+class InteractionMonitor : public properties::PropertyOwner {
+
+public:
+    InteractionMonitor();
+    ~InteractionMonitor() = default;
+
+    void setActivityState(bool isActive);
+    void setIdleTime(double time);
+
+
```
emiax: One empty line is enough here

include/openspace/interaction/interactionmonitor.h
```diff
@@ -0,0 +1,73 @@
...
+#ifndef __OPENSPACE_CORE___INTERACTIONMONITOR___H__
+#define __OPENSPACE_CORE___INTERACTIONMONITOR___H__
+
+#include <openspace/properties/propertyowner.h>
+#include <ghoul/logging/logmanager.h>
+
+#include <openspace/properties/scalar/boolproperty.h>
+#include <openspace/properties/scalar/doubleproperty.h>
+
+
```
emiax: One empty line is enough here

src/interaction/interactionmonitor.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+
```
emiax: Empty line can be removed

src/interaction/interactionmonitor.cpp
```diff
@@ -0,0 +1,85 @@
...
+        "IsInActiveStateInfo",
+        "Is State Active",
+        "Keeps track whether the interaction session is in active state or not. " 
+        "False if application is in idle state, true if it is in active state."
+    };
+} // namespace
+
+namespace openspace::interaction {
+
+    InteractionMonitor::InteractionMonitor()
```
emiax: The namespace should not be indented

src/interaction/interactionmonitor.cpp
liuloppan: I can remove it for now. I used it for debugging purposes earlier and thought  it could be useful but it does not serve any functionality at the moment.

include/openspace/interaction/interactionmonitor.h
liuloppan: fixed

src/interaction/interactionmonitor.cpp
liuloppan: fixed

modules/touch/touchmodule.cpp
```diff
@@ -114,9 +120,8 @@ bool TouchModule::hasNewInput() {
 }
 
 void TouchModule::hasNewWebInput(const std::vector<TuioCursor>& listOfContactPoints) {
```
GPayne: Suggest renaming this to `processNewWebInput` to be consistent with your new naming of `processNewInput` above.

liuloppan: Good point, fixed now.

PR #1001, opened 2019-10-28
Feature/get navigation state

emiax:


emiax:
@alexanderbock Let me know if you think this is ready for merge

src/interaction/navigationhandler.cpp
```diff
@@ -568,21 +572,20 @@ scripting::LuaLibrary NavigationHandler::luaLibrary() {
...
                 &luascriptfunctions::getNavigationState,
                 {},
-                "[string]",
-                "Return the current navigation state as a lua table. "
-                "The optional argument is the scene graph node to "
-                "use as reference frame. By default, the reference frame will picked "
-                "based on whether the orbital navigator is currently following the "
-                "anchor node rotation. If it is, the anchor will be chosen as reference "
-                "frame. If not, the reference frame will be set to the scene graph root."
+                "[table]",
```
emiax: This is the optional input argument, and not the return value, right? It should be [string].

alexanderbock: Yep.. you are right.. oopsie

emiax: Thanks for reverting

PR #1003, opened 2019-10-30
Feature/serve screenshots

emiax:
Remove trigger property `takeScreenShot`, and instead introduce lua function `takeScreenshot` returning the generated filename. 

Add an asset that adds the current `${SCREENSHOTS}` folder as an HTTP endpoint.

### DIFF
--- a/src/rendering/renderengine.cpp
+++ b/src/rendering/renderengine.cpp
@@ -124,16 +124,6 @@ namespace {
         "shown on the screen"
     };
 
-    constexpr openspace::properties::Property::PropertyInfo TakeScreenshotInfo = {
-        "TakeScreenshot",
-        "Take Screenshot",
-        "If this property is triggered, a screenshot is taken and stored in the current "
-        "working directory (which is the same directory where the OpenSpace.exe) is "
-        "located in most cases. The images are prefixed with 'SGCT' and postfixed with "
-        "the number of frames. This function will silently overwrite images that are "
-        "already present in the folder."
-    };
-
     constexpr openspace::properties::Property::PropertyInfo ApplyWarpingInfo = {
         "ApplyWarpingScreenshot",
         "Apply Warping to Screenshots",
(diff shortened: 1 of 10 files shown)


src/rendering/renderengine.cpp
```diff
@@ -1061,10 +1032,31 @@ void RenderEngine::setResolveData(ghoul::Dictionary resolveData) {
...
+
+    _latestScreenShotNumber = global::windowDelegate.takeScreenshot(_applyWarping);
+}
+
+/**
+ * Get the latest screenshot filename
+ */
+std::string RenderEngine::latestScreenShotFilename() const {
+    return fmt::format(
+        "OpenSpace_{:0>6}.png",
```
alexanderbock: This is not a happy solution;  this only works for the single configuration file without stereo rendering.  We won't get around asking SGCT for the correct screenshot name

src/rendering/renderengine.cpp
```diff
@@ -1061,10 +1032,31 @@ void RenderEngine::setResolveData(ghoul::Dictionary resolveData) {
...
+    // screenshots. So the first time we actually take one, we create the folder:
+
+    if (!FileSys.directoryExists(absPath("${SCREENSHOTS}"))) {
+        FileSys.createDirectory(
+            absPath("${SCREENSHOTS}"),
+            ghoul::filesystem::FileSystem::Recursive::Yes
+        );
+    }
+
+    _latestScreenShotNumber = global::windowDelegate.takeScreenshot(_applyWarping);
```
alexanderbock: This is also very brittle as, for example, in the case of stereo or two windows, two screenshots will be taken, but the function will only return the number for the first screenshot.  Same as below, we need to hook into SGCT for this rather than trying to reverse engineer how many screenshots were taken

emiax: When taking a screenshot with multiple windows, sgct uses the same screenshot number for all of them. I tested it now, and found a bug in master where two windows would try to write to the same file when taking a screenshot. Fixed it in this branch by adding the window number to the filename iff more than one window is used.

I suggest letting the screenshot function return the screenshot number instead and letting client code deal with reconstructing the file name if necessary. The alternative would be to return a list of filenames, which wont make client code happier anyways. Thoughts?

emiax:
Ok for merge?

PR #1007, opened 2019-11-12
Use a StaticTranslation to position the Milky Way

emiax:
This removes a hard coded value from the shader and makes it possible to focus on the center of the milky way


liuloppan:
I think this looks good, nothing controversial, looks ready for merge. I can merge if @alexanderbock approves.

PR #1017, opened 2019-12-02
Improvements in Performance due to adopted usage of CEF Message Loop

stefanseibert:
This improves performance especially when using the CEF web gui by using certain command line switches and by reducing the time intervall to pump the cef message loop about once every 10 milliseconds in the regular update loop.

### DIFF
--- a/modules/webbrowser/webbrowsermodule.cpp
+++ b/modules/webbrowser/webbrowsermodule.cpp
@@ -142,16 +142,6 @@ void WebBrowserModule::internalInitialize(const ghoul::Dictionary& dictionary) {
     _cefHost = std::make_unique<CefHost>(_webHelperLocation);
     LDEBUG("Starting CEF... done!");
 
-    global::callback::preSync.emplace_back([this]() {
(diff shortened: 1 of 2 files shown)


modules/webbrowser/webbrowsermodule.cpp
```diff
@@ -225,7 +215,7 @@ namespace webbrowser {
  * described in more detail in globalscallbacks.h.
  */
 
-std::chrono::microseconds interval = std::chrono::microseconds(1);
+std::chrono::microseconds interval = std::chrono::microseconds(10000);
```
liuloppan: Maybe a comment why this interval number was chosen. Has it been optimized somehow or is it an estimated value that may be possible to improve further?

stefanseibert: I added a comment that explains the magic number a bit better. It is recommended from CEF side to call that function "regularly" which is quite unspecific but from testing i feel it doesnt need more then 100fps. I wouldnt go less (higher number) since the gui will not be very responsive again

PR #1025, opened 2019-12-11
Feature/planetlabels

jccosta:
New Renderable type for labels. It can be added to any node (as a child).


liuloppan:
In assets each planet has a file named along the lines of "earth_labels.asset" which are globelabels (I assume) of different locations. However, would it not make sence to also put the planet label in this file? Currently they reside in the asset file of each planet. Alternatively rename the "..._labels.asset" files to "earth_globelabels.asset" or "earth_locationlabels.asset" to avoid confusion.

alexanderbock:
I merged it as it was with a few changes
 * By default the labels are disabled
 * L and Shift+L as keybinds to toggle all of the labels
 * Renamed _label to _globelabel assets as per @liuloppan's suggestion

Splitting the labels into their own asset files doesn't work unfortunately because of initialization order.  sun_label would be a child of sun, but also require the sun to be initialized before itself.  And the asset system will always initialize children before the parent.  So at the time of sun_labels initialization, the "Sun" node will not exist yet and the label generation will fail

PR #1027, opened 2019-12-11
Feature/ring shadows

jccosta:
Ring shadows!


alexanderbock:
OpenSpace_000000

Only one issue left with it I think ... there is an issue when rendering the RingsComponent with the background milkyway.  I'm assuming that the rings write into the depth buffer?  Maybe the depth testing or the updating of the depth buffer should be disabled for the rings as we want them to be transparent?

jccosta:
I fixed the transparency issue. I think it is ready for merging now.

alexanderbock:
Looks fantastic!!  Really well done!

:+1:

PR #1035, opened 2019-12-20
fix case sensitive include file for linux compile

sa5bke:
Just a quick fix to be able to compile on Linux which is case sensitive.


alexanderbock:
Thanks for the pull request..  that passed through the CI system.

PR #1037, opened 2019-12-27
support curl version < 7.32.0, i.e. rhel/centos 7

sa5bke:
This will make it possible to build OpenSpace on a Linux OS with an older curl version. The curl progress will not be shown on the splash screen but will instead be outputted in the console. I have verified that the download works on both cases on Linux.


alexanderbock:
Awesome, thanks!

PR #1040, opened 2020-01-09
Feature/distancelabels

liuloppan:
This PR contains:
* New vislab module
* RenderableDistanceLabel that inherits from RenderableLabel


modules/vislab/rendering/renderabledistancelabel.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include <modules/base/rendering/renderablenodeline.h>
+#include <openspace/documentation/documentation.h>
+#include <openspace/documentation/verifier.h>
+#include <openspace/engine/globals.h>
+#include <openspace/rendering/renderengine.h>
+#include <openspace/scene/scene.h>
+#include <string>
+
+namespace {
+    constexpr const char* _loggerCat = "RenderableDistanceLabel";
```
Mik-pe: I'm guessing this is unused since currently no logging is enabled in this file.

modules/vislab/rendering/renderabledistancelabel.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    if (dictionary.hasKey(NodeLineInfo.identifier)) {
+        _nodeLine = dictionary.value<std::string>(NodeLineInfo.identifier);
+        addProperty(_nodeLine);
+    }
+}
+
+void RenderableDistanceLabel::update(const UpdateData& data) {
+
+    if (global::renderEngine.scene()->sceneGraphNode(_nodeLine)) {
```
Mik-pe: Might as well assign to nodelineNode and check it for nullptr, since we're getting it on the row below.

modules/vislab/rendering/renderabledistancelabel.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        // Calculate distance
+        SceneGraphNode* nodelineNode = global::renderEngine.scene()->sceneGraphNode(_nodeLine);
+        RenderableNodeLine* nodeline = dynamic_cast<RenderableNodeLine*>(nodelineNode->renderable());
```
Mik-pe: is this a guaranteed cast? If so, consider static_cast? Otherwise consider nullptr-check on the cast.

modules/vislab/rendering/renderabledistancelabel.h
```diff
@@ -0,0 +1,51 @@
...
+class RenderableDistanceLabel : public RenderableLabels {
+public:
+     RenderableDistanceLabel(const ghoul::Dictionary& dictionary);
+
+     void update(const UpdateData& data) override;
+     static documentation::Documentation Documentation();
+
+     properties::StringProperty _nodeLine;
+
+private:
```
Mik-pe:
```suggestion
```

modules/vislab/rendering/renderabledistancelabel.cpp
```diff
@@ -0,0 +1,107 @@
...
+        std::string distanceText = std::to_string(std::round(myDistance / scale));
+        int pos = distanceText.find(".");
+        std::string subStr = distanceText.substr(pos);
+        distanceText.erase(pos, subStr.size());
+        std::string finalText = distanceText + " Km";
+        setLabelText(finalText);
+
+        // Update placement of label with transformation matrix
+        glm::dvec3 start = global::renderEngine.scene()->sceneGraphNode(nodeline->_start)->worldPosition();
+        glm::dvec3 end = global::renderEngine.scene()->sceneGraphNode(nodeline->_end)->worldPosition();
```
Mik-pe: Is nodeline->_start/_end a guaranteed scenegraphnode if _nodeLine exists?

modules/base/rendering/renderablenodeline.h
```diff
@@ -50,7 +51,18 @@ class RenderableNodeLine : public Renderable {
     ~RenderableNodeLine() = default;
 
     static documentation::Documentation Documentation();
-    
+
+    // Identifier of the optional distance label
+    std::string _distanceLabelId;
+    //SceneGraphNode* distanceLabelNode = nullptr;
+    //RenderableLabels* distanceLabel;
```
Mik-pe: These lines could probably be left out

modules/vislab/vislabmodule.cpp
```diff
@@ -0,0 +1,40 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include <modules/vislab/vislabmodule.h>
+
+namespace openspace {
+constexpr const char* _loggerCat = "VisLab Module";
```
Mik-pe: Same as above with _loggerCat

modules/vislab/vislabmodule.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef __OPENSPACE_MODULE_VISLAB___VISLABMODULE___H__
+#define __OPENSPACE_MODULE_VISLAB___VISLABMODULE___H__
+
+
+#include <openspace/util/openspacemodule.h>
+#include <openspace/util/factorymanager.h>
+#include <ghoul/misc/assert.h>
+#include <ghoul/misc/templatefactory.h>
+#include <ghoul/logging/logmanager.h>
```
Mik-pe: Unused include of logmanager.h

modules/vislab/vislabmodule.h
```diff
@@ -0,0 +1,52 @@
...
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_MODULE_VISLAB___VISLABMODULE___H__
+#define __OPENSPACE_MODULE_VISLAB___VISLABMODULE___H__
+
+
+#include <openspace/util/openspacemodule.h>
+#include <openspace/util/factorymanager.h>
+#include <ghoul/misc/assert.h>
+#include <ghoul/misc/templatefactory.h>
```
Mik-pe: These includes could probably be moved to .cpp-file

modules/base/rendering/renderablenodeline.h
```diff
@@ -27,6 +27,7 @@
 
 #include <openspace/rendering/renderable.h>
 
+#include <modules/base/rendering/renderablelabels.h>
```
Mik-pe: This include is unused, right?

modules/base/rendering/renderablenodeline.cpp
```diff
@@ -43,13 +43,13 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo StartNodeInfo = {
         "StartNode",
         "Start Node",
-        "The identifier of the node the line starts from. "
+        "The identifier of the node the line starts from. Defaults to 'Root' if not specified. "
```
Mik-pe: Long line - 
```suggestion
        "The identifier of the node the line starts from. "
        "Defaults to 'Root' if not specified."
```

modules/base/rendering/renderablenodeline.cpp
```diff
@@ -43,13 +43,13 @@ namespace {
...
     };
 
     constexpr openspace::properties::Property::PropertyInfo EndNodeInfo = {
         "EndNode",
         "End Node",
-        "The identifier of the node the line ends at. "
+        "The identifier of the node the line ends at. Defaults to 'Root' if not specified. "
```
Mik-pe: Long line - 
```suggestion
        "The identifier of the node the line ends at. "
        "Defaults to 'Root' if not specified."
```

modules/base/rendering/renderablenodeline.h
```diff
@@ -50,7 +51,18 @@ class RenderableNodeLine : public Renderable {
...
+    // Identifier of the optional distance label
+    std::string _distanceLabelId;
+    //SceneGraphNode* distanceLabelNode = nullptr;
+    //RenderableLabels* distanceLabel;
+
+    // Get the distance between the start and end node
+    double getDistance();
+
+    properties::StringProperty _start;
+    properties::StringProperty _end;
```
Mik-pe: Consider putting these behind private const getters since we never seem to set them outside this class?

modules/vislab/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+set(SOURCE_FILES
+    ${CMAKE_CURRENT_SOURCE_DIR}/rendering/renderabledistancelabel.cpp
+)
+source_group("Source Files" FILES ${SOURCE_FILES})
+
+create_new_module(
+    "VisLab"
+    vislab_module
+    ${HEADER_FILES} ${SOURCE_FILES}
+)
```
Mik-pe: Missing final newline of this file.

modules/vislab/rendering/renderabledistancelabel.cpp
liuloppan: That is correct, I can remove this, it can easily be added if we need it in the future.

modules/vislab/rendering/renderabledistancelabel.h
liuloppan: Agree

modules/base/rendering/renderablenodeline.h
liuloppan: Yes, my bad!

modules/base/rendering/renderablenodeline.cpp
liuloppan: fixed

modules/vislab/CMakeLists.txt
liuloppan: fixed

modules/base/rendering/renderablenodeline.h
liuloppan: yes, my bad! Old testing code....

modules/base/rendering/renderablenodeline.h
liuloppan: Technically they are set through luscripting via the GUI at some point, but this is still available even when they are private (through something we call the propertytree). If this is entirely kosher is debatable, but I will apply your suggestions with some getStart and getEnd functions and call these from 'renderabledistancelabel' instead.

modules/vislab/rendering/renderabledistancelabel.cpp
liuloppan: They are only validated on initialization by the nodeline and are always of type scenegraphnode. However, we could mess up the validity on runtime since we are changing the id with properties from the gui. But IF those ID:s are wrong this will become very apparent as they will mess up the nodeline itself. If a check should be made I think this should be validated in renderablenodeline itself, not on the label.

modules/vislab/rendering/renderabledistancelabel.cpp
liuloppan: Ok I will do this, and ake use of the loggerCat

PR #1046, opened 2020-01-13
Feature/touch improvements

Mik-pe:
This PR adds a Property (TouchActive) to the TouchModule which can be used to inactivate/reactivate the touch 3D navigation, without changing the GUI behavior.

It also fixes a bug which caused erratic zooming behaviour when touch navigation was active.  If the user focused a larger body, and switched back to a smaller one, the boundingbox of the large body would keep deciding the closest distance to the smaller object. This has been fixed.


GPayne:
Looks good to me.

PR #1050, opened 2020-01-24
Fixed issue 1049 and 554 (multiple viewport problems and lua console scales)

eriksunden:
Fixed issue 1049(using first viewport size as framebuffer size) and 554(loading and ending screen will be visible on the first viewport only). Also fixed such that Lua Console adapts to window size and not framebuffer size.
Tested with multiple configs, and zoom through the cosmos (milky way etc)

### DIFF
--- a/modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
+++ b/modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
@@ -1731,7 +1731,7 @@ void RenderableBillboardsCloud::renderToTexture(GLuint textureToRenderTo,
 
     glFramebufferTexture(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, textureToRenderTo, 0);
 
-    glViewport(0, 0, textureWidth, textureHeight);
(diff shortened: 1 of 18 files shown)


modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
```diff
@@ -1731,7 +1731,7 @@ void RenderableBillboardsCloud::renderToTexture(GLuint textureToRenderTo,
 
     glFramebufferTexture(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, textureToRenderTo, 0);
 
-    glViewport(0, 0, textureWidth, textureHeight);
+    glViewport(viewport[0], viewport[1], textureWidth, textureHeight);
```
alexanderbock: This actually looks like it should be 0,0 as it is rendering to an FBO?  @jccosta ?

alexanderbock:
Looks good to me;  Once @jccosta  signs off on the one in the RenderableBillboardsCloud:  Can merge

Issue #1062, opened 2020-01-30
PCF Shadows' shader files are not compiling on Mac


jccosta:
Mac (AMD) requires constant values for testing in loop parameters.

jccosta:
Fix added into #1063.

Issue #996, opened 2019-10-24
single_fisheye_gui.xml has errors


micahnyc:
needs border="true"

micahnyc:
tested and working in 0.15.1

PR #1063, opened 2020-01-30
macos compile and build (still needs ring shader const fix)

micahnyc:
macos compile fixes. 
Also Shadow-Mapping identifier fix for debug builds.


micahnyc:
addresses: #1062 #996

alexanderbock:
Can we inlude the ring shader fix into this as well? @jccosta ?

jccosta:
> Can we inlude the ring shader fix into this as well? @jccosta ?

Yes.

micahnyc:
I just updated this with fixes for the renderable trail. Trails and rings now working on mac. Ready to merge.

PR #1087, opened 2020-02-13
Changed webbrowser to use cef touch events

Mik-pe:
This PR changes the internal behaviour of the Cef webbrowser to treat touch events as true touch points (cef-wise) rather than as an injected mouse event.

This means that scrolling in menus, panning and potentially zooming of canvases should be possible through the means of touch.

For this to work, a newer CEF version than the one OSX currently uses is required. However, bumping CEF is not included in this PR, so touch is treated as injected mouse events on mac.

### DIFF
--- a/modules/webbrowser/include/browserinstance.h
+++ b/modules/webbrowser/include/browserinstance.h
@@ -84,7 +84,9 @@ public:
     void draw();
     void close(bool force = false);
 
+#ifdef WIN32
     void sendTouchEvent(const CefTouchEvent& event) const;
+#endif
 
     bool sendKeyEvent(const CefKeyEvent& event);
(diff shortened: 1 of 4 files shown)


micahnyc:
So if we up the cef version on mac, then we could have this code w/o all the #ifdef WIN32? 

...I would be ok to take this as is, since we have other windows specific touch code as well, and will make a note to look at this when upping the cef version on macos.

Mik-pe:
> 
> 
> So if we up the cef version on mac, then we could have this code w/o all the #ifdef WIN32?
> 
> ...I would be ok to take this as is, since we have other windows specific touch code as well, and will make a note to look at this when upping the cef version on macos.

Exactly! (mentioned in chat - putting it here for visibility)

PR #1089, opened 2020-02-18
Exposing Linear Flight Factor as Property

stefanseibert:
This exposes the linear flight factor as property. The factor is used to determine after what big of a percentage of the linear flight the flight is ended. It was defaulted to 0.0001 which leads to situation where the flight is perceived as taking too long to finish. By exposing the parameter we can set it to whatever value a project deems reasonable.


src/interaction/orbitalnavigator.cpp
```diff
@@ -232,6 +237,7 @@ OrbitalNavigator::OrbitalNavigator()
     , _velocitySensitivity(VelocityZoomControlInfo, 0.02f, 0.01f, 0.15f)
     , _applyLinearFlight(ApplyLinearFlightInfo, false)
     , _flightDestinationDistance(FlightDestinationDistanceInfo, 2e8f, 0.0f, 1e10f)
+    , _flightDestinationFactor(FlightDestinationFactorInfo, 1E-4)
```
liuloppan: May consider also adding a max and min value for the GUI property.

stefanseibert: Got it. Updated!

PR #1091, opened 2020-02-18
Updating SPVL Project branch with newest master updates

stefanseibert:
- Merged master into project/spvl
- Resolved merge conflicts and updated sub repos
- Tested the update branch by building it locally and running it with space pilot.


stefanseibert:
Please only squash merge!

liuloppan:
I think this is good for merge. I think squash merge is probably the best for us as this will generate one commit in our history which we can easily revert if we for some reason run into lots of issues. 
Unless @Mik-pe  has a different opinion I think it is ready for merge.

PR #1107, opened 2020-03-05
Added a way to customize renderable distance labels unit

stefanseibert:
This adds a way to customize the way how distances are shown on distance labels. It is possible to config the used unit (out of the existent openspace units, for example "gigameter"), and it is possible to set a custom descriptor. For SPVL for example we are using gigameter and the descriptor "million km" since we see that the average museum visitor has limited understanding to grasp "gigameter" and kilometer values are getting too big.

Capture

### DIFF
--- a/modules/vislab/rendering/renderabledistancelabel.cpp
+++ b/modules/vislab/rendering/renderabledistancelabel.cpp
@@ -42,6 +42,20 @@ namespace {
         "Property to track a nodeline. When tracking the label text will be updating the "
         "distance from the nodeline start and end."
     };
+
+    constexpr openspace::properties::Property::PropertyInfo DistanceUnitInfo = {
+        "DistanceUnit",
+        "Distance Unit",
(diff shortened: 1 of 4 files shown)


modules/vislab/rendering/renderabledistancelabel.cpp
```diff
@@ -76,6 +104,16 @@ RenderableDistanceLabel::RenderableDistanceLabel(const ghoul::Dictionary& dictio
         _nodelineId = dictionary.value<std::string>(NodeLineInfo.identifier);
         addProperty(_nodelineId);
     }
+    if (dictionary.hasKey(DistanceUnitInfo.identifier)) {
+        _distanceUnit = static_cast<int>(
+            dictionary.value<double>(DistanceUnitInfo.identifier)
+            );
```
liuloppan: Move end parentheis so it aligns with' _distanceUnit =...'

PR #1109, opened 2020-03-06
Added a way to customize renderable distance labels unit

stefanseibert:
This adds a way to customize the way how distances are shown on distance labels. It is possible to config the used unit (out of the existent openspace units, for example "gigameter"), and it is possible to set a custom descriptor. For SPVL for example we are using gigameter and the descriptor "million km" since we see that the average museum visitor has limited understanding to grasp "gigameter" and kilometer values are getting too big.
75972698-24b2d980-5ed4-11ea-9d91-1f0dd9103e9f

### DIFF
--- a/modules/base/rendering/renderablelabels.cpp
+++ b/modules/base/rendering/renderablelabels.cpp
@@ -784,4 +784,22 @@ float RenderableLabels::unit(int unit) const {
     }
 }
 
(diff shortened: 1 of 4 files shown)


Mik-pe:
Code looks good to me. 
Featurewise - since it's opt-in I feel this can be merged without any issues!

PR #1113, opened 2020-03-09
Added possibility to listen to latest type of touch interaction

stefanseibert:
This adds a way to listen to the latest touch interaction type. We want to log this information to study the usage of our application better.


stefanseibert:
Closed this, as it is not needed at the time being.

PR #1115, opened 2020-03-10
Feature/continuous pinch

Mik-pe:
This PR fixes some issues in the win32_touch hook (timestamps are now based on touchscreen driver rather than OpenSpace message pumping frequency).

We also now default the pinch gesture to use continuous zooming, meaning that if we pinch and stop our fingers, we will keep on zooming in. Letting go of one finger will let the pinch motion continue, and completely letting go will stop the zoom.

No values are tweaked to perfection as of this PR, but in general this improves the touch experience.

### DIFF
--- a/include/openspace/util/touch.h
+++ b/include/openspace/util/touch.h
@@ -80,6 +80,7 @@ public:
 private:
     //A deque of recorded inputs. Adding newer points to the front of the queue
     std::deque<TouchInput> _inputs;
+    TouchInput _firstInput;
 
     size_t _touchDeviceId;
     size_t _fingerId;
--- a/include/openspace/util/touch.h
+++ b/include/openspace/util/touch.h
@@ -46,7 +46,7 @@ struct TouchInput {
     float y;
     float dx = 0.f;         // movement in x direction since last touch input
     float dy = 0.f;         // movement in y direction since last touch input
-    double timestamp; // timestamp in seconds from global touch initialization
+    double timestamp;       // timestamp in seconds from global touch initialization
 };
 
 class TouchInputHolder {
@@ -56,6 +56,7 @@ public:
     // tryAddInput:
     // Succeeds upon a different input than last.
     // Fails upon a too similar input as last.
+    // Updates time for the last input if same position.
     bool tryAddInput(TouchInput input);
     void clearInputs();
 
@@ -72,6 +73,7 @@ public:
     double gestureTime() const;
 
     size_t numInputs() const;
+    const TouchInput& firstInput() const;
     const TouchInput& latestInput() const;
     const std::deque<TouchInput>& peekInputs() const;
 
--- a/include/openspace/util/touch.h
+++ b/include/openspace/util/touch.h
@@ -32,6 +32,9 @@
 
 namespace openspace {
 
+// The TouchInput represents a single finger/device-input at a specific point in time.
+// the fingerId and touchDeviceId coupled with the timestamp allows this to be compared
+// with other TouchInputs in order to calculate gesture-like behaviour.
 struct TouchInput {
     TouchInput(size_t touchDeviceId, size_t fingerId, float x, float y, double timestamp);
     glm::vec2 screenCoordinates(glm::vec2 resolution) const;
@@ -49,17 +52,22 @@ struct TouchInput {
     double timestamp;       // timestamp in seconds from global touch initialization
 };
 
+// The TouchInputHolder holds one or many TouchInputs, in order to track the history of 
+// the finger/input device
 class TouchInputHolder {
 public:
     TouchInputHolder(TouchInput input);
 
-    // tryAddInput:
     // Succeeds upon a different input than last.
     // Fails upon a too similar input as last.
     // Updates time for the last input if same position.
     bool tryAddInput(TouchInput input);
+    
     void clearInputs();
 
+    // Checks whether or not this Holder actually holds a specific input (based on IDs)
+    // Succeeds when `input` is held by this Holder
+    // Fails if `input` is not held by this Holder
     bool holdsInput(const TouchInput &input) const;
 
     size_t touchDeviceId() const;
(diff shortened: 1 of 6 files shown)


include/openspace/util/touch.h
```diff
@@ -72,12 +73,14 @@ class TouchInputHolder {
     double gestureTime() const;
 
     size_t numInputs() const;
+    const TouchInput& firstInput() const;
```
liuloppan: Why is it needed to keep track of the firstInput?

modules/touch/src/touchinteraction.cpp
```diff
@@ -829,54 +853,31 @@ void TouchInteraction::computeVelocities(const std::vector<TouchInputHolder>& li
...
-                        static_cast<float>(_zoomSensitivityExponential)
-                    );
-                }
-            }
-            else {
-                zoomFactor = 1.0;
-            }
-            _vel.zoom = zoomFactor * _zoomSensitivityProportionalDist *
-                         std::max(_touchScreenSize.value() * 0.1, 1.0);
+            _constTimeDecayCoeff.zoom = 1.0;
```
liuloppan: _constTimeDecayCoeff is this really a needed coeff if it is always set to 1?

modules/touch/src/win32_touch.cpp
```diff
@@ -213,8 +218,10 @@ Win32TouchHook::Win32TouchHook(void* nativeWindow)
 
     if (!gStarted) {
         gStarted = true;
-        gStartTime = std::chrono::duration_cast<std::chrono::microseconds>(
-            std::chrono::high_resolution_clock::now().time_since_epoch()
+        LARGE_INTEGER count;
```
liuloppan: This seems like a smart thing, but what makes it better than the previous solution with high-resolution-clock.time-since-epoch? Is it something we want to incorporate in more places where we currently use high-res-clock?

include/openspace/util/touch.h
Mik-pe: For continuous touch I basically measure the distance between the current distance from the fingers to their centroid point against the distance from the first ever pinch-input against the first ever centroid between those two fingers, if that makes sense.

The difference between current and first distance from centroid is what gives the current zoom velocity.

modules/touch/src/touchinteraction.cpp
Mik-pe: For continuous zoom I set this 'coeff' to 1, since i don't want the zoom to have any falloff while stepping the position. Once the user releases both fingers, the coeff is set to whatever is specified based on the current zoom velocity in the `resetAfterInput` function

modules/touch/src/win32_touch.cpp
Mik-pe: Good question!

std::chrono might use the built-in windows call under-the-hood, but for this specific windows touch hook we get the value for each touch from the touch display driver as a windows-supplied time, meaning that if we want to measure against the time received from the touch input I felt like it would be best to get the "startTime" from the windows API as well, so they match.

I don't think this should spread to any other places, it might be relevant if we want to provide our own timer classes and not rely on std::chrono, but I feel like that's an issue that needs another problem than this one.

Mik-pe: Looking at https://github.com/microsoft/STL/blob/master/stl/inc/chrono it seems like microsoft STL uses basically the exact same code, so it shouldn't matter, want me to change it back?

liuloppan: Then I would suggest to change it back to keep it more consistent with other parts of the app that uses high_resolution_clock.

modules/touch/src/touchinteraction.cpp
liuloppan: Wouldn't it be better to not include it in the calulations in this case instead of setting and resetting it back and forth? But if that is too complicated this solution also works.

include/openspace/util/touch.h
liuloppan: Ok so the speed of the initial pinch is the speed that will be held?

Mik-pe: Yeah, basically:
- initial touch with 2 fingers: save the location of these two fingers and give 0 zoom-speed since we haven't moved
- move out from center with both fingers: increase zoom speed outwards based on finger-distance travelled from initial touch
- move back to initial touch location: decrease zoom once more to 0 speed
- move in towards center with both fingers: increase zoom speed inwards based on finger-distance travelled from intitial touch

In the future I would like to use finger_distance_travelled in millimeters or such, so that we can have some scaling factor based on display size that controls all of this.

modules/touch/src/touchinteraction.cpp
Mik-pe: So, the touch-stuff has a `step` function which steps the camera with delta_time each frame based on velocities saved in some structure (orbit, zoom, roll, pan velocities). It also calls a `decelerate` function with this same delta_time. 

In this case, as a "pinch-gesture" happens, we don't want the zoom velocity to change until we release our fingers. To overcome this I set the decay-coeff of `zoom` to 1, which basically says "don't decelerate this zoom-field when you're decelerating all the velocity-stuff".

All the other velocities(such as orbit and potentially panning?) should still be decelerated, so that we get some friction-motion going on while spinning around, so I don't touch the other coefficients. 

When resetting the input (which happens when the user removes all fingers from the screen) we then set the decay-coeff for zoom, which tells the decelerate function to start decelerating the zoom-velocity. So I feel like it's the most straight-forward solution

liuloppan: Ok fair enough, That is a good explanation. :)

include/openspace/util/touch.h
```diff
@@ -56,6 +56,7 @@ class TouchInputHolder {
     // tryAddInput:
     // Succeeds upon a different input than last.
     // Fails upon a too similar input as last.
+    // Updates time for the last input if same position.
```
GPayne: I would prefer to see more comments like these in this file for the TouchInput and TouchInputHolder classes. Not every method needs an explanation, but it would help for some like holdsInput().

Mik-pe: This was a good idea, I added some explanation to the TouchInput and the TouchInputHolder as well to differentiate them.

PR #1117, opened 2020-03-11
Feature/spvl continuous pinch

Mik-pe:
This enables continuous pinch for the project/spvl branch. Should basically be the same as https://github.com/OpenSpace/OpenSpace/pull/1115

### DIFF
--- a/modules/touch/src/touchinteraction.cpp
+++ b/modules/touch/src/touchinteraction.cpp
@@ -118,8 +118,10 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo PinchZoomFactorInfo = {
-            "PinchZoomFactor", "Scaling distance travelled on pinch",
-            ""  // @TODO Missing documentation
+            "PinchZoomFactor",
+            "Scaling distance travelled on pinch",
+            "This value is used to reduce the amount of pinching needed. A linear"
+            "kind of sensitivity that will alter the pinch-zoom speed."
     };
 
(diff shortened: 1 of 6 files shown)


modules/touch/src/touchinteraction.cpp
```diff
@@ -1103,11 +1104,30 @@ void TouchInteraction::step(double dt) {
...
             }
+            const double currentPosDistance = length(centerToCamera);
 
             //Apply the velocity to update camera position
-            glm::dvec3 zoomDistanceIncrement = directionToCenter * _vel.zoom * dt;
+            double zoomVelocity = _vel.zoom;
+            if (!directTouch) {
+                const double distanceFromSurface =
+                    length(currentPosDistance) - anchor->boundingSphere();
+                if (distanceFromSurface > 0.1) {
```
stefanseibert: Are we never getting closer then 0.1? Because if that check is only to avoid division by zero wouldnt it be sufficient to do > 0.0? Just thinking if this might cause situations where velocity is smooth but then has a jump to the value of 1 if we get really close.

Mik-pe: This code was here before so I can't speak for it directly, some thoughts:

The distanceFromSurface is only when we are not in "directTouch"-mode, so this is basically an always-true if statement. I guess this is meant for some case where the user would travel really really fast or hit a framedrop while going close to the planet..? Do you think it's safe to remove it?

stefanseibert: I see. I think it is fine then to keep it as is.

PR #1118, opened 2020-03-11
cef touch events for spvl project branch

Mik-pe:
Enabled cef touch events instead of injected mouse events for the CEF-part. 
Should be the same as https://github.com/OpenSpace/OpenSpace/pull/1087

### DIFF
--- a/modules/webbrowser/src/eventhandler.cpp
+++ b/modules/webbrowser/src/eventhandler.cpp
@@ -208,15 +208,11 @@ void EventHandler::initialize() {
             }
 
             if (_validTouchStates.empty()) {
-                _mousePosition.x = windowPos.x;
-                _mousePosition.y = windowPos.y;
-                _leftButton.down = true;
-                _browserInstance->sendMouseClickEvent(
-                    mouseEvent(),
-                    MBT_LEFT,
-                    false,
-                    BrowserInstance::SingleClick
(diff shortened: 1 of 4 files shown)


modules/webbrowser/src/eventhandler.cpp
```diff
@@ -438,6 +463,23 @@ CefMouseEvent EventHandler::mouseEvent(KeyModifier mods) {
...
+CefTouchEvent EventHandler::touchEvent(const TouchInput& input,
+    const cef_touch_event_type_t eventType) const
+{
+    const glm::vec2 windowPos = input.currentWindowCoordinates();
+    CefTouchEvent event = {};
+    event.id = static_cast<int>(input.fingerId);
+    event.x = windowPos.x;
+    event.y = windowPos.y;
+    event.type = eventType;
+    //TODO: We should probably use key mods for touch as well:
```
stefanseibert: Even if they are not used right now wouldnt it make more sense to directly add them? Seems like all the mouseEvents mostly do not use them either but having KeyModifier mods = KeyModifier::NoModifier as default parameter and then event.modifiers |= static_cast<uint32_t>(mapToCefModifiers(mods)); in case someone later on wants to add functionality around it?

Mik-pe: Yeah this sounds like a good suggestion

PR #1123, opened 2020-03-21
Feature/asteroids

GPayne:
- Added new asteroid/comet assets.
- Combined satellite and sbdb renderable into a base class.
- Better error handling for data files, ignores individual problem lines while keeping remaining data.
- Implemented upperLimit property in order to limit very large sbdb data sets.
- Removed 3 data sets that are incompatible with kepler orbits.


GPayne:
Recommend using this github compare link:
https://github.com/OpenSpace/OpenSpace/compare/4afee46..bdf859c
for a "squashed" diff of all changes in this PR. Here's a basic summary of this diff:
1. New asteroids.scene file, pretty basic.
2. There's a new .asset file for each asteroid/comet data set, also very basic.
3. Modification to sssb_shared.asset for GUI-friendly name for each data set.
4. New abstract base class in renderableorbitalkepler.cpp. Basically lots of stuff removed from renderablesatellites.cpp and moved to here. Now, renderablesatellites and renderablesmallbody are derived classes and only have their functionality unique to the orbit type.

data/assets/scene/solarsystem/sssb/sssb_shared.asset
```diff
@@ -27,10 +27,11 @@ local createSssbGroupObject = function(filename, sssbFolder, trailColor)
             Path = sssbFolder.."/"..filename,
             Segments = 200,
             Color = trailColor,
-            Fade = 0.5,
+            TrailFade = 0.5,
         },
         GUI = {
-            Path = "/Solar System/Small Solar System Body"
+            Path = "/Solar System/Small Solar System Body",
```
micahnyc:
```suggestion
            Path = "/Solar System/Small Bodies",
```
just for menu ease

data/assets/scene/solarsystem/sssb/sssb_shared.asset
```diff
@@ -27,10 +27,11 @@ local createSssbGroupObject = function(filename, sssbFolder, trailColor)
...
+			Name = guiName
```
micahnyc:
```suggestion
        Name = guiName
```

micahnyc:
```suggestion
            Name = guiName
```

data/assets/scene/solarsystem/sssb/pha.asset
```diff
@@ -2,7 +2,9 @@ local assetHelper = asset.require('util/asset_helper')
 local sharedSssb = asset.require('./sssb_shared')
 
 local filepath = sharedSssb.downloadSssbDatabaseFile(asset, 'pha', 'sssb_data_pha')
-local object = sharedSssb.createSssbGroupObject("sssb_data_pha.csv", filepath, { 0.75, 0.2, 0.2 })
+local object = sharedSssb.createSssbGroupObject('sssb_data_pha.csv', "Potentially Hazardous Asteroids", filepath, { 0.75, 0.2, 0.2 })
```
micahnyc:
```suggestion
local object = sharedSssb.createSssbGroupObject('sssb_data_pha.csv', "Potentially Hazardous Asteroid", filepath, { 0.75, 0.2, 0.2 })
```
just making plurality consistant

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -0,0 +1,561 @@
...
+        int year;
+        int dayOfYear;
+        bool operator<(const LeapSecond& rhs) const {
+            return std::tie(year, dayOfYear) < std::tie(rhs.year, rhs.dayOfYear);
+        }
+    };
+
+    const LeapSecond Epoch = { 2000, 1 };
+
+    // List taken from: https://www.ietf.org/timezones/data/leap-seconds.list
```
micahnyc: This is making me thing there is probably something interesting we can do with cmake to download the list and generate it? overkill for right now, but a thought for the future

modules/space/rendering/renderablesatellites.cpp
```diff
@@ -76,197 +75,16 @@ namespace {
         "Trail Fade",
         "This value determines how fast the trail fades and is an appearance property. "
     };
-    
-    constexpr const char* KeyFile = "Path";
-    constexpr const char* KeyLineNum = "LineNumber";
+    static const openspace::properties::Property::PropertyInfo UpperLimitInfo = {
```
micahnyc: is this needed since it's in the parent class?
 same for trail fade?

modules/space/rendering/renderablesmallbody.cpp
```diff
@@ -67,131 +66,44 @@ namespace {
...
-    };
     constexpr openspace::properties::Property::PropertyInfo LineColorInfo = {
         "Color",
         "Color",
         "This value determines the RGB main color for the lines and points of the trail."
     };
-    
-    constexpr const char* KeyFile = "Path";
-    constexpr const char* KeyLineNum = "LineNumber";
+    constexpr openspace::properties::Property::PropertyInfo TrailFadeInfo = {
```
micahnyc: same q as above

openspace.cfg
```diff
@@ -45,7 +45,8 @@ SGCTConfig = sgct.config.single{}
 
 -- Sets the scene that is to be loaded by OpenSpace. A scene file is a description
 -- of all entities that will be visible during an instance of OpenSpace
-Asset = "default"
+Asset = "asteroids"
```
micahnyc:
```suggestion
Asset = "default"
```

openspace.cfg
```diff
@@ -45,7 +45,8 @@ SGCTConfig = sgct.config.single{}
...
+-- Asset = "default"
```
micahnyc:
```suggestion
Asset = "default"
```

data/assets/scene/solarsystem/sssb/pha.asset
GPayne: Good catch, but actually I decided to do the opposite and make the other names plural as they should be. I had auto-generated the gui names based on their filenames, but now it's better.

openspace.cfg
GPayne: Combining this step with next code change

openspace.cfg
GPayne: Done

modules/space/rendering/renderableorbitalkepler.cpp
GPayne: Yes, a good idea for a URL sync in the future

modules/space/rendering/renderablesatellites.cpp
GPayne: All of the PropertyInfo declarations are actually not in the parent class. Throughout OpenSpace these are declared in a separate namespace rather than in the class, so I just stayed with that convention. The downside of this is that these are repeated in base classes.

However, I deleted this particular `upperLimitInfo` variable just now because I had already removed the upper limit from satellite (only smallbody uses it because we don't have files with huge numbers of satellites) but forgot to remove it here.

modules/space/rendering/renderablesmallbody.cpp
GPayne: See related answer in renderablesatellites.cpp

alexanderbock:
There are two astroids in the "Chiron-type Comets" that are coming from outside?  Is that correct or a problem in the dataset?
OpenSpace_000000


The Inner Main Astroid Belt looks funky with this torus.. Some question as above?
OpenSpace_000001


The transneptunian object astroids also look a bit weird:

OpenSpace_000002

GPayne:
Screenshots with new algorithm that varies number of orbital segments based on orbit eccentricity.

Same views as above for Chiron comets:
OpenSpace_000000

...and Trans-Neptunian asteroids:
OpenSpace_000001

PR #1158, opened 2020-05-19
Feature/profiles

GPayne:
Feature for saving current settings to a new profile.


data/assets/default.profile
```diff
@@ -3,10 +3,10 @@
 
 #Asset
 scene/solarsystem/planets/earth/earth	required
-scene/solarsystem/planets/earth/satellites/satellites   required
+scene/solarsystem/planets/earth/satellites/satellites	required
 
 #Property
-setPropertyValue  {earth_satellites}.Renderable.Enabled false
+setPropertyValue	{earth_satellites}.Renderable.Enabled	false
```
GPayne, line 9: This is a tab-delimited file. If these are replaced with spaces, then the parsing won't work.

openspace.cfg
```diff
@@ -64,7 +64,7 @@ Asset = "default"
 -- Sets the profile that should be loaded by OpenSpace.  Profiles are going to replace
 -- assets in a future versions and shouldn't be used at the same time as the 'Asset'
 -- setting above
--- Profile = "default"
+Profile = "default"
```
GPayne, line 67: This will make OpenSpace master run in profile mode by default!

data/assets/default.profile
alexanderbock: They were space-delimited in the repository and I replaced them with tabs on my end

openspace.cfg
alexanderbock: Yea, I left there for now as I had to quickly commit everything when I left for the office this morning.  Discussion pending if we want to make it the default

alexanderbock:
Punting the review back to you @GPayne

PR #1190, opened 2020-06-12
Fix Linux Fixes

TomSchober:
@jccosta I think this just ended up here by accident.  Was unable to compile without removing the superfluous #endif


jccosta:
Yes, you are right.
I fixed in a different commit but didn't push it yet.
Thanks!

PR #1192, opened 2020-06-13
Feature/mars 2020

micahnyc:
Adds model and trail for m2020. Created mars scene which includes insight as well.


micahnyc:
sync files: https://drive.google.com/file/d/15axs0V2a-2SKb7I1RWT9nfFW_Kvkhn7z/view?usp=sharing

PR #1215, opened 2020-06-24
Removed fade out distance for labels and updated asset files with labels

GPayne:
My solution was to remove the fade-out property, which seems redundant. All planets/moon asset files with labels were updated with settings that work for the diameter & resolution of each.


liuloppan:
Imo the fadeoutdistance is not redundant. The use case I see is if you want to see something from far away with a label and you have it fade in at a certain distance and as you approach the object/3D model the label will fade out as to not obstruct the view and you have the object in clear sight. However for this case, the fadeoutdistance is always smaller than the fadeindistance. Not sure that is how it is implemented.

GPayne:
Will re-do solution to this

PR #1228, opened 2020-06-29
Fixed issue with transparent definition in renderbin.

jccosta:
Fixed wrong definition for fully transparent renderable.


WeirdRubberDuck:
Who reviews this pull request? Would like this change asap :D 

Also, with this change I assume that tests like the following one in `dumesh_fs.glsl`:
```
    if (alphaValue == 0.0) {
        discard;
    }
```
is no longer needed? At least if the renderbin is set from the opacity value

alexanderbock:
@jccosta Are they needed or does the GBuffer fragments need to be handled differently when fully transparent?

alexanderbock:
And I thought I had already merged that PR into master yesterday? I’m confused

jccosta:
The tests are still needed. When you discard a fragment, you avoid the depth comparison with the current fragment in the main buffer and, in our case, writing to multiple buffers at the same time, saving precious milliseconds. :-)

jccosta:
@alexanderbock, the PR you merged into master was related to the new bin; this one is related to the issue with transparent objects still being seen (black pixels on top of the Milk Way) when the alpha value is equal 0.
The correction made in the PR you merged had, by my mistake, this correction too.
So @WeirdRubberDuck, the correction is already in master, and the optimization should still be there, in my opinion.
I think we can close this one then.

PR #1258, opened 2020-07-15
Issue/1238

### LINKED ISSUE #1257: Better handling of ComputeScreenSpaceData for Scene graph nodes
Right now all Scene graph nodes get a handful of properties to handle screen space positioning.
The screen space positioning is currently only used for the info markers in the touch profile.

The problem is that they clutter the gui for all nodes while only being used for a very small thing.
Suggestion: Have them invisible to the GUI by default. Only "ComputeScreenSpaceData" should be visible in GUI. However the other properties ("ScreenSpacePosition", "ScreenVisibility" etc)  still need to be exposed to the web-gui for touch, or else the markers won't work.

liuloppan:
To further improve the possibilites of automatic flight there has been a request to be able to have a bounding shape on scene graph nodes, where they previously where limited to renderable types. 

This PR contains:
* A boundingSphere-property being added to scenegraph nodes. It will be 0 by default. If it has a renderable it will compare the bounding sphere of the renderable and pick the largest out of the two bounding spheres as the bounding sphere of the node.  The idea here is that if a renderable child is attached to the parent, it will automatically respect that bounding sphere without additional changes having to be applied. Unless explicitly provided in the dictionary, this boundingsphere property will be hidden in the gui.
* Added a small bounding sphere to ISS node
* Updated documentation and hiding of a number of screen space properties that are currently only used for touch gui markers, Also closes #1257 

Questions: 
* Can it lead to confusion that a scene graph node can have a boundingSphere both on its node and on its renderable child?
* Are there issues with the automatically picking the largest of the two bounding spheres?
* Are there issues having a duplicate property existing on both the renderable and the scenegraph node?

### DIFF
--- a/src/scene/scenegraphnode.cpp
+++ b/src/scene/scenegraphnode.cpp
@@ -86,13 +86,14 @@ namespace {
         "VisibilityDistance",
         "" // @TODO Missing documentation
     };
-    constexpr const char* KeyFixedBoundingSphere = "FixedBoundingSphere";
 
     constexpr openspace::properties::Property::PropertyInfo BoundingSphereInfo = {
         "BoundingSphere",
         "Bounding Sphere",
         "The bounding sphere of the scene graph node. This can be the "
-        "bounding sphere of a renderable or a fixed bounding sphere. "
+        "bounding sphere of an attached renderable or directly specified to the node. "
+        "If there is a boundingsphere on both the renderable and the ndoe, the largest number will be picked. ",
+        openspace::properties::Property::Visibility::Hidden
     };
 
     constexpr openspace::properties::Property::PropertyInfo GuiPathInfo = {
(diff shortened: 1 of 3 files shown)


liuloppan:
Also closes #1257

src/scene/scenegraphnode.cpp
```diff
@@ -56,43 +56,50 @@ namespace {
...
     };
-    constexpr const char* KeyFixedBoundingSphere = "FixedBoundingSphere";
 
     constexpr openspace::properties::Property::PropertyInfo BoundingSphereInfo = {
         "BoundingSphere",
         "Bounding Sphere",
         "The bounding sphere of the scene graph node. This can be the "
-        "bounding sphere of a renderable or a fixed bounding sphere. "
+        "bounding sphere of an attached renderable or directly specified to the node. "
+        "If there is a boundingsphere on both the renderable and the ndoe, the largest number will be picked. ",
```
WeirdRubberDuck, line 101: Misspelled 'node'

src/scene/scenegraphnode.cpp
```diff
@@ -246,6 +258,11 @@ std::unique_ptr<SceneGraphNode> SceneGraphNode::createFromDictionary(
         LDEBUG(fmt::format(
             "Successfully created renderable for '{}'", result->identifier()
         ));
+
+        // If the renderable child has a bounding sphere that is larger, we allow it to override
+        result->_boundingSphere.setValue(std::max(result->_renderable->boundingSphere(),
+            result->_boundingSphere.value()
```
WeirdRubberDuck, line 264: I do think we should present a warning if we override a value specified by the user in the dictionary

liuloppan: Added warning :)

Issue #997, opened 2019-10-24
Assetize globebrowsing files


micahnyc:
We would like the extra globebrowsing layers that are added via the customization asset (downloaded from osp.com) to be added as assets the way the apollo globebrowsing is added.

micahnyc:
Some ones that Carter want's to add are here: 

Tycho Central Peak:
On Moon Trek: search "tycho"
LRO LROC DEM, Tycho Crater
LRO LROC Image Mosaic, Tycho Crater

Surveyor 7, which landed on the northwestern flank of Tycho:
http://wms.lroc.asu.edu/lroc/view_rdr/NAC_DTM_SURVEYOR7   array of NAC's, single DEM

alexanderbock:
Potential connection with the license merging;  how do we handle different licenses for different image layers of the `.info` files?

PR #1259, opened 2020-07-16
issue/997 - Assetize globe layers (sun, planets, dwarf planets and moons)

WeirdRubberDuck:
All existing layers have been moved to separate files and are added to the node assets in the `base.asset` file instead of directly in the asset for the globe's scene graph node. The layers that we provide are grouped together in a `default_layers.asset` file. The layers are added and enabled in an `add_default_layers.asset` file, possibly also including the name of the globe, for clarity. 

The layers are added into a `layers` folder for each globe asset, with a sub-folder for each layer group. Globes that only have a single color layer also follow this file structure, for consistency and to make it clearer where to add new layers.

Lots of changed files... But they follow the same structure for all globes. **I would really appreciate to get an opinion on the structure of adding the layers and the general files structure.** For example, for Earth the relevant files are: `add_earth_default_layers.asset` and `layers/default_layers.asset`. 

Some questions:
- Does this solution add to much complexity when adding layers? 
- Does the file structure make sense, for example when using the launcher to create a new profile?


WeirdRubberDuck:
OBS! Should probably add the Layers that Carter wanted as well (see [issue](https://github.com/OpenSpace/OpenSpace/issues/997))

GPayne:
I like this approach overall. It does add some complexity, but I don't think too much. The downside is more files and directories, but the benefit is improved readability of each individual asset file (some assets were getting too big).

I prefer some name changes for simplicity:
- Changing `<globe>/layers/default_layers.asset` to `<globe>/layers/default.asset`
- Removing verbs from asset files; changing `<globe>/add_default_layers.asset` to `<globe>/default_layers.asset`.

Also, does this affect how layers appear under a globe in the GUI (e.g. _ColorLayers_, _HeightLayers_, _NightLayers_, etc.? It does look like `util/layer_helper.asset` takes care of this, but I would verify that it stays that way in the GUI.

alexanderbock:
I agree with the removal of the verbs and I'll add that to the PR before merging it.

@GPayne   No, the GUI part will be completely unaffected

PR #1261, opened 2020-07-22
issue/1260 correct capitalization of asset file references

johnriedel:
This fixes an error encountered during initial run of bin/OpenSpace on Ubuntu 20.04.


alexanderbock:
Thanks for that!

PR #1276, opened 2020-08-14
adding option to not reset velocity on anchor change

micahnyc:


micahnyc:
In conjunction with a gui update, will allow for changing of focus during a presentation without stopping the camera.

micahnyc:
this commit also updates the flightcontrollertopic to support anchor and aim as it was developed with focus only.

PR #1280, opened 2020-08-17
Feature/remove asset request

alexanderbock:


alexanderbock:
Removes the `request` function on the `asset` object
Replace all `request` in existing assets with `require`

PR #1282, opened 2020-08-18
Issue/1174 - step between delta time steps using GUI or keybindings

WeirdRubberDuck:
Add a list of delta time steps to TimeManager and send next and previous delta time steps via the server to the GUI. The delta time steps list can be set from a separate section in a profile, e.g. like this: 
```
#DeltaTimes
1
60
21600
604800
```
Negative versions of each value will be added as well and the list will be sorted to be in increasing order. 

The next/previous delta time step is determined as the first value in the list that is larger/smaller than the current delta time value. 

This PR is connected to another PR in the GUI frontend repo (OpenSpace/OpenSpace-WebGuiFrontend#18), that includes buttons for stepping between delta time steps in the TimePicker GUI: 
GUI_marked_change

Some remaining work:
- Documentation for profiles should include delta time steps as well
- Add relevant delta time steps to all our existing profiles


liuloppan:
Found one issue: 

1.Use "right" and "left" keyboards to change deltatimestep (works)
2. Change"Display Unit"
3. Use "right" and "left" keyboards to change deltatimestep (does not work)

If I press pause and play in the GUI it works again.

WeirdRubberDuck:
> Found one issue:
> 
> 1.Use "right" and "left" keyboards to change deltatimestep (works)
> 2. Change"Display Unit"
> 3. Use "right" and "left" keyboards to change deltatimestep (does not work)
> 
> If I press pause and play in the GUI it works again.

Good find! This seems to be a problem with the "Display units" select not stopping to listen to keyboard events once an item has been selected. No keybinding works immediately after selecting an item in the list... Not until you click somewhere in OpenSpace

I'll make an issue for this and the other bug with the Display units that we talked about and have a look at it

WeirdRubberDuck:
Added a couple of test cases. If this all looks good to both of you @liuloppan and @GPayne then maybe this PR is ready for merge?

micahnyc:
wondering if the delta times need their own topic or if they should just be part of the timetopic included in the sendFullTimeData function.

WeirdRubberDuck:
> wondering if the delta times need their own topic or if they should just be part of the timetopic included in the sendFullTimeData function.

Hm, a bit unsure about this. The initial idea of creating a separate topic was that the delta time steps don't need to be updated as frequently as the time. However, since in the end only the next and previous time step is being sent in the topic (rather than the whole list, as I initially planned), maybe it would not be a problem to instead just include them in the timetopic... 

I guess it would be possible to move the delta time callbacks into the time topic, and keep the separate "send" function for the next and previous delta time, instead of including it with the `sendFullTimeData`. That way we don't send the time steps every 50 milliseconds but still reduce the number of files and keep all the time data in the same topic. Does that sound like a good solution, @micahnyc ?

WeirdRubberDuck:
> > wondering if the delta times need their own topic or if they should just be part of the timetopic included in the sendFullTimeData function.
> 
> Hm, a bit unsure about this. The initial idea of creating a separate topic was that the delta time steps don't need to be updated as frequently as the time. However, since in the end only the next and previous time step is being sent in the topic (rather than the whole list, as I initially planned), maybe it would not be a problem to instead just include them in the timetopic...
> 
> I guess it would be possible to move the delta time callbacks into the time topic and keep the separate "send" function for the next and previous delta time, instead of including it with the `sendFullTimeData`. That way we don't send the time steps every 50 milliseconds but still reduce the number of files and keep all the time data in the same topic. Does that sound like a good solution, @micahnyc ?

Realized that my interpretation of how often the `sendFullTimeData` is called was wrong, after looking at calls in Tracy. However, the delta times does not need to be updated as often as the the "full time" data (that's sent every time the target delta time changes). 

Although, calling sendJson twice sometimes might be more expensive then including the delta times in the full data json... Gonna use Tracy to try to figure out what's most efficient :)

PR #1287, opened 2020-08-24
Feature/general performance

alexanderbock:
Pull request for the "general performance" branch that does a lot of changes in a large number of places to improve the overall performance.
The test case was the standard view when starting the default scene


include/openspace/scene/profile.h
```diff
@@ -121,7 +121,7 @@ class Profile {
      * and all of the property & asset changes that were made since startup.
      */
     void saveCurrentSettingsToProfile(const properties::PropertyOwner& rootOwner,
-        const std::string& currentTime,
+        std::string currentTime,
```
GPayne, line 124: Did you intend to change this to a std::string_view?

alexanderbock: No, it can be a `std::string` because it's value is saved in the `Time t` parameter.  The const reference before would have needed a copy in some situations which can be prevented now:

A (before):
```
std::string t = "timestring";
profile.saveCurrentSettingsToProfile(owner, t);   // passed by reference

//
t.time = currentTime; /// < Makes a copy of the string
```

B (now):
```
std::string t = "timestring";
profile.saveCurrentSettingsToProfile(owner, std::move(t));   // moving into the parameter, not copying

//
t.time = std::move(currentTime); /// < Moving and also not making a copy
```

modules/globebrowsing/src/timequantizer.cpp
```diff
@@ -215,36 +209,35 @@ void DateTime::incrementOnce(int value, char unit) {
     bool inBounds = true;
     switch (unit) {
         case 'm':
-            if (singleIncrement(_minute, value, 0, 59))
+            if (singleIncrement(_minute, value, 0, 59)) {
                 break;
-            // else fall-through if overflow...
-
+            }
+            //[[ fallthrough ]]
```
GPayne, line 215: Why add the comment to these lines? Without it compiler warnings about fallthrough should avoided.

src/rendering/renderengine.cpp
```diff
@@ -446,6 +449,22 @@ void RenderEngine::initialize() {
...
+
+    _versionString = OPENSPACE_VERSION_STRING_FULL;
+    if (global::versionChecker.hasLatestVersionInfo()) {
+        VersionChecker::SemanticVersion latest = global::versionChecker.latestVersion();
+
+        VersionChecker::SemanticVersion current{
+            OPENSPACE_VERSION_MAJOR,
+            OPENSPACE_VERSION_MINOR,
+            OPENSPACE_VERSION_PATCH
+};
```
GPayne, line 461: indent

src/rendering/renderengine.cpp
```diff
@@ -1456,18 +1475,23 @@ void RenderEngine::renderScreenLog() {
...
-            *_fontLog,
-            glm::vec2(
-                10 + 30 * _fontLog->pointSize(),
-                _fontLog->pointSize() * nr * 2 + fontRes.y * _verticalLogOffset
-            ),
-            fmt::format("({})", lvl),
-            color
-        );
+                }
+            }(it->level);
```
GPayne, line 1479: Is this lambda here for performance reasons?

modules/base/dashboard/dashboarditemdistance.cpp
```diff
@@ -440,18 +445,20 @@ void DashboardItemDistance::render(glm::vec2& penPosition) {
...
     penPosition.y -= _font->height();
-
-    RenderFont(
-        *_font,
-        penPosition,
-        fmt::format(
-            "Distance from {} to {}: {:f} {}",
-            sourceInfo.second, destinationInfo.second, dist.first, dist.second
-        )
+    std::fill(_buffer.begin(), _buffer.end(), 0);
```
alexanderbock, line 448: It would be good to have a limitation in place to keep the length of a scene graph node name/identifier to < 100 characters to be able to have a reasonable limit on the length of the formatted string

modules/base/dashboard/dashboarditemdistance.cpp
```diff
@@ -358,6 +359,8 @@ DashboardItemDistance::DashboardItemDistance(const ghoul::Dictionary& dictionary
     addProperty(_requestedUnit);
 
     _font = global::fontManager.font(_fontName, _fontSize);
+
+    _buffer.resize(256);
```
alexanderbock, line 363: This should probably become a std::array instead of a vector since we are not using the resizing?

src/rendering/renderengine.cpp
alexanderbock: I don't think it has any impact on the performance, but it allows us to make the `color` variable `const` and ensures that it is initialized.  IIFE ("iffy") is one of those JavaScript best-practices that i really like:
https://en.wikipedia.org/wiki/Immediately_invoked_function_expression

modules/globebrowsing/src/timequantizer.cpp
alexanderbock: Oopsie

PR #1289, opened 2020-08-25
Delta time steps server code cleanup

WeirdRubberDuck:
Don't have a separate server topic to delta times. Instead, include next and prev time step in `sendFullTimeData` in timetopic. 

It actually seems to be slightly "cheaper" to include the next and prev delta time steps in the same topic since sending json objects over the server is expensive... and the time update will happen anyways. :)

Also, send the delta time steps list over the server as well, when it is changed. In case anyone want to create a GUI using this information.


modules/server/include/topics/timetopic.h
```diff
@@ -41,16 +41,21 @@ class TimeTopic : public Topic {
 private:
     const int UnsetOnChangeHandle = -1;
 
+    nlohmann::json getNextPrevDeltaTimeStepJson();
```
alexanderbock, line 44: This can probably be marked as `const`?

modules/server/src/topics/timetopic.cpp
```diff
@@ -94,6 +99,37 @@ void TimeTopic::handleJson(const nlohmann::json& json) {
             sendFullTimeData();
         }
     });
+
+    _deltaTimeStepsCallbackHandle = global::timeManager.addDeltaTimeStepsChangeCallback(
+        [this]() {
+            const std::vector<double> steps = global::timeManager.deltaTimeSteps();
```
alexanderbock, line 105: I'd say either `std::vector<double>` or `const std::vector<double>&`  If using the first, there will be no copy since the optimizer will get rid of it probably, the second is a bit more verbose and makes it clear that it's constant

PR #1300, opened 2020-09-16
Thesis/2018/exoplanets

WeirdRubberDuck:
Branch with exoplanet systems, where the host star is rendered as a RenderableGlobe with a color matching its BV-index. Also, the uncertainty of the major axis of the planet orbits is visualized using a colored disk. 

For now, to add a planet system the name of the hosting must be known and passet to the `openspace.exoplanets.addExoplanetSystem(...)` function. Later on a GUI for adding exoplanet systems will be added.


micahnyc:
I believe this is the new source for exoplanet data: https://exoplanetarchive.ipac.caltech.edu/index.html

The csv table download is here: https://exoplanetarchive.ipac.caltech.edu/cgi-bin/TblView/nph-tblView?app=ExoTbls&config=PS

but they also have an API: https://exoplanetarchive.ipac.caltech.edu/docs/program_interfaces.html which could potentially be connected with our task or with urlsync.

Example api request: https://exoplanetarchive.ipac.caltech.edu/cgi-bin/nstedAPI/nph-nstedAPI?table=exoplanets

alexanderbock (changes requested):
I think it looks good to me; though I have two suggestions:

1. Adding a simple Lua function that prints the names of all Exoplanet systems that could be added would make a first step towards usability
2. When I do `openspace.exoplanets.addExoplanetSystem("Kepler-119")`, I get
```
Lua (Error)     Error loading script '': [string "openspace.addSceneGraphNode({Identifier = 'Ke..."]:1: unfinished string near <eof>
```

WeirdRubberDuck:
@alexanderbock 

1. Adding a Lua function that prints the names of the existing exoplanets sounds like a great idea! Will do (Y) 
2. that <eof> error is caused by the `string_view` bug we discussed yesterday.

Thanks for the coding style input! 

About the updated data: I was not aware of that there is a new source for the data until @micahnyc mentioned it yesterday. To me it makes sense to use the most updated data, since the research about exoplanets is highly ongoing. So before this PR is merged, maybe we should make sure to update the data :) do you agree, @alexanderbock ?

alexanderbock:
Jep; completely agree with that :)

modules/exoplanets/exoplanetsmodule.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <openspace/scene/scene.h>
+#include <openspace/util/factorymanager.h>
+#include <thread>
+#include <chrono>
+
+#include "exoplanetsmodule_lua.inl"
+
+namespace openspace {
+
+constexpr const char* _loggerCat = "exoplanets";
```
alexanderbock, line 43: We normally put these into an anonymous namespace to make sure that the symbol doesn't get exported:
```
namespace {
  constexpr const char* _loggerCat = "exoplanets";
} // namespace
```

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -0,0 +1,558 @@
...
+#include <glm/gtc/quaternion.hpp>
+#include <glm/gtx/quaternion.hpp>
+#include <glm/gtx/transform.hpp>
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
+namespace openspace::exoplanets::luascriptfunctions {
+
+constexpr const char* _loggerCat = "ExoplanetsModule";
```
alexanderbock, line 49: Same above with the anonymous namespace

modules/exoplanets/rendering/renderableorbitdisc.cpp
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    if (_textureIsDirty) {
+        loadTexture();
+        _textureIsDirty = false;
+    }
+}
+
+void RenderableOrbitDisc::loadTexture() {
+    if (_texturePath.value() != "") {
```
alexanderbock, line 261: `!_texturePath.value().empty()` would not have to create a temporary empty std::string for the comparison

modules/exoplanets/rendering/renderableorbitdisc.cpp
```diff
@@ -0,0 +1,325 @@
...
+            _textureFile = std::make_unique<ghoul::filesystem::File>(_texturePath);
+            _textureFile->setCallback(
+                [&](const ghoul::filesystem::File&) { _textureIsDirty = true; }
+            );
+        }
+    }
+}
+
+void RenderableOrbitDisc::createPlane() {
+    const GLfloat size = _size.value() * (1.0 + _eccentricity.value());
```
alexanderbock, line 284: This line should work without the `value()`  calls as well as the properties should automatically convert to their underlying type:

```
FloatProperty p(...);
float f = p;

p = 2.f;
```

modules/exoplanets/shaders/orbitdisc_fs.glsl
```diff
@@ -0,0 +1,114 @@
...
+uniform float opacity;
+uniform float eccentricity;
+uniform float semiMajorAxis;
+
+const float AstronomicalUnit = 149597870700.0; // m
+const float Epsilon = 0.0000001;
+
+// Compute semi minor axis from major axis, a, and eccentricity, e
+float semiMinorAxis(float a, float e) {
+    return a * sqrt(1.0 - pow(e, 2.0));
```
alexanderbock, line 42: `pow(e, 2.0) == e * e`

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+
+namespace {
+    constexpr const char* KeyInputCsv = "InputCSV";
+    constexpr const char* KeyInputSpeck = "InputSPECK";
+    constexpr const char* KeyOutputBin = "OutputBIN";
+    constexpr const char* KeyOutputLut = "OutputLUT";
+
+    constexpr const char* _loggerCat = "CsvToBinTask";
+
+    constexpr const char* TeffBvPath = "${SYNC}/http/exoplanets_data/1/teff_bv.txt";
```
alexanderbock, line 45: It might be better to require the TeffBvpath to be passed in via the dictionary here;  that would remove the hardcoded location of the `http/exoplanets_data` path.  If it is in the dictionary, the task specification can set it and it has explicit access to the location file

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    );
+
+    _inputCsvPath = absPath(dictionary.value<std::string>(KeyInputCsv));
+    _inputSpeckPath = absPath(dictionary.value<std::string>(KeyInputSpeck));
+    _outputBinPath = absPath(dictionary.value<std::string>(KeyOutputBin));
+    _outputLutPath = absPath(dictionary.value<std::string>(KeyOutputLut));
+}
+
+std::string ExoplanetsCsvToBinTask::description() {
+    return "Extract metadata from csv-file " + _inputCsvPath +
```
alexanderbock, line 64: Might be nicer to read as:
```
fmt::format("Extract metadata from csv-file '{}' and write as bin to '{}'", _inputCsvPath, _outputBinPath);
```

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+
+    std::ofstream binFile(_outputBinPath, std::ios::out | std::ios::binary);
+    std::ofstream lutFile(_outputLutPath);
+
+    int version = 1;
+    binFile.write(reinterpret_cast<char*>(&version), sizeof(int));
+
+    Exoplanet p;
+
+    std::string planetName;
```
alexanderbock, line 83: This variable is only needed further down, so it should be moved there

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    std::string component;
```
alexanderbock, line 84: Same here

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    std::string planetRow;
+    getline(csvFile, planetRow); // The first line, containing the data names
+
+    bool isKeplerObject = false;
```
alexanderbock, line 88: And here

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    int total = 0;
+    while (getline(csvFile, planetRow)) {
+        ++total;
+    }
+    csvFile.clear();
+    csvFile.seekg(0);
+    getline(csvFile, planetRow); // The first line, containing the data names
+    LINFOC("CSVTOBIN", fmt::format("Loading {} stars", total));
+
+    auto readFloatData = [](const std::string& data) -> float {
```
alexanderbock, line 98: If you want to be a bit C++-fancier, you can use `from_chars` from the `<utility> header:
```
int result;
auto [p, ec] = std::from_chars(data.begin(), data.end(), result);
if (ec == std::errc()) {
  return result;
}
else {
   return NAN;
}
```

alexanderbock: https://en.cppreference.com/w/cpp/language/structured_binding

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+}
+
+glm::vec3 ExoplanetsCsvToBinTask::starPosition(const std::string& starName) {
+    glm::vec3 position;
+    position[0] = NAN;
+    position[1] = NAN;
+    position[2] = NAN;
+    std::ifstream exoplanetsFile(_inputSpeckPath);
+    if (!exoplanetsFile) {
+        LERROR(fmt::format("Error opening file expl.speck."));
```
alexanderbock, line 548: We normally don't put '.' at the end of error messages

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    }
+
+    std::string line;
+    std::string data; // data
+    std::string name;
+    while (getline(exoplanetsFile, line)) {
+        bool shouldSkipLine = (
+            line.empty() || line[0] == '#' || line.substr(0, 7) == "datavar" ||
+            line.substr(0, 10) == "texturevar" || line.substr(0, 7) == "texture"
+            );
```
alexanderbock, line 558: One indentation too much

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    position[0] = NAN;
+    position[1] = NAN;
+    position[2] = NAN;
+    std::ifstream exoplanetsFile(_inputSpeckPath);
+    if (!exoplanetsFile) {
+        LERROR(fmt::format("Error opening file expl.speck."));
+    }
+
+    std::string line;
+    std::string data; // data
```
alexanderbock, line 552: These variables should be a bit further down where they are used

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+        std::istringstream linestream(line);
+        getline(linestream, data, '#');
+        getline(linestream, name);
+        name.erase(0, 1);
+
+        std::string coord;
+        if (name == starName) {
+            std::stringstream dataStream(data);
+            getline(dataStream, coord, ' ');
+            position[0] = std::stof(coord.c_str(), nullptr);
```
alexanderbock, line 573: See above

modules/exoplanets/tasks/exoplanetscsvtobintask.cpp
```diff
@@ -0,0 +1,677 @@
...
+    // Apply transformation matrix to pos
+    glm::dmat4 _transformationMatrix = glm::dmat4(1.0);
+    glm::vec3 transformedPosition = glm::vec3(
+        _transformationMatrix * glm::dvec4(position, 1.0)
+    );
+
+    return transformedPosition;
+}
+
+float ExoplanetsCsvToBinTask::bvFromTeff(const float teff) {
```
alexanderbock, line 591: There is no real reason to declare an in-parameter as `const`  there is no optimization that can happen because of it and it makes too strong of a guarantee for when we wan to change the function declaration later;  let's say we change this function and all of a sudden we want to constrain `teff`.  So we either would have to make a copy that is `clamp`ed, or we would need to change the function definition toe `float teff` which might break client code unnecessarily

alexanderbock (approved):

PR #1302, opened 2020-09-21
Issue/1285 - Add delta time steps to profiles and generate keybindings for number keys

### LINKED ISSUE #1285: Add relevant delta time steps for all existing profiles
Once PR #1282 is merged. 

The delta times currently used for the number keybindings is probably good for most profiles. Some profiles that will most likely require other time steps are `gaia` and `voyager`. 

This task could also involve redefining the number keybindings to match the list of delta time steps?

WeirdRubberDuck:
This PR adds:
- Delta time steps to all profiles, either the previous default ones or the ones that were added through keybindings in the profile, if any
- Generated keybindings to the number keys for up to 30 time steps, using the SHIFT and then CTRL keys. Negative versions are accessed using the ALT key. 

The `TimeManager` is responsible for creating and removing the keybindings when necessary. Note that other commands that are bound to the number keys will potentially be deleted when the keybinds for the delta times are reset (e.g. when the delta time steps are set during run time). later on, the keybinding system should probably be updated to not silently remove keybindings that were not intended to be removed. 

Closes #1285


WeirdRubberDuck (commented):
Note: The delta times in the profiles need to be adapted to the new json format before merge

data/profiles/apollo_sites.profile
```diff
@@ -33,3 +33,35 @@ Apollo11LemModel
 Apollo17LemModel
 Apollo11
 Apollo11LunarLander
+
+#DeltaTimes
```
alexanderbock, line 37: There is something awefully visually pretty when scrolling past the numbers

include/openspace/util/timemanager.h
```diff
@@ -25,22 +25,17 @@
 #ifndef __OPENSPACE_CORE___TIMEMANAGER___H__
 #define __OPENSPACE_CORE___TIMEMANAGER___H__
 
+#include <openspace/properties/propertyowner.h>
+#include <openspace/properties/scalar/floatproperty.h>
 #include <openspace/util/syncdata.h>
+#include <openspace/util/keys.h>
```
alexanderbock, line 31: `k` should be sorted before `s`

src/util/timemanager.cpp
```diff
@@ -412,6 +418,96 @@ void TimeManager::setDeltaTimeSteps(std::vector<double> deltaTimes) {
...
+        Key::Num6,
+        Key::Num7,
+        Key::Num8,
+        Key::Num9,
+        Key::Num0
+    };
+
+    // First, clear any previous keybinds
+    for (const KeyWithModifier& kb : _deltaTimeStepKeybindings) {
+        global::keybindingManager.removeKeyBinding(kb);
```
alexanderbock, line 441: Could this remove user-keybindings?  If so, it might be better to check if there is a keybind and warn if there is, before removing it?

src/util/timemanager.cpp
```diff
@@ -412,6 +418,96 @@ void TimeManager::setDeltaTimeSteps(std::vector<double> deltaTimes) {
...
+        _deltaTimeSteps.end(),
+        std::back_inserter(steps),
+        [](double value) { return value >= 0.0; }
+    );
+
+    const int nKeys = static_cast<int>(Keys.size());
+    const int nSteps = static_cast<int>(steps.size());
+    const int maxIterations = (nSteps >= nKeys) ? nKeys : nSteps;
+
+    const char* guiPath = DeltaTimeStepsKeybindsGuiPath;
```
alexanderbock, line 461: if you make this `constexpr const char*` you don't need to explicitly capture it in the lambda.    It might need to be `static`, too 🤔

src/util/timemanager.cpp
```diff
@@ -412,6 +418,96 @@ void TimeManager::setDeltaTimeSteps(std::vector<double> deltaTimes) {
...
+
+    auto addDeltaTimeKeybind = [&guiPath, this](Key key, KeyModifier mod, double step) {
+        const std::string s = fmt::format("{:.0f}", step);
+        global::keybindingManager.bindKeyLocal(
+            key,
+            mod,
+            "openspace.time.interpolateDeltaTime(" + s + ")",
```
alexanderbock, line 468: More `fmt::format`?

src/util/timemanager.cpp
```diff
@@ -412,6 +418,96 @@ void TimeManager::setDeltaTimeSteps(std::vector<double> deltaTimes) {
...
+    }
+
+    LINFO("Added keybindings for specified delta time steps.");
+    const int maxKeyBinds = 3 * nKeys;
+    if (nSteps > maxKeyBinds) {
+        LWARNING(fmt::format(
+            "Error settings delta time keys: Too many delta times, so not all could be "
+            "mapped to a key. Total: {} steps, which is {} more than the number of "
+            "available keybindings.",
+            nSteps,
```
alexanderbock, line 507: For the logging, etc, the "one parameter per line rule" is relaxed and you can put:
```
       ....
        nSteps, nSteps - maxKeybinds
```

alexanderbock (approved):

PR #1311, opened 2020-10-09
Update to a snapshot of the new exoplanets data

WeirdRubberDuck:
Update exoplanets data to a one that is still being maintained and updated. This PR contains a task that prepares data from the NASA Exoplanet archive's [Planetary Systems Composite datase](https://exoplanetarchive.ipac.caltech.edu/cgi-bin/TblView/nph-tblView?app=ExoTbls&config=PSCompPars). This is a dataset where several references have been combined to provide one entry per exoplanet, as complete as possible. 

Later on, my plan is to refactor the rest of the exoplanets code to better utilize the new dataset and also add functionality to automatically resynchronize and preprocess the data in a given time interval. So, issue 1303 should not be closed by this PR. 

OBS! For this PR the content of the exoplanets_data synced folder has been updated, but the version is still the same. Developers will have to clear and resync this folder. 

P.S: If anyone knows a better way of reading a number of specific columns from a csv file, please let me know! (exoplanetsdatapreparationtask.cpp)


include/openspace/util/coordinateconversion.h
```diff
@@ -0,0 +1,44 @@
...
+
+#ifndef __OPENSPACE_CORE___COORDINATECONVERSION___H__
+#define __OPENSPACE_CORE___COORDINATECONVERSION___H__
+
+#include <ghoul/glm.h>
+
+namespace openspace {
+
+/**
+* Converts from ICRS coordinates to galactic cartesian coordinates.
```
alexanderbock, line 33: Nitpicky:   We'd align the `*` with the first `*` of the `/**`, not the baseline

include/openspace/util/coordinateconversion.h
```diff
@@ -0,0 +1,44 @@
...
+* \param ra Right ascension, given in decimal degrees
+* \param dec Declination, given in decimal degrees
+* \param distance The distance, or radius, to the position given in any unit.
+* \return A position in galactic cartesian coordinates, given in the same unit as the
+* distance parameter.
```
alexanderbock, line 38: Nitpicky:  the second line would be aligned with the first word of the return value, so:
```
\return A position in galactic....
        distance parameter
```

modules/exoplanets/exoplanetshelper.h
```diff
@@ -33,42 +33,33 @@ struct Exoplanet {
     float a;            // Orbital semi-major axis in AU
     double aUpper;      // Upper uncertainty of orbital semi-major axis
     double aLower;      // Lower uncertainty of orbital semi-major axis
-    double ua;          // Uncertainty of orbital semi-major axis
-    float bigOm;        // Longitude of ascending node in degrees
-    float bigOmUpper;   // Upper uncertainty of longitude of ascending node
-    float bigOmLower;   // Lower uncertainty of longitude of ascending node
-    float uBigOm;       // Uncertainty of longitude of ascending node
+    float bigOmega;     // Longitude of ascending node in degrees
+    float bigOmegaUpper;// Upper uncertainty of longitude of ascending node
```
alexanderbock, line 37: Nitpicky:  an extra space here would make it a bit clearer optically where the comments begin

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -0,0 +1,449 @@
...
+    LINFO(fmt::format("Loading {} exoplanets", total));
+
+    auto readFloatData = [](const std::string& str) -> float {
+    #ifdef WIN32
+        float result;
+        auto [p, ec] = std::from_chars(str.data(), str.data() + str.size(), result);
+        if (ec == std::errc()) {
+            return result;
+        }
+        return NAN;
```
alexanderbock, line 116: Instead of using the C style NAN, which is a `#define` in <cmath>, it might be better to use `std::numeric_limits<float>::quiet_NaN()` instead?
It might be very much more verbose, but more accuratly reflect that it's an "allowed" NaN, rather than a `signaling_NaN`

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -0,0 +1,449 @@
...
+            }
+            // Number of planets in the system
+            else if (column == "sy_pnum") {
+                p.nComp = readIntegerData(data);
+            }
+        }
+
+        // @TODO (emmbr 2020-10-05) Currently, the dataset has no information about the
+        // longitude of the ascending node, but maybe it might in the future
+        p.bigOmega = NAN;
```
alexanderbock, line 290: See above

alexanderbock (commented):
Looks great :+1:

alexanderbock:
yeey

PR #1354, opened 2020-10-23
Issue/1342

jccosta:
The new solution will require a full pass on assets using the rendering billboards cloud in order to adjust the new max and min billboards' sizes when the sizing control is enabled.


alexanderbock:
@micahnyc @carteremmart    We should add these changes to this branch before merging it in

PR #1368, opened 2020-11-04
Issue/1185

Roxeena:
This fixes a bug that reads the horizon files twice instead of once, speeding up the reading of the files a lot. Caching has also been added to make it even faster.


modules/space/translation/horizonstranslation.cpp
```diff
@@ -130,13 +128,48 @@ glm::dvec3 HorizonsTranslation::position(const UpdateData& data) const {
...
+}
+
+void HorizonsTranslation::readHorizonsTextFile() {
+    std::ifstream fileStream(_horizonsTextFile);
 
     if (!fileStream.good()) {
         LERROR(fmt::format(
-            "Failed to open Horizons text file '{}'", horizonsTextFilePath
-        ));
+            "Failed to open Horizons text file '{}'", _horizonsTextFile));
```
alexanderbock, line 172: Nitpicky:  the `));` would be on the next line, starting on the same column as `LERROR`

modules/space/translation/horizonstranslation.cpp
```diff
@@ -186,4 +219,77 @@ void HorizonsTranslation::readHorizonsTextFile(const std::string& horizonsTextFi
...
+bool HorizonsTranslation::loadCachedFile(const std::string& file) {
+    std::ifstream fileStream(file, std::ifstream::binary);
+
+    if (!fileStream.good()) {
+        LERROR(fmt::format("Error opening file '{}' for loading cache file", file));
+        return false;
+    }
+
+    // Read how many values to read
+    int32_t nKeyframes = 0;
```
alexanderbock, line 231: I guess this would be `nValues` instead of `nKeyframes`?

modules/space/translation/horizonstranslation.cpp
```diff
@@ -186,4 +219,77 @@ void HorizonsTranslation::readHorizonsTextFile(const std::string& horizonsTextFi
...
+
+void HorizonsTranslation::saveCachedFile(const std::string& file) const {
+    std::ofstream fileStream(file, std::ofstream::binary);
+    if (!fileStream.good()) {
+        LERROR(fmt::format("Error opening file '{}' for save cache file", file));
+        return;
+    }
+
+    // Write how many values are to be written
+    int32_t nKeyframes = static_cast<int32_t>(_timeline.nKeyframes());
```
alexanderbock, line 269: Same as above

modules/space/translation/horizonstranslation.cpp
```diff
@@ -186,4 +219,77 @@ void HorizonsTranslation::readHorizonsTextFile(const std::string& horizonsTextFi
...
+    if (nKeyframes == 0) {
+        throw ghoul::RuntimeError("Error writing cache: No values were loaded");
+    }
+    fileStream.write(reinterpret_cast<const char*>(&nKeyframes), sizeof(int32_t));
+
+    // Write data
+    std::deque<Keyframe<glm::dvec3>> keyframes = _timeline.keyframes();
+    for (int i = 0; i < nKeyframes; i++) {
+        // First write timestamp
+        fileStream.write(reinterpret_cast<const char*>(&keyframes[i].timestamp),
```
alexanderbock, line 279: In this case (and the ones below),
the `&keyframes[i]...` part would be on a new line (like the `sizeof`)

alexanderbock (approved):

modules/space/translation/horizonstranslation.cpp
Roxeena: I actually mean "read how many keyframes to read" so i adjusted the comment instead

PR #1405, opened 2020-11-23
Issue/1366

GPayne:
Fixes issue 1366 for new recordings but does not solve compatibility with legacy versions.


GPayne:
Updated with fix for support of legacy versions.

PR #1428, opened 2020-12-08
Exoplanets cleanup

WeirdRubberDuck:
- Exoplanet code cleanup (ugly matching of names in speck file)
- Now possible to add exoplanet systems where one or a few planets are having insufficient data. Good for example for the Kepler-80 system, where there is enough data for 5/6 exoplanets (previously this one could not be added at all)

### DIFF
--- a/modules/exoplanets/exoplanetshelper.cpp
+++ b/modules/exoplanets/exoplanetshelper.cpp
@@ -42,46 +42,6 @@ namespace {
 
 namespace openspace::exoplanets {
 
-std::string_view speckStarName(std::string_view csvName) {
-    if (csvName == "MOA-2009-BLG-387L") { return "MOA 2009-BLG-387L"; }
-    if (csvName == "OGLE-2007-BLG-368L") { return "OGLE 2007-BLG-368L"; }
-    if (csvName == "OGLE-2005-BLG-169L") { return "OGLE 2005-BLG-169L"; }
-    if (csvName == "OGLE-2005-BLG-071L") { return "OGLE 2005-BLG-71L"; }
-    if (csvName == "OGLE-2003-BLG-235L") { return "OGLE 2003-BLG-235L"; }
-    if (csvName == "MOA-2008-BLG-310L") { return "MOA 2008-BLG-310L"; }
-    if (csvName == "OGLE-2006-BLG-109L") { return "OGLE 2006-BLG-109L"; }
-    if (csvName == "HD 137388") { return "HD 137388 A"; }
-    if (csvName == "MOA-2010-BLG-477L") { return "MOA 2010-BLG-477L"; }
-    if (csvName == "MOA-2009-BLG-266L") { return "MOA 2009-BLG-266L"; }
-    if (csvName == "iot Dra") { return "HIP 75458"; }
-    if (csvName == "MOA-2007-BLG-400L") { return "MOA 2007-BLG-400L"; }
-    if (csvName == "OGLE-2011-BLG-0251L") { return "OGLE 2011-BLG-251L"; }
-    if (csvName == "OGLE-2005-BLG-390L") { return "OGLE 2005-BLG-390L"; }
-    if (csvName == "MOA-2007-BLG-192L") { return "MOA 2007-BLG-192L"; }
-    if (csvName == "MOA-2009-BLG-319L") { return "MOA 2009-BLG-319L"; }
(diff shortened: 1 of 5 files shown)


modules/exoplanets/exoplanetshelper.cpp
```diff
@@ -42,66 +42,29 @@ namespace {
...
-    if (csvName == "MOA-2010-BLG-477L") { return "MOA 2010-BLG-477L"; }
-    if (csvName == "MOA-2009-BLG-266L") { return "MOA 2009-BLG-266L"; }
-    if (csvName == "iot Dra") { return "HIP 75458"; }
-    if (csvName == "MOA-2007-BLG-400L") { return "MOA 2007-BLG-400L"; }
-    if (csvName == "OGLE-2011-BLG-0251L") { return "OGLE 2011-BLG-251L"; }
-    if (csvName == "OGLE-2005-BLG-390L") { return "OGLE 2005-BLG-390L"; }
-    if (csvName == "MOA-2007-BLG-192L") { return "MOA 2007-BLG-192L"; }
-    if (csvName == "MOA-2009-BLG-319L") { return "MOA 2009-BLG-319L"; }
-    return csvName;
+bool isValidPosition(const glm::vec3 pos) {
```
alexanderbock, line 45: The pos can be a const reference to skip a potential copy:  `const glm::vec3&`

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -56,120 +56,125 @@ constexpr const char* NoDataTextureFile =
...
-    std::vector<Exoplanet> planetSystem;
-    std::vector<std::string> planetNames;
-
     while (getline(lut, line)) {
         std::istringstream ss(line);
         std::string name;
         getline(ss, name, ',');
 
-        if (name.compare(0, name.length() - 2, starNameSpeck) == 0) {
+        if (name.compare(0, name.length() - 2, starName) == 0) {
```
alexanderbock, line 88: I generally dislike the `compare` function since it's not super obvious which way the comparsion goes (does `== 0` mean that the strings are equal, unequal  since 0 normally means `false` but `==` means they they are equal), so i would suggest:
`if (name.substr(0, name.length() - 2) == starName) {`
instead

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -56,120 +56,125 @@ constexpr const char* NoDataTextureFile =
...
+    sanitizeNameString(sanitizedStarName);
+
+    const std::string guiPath = ExoplanetsGuiPath + sanitizedStarName;
 
-    if (!found) {
-        LERROR(fmt::format("No star with the provided name was found: '{}'", starName));
+    SceneGraphNode* existingStarNode = sceneGraphNode(starIdentifier);
+    if (existingStarNode) {
+        LERROR(fmt::format(
+            "Adding of exoplanet system '{}' failed. The system has already been added.",
```
alexanderbock, line 137: We leave out the `.` at the end of warning/error messages

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -351,8 +349,9 @@ void createExoplanetSystem(std::string_view starName) {
         "}";
 
         openspace::global::scriptEngine->queueScript(
-            "openspace.addSceneGraphNode(" + planetTrailNode + ");",
-            openspace::scripting::ScriptEngine::RemoteScripting::Yes
+            "openspace.addSceneGraphNode(" + planetTrailNode + ");"
```
alexanderbock, line 352: Could use fmt::format here for a bit more compact way of writing this

modules/exoplanets/exoplanetsmodule_lua.inl
WeirdRubberDuck: Even multiple sentence ones? Okay!

modules/exoplanets/exoplanetsmodule_lua.inl
WeirdRubberDuck: Please elaborate on how you mean here :) I've been considering using fmt::format for these `queueScript` calls but personally thought this was the more readable option

Also, I prefer these things to be done somewhat consistent at least throughout the file. So maybe let me know how you prefer it to look like and I can change it for all places in the file later.

PR #1436, opened 2020-12-14
Add habitable zone for exoplanet systems

WeirdRubberDuck:


modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -46,16 +46,20 @@ namespace openspace::exoplanets::luascriptfunctions {
...
-    "${SYNC}/http/exoplanets_data/1/exoplanets_data.bin";
+    "${SYNC}/http/exoplanets_data/2/exoplanets_data.bin";
 
 constexpr const char* StarTextureFile = "${SYNC}/http/exoplanets_textures/1/sun.jpg";
 constexpr const char* NoDataTextureFile =
     "${SYNC}/http/exoplanets_textures/1/grid-32.png";
 constexpr const char* DiscTextureFile =
     "${SYNC}/http/exoplanets_textures/1/disc_texture.png";
 
+const float AU = static_cast<float>(distanceconstants::AstronomicalUnit);
```
alexanderbock, line 59: `constexpr const`-able?

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -236,7 +245,7 @@ void createExoplanetSystem(const std::string& starName) {
 
     // Planets
     for (size_t i = 0; i < system.planetNames.size(); i++) {
-        ExoplanetDataEntry planet = system.planetsData[i];
+        ExoplanetDataEntry& planet = system.planetsData[i];
```
alexanderbock, line 248: seems like this could be `const ExoplanetDataEntry&`?

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -400,6 +407,79 @@ void createExoplanetSystem(const std::string& starName) {
...
+        const glm::vec2 zone = computeHabitableZone(
+            system.starData.teff,
+            system.starData.luminosity
+        );
+
+        float meanInclination = 0.f;
+        for (const ExoplanetDataEntry& p : system.planetsData) {
+            meanInclination += p.i;
+        }
+        meanInclination /= static_cast<float>(system.planetsData.size());
```
alexanderbock, line 425: Could it ever happen that the planetsData.size() is 0 here?  In that case, this would lead to an exception

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -88,25 +88,36 @@ void ExoplanetsDataPreparationTask::perform(
...
     std::string colName;
     while (getline(sStream, colName, ',')) {
         columnNames.push_back(colName);
     }
 
     // Read total number of items
     int total = 0;
-    while (getline(inputDataFile, planetRow)) {
+    std::string row;
+    while (getline(inputDataFile, row)) {
```
alexanderbock, line 113: `std::getline`  otherwise you're relying on the namespace deduction based on the input argument to detect the right `getline` function, which seems unnecessary in this case

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -156,7 +167,7 @@ void ExoplanetsDataPreparationTask::perform(
     ExoplanetDataEntry p;
     std::string data;
     int exoplanetCount = 0;
-    while (getline(inputDataFile, planetRow)) {
+    while (getline(inputDataFile, row)) {
```
alexanderbock, line 170: see above

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -167,7 +178,7 @@ void ExoplanetsDataPreparationTask::perform(
         float dec = std::numeric_limits<float>::quiet_NaN();    // decimal degrees
         float distanceInParsec = std::numeric_limits<float>::quiet_NaN();
 
-        std::istringstream lineStream(planetRow);
+        std::istringstream lineStream(row);
         int columnIndex = 0;
         while (getline(lineStream, data, ',')) {
```
alexanderbock, line 183: ^^

src/engine/openspaceengine_lua.inl
```diff
@@ -288,6 +292,71 @@ int downloadFile(lua_State* L) {
...
+* createPixelImage():
+* Creates a one pixel image with a given color and returns the p
+*/
+int createPixelImage(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, 2, "lua::createPixelImage");
+
+    const std::string& name = ghoul::lua::value<std::string>(L, 1);
+    const ghoul::Dictionary& d = ghoul::lua::value<ghoul::Dictionary>(L, 2);
+
+    using namespace openspace::documentation;
```
alexanderbock, line 306: Maybe you can put a @TODO comment in here that we want to clean this up with a more direct use of the Verifier in the future

src/engine/openspaceengine_lua.inl
```diff
@@ -288,6 +292,71 @@ int downloadFile(lua_State* L) {
...
+    const std::string& key = "color";
+    ghoul::Dictionary colorDict = {{ key, d }};
+    TestResult res = DoubleVector3Verifier()(colorDict, key);
+
+    if (!res.success) {
+        return ghoul::lua::luaError(L,
```
alexanderbock, line 312: Nitpicky:  The `L` should be on a new line

alexanderbock (approved):

modules/exoplanets/exoplanetshelper.h
```diff
@@ -71,9 +79,11 @@ struct ExoplanetDataEntry {
 };
 
 struct StarData {
-    glm::vec3 position  = glm::vec3(std::numeric_limits<float>::quiet_NaN()); // In parsec
-    float radius        = std::numeric_limits<float>::quiet_NaN(); // In solar radii
-    float bvColorIndex  = std::numeric_limits<float>::quiet_NaN();
+    glm::vec3 position = glm::vec3(std::numeric_limits<float>::quiet_NaN()); // In parsec
```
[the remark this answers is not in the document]
WeirdRubberDuck: ^^ Don't know who is on the other side of that fight, but I like doing so sometimes when I think it helps with code readability. Realize that I don't have a concrete strategy for it though, apart from gut feeling... So you can get as you want this time ;)

modules/exoplanets/exoplanetsmodule_lua.inl
WeirdRubberDuck: That should not be possible, since a system without planets are not considered. Previously there is a check for this
```
    if (system.planetNames.empty()) {
        LERROR(fmt::format("Exoplanet system '{}' could not be found", starName));
        return;
    }
```
However, I can make it more clear by checking agains `planetsData` instead of `planetNames` (these vectors should always have the same length)

modules/exoplanets/exoplanetsmodule_lua.inl
WeirdRubberDuck: It can not, since some member values of `planet` are replaced with default values later in the code if they are nan. 

This will be improved after my data reading rewrite, after which I hope to not have to handle nan values explicitly in the scene graph node creation code anymore :)

PR #1468, opened 2021-01-26
Add habitable zone for our Sun and some useful generic classes

WeirdRubberDuck:
* Add a simple disc renderable
* Generic plane geometry class
* Generic texture component that can be used by renderables to simplify loading code **
* Add a specialized habitable zone renderable that inherits from RenderableDisc, but computes the size based on given star properties
* Add habitable zone to our solar system
* Update textures for habitable zones

** Would like to get some input on if this class is actually useful, or if it will become too much of just a 'pass through' class that duplicated the interface to the `Texture` class... The more options one want to add to the texture, the more complex this `TextureComponent` class will have to become


modules/base/rendering/renderabledisc.cpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+    setBoundingSphere(_size);
+    _size.onChange([&]() { _planeIsDirty = true; });
+    addProperty(_size);
+
+    if (dictionary.hasKey(WidthInfo.identifier)) {
+        _width = static_cast<float>(dictionary.value<double>(WidthInfo.identifier));
+    }
+    addProperty(_width);
+    addProperty(_opacity);
```
micahnyc, line 125: feels like opacity should be read from the dictionary also

modules/exoplanets/exoplanetsmodule.cpp
```diff
@@ -94,7 +94,7 @@ void ExoplanetsModule::internalInitialize(const ghoul::Dictionary&) {
 std::vector<documentation::Documentation> ExoplanetsModule::documentations() const {
     return {
         ExoplanetsDataPreparationTask::documentation(),
-        RenderableOrbitDisc::Documentation()
+        RenderableOrbitDisc::Documentation(),
```
micahnyc, line 97: extra

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -446,17 +438,11 @@ void createExoplanetSystem(const std::string& starName) {
...
-                "Offset = { " +
-                    std::to_string(relativeOffset) + ", " +
-                    std::to_string(relativeOffset) +
-                "}," //min / max extend
-                "Opacity = 0.05"
+                "Type = 'RenderableHabitableZone',"
+                "Texture = openspace.absPath('" + HabitableZoneTextureFile + "'),"
+                "Luminosity = " + std::to_string(system.starData.luminosity) + ","
+                "EffectiveTemperature = " + std::to_string(system.starData.teff) + ","
+                "Opacity = 0.07"
```
micahnyc, line 445: seems like all our zones have an Opacity of < 0.1 I wonder if this is something that can be put into the logic of the class instead of the user always setting such small values

src/rendering/texturecomponent.cpp
```diff
@@ -0,0 +1,111 @@
+/*****************************************************************************************
```
micahnyc, line 1: excited for these nice helper functions! Do we consider this like and OpenSpaceTextureComponent with stuff specific to OpenSpace? or should this just maybe be a class in ghoul?

modules/space/rendering/renderablehabitablezone.h
```diff
@@ -0,0 +1,71 @@
...
+    static documentation::Documentation Documentation();
+
+private:
+    void initializeShader() override;
+    void updateUniformLocations() override;
+    void computeZone();
+
+    /**
+     * Compute the inner and outer boundary of the habitable zone of a star, accordring to
+     * formula and coefficients by Kopparapu et al. (2015) https://arxiv.org/abs/1404.5292
```
micahnyc, line 50: yay for citing the paper!

modules/space/rendering/renderablehabitablezone.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+glm::vec4 RenderableHabitableZone::computeKopparapuZoneBoundaries(float teff,
+                                                                  float luminosity)
+{
+    // Kopparapu's formula only considers stars with teff in range [2600, 7200] K.
+    // However, we want to use the formula for more stars, so add some flexibility to
+    // the teff boundaries.
+    // OBS! This also prevents problems with too large values in the distance computation
+    if (teff > 8000.f || teff < 2000.f) {
+        // For the other stars, use a method by Tom E. Morris:
```
micahnyc, line 235: Is this something we can communicate to the user through the interface? I.e a property that is changed by  the renderable so they know? Is this not worth it b/c it will wind up looking the same? Thinking if some will look different then others, it might be nice to explain why.

data/assets/scene/solarsystem/sun/habitablezone.asset
```diff
@@ -0,0 +1,39 @@
...
+        Description = "Habitable zone for the sun in our solar system."
+    }
+}
+
+assetHelper.registerSceneGraphNodesAndExport(asset, { HabitableZone })
+
+asset.meta = {
+    Name = "Sun Habitable Zone",
+    Version = "1.0",
+    Description = [[ The habitable zone around our sun]],
```
micahnyc, line 34: maybe here we can add info about the paper/data source.

modules/base/rendering/renderabledisc.cpp
WeirdRubberDuck: That is done in the constructor of the base class ( `Renderable` ) :)

modules/space/rendering/renderablehabitablezone.cpp
WeirdRubberDuck: That's a good point. A drawback with this method is the we don't get the optimistic boundaries, which will affect the appearance of the disc depending on the transfer function/texture. I think a property would make sense, good suggestion!

src/rendering/texturecomponent.cpp
WeirdRubberDuck: Good question. I guess it might actually make more sense for it to live in Ghoul, since it's sort of a simplified interface to the `ghoul::opengl::Texture` class. Maybe @alexanderbock has an opinion?

modules/base/rendering/renderabledisc.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include <openspace/scene/scene.h>
+#include <openspace/util/updatestructures.h>
+#include <ghoul/filesystem/filesystem.h>
+#include <ghoul/logging/logmanager.h>
+#include <ghoul/opengl/openglstatecache.h>
+#include <ghoul/opengl/programobject.h>
+#include <ghoul/opengl/textureunit.h>
+
+namespace {
+    constexpr const char* _loggerCat = "RenderableDisc";
```
alexanderbock, line 40: I wanna start changing this in more places, but it will be better to declare it as
`constexpr const char _loggerCat[] = "RenderableDisc";`   since that means the length of the string is encoded in the type and doesn't need a `strlen` copy.  Even if it happens at compile time, it would still slow things down

modules/base/rendering/renderabledisc.cpp
```diff
@@ -0,0 +1,230 @@
...
+        absPath("${MODULE_BASE}/shaders/disc_fs.glsl")
+    );
+    updateUniformLocations();
+}
+
+void RenderableDisc::updateUniformLocations() {
+   ghoul::opengl::updateUniformLocations(*_shader, _uniformCache, UniformNames);
+}
+
+float RenderableDisc::planeSize() {
```
alexanderbock, line 226: This function can be declared `const`

modules/base/shaders/disc_fs.glsl
```diff
@@ -0,0 +1,60 @@
...
+in vec2 vs_st;
+in float vs_screenSpaceDepth;
+
+uniform sampler1D colorTexture;
+uniform float width;
+uniform float opacity;
+
+Fragment getFragment() {
+    // Moving the origin to the center
+    vec2 st = (vs_st - vec2(0.5)) * 2.0;
```
alexanderbock, line 36: Since the original value of `vs_st` will not be used, it might make sense to move this calculation into the vertex shared.  That way, it will only be done once per vertex (and the resulting value interpolated), rather than once per fragment.   Not sure if it will actually make a difference

modules/base/shaders/disc_fs.glsl
```diff
@@ -0,0 +1,60 @@
...
+
+    // The length of the texture coordinates vector is our distance from the center
+    float radius = length(st);
+
+    // We only want to consider ring-like objects so we need to discard everything else
+    if (radius > 1.0)
```
alexanderbock, line 42:
```
if (radius > 1.0) {
    discard;
}

File: modules/exoplanets/exoplanetsmodule_lua.inl [hunk 12]
```diff
@@ -55,6 +55,8 @@ constexpr const char* NoDataTextureFile =
     "${SYNC}/http/exoplanets_textures/1/grid-32.png";
 constexpr const char* DiscTextureFile =
     "${SYNC}/http/exoplanets_textures/1/disc_texture.png";
+constexpr const char* HabitableZoneTextureFile =
```
alexanderbock, line 58: Same as above:
`constexpr const char HabitableZoneTextureFile[] = "..."`

modules/exoplanets/exoplanetsmodule_lua.inl
alexanderbock: I think it will be tricky since the opacity is a property of the Renderable class.  It has it's attractions to have that property be guaranteed to be the same everywhere imho

include/openspace/rendering/texturecomponent.h
```diff
@@ -0,0 +1,76 @@
...
+
+    // Loads a texture from a file on disk
+    void loadFromFile(const std::string& path);
+
+    // Function to call in a renderable's update function to make sure
+    // the texture is kept up to date
+    void update();
+
+private:
+    std::unique_ptr<ghoul::filesystem::File> _textureFile = nullptr;
```
alexanderbock, line 62: No need for the `= nullptr`.  The default constructor already does that for us

modules/space/rendering/renderablehabitablezone.cpp
```diff
@@ -0,0 +1,278 @@
...
+#include <openspace/util/distanceconstants.h>
+#include <openspace/util/updatestructures.h>
+#include <ghoul/logging/logmanager.h>
+#include <ghoul/filesystem/filesystem.h>
+#include <ghoul/opengl/openglstatecache.h>
+#include <ghoul/opengl/programobject.h>
+#include <ghoul/opengl/textureunit.h>
+
+namespace {
+    constexpr const char* _loggerCat = "RenderableHabitableZone";
```
alexanderbock, line 42: See above

modules/space/rendering/renderablehabitablezone.cpp
```diff
@@ -0,0 +1,278 @@
...
+    ghoul::opengl::updateUniformLocations(*_shader, _uniformCache, UniformNames);
+}
+
+void RenderableHabitableZone::updateUniformLocations() {
+    ghoul::opengl::updateUniformLocations(*_shader, _uniformCache, UniformNames);
+}
+
+void RenderableHabitableZone::computeZone() {
+    glm::vec4 distancesInAu = computeKopparapuZoneBoundaries(_teff, _luminosity);
+    constexpr float AU = static_cast<float>(distanceconstants::AstronomicalUnit);
```
alexanderbock, line 208: `constexpr const float`?

alexanderbock: Also;  I'd check if the `float` is actually sufficient.  It only has 7 or 8 significant digits and these values might run out of precision.  Maybe better to do the calculations in `double` and then cast to `float` right at the end

modules/space/rendering/renderablehabitablezone.cpp
```diff
@@ -0,0 +1,278 @@
...
+    const float innerConservative = distancesInAu[1] * AU;
+    const float outerConservative = distancesInAu[2] * AU;
+    const float outer = distancesInAu[3] * AU;
+
+    float discWidth = 0.f;
+    if (outer > 0.f) {
+        discWidth = (outer - inner) / outer;
+    }
+
+    _size.set(outer);
```
alexanderbock, line 219: `_size = outer;`

modules/space/shaders/habitablezone_fs.glsl
```diff
@@ -0,0 +1,90 @@
...
+// is treated as a linear scale where the color represent too cold to too hot. Account
+// for the conservative bounds my mapping one third of the texture ouside each boundary.
+// All parameters \in [0,1], where 1.0 corresponds to the max radius.
+float computeTextureCoord(float radius, float innerRadius,
+                          float conservativeInner, float conservativeOuter)
+{
+    const float t1 = 1.0 / 3.0;
+    const float t2 = 2.0 / 3.0;
+
+    if(radius < conservativeInner) {
```
alexanderbock, line 46: Missing space between `if` and `(`

modules/space/shaders/habitablezone_fs.glsl
```diff
@@ -0,0 +1,90 @@
...
+    }
+    else {
+        float t = (radius - conservativeInner) / (conservativeOuter - conservativeInner);
+        return mix(t1, t2, t);
+    }
+}
+
+Fragment getFragment() {
+    // Moving the origin to the center
+    vec2 st = (vs_st - vec2(0.5)) * 2.0;
```
alexanderbock, line 62: Same argument as above

src/rendering/texturecomponent.cpp
alexanderbock: I think having it in OpenSpace makes more sense since we than have some additional options.  For example, this file uses the  ghoul::filesystem::File  watch mechanism, which requires a function call to happen every frame to trigger these changes.  If this class would be in Ghoul and a person using Ghoul and this class doesn't make the call, it would work unexpected.

src/rendering/texturecomponent.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include <openspace/rendering/texturecomponent.h>
+
+#include <ghoul/filesystem/file.h>
+#include <ghoul/filesystem/filesystem.h>
+#include <ghoul/io/texture/texturereader.h>
+#include <ghoul/logging/logmanager.h>
+
+namespace {
+    constexpr const char* _loggerCat = "TextureComponent";
```
alexanderbock, line 33: See above

src/rendering/texturecomponent.cpp
```diff
@@ -0,0 +1,111 @@
...
+} // namespace
+
+namespace openspace {
+
+ghoul::opengl::Texture* TextureComponent::texture() const {
```
alexanderbock, line 38: If the return value cannot be `nullptr`, would it make sense to return it as `ghoul::opengl::Texture&` instead?  Also I'm surprised that this compiles.  I would have expected that it needs to be `const ghoul::opengl::Texture*` since it is a const function.  Might be good to have both options

src/util/planegeometry.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+namespace {
+    constexpr const char* _loggerCat = "PlaneGeometry";
+} // namespace
+
+namespace openspace {
+
+PlaneGeometry::PlaneGeometry(glm::vec2 size) : _size(std::move(size)) {}
+
+PlaneGeometry::PlaneGeometry(float size) : _size(size, size) {}
```
alexanderbock, line 38: `PlaneGeometry::PlaneGeometry(float size) : PlaneGeometry(glm::vec2(size, size)) {}` ?

src/util/planegeometry.cpp
```diff
@@ -0,0 +1,123 @@
...
+    VertexData vertices[] = {
+        { -size.x, -size.y, 0.f, 0.f },
+        {  size.x,  size.y, 1.f, 1.f },
+        { -size.x,  size.y, 0.f, 1.f },
+        { -size.x, -size.y, 0.f, 0.f },
+        {  size.x, -size.y, 1.f, 0.f },
+        {  size.x,  size.y, 1.f, 1.f },
+    };
+
+    if (_vaoId == 0) {
```
alexanderbock, line 65: I think I would prefer a check in the beginning of the `initialize` (could just be a `ghoul_assert` too) that makes sure we don't call this function twice.  Since it would most likely be a bug if that happens?

src/util/planegeometry.cpp
```diff
@@ -0,0 +1,123 @@
...
+        glGenVertexArrays(1, &_vaoId);
+    }
+
+    if (_vBufferId == 0) {
+        glGenBuffers(1, &_vBufferId);
+
+        if (_vBufferId == 0) {
+            LERROR("Could not create vertex buffer");
```
alexanderbock, line 73: How about throwing an exception instead and removing the `bool` return value?  There is not much use to recover if the buffer creation fails?

src/util/planegeometry.cpp
alexanderbock: I see now that the `initialize` function is called down below.  In that case, I think it would be better to split the function up into the actual `initialize` part and the `update` part.  Agree that the `initialize` function would be pretty short

alexanderbock (changes requested):

src/util/planegeometry.cpp
WeirdRubberDuck: I agree. Created an `updateGeometry` function instead so now this class more resembles the renderables. 

A related question: Do you think it is confusing that this class has an `initialize` and `deinitialize` function that are responsible for calling openGL function? Would it be better to name them `initializeGL` for example, to increase the similarity with how the renderables work?

alexanderbock: I think it might help, but I wouldn't see it as absolutely necesasry.. we have a few places where that is the case.  I see it as:
`initialize` -> may need OpenGL or may not
`initializeGL` -> definitely needs OpenGL

modules/base/rendering/renderabledisc.cpp
micahnyc: Then I vote that the `addProperty(_opacity);` goes to the base class for consistency.

src/rendering/texturecomponent.cpp
WeirdRubberDuck: Oops, true that the return value should be const in this case, since it's a const function.. Thanks. Adding both options, in case someone wants to access and change the texture object somehow. 

To me it makes sense to return it as a pointer, since `_texture` can be a `nullptr` if the texture is never loaded from the file, for any reason 🤔

data/assets/examples/discs.asset
```diff
@@ -0,0 +1,22 @@
+local assetHelper = asset.require('util/asset_helper')
+
+local singeColorTexturePath = openspace.createPixelImage("example_ring_color", {0.0, 1.0, 1.0})
```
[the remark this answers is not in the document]
WeirdRubberDuck: What is the potential danger with that?

data/assets/scene/solarsystem/sun/habitablezone.asset
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+assetHelper.registerSceneGraphNodesAndExport(asset, { HabitableZone })
+
+asset.meta = {
+    Name = "Sun Habitable Zone",
+    Version = "1.0",
+    Description = [[ The habitable zone around our sun, computed using formula and 
+    coefficients by Kopparapu et al. (2015) https://arxiv.org/abs/1404.5292]],
```
WeirdRubberDuck, line 29: @micahnyc does this look ok? Or should the Url to the paper be in`URL` part of the `asset.meta`?

micahnyc: This is good. A lot of the DU stuff has the source in the description so it will be consistent with what we have.

Issue #1209, opened 2020-06-22
Asset values for labeling for different bodies are wrong


alexanderbock:
Report from Alex @ NCMNS
•	Labeling of surface features on Venus, Mars, Enceladus, Titan, Pluto, and Charon does not appear unless the “Fade In Starting Distance for Labels” value is increased to be higher than the “Fade Out Starting Distance for Labels” value. These settings have proper values for Mercury and Jupiter’s Galilean moons.

@jccosta  Probably knows and can instruct

GPayne:
Created PR#1215 for this.

Roxeena:
Was this fixed in #1476? If so I suggest we close this issue.

GPayne:
Yes, this was fixed in PR #1476

PR #1476, opened 2021-01-29
Globebrowsing Fade-In/Out Fix and Assets Updates (Issue #1209)

GPayne:
Fixed fade out of globebrowsing labels so that fading-out labels when close to the surface works with fade-in as well.
Also updated all asset files that use the labels so that the fade in/out ranges, culling, and font sizes look good.


data/assets/scene/solarsystem/planets/earth/magnetosphere.asset
```diff
@@ -0,0 +1,67 @@
+local assetHelper = asset.require('util/asset_helper')
```
alexanderbock, line 1: I assume this file made it in by accident? :D

modules/globebrowsing/src/globelabelscomponent.cpp
```diff
@@ -50,7 +50,7 @@ namespace {
 
     constexpr const char* KeyLabelsFileName = "FileName";
 
-    constexpr const double LabelFadeRangeConst = 1500.0;
+    constexpr const double LabelFadeOutLimitAltitude_meters = 25000.0;
```
alexanderbock, line 53: Would prefer: `LabelFadeOutLimitAltitudeMeters` to keep the same camelCase naming convention

alexanderbock (changes requested):

data/assets/scene/solarsystem/planets/earth/magnetosphere.asset
GPayne: Oops! Removed it.

modules/globebrowsing/src/globelabelscomponent.cpp
GPayne: Changed this.

alexanderbock:
👍

PR #1478, opened 2021-01-29
Optimized Images

vardaan-raj:


alexanderbock:
Hi.  What does "optimized" mean?  Optimized in which regard?

PR #1488, opened 2021-02-03
Feature/multitexturing (closes #112)

### LINKED ISSUE #112: Using AssImp to load models that have multiple textures and objects
<a href="https://github.com/alexanderbock"></img></a> **Issue by [alexanderbock](https://github.com/alexanderbock)**
_Friday Jul 08, 2016 at 09:06 GMT_
_Originally opened as https://github.com/OpenSpace/OpenSpace-Development/issues/114_

---

For example Rosetta:
https://www.dropbox.com/s/1dozz14wnsth59s/rosetta.zip

Roxeena:
Improved model loading that supports models with multiple parts, materials and textures.


data/assets/scene/milkyway/objects/orionnebula/nebula.asset
```diff
@@ -59,11 +59,7 @@ local OrionNebulaModel = {
     },
     Renderable = {
         Type = "RenderableModel",
-        Geometry = {{
-            Type = "MultiModelGeometry",
-            GeometryFile = sync .. "/orion_nebula.obj",
-            ColorTexture = sync .. "/heic0601a_masked.png"
-        }},
+        GeometryFile = sync .. "/orion_nebula.obj",
```
alexanderbock, line 62: So much nicer <3

modules/base/rendering/renderablemodel.cpp
```diff
@@ -263,13 +267,67 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+                // Handle each file
+                file = absPath(fileDictionary.value<std::string>(k));
+                geometries.push_back(ghoul::io::ModelReader::ref().loadModel(
+                    file,
+                    _forceRenderInvisible,
+                    _notifyInvisibleDropped
+                ));
+            }
+
+            if (geometries.size() > 0) {
```
alexanderbock, line 308: = `!geometries.empty()`

modules/base/rendering/renderablemodel.cpp
```diff
@@ -263,13 +267,67 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+                ghoul::modelgeometry::ModelGeometry combinedGeometry =
+                std::move(*geometries[0].release());
```
alexanderbock, line 310: I think I would prefer this to be split up into two lines.  The `release` might trigger a `delete` before we can make the `move` of the ModelGeometry.  Would be better to save the ModelGeometry into the `combinedGeometry` first and then do the `release` (if that is even necessary)

modules/base/rendering/renderablemodel.cpp
```diff
@@ -263,13 +267,67 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+
+                 // Combine all models into one ModelGeometry
+                 for (unsigned int i = 1; i < geometries.size(); ++i) {
```
alexanderbock, line 313:
```cpp
for (const std::unique_ptr<ghoul::modelgeometry::ModelGeometry>& geometry : geometries) {
    for (const ghoul::io::ModelMesh& mesh : geometry->meshes()) {
        ...
    }
}
```

modules/base/rendering/renderablemodel.cpp
```diff
@@ -263,13 +267,67 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+                    for (unsigned int m = 0; m < geometries[i]->meshes().size(); ++m) {
+                        combinedGeometry.meshes().push_back(
+                            std::move(geometries[i]->meshes()[m])
+                        );
+                    }
+
+                    for (unsigned int t = 0; t < geometries[i]->textureStorage().size(); ++t) {
```
alexanderbock, line 320: same here

modules/base/shaders/model_fs.glsl
```diff
@@ -28,39 +28,99 @@ in vec2 vs_st;
...
 uniform float lightIntensities[8];
 
 uniform float opacity = 1.0;
 
-const vec3 SpecularAlbedo = vec3(1.0);
-
 Fragment getFragment() {
+
+	// Render invisible mesh with flashy procedural material
+	if(use_forced_color) {
```
alexanderbock, line 62: Missing space between the `if` and the `(`

modules/base/shaders/model_fs.glsl
```diff
@@ -28,39 +28,99 @@ in vec2 vs_st;
...
+		Fragment frag;
+
+		vec3 adjustedPos = floor(vs_positionCameraSpace.xyz * 3.0);
```
alexanderbock, line 65: This looks like an indentation width of 8?

modules/base/shaders/model_fs.glsl
```diff
@@ -28,39 +28,99 @@ in vec2 vs_st;
...
     Fragment frag;
 
     if (performShading) {
+
+		vec3 specularAlbedo;
+		if (has_texture_specular) {
+			specularAlbedo = texture(texture_specular, vs_st).rgb;
+		}
+		else {
+			if(has_color_specular) {
```
alexanderbock, line 102: Missing space between `if` and `(`

modules/base/shaders/model_fs.glsl
```diff
@@ -28,39 +28,99 @@ in vec2 vs_st;
...
+			}
+		}
+
         // Some of these values could be passed in as uniforms
         const vec3 lightColorAmbient = vec3(1.0);
         const vec3 lightColor = vec3(1.0);
         
-        vec3 n = normalize(vs_normalViewSpace);
+        vec3 n;
+		if(has_texture_normal) {
```
alexanderbock, line 115: Missing space between `if` and `(`

modules/base/shaders/model_fs.glsl
```diff
@@ -73,12 +133,12 @@ Fragment getFragment() {
             float specularCosineFactor = dot(c,r);
             const float specularPower = 100.0;
 
-            vec3 diffuseColor =
-                diffuseIntensity * lightColor * diffuseAlbedo *
-                max(diffuseCosineFactor, 0);
+			vec3 diffuseColor =
```
alexanderbock, line 136: ?  Indentation?

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
```diff
@@ -87,10 +90,10 @@ documentation::Documentation RenderableModelProjection::Documentation() {
                 Optional::No
             },
             {
-                keyGeometry,
-                new ReferencingVerifier("base_geometry_model"),
+                KeyGeomModelFile,
+                new OrVerifier({ new StringVerifier, new StringListVerifier }),
                 Optional::No,
-                "The geometry that is used for rendering this model."
+                "The file or files that that is used for rendering this model"
```
alexanderbock, line 96: "that that is"  -> "that are"?

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
```diff
@@ -126,9 +129,56 @@ RenderableModelProjection::RenderableModelProjection(const ghoul::Dictionary& di
...
-    Dictionary geometryDictionary = dictionary.value<Dictionary>(keyGeometry);
-    _geometry = modelgeometry::ModelGeometry::createFromDictionary(geometryDictionary);
+
+    if (dictionary.hasKey(KeyGeomModelFile)) {
+        std::string file;
+
+        if (dictionary.hasValue<std::string>(KeyGeomModelFile)) {
+            // Handle single file
+            file = absPath(dictionary.value<std::string>(KeyGeomModelFile));
+            _geometry = ghoul::io::ModelReader::ref().loadModel(file, false, true);
```
alexanderbock, line 139: For these kinds of case we have the `ghoul::Boolean` type where you can give the parameters proper names and use `Yes` and `No` instead.  If I'm just reading this, I have to go to the function declaration to see what the boolean parameters actually mean

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
```diff
@@ -126,9 +129,56 @@ RenderableModelProjection::RenderableModelProjection(const ghoul::Dictionary& di
...
+                geometries.push_back(ghoul::io::ModelReader::ref().loadModel(
+                    file,
+                    false,
+                    true
+                ));
+            }
+
+            if (geometries.size() > 0) {
+                ghoul::modelgeometry::ModelGeometry combinedGeometry =
+                    std::move(*geometries[0].release());
```
alexanderbock, line 159: Same comment as above... I'm afraight that the unique_ptr might trigger it's `delete` before the `std::move()` can yeet the memory/variables

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
```diff
@@ -126,9 +129,56 @@ RenderableModelProjection::RenderableModelProjection(const ghoul::Dictionary& di
...
+
+                // Combine all models into one ModelGeometry
+                for (unsigned int i = 1; i < geometries.size(); ++i) {
```
alexanderbock, line 162: See above

alexanderbock (approved):
:+1:

modules/base/rendering/renderablemodel.cpp
Roxeena: Note that the loop starts at index 1 and not 0 for the geometries. Will change for the others though.

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
Roxeena: See above (:

modules/base/rendering/renderablemodel.cpp
alexanderbock: Ha!  Yea, you're right  I didn't notice that

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
alexanderbock: Good point!

modules/base/rendering/renderablemodel.cpp
Roxeena: I noticed now that for both the `ModelMesh` and the `TextureEntry` this will call the copy constructor, which is deleted so I think I’m gonna keep it as it is.

alexanderbock: Huh.. it shouldn't call the copy constructor if you bind the for loop to a `const ghoul::io::ModelMesh&`;  if it is `const ghoul::io::ModelMesh` then it would, but not with the reference 🤔

Roxeena: Well that is what the compiler told me when I tried it. Could it be because the mesh array is accessed with the function `meshes()` and not directly? Because it works fine in all other places.

alexanderbock: Ah.. yea, it was the second `const` since ofc we can't move out of a `const T&`:
This works:
```cpp
            if (!geometries.empty()) {
                using namespace ghoul::modelgeometry;
                ModelGeometry combinedGeometry = std::move(*geometries[0].release());

                 // Combine all models into one ModelGeometry
                for (const std::unique_ptr<ModelGeometry>& geometry : geometries) {
                    for (ghoul::io::ModelMesh& mesh : geometry->meshes()) {
                        combinedGeometry.meshes().push_back(std::move(mesh));
                    }

                    for (ModelGeometry::TextureEntry& s : geometry->textureStorage()) {
                        combinedGeometry.textureStorage().push_back(std::move(s));
                    }
                }
                _geometry = std::make_unique<ghoul::modelgeometry::ModelGeometry>(
                    std::move(combinedGeometry)
                );
                _geometry->calculateBoundingRadius();
            }
```

alexanderbock:
@micahnyc   Could you have a look through the models?  We need  to update them either by adding material files or replacing them with a single model file or else all models will turn grey after the merge

PR #1496, opened 2021-02-07
Feature/user profiles assets

micahnyc:
Added USER_ASSETS and USER_PROFILE tokens. 

New profiles are now saved in USER_PROFILE dir instead of PROFILE. 

USER_ASSETS are added to the end of the asset tree for the profile editing, then the token is expanded at load time.


micahnyc:
One thing I didn't like about this was that I was putting "${USER_ASSETS}" through out the code. I went with it because we also have ${DATA} ${ASSETS} ${RECORDINGS} ${PROFILES} in lots of other places so it at least felt consistent. It would be nice to define these somewhere.

alexanderbock:
This solution would split the user-generated profiles and the user-generated assets into two very different directories.  How about just having a `${USER_CONTENT}` that has a hard-coded `profile` and `asset` folder in which the custom files live?  That way, users would only have a single folder "user_data" to copy between installations (or map it to the `OpenSpaceData` folder.

I wouldn't like to have the folder by default outside the OpenSpace folder as users might inadvertently change a file that some other OpenSpace installation depends on

alexanderbock:
/config
/data/assets
/data/profiles
/recordings
/screenshots

micahnyc:
I feel like we talked about less ${PATH_TOKENS} and here I have made more 

`    USER = os.getenv("OPENSPACE_USER") or "${BASE}/user",
    USER_ASSETS = "${USER}/data/assets",
    USER_PROFILES = "${USER}/data/profiles",
    USER_CONFIG = "${USER}/config",`

but I just felt to dirty putting the strings /data/assets and /data/profiles and after we said to keep ${RECORDINGS} and ${SCREENSHOTS} i thought to make these. 

open for changes thou

alexanderbock:
I'm not that opposed to having more path tokens, even though in reality the vast majority will be kept unchanged, it might be nice to keep that flexibility

apps/OpenSpace/ext/launcher/include/launcherwindow.h
```diff
@@ -81,16 +81,21 @@ Q_OBJECT
     QWidget* createCentralWidget();
     void setBackgroundImage(const std::string& syncPath);
 
-    void openProfileEditor(const std::string& profile);
+    void openProfileEditor(const std::string& profile, const bool isUserProfile);
```
alexanderbock, line 84: The parameter doesn't have to be declared as `const`

apps/OpenSpace/ext/launcher/src/filesystemaccess.cpp
```diff
@@ -49,7 +49,8 @@ std::string FileSystemAccess::useQtFileSystemModelToTraverseDir(std::string dir)
 
 void FileSystemAccess::parseChildDirElements(QFileInfo fileInfo, std::string space,
                                          int level, std::vector<std::string>& dirNames,
-                                         std::vector<std::string>& output)
+                                         std::vector<std::string>& output,
+                                         bool userAsssets)
```
alexanderbock, line 53: Nitpicky;  seems like the `bool userAssets` could be on the same line as the `output` parameter.
And:  userAsssset has one s too many

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -330,14 +365,40 @@ void LauncherWindow::populateProfilesList(std::string preset) {
...
     }
 }
 
 void LauncherWindow::populateWindowConfigsList(std::string preset) {
     namespace fs = std::filesystem;
 
     _windowConfigBox->clear();
+
+    if (std::filesystem::exists(_userConfigPath)) {
+
```
alexanderbock, line 377: Extra empty line

apps/OpenSpace/main.cpp
```diff
@@ -1005,7 +1005,7 @@ std::string selectedSgctProfileFromLauncher(LauncherWindow& lw, bool hasCliSGCTC
             }
         }
         else {
-            config = "${CONFIG}/" + config + xmlExt;
+            config = config + xmlExt;
```
alexanderbock, line 1008: -> `config += xmlExt;`

src/interaction/sessionrecording.cpp
```diff
@@ -1713,7 +1713,10 @@ std::vector<std::string> SessionRecording::playbackList() const {
     std::vector<std::string> allInputFiles = currentDir.readFiles();
     for (const std::string& f : allInputFiles) {
         // Remove path and keep only the filename
-        fileList.push_back(f.substr(path.length() + 1, f.length() - path.length() - 1));
+        const std::string filename = f.substr(path.length() + 1, f.length() - path.length() - 1);
+        if (filename.rfind(".", 0) != 0) { //Don't add hidden files
```
alexanderbock, line 1717: a) Why the rfind?  I don't think this will work if you have a hidden file that has a file extension since it will find the last . in the file name that is the separator
b) This is a unix-way of finding hidden files.  On `WIN32` you'll need to instead check for:
```
DWORD attributes = GetFileAttributes(filename);
bool isHidden = attributes & FILE_ATTRIBUTE_HIDDEN;
```

src/scene/profile_lua.inl
```diff
@@ -71,7 +71,10 @@ int saveSettingsToProfile(lua_State* L) {
             absPath("${PROFILES}") + '/' + global::configuration->profile + ".profile";
         std::string destPath =
             absPath("${PROFILES}") + '/' + newFile + ".profile";
-
+        if (!FileSys.fileExists(sourcePath)) {
+            sourcePath = absPath("${USER_PROFILES}")
```
alexanderbock, line 75: Suggestion:
```cpp
sourcePath = fmt::format("{}/{}.profile", absPath("${USER_PROFILES}"), global::configuration->profile);
```

src/scene/profile_lua.inl
```diff
@@ -99,8 +102,10 @@ int saveSettingsToProfile(lua_State* L) {
     else if (saveFilePath.find('.') != std::string::npos) {
         return luaL_error(L, "Only provide the filename to save without file extension");
     }
-    const std::string absFilename = absPath("${PROFILES}/" + saveFilePath + ".profile");
-
+    std::string absFilename = absPath("${PROFILES}/" + saveFilePath + ".profile");
+    if (!FileSys.fileExists(absFilename)) {
+        absFilename = absPath("${USER_PROFILES}/" + saveFilePath + ".profile");
```
alexanderbock, line 107: See above

alexanderbock (approved):

PR #1497, opened 2021-02-09
Add drag and drop callback

alexanderbock:


scripts/drag_drop_handler.lua
```diff
@@ -0,0 +1,37 @@
+-- this script gets two parameters in its global scope:
```
WeirdRubberDuck, line 1: Nitpicky, but this sentence should start with a capital letter

WeirdRubberDuck (approved):
😍

scripts/drag_drop_handler.lua
alexanderbock: not need for "but";  "nitpicky" and "right" are not opposites

scripts/drag_drop_handler.lua
```diff
@@ -0,0 +1,37 @@
...
+  do return "" end
+end
+
+-- Lua doesn't enjoy \ that are used by Windows extensively. So we convert all \ into /
+filename = filename:gsub("\\", "/")
+basename = basename:gsub("\\", "/")
+basename_without_extension = basename:sub(0, #basename - extension:len())
+
+if     extension == ".jpg" or extension == ".jpeg" or extension == ".tif" or
+       extension == ".png" then
```
GPayne, line 26: I could see this section being added-to in the future with new file types. How about adding functions to check for file types, like:
`if isImage() then`
The function itself would test for the 4 file image types (a list which may expand) and make the evaluation of file types cleaner.

GPayne (approved):

scripts/drag_drop_handler.lua
alexanderbock: Good idea!

scripts/drag_drop_handler.lua
```diff
@@ -0,0 +1,43 @@
+-- This script gets two parameters in its global scope:
+-- filename:   The full path for the file that was dropped on the application.
+--             Example:  C:/OpenSpace/openspace.cfg
+-- basename:   Only the name of the actual file with extension, but without the full rest
+--             of the path.
+--             Example:  openspace.cfg
+-- extension:  The extention of the file
+--             Example: cfg
```
Roxeena, line 8: Shouldn't this be .cfg instead of cfg?

Roxeena (approved):

scripts/drag_drop_handler.lua
alexanderbock: Good eye!  It definitely should be

src/engine/openspaceengine.cpp
```diff
@@ -1444,6 +1445,52 @@ void OpenSpaceEngine::touchExitCallback(TouchInput input) {
     }
 }
 
+void OpenSpaceEngine::handleDragDrop(const std::string& file) {
+    std::filesystem::path f(file);
+
+    ghoul::lua::LuaState s(ghoul::lua::LuaState::IncludeStandardLibrary::Yes);
+    std::string absolutePath = absPath("${BASE}/scripts/drag_drop_handler.lua");
```
micahnyc, line 1452: ${SCRIPTS} vs ${BASE}/scripts

alexanderbock: Good point. Fixed

PR #1499, opened 2021-02-10
Expose exoplanet creation settings to user as module properties

WeirdRubberDuck:
Also rename 1 Au Ring to 1 Au Circle

### DIFF
--- a/modules/exoplanets/exoplanetsmodule.h
+++ b/modules/exoplanets/exoplanetsmodule.h
@@ -40,15 +40,15 @@ public:
     ExoplanetsModule();
     virtual ~ExoplanetsModule() = default;
 
-    const std::string exoplanetsDataPath() const;
-    const std::string lookUpTablePath() const;
-    const std::string starTexturePath() const;
-    const std::string noDataTexturePath() const;
-    const std::string orbitDiscTexturePath() const;
-    const std::string habitableZoneTexturePath() const;
-    const bool showComparisonCircle() const;
-    const bool showHabitableZone() const;
-    const bool useOptimisticZone() const;
+    std::string exoplanetsDataPath() const;
+    std::string lookUpTablePath() const;
+    std::string starTexturePath() const;
+    std::string noDataTexturePath() const;
+    std::string orbitDiscTexturePath() const;
+    std::string habitableZoneTexturePath() const;
+    bool showComparisonCircle() const;
+    bool showHabitableZone() const;
+    bool useOptimisticZone() const;
 
     scripting::LuaLibrary luaLibrary() const override;
     std::vector<documentation::Documentation> documentations() const override;
--- a/modules/exoplanets/exoplanetsmodule.h
+++ b/modules/exoplanets/exoplanetsmodule.h
@@ -28,6 +28,8 @@
 #include <openspace/util/openspacemodule.h>
 
 #include <openspace/documentation/documentation.h>
+#include <openspace/properties/scalar/boolproperty.h>
+#include <openspace/properties/stringproperty.h>
 
 namespace openspace {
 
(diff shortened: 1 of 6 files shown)


modules/exoplanets/exoplanetsmodule.h
```diff
@@ -38,11 +40,30 @@ class ExoplanetsModule : public OpenSpaceModule {
     ExoplanetsModule();
     virtual ~ExoplanetsModule() = default;
 
+    const std::string exoplanetsDataPath() const;
```
alexanderbock, line 43: No need to impose the `const`ness on the caller, since it doesn't do anything as a copy is returned anyway and that copy doesn't need to be `const`.  It should be up to the caller to decide whether they want to store it as `const` variable or not;
```
  std::vector a = module->exoplanetsDataPath();
  const std::vector b = module->exoplanetsDataPath();
```
both compile and do the same thing

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -419,14 +410,16 @@ void createExoplanetSystem(const std::string& starName) {
     const glm::dmat4 rotation = computeOrbitPlaneRotationMatrix(meanInclination);
     const glm::dmat3 meanOrbitPlaneRotationMatrix = static_cast<glm::dmat3>(rotation);
 
-    // 1 AU Size Comparison Ring
-    const std::string ringIdentifier = starIdentifier + "_1AU_Ring";
-    const std::string ring = "{"
-        "Identifier = '" + starIdentifier + "_1AU_Ring',"
+    bool isEnabled = module->showComparisonCircle();
+    const std::string isEnabledString = isEnabled ? "true" : "false";
```
alexanderbock, line 414: can be a `std::string_view` since the text assigned to it is a literal

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -463,16 +456,24 @@ void createExoplanetSystem(const std::string& starName) {
             "the greenhouse effect would not be able to maintain surface temperature "
             "above freezing anywhere on the planet.";
 
+        const std::string hzTexture = module->habitableZoneTexturePath();
+
+        bool isEnabled = module->showHabitableZone();
+        const std::string isEnabledString = isEnabled ? "true" : "false";
```
alexanderbock, line 462: See above

alexanderbock (approved):

modules/exoplanets/exoplanetsmodule_lua.inl
WeirdRubberDuck: Hmm, but `std::string_view` cannot be concatenated with an `std::string` using the `operator+` 🤔 which is done below

alexanderbock: Oh right; i forgot you used the "regular" concatenation... with fmt::format it would have worked, but that makes the whole line much more unwieldy...

So nevermind!

openspace.cfg
```diff
@@ -167,6 +167,19 @@ ModuleConfigurations = {
...
         ShowExceptions = false
+    },
+    Exoplanets = {
+        ShowComparisonCircle = false,
+        ShowHabitableZone = true,
+        UseOptimisticZone = true,
+        -- OBS! All these paths must be set
+        -- @TODO (emmbr 2021-02-10) this comment can be removed once we have
+        -- documentation for modules
+        DataFolder = "${SYNC}/http/exoplanets_data/2/",
```
micahnyc, line 178: I would suggest putting the setting of these values to be a lua script that is run in the onInitalize of the exoplanets_data.asset and exoplanets_textures.asset files. Right now these feel so hard coded to the asset being sync'd and never changing it's folder names, version numbers etc.

This would be similar to how the webgui.asset sets its cef route after it has synced the download of the html.

WeirdRubberDuck: Good point! I totally agree that setting them from the asset would be better. 

However, since the paths are not set to optional, not setting them in the `openspace.cfg` results in an uncaught SpecificationError (see issue #1502). Could get around this by setting the paths to optional, but the exoplanets stuff cannot be used without them, so there might be a risk of problems if a user tries to change things

 Also, if a user sets a custom path in the cfg file (to for example another texture), it will later be overwritten by the asset's `onInitialize`? 🤔 Not super clear to the user what's going on here, I think

micahnyc: Having them not optional just to satisfy the SpecificationError seems not the correct way to go. Is the code even checking that these files exists, or just that the specification is met. It feels like "oh the rules say this so I'm doing this"

The nice thing about them being optional is that then we wouldn't have to even put them in our .cfg file. 

The asset could check the module properties with getPropertyValue to see if they are set before it sets them itself. 

I believe is the first time we are trying to consider replacing  sync/ asset via a variable in the module configurations which is why we are coming to this unsolved issue.

WeirdRubberDuck: The specification is only checking that the path is set, but I agree that it would be better to actually check that the files actually exist as well. Maybe a specialized `StringVerifier` that does this could be added for files

And yes, seems like the first time. But it seems to me like the nicer option, for now, is to make the properties optional and do as you suggested: check if there is a value with `getPropertyValue` before setting the paths. Maybe it's also worth discussing at our next meeting, to see how we want to handle these scenarios in the future

WeirdRubberDuck: Alex also pointed out that making the properties optional might actually make sense. "if I run OpenSpace without the exoplanets module or I am certain I never want to create any exoplanets it is an optional value since I can run happily without it"

PR #1504, opened 2021-02-12
Issue/1466

Roxeena:


alexanderbock:
I think the expansion of the wildcard character needs to be a bit more generous.  It currently only applies to the PropertyValue part,  not the intermediate property owners.  The following examples don't currently work:
```
openspace.setPropertyValue("Scene.*.Renderable.Enabled", false);
openspace.setPropertyValue("*.Renderable.Enabled", false);
openspace.setPropertyValue("*Trail.*.Enabled", false);
```

Roxeena:
Oops, my previous commit apparently took away too many matches it seems like. I have fixed it now so those commands work now. I added a warning if more than one wildcard is used since I am not sure how to support an arbitrary number of wildcards :thinking:

alexanderbock:
Found another corner case that behaves differently:
```lua
openspace.setPropertyValue("Scene.MilkyWayVolum*.Renderable.Enabled", true)
```
matches against `Scene.MilkyWayVolume.Renderable.EnabledPointsRatio`  where as 
```lua
openspace.setPropertyValue("Scene.MilkyWayVolume.Renderable.Enabled", true)
```
correctly produces an error

Roxeena:
Hopefully it should work now

alexanderbock:
Yeey.. I can't break it anymore :)

PR #1511, opened 2021-02-25
Issue/1305 - Add a colored glare to exoplanet stars

### LINKED ISSUE #1305: Add a colored glare to the stars of exoplanet systems
Add a glare similar to the Sun's to the star of each exoplanet system. The glare should be tinted in the same color as the star.

The tricky bit is to figure out how to change the color of a texture from within OpenSpace, since that functionality does not exist yet

WeirdRubberDuck:
Been wanting this for a while and went for a pretty simple but sort of satisfying solution. The size of the glare scales depending on the radius and temperature of the star. 

With this change, all `RenderablePlane` subclasses can be tinted with a color, which makes sense to me and also seemed like the cleanest implementation (considered making a subclass specific for colored tinted planes but that just becomes messy really quickly with the current `RenderablePlane` implementation)

Some example images: 
| Trappist 1 (low luminosity) | HD 81817 (more luminous) |
| ------------- | ------------- |
| Trappist glare | Glare HD81817 |

Fixes #1305


jccosta:
Hi @WeirdRubberDuck. Did you test it in fisheye rendering mode?
If the plane is between two different frustums, you can have culling issues. You may need to increase the size of the plane while maintaining the glow size.

WeirdRubberDuck:
Thank for the tip @jccosta . Tested it now and cannot notice any issues. 

Note however that I've used the exact same rendering and texture as for the already existing Sun Glare. Is the risk of culling issues the reason why the Sun globe is not enabled per default? 🤔

alexanderbock:
The reason the sun globe is disabled by default is that it is rendered on top of the glare, which looks odd with a white glare and an orange dot with it

data/assets/scene/milkyway/exoplanets/exoplanets_textures.asset
```diff
@@ -22,6 +25,11 @@ asset.onInitialize(function ()
         openspace.setPropertyValueSingle(p, starTexture)
     end
 
+    local p = "Modules.Exoplanets.StarGlareTexture";
+    if(openspace.getPropertyValue(p) == "") then
```
alexanderbock, line 29: Adding an extra space between `if` and `(`

alexanderbock (approved):

micahnyc (approved):
works and looks great!

Issue #1484, opened 2021-02-01
adjusting RenderableTrail opacity setting creates rendering conflict with atmosphere.


micahnyc:
Setting the opacity of the ISS trail creates a rendering issue with the earth atmosphere, even if you set the opacity back to 1.0 the issue still remains.

![openspace_000002](https://user-images.githubusercontent.com/708535/106509851-2007f400-649c-11eb-8439-c30a2cc02108.png)

(As reported by Dan Tell)

micahnyc:
close with PR merge

PR #1519, opened 2021-03-03
prevent renderbin change for overlay renderables

micahnyc:
Addressing https://github.com/OpenSpace/OpenSpace/issues/1484 
When changing the opacity of a trail, it's renderbin was also being changed by the default renderable behavior. 

we already had a case to prevent this default behavior to renderables that are marked as RenderBin::PostDeferredTransparent
 so I add ones that are marked as RenderBin::Overlay to this case. 

This changes ultimately changes the order in which the trail will be rendered here: https://github.com/OpenSpace/OpenSpace/blob/cdbc1c2695b90fe1b9af1b40e5fd6ed8b577ecf1/src/rendering/framebufferrenderer.cpp#L1137 and thus making it 'not interact' with the atmosphere

### DIFF
--- a/src/rendering/renderable.cpp
+++ b/src/rendering/renderable.cpp
@@ -241,7 +241,7 @@ void Renderable::setRenderBinFromOpacity() {
 
 void Renderable::registerUpdateRenderBinFromOpacity() {
     _opacity.onChange([this](){
-        if (_renderBin != Renderable::RenderBin::PostDeferredTransparent) {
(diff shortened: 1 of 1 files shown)


alexanderbock:
Which renderable do we actually have that are rendered in the overlay bin?

WeirdRubberDuck:
> Which renderable do we actually have that are rendered in the overlay bin?

I believe it's only the trails that are per default, and that the RenderableImagePlane has an option to be rendered in overlay

GPayne:
This code change looks good to me. I assume you have tested this by viewing trails through an atmosphere?

PR #1531, opened 2021-03-12
Remove hardcoded path to B-V colormap

WeirdRubberDuck:
Also prevent future problems in reading the file (e.g. if number of lines of comments would change)

### DIFF
--- a/data/assets/scene/digitaluniverse/stars_colormap.asset
+++ b/data/assets/scene/digitaluniverse/stars_colormap.asset
@@ -15,7 +15,7 @@ asset.meta = {
     Description = [[A lookup table that maps a B-V color index to an RGB color. 
         The B-V values are in the range (-0.4, 2.0) and each line maps a value 
         in that range to a color]],
-    Author = "Brian Abbott (AMNH)",
+    Author = "OpenSpace Team",
     URL = "https://www.amnh.org/research/hayden-planetarium/digital-universe",
     License = "AMNH Digital Universe"
 }
--- a/data/assets/scene/digitaluniverse/stars_colormap.asset
+++ b/data/assets/scene/digitaluniverse/stars_colormap.asset
@@ -1,0 +1,22 @@
+local assetHelper = asset.require('util/asset_helper')
+
+local BvColorLUT = asset.syncedResource({
+    Name = "Stars Color Table",
+    Type = "HttpSynchronization",
+    Identifier = "stars_colormap",
+    Version = 2
+})
+
+asset.export("BvColorLUT", BvColorLUT)
+
+asset.meta = {
+    Name = "Stars B-V Colormap",
+    Version = "2.0",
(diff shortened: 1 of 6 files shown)


data/assets/scene/digitaluniverse/stars_colormap.asset
```diff
@@ -0,0 +1,22 @@
...
+
+asset.export("BvColorLUT", BvColorLUT)
+
+asset.meta = {
+    Name = "Stars B-V Colormap",
+    Version = "2.0",
+    Description = [[A lookup table that maps a B-V color index to an RGB color. 
+        The B-V values are in the range (-0.4, 2.0) and each line maps a value 
+        in that range to a color]],
+    Author = "Brian Abbott (AMNH)",
```
WeirdRubberDuck, line 18: @micahnyc what would be a good author here?

modules/exoplanets/exoplanetsmodule.cpp
```diff
@@ -123,6 +129,9 @@ namespace {
         // [[codegen::verbatim(DataFolderInfo.description)]]
         std::optional<std::filesystem::path> dataFolder [[codegen::directory()]];
 
+        // [[codegen::verbatim(StarTextureInfo.description)]]
```
WeirdRubberDuck, line 132: Oops, copy paste error. Should be "BvColorMapInfo"

data/assets/scene/digitaluniverse/stars_colormap.asset
micahnyc: I would just put "OpenSpace Team". That's what I've been putting when there isn't a specific author in mind/known.

It might have been Brian in this case but maybe he got it from somewhere else...

micahnyc (approved):

data/assets/scene/milkyway/exoplanets/exoplanets_data.asset
```diff
@@ -10,6 +12,11 @@ asset.onInitialize(function ()
     if(openspace.getPropertyValue(p) == "") then
```
alexanderbock, line 12: Nity picky;  should be `if (` with a space

modules/exoplanets/exoplanetshelper.cpp
```diff
@@ -57,19 +58,37 @@ bool hasSufficientData(const ExoplanetDataEntry& p) {
...
         ));
-        return glm::vec3(0.f, 0.f, 0.f);
+        return glm::vec3(0.f);
     }
 
-    const int t = static_cast<int>(round(((bv + 0.4) / (2.0 + 0.4)) * 255));
+    // Interpret the colormap cmap file
+    std::string line;
+    while (std::getline(colorMap, line)) {
+        if ((line[0] == '#') || line.empty()) {
```
alexanderbock, line 77: You should check the other way around;  otherwise if the line is empty, the `line[0]` will lead to a crash

alexanderbock (approved):

PR #1550, opened 2021-03-29
Update Apollo models

Roxeena:


alexanderbock:
Looks good!  There is currently a bug with the HTTPSynchronization that means that if multiple independent assets do an HTTPSyncrhonization on the same identifier, the files won't get downloaded correctly.   For that reason, we put the HTTPSynchronization into its own assets and `require` it from everywhere else.  I think we should do the same here

alexanderbock (approved):

PR #1552, opened 2021-03-31
Feature/model animation

Roxeena:


modules/base/rendering/renderablemodel.cpp
```diff
@@ -62,6 +62,12 @@ namespace {
         { "Color Adding", ColorAddingBlending }
     };
 
+    constexpr openspace::properties::Property::PropertyInfo enableAnimationInfo = {
```
alexanderbock, line 65: Since it is a constant, the first letter would be capitalized

modules/base/rendering/renderablemodel.cpp
```diff
@@ -62,6 +62,12 @@ namespace {
...
+        "EnableAnimation",
+        "Enable Animation",
+        "Enable Animation"
```
alexanderbock, line 68: Maybe some more comprehensive description what the property does?

modules/base/rendering/renderablemodel.cpp
```diff
@@ -126,7 +132,7 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo BlendingOptionInfo = {
-        "BledingOption",
+        "BlendingOption",
         "Blending Options",
         "Debug option for blending colors."
```
alexanderbock, line 137: Same here.  Would be good to have some more comprehensive description

modules/base/rendering/renderablemodel.cpp
```diff
@@ -136,110 +142,111 @@ namespace {
         "Enable Opacity Blending",
         "Enable Opacity Blending."
     };
+
+    struct [[codegen::Dictionary(RenderableModel)]] Parameters {
+        // The file or files that should be loaded in this RenderableModel. The file can
+        // contain filesystem tokens or can be specified relatively to the
+        // location of the .mod file.
```
alexanderbock, line 149: .mod -> asset

modules/base/rendering/renderablemodel.cpp
```diff
@@ -136,110 +142,111 @@ namespace {
...
+        // Set if invisible parts (parts with no textures or materials) of the model
+        // should be forced to render or not.
+        std::optional<bool> forceRenderInvisible;
+
+        // [[codegen::verbatim(enableAnimationInfo.description)]]
+        std::optional<bool> enableAnimation;
+
+        // The date and time that the model animation should start.
+        // In format 'YYYY MM DD hh:mm:ss'.
+        std::optional<std::string> animationStartTime [[codegen::dateTime()]];
```
alexanderbock, line 176: Currently, all of the attributes (with the exception of the `Dictionary` are all lowercaps.  I think it would be good to have that here as well and use `codegen::datetime` instead for consistency

modules/base/rendering/renderablemodel.cpp
```diff
@@ -136,110 +142,111 @@ namespace {
...
+        std::optional<bool> disableFaceCulling;
+
+        // [[codegen::verbatim(ModelTransformInfo.description)]]
+        std::optional<glm::dmat3x3> modelTransform;
+
+        // [[codegen::verbatim(RotationVecInfo.description)]]
+        std::optional<glm::dvec3> rotationVector;
+
+        // [[codegen::verbatim(LightSourcesInfo.description)]]
+        std::optional<std::vector<ghoul::Dictionary>> lightSources [[codegen::reference("core_light_source")]];
```
alexanderbock, line 225: This looks like it exceeds the line limit

modules/base/rendering/renderablemodel.cpp
```diff
@@ -279,103 +282,173 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            _scaleVector = glm::dvec3(convertUnit(_modelScale, DistanceUnit::Meter));
         }
-        else if (dictionary.hasValue<ghoul::Dictionary>(KeyGeomModelFile)) {
-            LWARNING("Loading a model with several files is deprecated and will be "
-                "removed in a future release"
-            );
+        else if (std::holds_alternative<float>(*p.modelScale)) {
+            _scaleVector = glm::dvec3(std::get<float>(*p.modelScale));
+        }
+    }
```
alexanderbock, line 328: I started to put a 
```cpp
else {
}
```
in here as well so that there will be a hard crash when someone adds something to the variant but forgets to check it here

modules/base/rendering/renderablemodel.cpp
```diff
@@ -279,103 +282,173 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
-                        );
-                    }
-                }
-                 _geometry = std::move(combinedGeometry);
-                _geometry->calculateBoundingRadius();
+    if (p.enableAnimation.has_value()) {
+        if (!_geometry->hasAnimation()) {
+            LWARNING("Attempting to enable animation for a model that does not have any");
+        }
+        else if (*p.enableAnimation &&_animationStart == "") {
```
alexanderbock, line 341: Nitpicky:  missing space after the `&&`

For std::string it's better to check `_animationStart.empty()`  as that way the `""` doesn't need to be converted to a `std::string` that then gets passed to the `operator==` function

modules/base/rendering/renderablemodel.cpp
```diff
@@ -279,103 +282,173 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+                    break;
+                case Parameters::AnimationTimeUnit::Minute:
+                    timeUnit = TimeUnit::Minute;
+                    break;
+                default:
+                    throw ghoul::MissingCaseException();
             }
+
+            _geometry->setTimeScale(convertTime(1.0, timeUnit, TimeUnit::Second));
+        }
```
alexanderbock, line 383: Same as above re ghoul::MissingCaseException

modules/base/rendering/renderablemodel.cpp
```diff
@@ -279,103 +282,173 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            case Parameters::AnimationMode::Once:
+            default:
+                _animationMode = AnimationMode::Once;
+                break;
         }
     }
 
-    if (dictionary.hasKey(ModelTransformInfo.identifier)) {
-        _modelTransform = dictionary.value<glm::dmat3>(ModelTransformInfo.identifier);
+    if (p.modelTransform.has_value()) {
```
alexanderbock, line 411: More concise to write:
`_modelTransform = p.modelTransform.value_or(_modelTransform);`
and the same for the rest here

modules/base/rendering/renderablemodel.cpp
```diff
@@ -390,9 +463,22 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
         _modelTransform = glm::mat4_cast(glm::quat(glm::radians(_rotationVec.value())));
     });
 
+    _enableAnimation.onChange([this]() {
+        if (!_geometry->hasAnimation()) {
+            LWARNING("Attempting to enable animation for a model that does not have any");
+        }
+        else if (_enableAnimation.value() && _animationStart == "") {
```
alexanderbock, line 470: See above re std::string::empty

alexanderbock: Also the `.value()` call shouldn't be necessary as the BoolProperty has an `operator bool()` function overload

modules/base/rendering/renderablemodel.cpp
```diff
@@ -390,9 +463,22 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            LWARNING("Cannot enable animation without a given start time");
+            _enableAnimation = false;
+        }
+        else {
+            _geometry->enableAnimation(_enableAnimation.value());
```
alexanderbock, line 475: same as above

modules/base/rendering/renderablemodel.cpp
```diff
@@ -279,103 +282,173 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+    if (p.enableAnimation.has_value()) {
+        if (!_geometry->hasAnimation()) {
+            LWARNING("Attempting to enable animation for a model that does not have any");
+        }
+        else if (*p.enableAnimation &&_animationStart == "") {
+            LWARNING("Cannot enable animation without a given start time");
+        }
+        else {
+            _enableAnimation = *p.enableAnimation;
+            _geometry->enableAnimation(_enableAnimation.value());
```
alexanderbock, line 346: No need for the `.value()` call as the BoolProperty has an overload that returns a `bool` implicitly  (all TemplatePropertys do)

modules/base/rendering/renderablemodel.cpp
```diff
@@ -390,9 +463,22 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
         _modelTransform = glm::mat4_cast(glm::quat(glm::radians(_rotationVec.value())));
```
alexanderbock, line 463: Same here re `.value()`

modules/base/rendering/renderablemodel.cpp
```diff
@@ -390,9 +463,22 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+        }
+        else {
+            _geometry->enableAnimation(_enableAnimation.value());
+        }
+    });
+
 
-    if (dictionary.hasKey(RotationVecInfo.identifier)) {
-        _rotationVec = dictionary.value<glm::dvec3>(RotationVecInfo.identifier);
+    if (p.rotationVector.has_value()) {
```
alexanderbock, line 480: See above

modules/base/rendering/renderablemodel.cpp
```diff
@@ -426,6 +508,15 @@ bool RenderableModel::isReady() const {
 void RenderableModel::initialize() {
     ZoneScoped
 
+    if (_geometry->hasAnimation() && _enableAnimation.value() && _animationStart == "") {
```
alexanderbock, line 511: see above re `.value()` and `std::string::empty`

modules/base/rendering/renderablemodel.cpp
```diff
@@ -426,6 +508,15 @@ bool RenderableModel::isReady() const {
...
+        LWARNING("Model with animation not given any start time");
+    }
+    else if (_geometry->hasAnimation() && !_enableAnimation.value()) {
```
alexanderbock, line 514: ^^

modules/base/rendering/renderablemodel.cpp
```diff
@@ -449,7 +540,7 @@ void RenderableModel::initializeGL() {
 
     _geometry->initialize();
     _geometry->calculateBoundingRadius();
-    setBoundingSphere(glm::sqrt(_geometry->boundingRadius()));
+    setBoundingSphere(glm::sqrt(_geometry->boundingRadius()) * _scaleVector.x);
```
alexanderbock, line 543: Wouldn't it be possible to specify different scale values for each direction?  In which case multiplying with 
`glm::compMax(_scaleVector)` (https://glm.g-truc.net/0.9.5/api/a00191.html) would give the better pessimistic bounding sphere

modules/base/rendering/renderablemodel.cpp
```diff
@@ -466,74 +557,87 @@ void RenderableModel::deinitializeGL() {
...
+    );
 
-    _program->setUniform(_uniformCache.opacity, _opacity);
+    // This distance will be enough to render the globe as one pixel if the field of
+    // view is 'fov' radians and the screen resolution is 'res' pixels.
+    //constexpr double fov = 2 * glm::pi<double>() / 6; // 60 degrees
+    //constexpr double tfov = tan(fov / 2.0); // doesn't work unfortunately
+    constexpr double tfov = 0.5773502691896257;
+    constexpr int res = 2880;
+    const double maxDistance = res * boundingSphere()/ tfov;
```
alexanderbock, line 571: nitpicky:  missing space before `/`

modules/base/rendering/renderablemodel.cpp
```diff
@@ -466,74 +557,87 @@ void RenderableModel::deinitializeGL() {
...
+
+        if (_disableFaceCulling) {
+            glDisable(GL_CULL_FACE);
+        }
 
-    glEnablei(GL_BLEND, 0);
-    switch (_blendingFuncOption) {
+        glEnablei(GL_BLEND, 0);
+        switch (_blendingFuncOption) {
         case DefaultBlending:
```
alexanderbock, line 641: Nitpicky;  `case` label should be indented one level deeper

modules/base/rendering/renderablemodel.cpp
```diff
@@ -549,31 +653,68 @@ void RenderableModel::render(const RenderData& data, RendererTasks&) {
...
     }
+
+    if (_geometry->hasAnimation() && _animationStart != "") {
+        double realtiveTime;
+        double now = data.time.j2000Seconds();
+        double startTime = data.time.convertTime(_animationStart);
+        double duration = _geometry->animationDuration();
+
+        switch (_animationMode) {
+            case AnimationMode::LoopFromStart:
```
alexanderbock, line 690: Maybe a comment here describing the idea behind remapping the relative time would be good for future-us that will come back to this code at some point

modules/base/rendering/renderablemodel.cpp
```diff
@@ -549,31 +653,68 @@ void RenderableModel::render(const RenderData& data, RendererTasks&) {
...
+            case AnimationMode::BounceInfinitely: {
+                double modulo = fmod(now - startTime, 2 * duration);
+                if (modulo < 0) {
+                    modulo += 2 * duration;
+                }
+                realtiveTime = duration - abs(modulo - duration);
+                break;
+            }
+            case AnimationMode::Once:
+            default:
```
alexanderbock, line 712: Maybe better to move the code actually into the `Once` case and throw a `ghoul::MissingCaseException` here in the `default` label

modules/base/rendering/renderablemodel.cpp
```diff
@@ -279,103 +282,173 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+                _animationMode = AnimationMode::LoopInfinitely;
+                break;
+            case Parameters::AnimationMode::BounceFromStart:
+                _animationMode = AnimationMode::BounceFromStart;
+                break;
+            case Parameters::AnimationMode::BounceInfinitely:
+                _animationMode = AnimationMode::BounceInfinitely;
+                break;
+            case Parameters::AnimationMode::Once:
+            default:
```
alexanderbock, line 405: Maybe better to move the code actually into the `Once` case and throw a `ghoul::MissingCaseException` here in the `default` label

alexanderbock: Otherwise, there is the danger of someone adding a new case and then needs to figure out where the missing label is, rather than getting a proper crash

modules/base/rendering/renderablemodel.cpp
```diff
@@ -549,31 +653,68 @@ void RenderableModel::render(const RenderData& data, RendererTasks&) {
...
 
-void RenderableModel::update(const UpdateData&) {
+void RenderableModel::update(const UpdateData& data) {
     if (_program->isDirty()) {
         _program->rebuildFromFile();
         ghoul::opengl::updateUniformLocations(*_program, _uniformCache, UniformNames);
     }
+
+    if (_geometry->hasAnimation() && _animationStart != "") {
+        double realtiveTime;
```
alexanderbock, line 684: `realtiveTime` -> `relativeTime`

modules/base/rendering/renderablemodel.cpp
```diff
@@ -549,31 +653,68 @@ void RenderableModel::render(const RenderData& data, RendererTasks&) {
...
 }
 
-void RenderableModel::update(const UpdateData&) {
+void RenderableModel::update(const UpdateData& data) {
     if (_program->isDirty()) {
         _program->rebuildFromFile();
         ghoul::opengl::updateUniformLocations(*_program, _uniformCache, UniformNames);
     }
+
+    if (_geometry->hasAnimation() && _animationStart != "") {
```
alexanderbock, line 683: See above re `_animationStart.empty()`

modules/base/rendering/renderablemodel.h
```diff
@@ -70,9 +71,22 @@ class RenderableModel : public Renderable {
...
+        BounceFromStart,
+        BounceInfinitely
+    };
+
     std::unique_ptr<ghoul::modelgeometry::ModelGeometry> _geometry;
+    DistanceUnit _modelScale;
+    glm::dvec3 _scaleVector = glm::dvec3(1.0, 1.0, 1.0);
     bool _forceRenderInvisible = false;
     bool _notifyInvisibleDropped = true;
+    std::string _animationStart = "";
```
alexanderbock, line 87: No need to initialize a `std::string` with `""` as the default constructor of `std::string` does that already

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
```diff
@@ -73,131 +75,66 @@ namespace {
...
     _projectionComponent.initialize(
         identifier(),
-        dictionary.value<ghoul::Dictionary>(keyProjection)
+        p.projection
     );
 
     double boundingSphereRadius = 1.0e9;
-    if (dictionary.hasValue<double>(keyBoundingSphereRadius)) {
-        boundingSphereRadius = dictionary.value<double>(keyBoundingSphereRadius);
+    if (p.boundingSphereRadius.has_value()) {
```
alexanderbock, line 131: More concise version to write this:
`double boundingSphereRadius = p.boundingSphereRadius.value_or(1.0e9);`

modules/spacecraftinstruments/rendering/renderablemodelprojection.cpp
```diff
@@ -73,131 +75,66 @@ namespace {
...
+        boundingSphereRadius = *p.boundingSphereRadius;
     }
     setBoundingSphere(boundingSphereRadius);
 
-    if (dictionary.hasValue<bool>(PerformShadingInfo.identifier)) {
-        _performShading = dictionary.value<bool>(PerformShadingInfo.identifier);
+    if (p.performShading.has_value()) {
```
alexanderbock, line 136: Same here:
`_performShading = p.performShading.value_or(_performShading);`

src/documentation/verifier.cpp
```diff
@@ -271,6 +273,57 @@ std::string DirectoryVerifier::type() const {
     return "Directory";
 }
 
+TestResult DateTimeVerifier::operator()(const ghoul::Dictionary& dict,
+    const std::string& key) const
```
alexanderbock, line 277: Nitpicky:  the second line should be aligned with the opening `(` of the parameter list on the line above

src/documentation/verifier.cpp
```diff
@@ -271,6 +273,57 @@ std::string DirectoryVerifier::type() const {
...
+        off.reason = TestResult::Offense::Reason::Verification;
+        off.explanation = "Not a valid format, should be: YYYY MM DD hh:mm:ss";
+        res.offenses.push_back(off);
+    }
+    // then check if valid date
+    else {
+        // normalize e.g. 29/02/2013 would become 01/03/2013
+        std::tm t_copy(t);
+        time_t when = mktime(&t_copy);
+        std::tm *norm = localtime(&when);
```
alexanderbock, line 305: We attach the pointer to the type as it is a clarifier of the type and not the name

alexanderbock (changes requested):

modules/base/rendering/renderablemodel.cpp
Roxeena: See above :)

modules/base/rendering/renderablemodel.cpp
Roxeena: I tried to take away the `.value()` here but it didn't work :(

modules/base/rendering/renderablemodel.cpp
Roxeena: I agree because I'm not sure what this property does :sweat_smile:

tests/test_distanceconversion.cpp
```diff
@@ -0,0 +1,258 @@
...
+    REQUIRE(res == Approx(meters / (10.0 * 20.1168)));
+
+    res = convertDistance(meters, DistanceUnit::Meter, DistanceUnit::Mile);
+    REQUIRE(res == Approx(meters / 1609.344));
+
+    res = convertDistance(meters, DistanceUnit::Meter, DistanceUnit::League);
+    REQUIRE(res == Approx(meters / (3.0 * 1609.344)));
+}
+
+TEST_CASE("DistanceConversion: Cross convertions", "[distanceconversion]") {
```
WeirdRubberDuck, line 186: Should be "Cross conversion"

src/documentation/verifier.cpp
```diff
@@ -271,6 +273,57 @@ std::string DirectoryVerifier::type() const {
     return "Directory";
 }
 
+TestResult DateTimeVerifier::operator()(const ghoul::Dictionary& dict,
```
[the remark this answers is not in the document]
Roxeena: #1563 :relaxed:

alexanderbock (approved):
Looks good to me :)

PR #1561, opened 2021-04-15
Feature/interactionsphere

alexanderbock:


alexanderbock:
- This pull request separates the boundingsphere into a new boundingsphere that is the limit of all vertices for a scenegraph node and an interaction sphere that is used for the camera movement restrictions.
 - There are also some places where the boundingsphere was not computed correctly
 - Added a debug rendering option to show the bounding sphere for a scene graph node

modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
```diff
@@ -1531,17 +1494,24 @@ void RenderableBillboardsCloud::createDataSlice() {
...
             1.0
-        );
-        // W-normalization
-        transformedPos /= transformedPos.w;
-        glm::vec4 position(glm::vec3(transformedPos), static_cast<float>(_unit));
+        ));
+        glm::vec4 position(transformedPos, static_cast<float>(_unit));
+
+        const double unitMeter = unitToMeter(_unit);
+        glm::dvec3 p = glm::dvec3(glm::dvec3(position) * unitMeter);
```
GPayne, line 1510: You could also just do:
`glm::dvec3 p = glm::dvec3(position * unitMeter);`

modules/digitaluniverse/rendering/renderableplanescloud.cpp
```diff
@@ -1099,12 +1099,43 @@ void RenderablePlanesCloud::createPlanes() {
...
+                    scale = static_cast<float>(1e6 * PARSEC);
+                    break;
+                case Gigaparsec:
+                    scale = static_cast<float>(1e9 * PARSEC);
+                    break;
+                case GigalightYears:
+                    scale = static_cast<float>(306391534.73091 * PARSEC);
+                    break;
+            }
+
```
GPayne, line 1128: I like the switch case format above in RenderableBillboardsCloud::unitToMeter. How about doing it that way leaving scale as a double, and then casting to a float after the switch case?

modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
alexanderbock: Had to do `glm::dvec3(position) * unitMeter` as `glm::dvec3(position * unitMeter)` doesn't compile since position is glm::vec3 and glm::vec3 * double doesn't work

modules/digitaluniverse/rendering/renderableplanescloud.cpp
alexanderbock: Good point  -> changed

src/rendering/renderable.cpp
```diff
@@ -179,12 +167,16 @@ double Renderable::boundingSphere() const {
...
+    return _interactionSphere;
+}
+
 SurfacePositionHandle Renderable::calculateSurfacePositionHandle(
                                                  const glm::dvec3& targetModelSpace) const
 {
     const glm::dvec3 directionFromCenterToTarget = glm::normalize(targetModelSpace);
     return {
-        directionFromCenterToTarget * boundingSphere(),
+        directionFromCenterToTarget * _parent->interactionSphere(),
```
WeirdRubberDuck, line 179: Changing to the interaction sphere here makes sense, as this function is mainly used for navigation. Although, I think the name of the function becomes a little misguiding, as we are no longer computing a handle to the surface of the renderable (the interaction sphere could theoretically be very different from the actual "surface" of the renderable). 

Don't have a good idea about a better name though, so maybe it's enough to add an explaining comment for the function in the header file? Pointing out that it is a handle to a point on the interaction sphere

WeirdRubberDuck (changes requested):
Looks good, and I'm excited about this change! 

Although, there seems to be a problem with the Milky Way Galaxy Image no longer fading out close to our solar system (it's bounding sphere is set to -1). This should be fixed before merge

image

src/rendering/renderable.cpp
alexanderbock: I added a comment to it'  the SurfacePositionHandle was already documented generally enough:
```
 * Defines the position of an object relative to a surface. The surface is defined as
 * a reference surface together with a height offset from that reference surface.
 *
```

alexanderbock:
Addressed the comments;  the milkyway was a wrong "fix" for the previous PRs and was fixed

WeirdRubberDuck (approved):

PR #1568, opened 2021-04-23
Fix that allows all days of month except monthly increment still at 1-28

GPayne:
Timequantizer still requires start/end days to be in 1-28 range for monthly increment, but allows the full day range for any other increment.

### DIFF
--- a/modules/globebrowsing/src/timequantizer.cpp
+++ b/modules/globebrowsing/src/timequantizer.cpp
@@ -371,17 +371,31 @@ void TimeQuantizer::setResolution(const std::string& resolutionString) {
 
 void TimeQuantizer::verifyStartTimeRestrictions() {
     //If monthly time resolution then restrict to 28 days so every month is consistent
-    unsigned int dayUpperLimit = (_resolutionUnit == 'M') ? 28 : 31;
+    unsigned int dayUpperLimit;
+    std::string helpfulDescription = "the selected month";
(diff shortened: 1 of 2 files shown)


modules/globebrowsing/src/timequantizer.cpp
```diff
@@ -366,13 +366,17 @@ void TimeQuantizer::setStartEndRange(const std::string& start, const std::string
 
 void TimeQuantizer::setResolution(const std::string& resolutionString) {
     _resolution = parseTimeResolutionStr(resolutionString);
+    verifyStartTimeRestrictions();
 }
 
 void TimeQuantizer::verifyStartTimeRestrictions() {
-    if (_start.day() < 1 || _start.day() > 28) {
+    //If monthly time resolution then restrict to 28 days so every month is consistent
+    unsigned int dayUpperLimit = (_resolutionUnit == 'M') ? 28 : 31;
```
alexanderbock, line 374: I'd also check for yearly;  otherwise someone might want to ask for a start date of february 29th in a leap year and we have same issue.
Super edge case, but still

alexanderbock (approved):

modules/globebrowsing/src/timequantizer.cpp
GPayne: Added a case for yearly in commit dd6aba5

micahnyc (approved):
tested and working

PR #1593, opened 2021-05-18
Feature/space mouse

Roxeena:


include/openspace/interaction/joystickcamerastates.h
```diff
@@ -60,7 +61,15 @@ class JoystickCameraStates : public CameraInteractionStates {
         AxisInvert invert = AxisInvert::No;
         AxisNormalize normalize = AxisNormalize::No;
 
+        // The axis values can either go back to 0 when the joystick is released or it can
+        // stay at the value it was before the joystick was released.
+        // The latter is called a static axis, when the values don't go back to 0.
+        bool isStatic = false;
```
alexanderbock, line 67: I'd imagine that a "static axis" is an axis that doesn't change..  Maybe a different term might be better?   Not really sure what a good term would be.. .  "integrated axis"?

src/interaction/joystickcamerastates.cpp
```diff
@@ -67,49 +72,55 @@ void JoystickCameraStates::updateStateFromInput(const InputState& inputState,
             value *= -1.f;
         }
 
-        value = static_cast<float>(value * _sensitivity);
+        if (std::fabs(t.sensitivity) > std::numeric_limits<double>().epsilon()) {
```
alexanderbock, line 75: `std::abs` should be fine here since `t.sensitivity` is already a `float`?  And should probably be compared to `std::numeric_limits<float>::epsilon()`   Checking against `double` would be the same as checking against > as the double-epsilon will be smaller than the smaller representable float.   and b) no need to create an object of numeric_limits and then call `epsilon()` on that object since `epsilon` is a static function

alexanderbock (approved):
Looks great to me!

include/openspace/interaction/joystickcamerastates.h
Roxeena: I agree that it is not a perfect term. How about sticky? It feels more intuitive that it can be moved but it will stick to the previous value when let go.

src/interaction/joystickcamerastates.cpp
Roxeena: Both `t.sensitivity` and `_sensitivity` are doubles, so then it should still be `std::fabs` and `std::numeric_limits<double>().epsilon()` right?

include/openspace/interaction/joystickcamerastates.h
alexanderbock: I can live with "sticky" :)

src/interaction/joystickcamerastates.cpp
alexanderbock: Ah, yes to `double` in numeric limits, but you can still use `std::abs`  (see (2) in https://en.cppreference.com/w/cpp/numeric/math/fabs).  No need to worry about the type of the value then.

And `std::numeric_limits<double>::epsilon()`, not `std::numeric_limits<double>().epsilon()` :)

PR #1604, opened 2021-05-22
Help menu to reduce learning curve

CreaticDD:
Currently Windows platform only. Is hidden on other platforms until added.

I don't have Mac or Linux so will depend on someone else to add these.


alexanderbock:
I'm closing this in favor of a solution to open these webpages on the UI side instead of going through the core

PR #1609, opened 2021-05-25
Feature/numeric slider updates

### LINKED ISSUE #1572: Logarithmic sliders should use exponent from NumericalProperty
NumericalProperties have an exponent parameter that currently isn't used anywhere. It could be used by the GUI to set the exponent for the exponential scaling of the slider. That way, each renderable could set which exponent it finds suitable for its specific property.

### LINKED ISSUE #1581: Min-max slider ViewOption for Vec2Properties
Something like this, but OpenSpace style: 
image

WeirdRubberDuck:
Sorry for the huge PR... Here's a summary of the changes:

* Remove Logarithmic ViewOption and instead always use the NumericProperties' exponent (closes #1572)
* Add exponent to some relevant numeric properties, to make their sliders logarithmic
* Add ViewOption for MinMaxRange slider for vec2 properties (closes #1581)
* Update the properties of some renderables in order to utilize the MinMaxRange view option (**breaking change!**). Also updates the relevant assets
* Add more detailed documentation for labels functionality
* Slight refactoring of labels code

Related PR in the WebGui repo: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/48


modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
```diff
@@ -290,11 +280,8 @@ namespace {
         // objects being rendered
         std::optional<std::string> labelFile;
```
alexanderbock, line 281: Would it make sense to move all of the label-related properties into a subtable instead?
Making:
```lua
  Renderable = {
    Type = "RenderableBillboardsCloud",
    Enabled = false,
    Color = { 0.4, 0.4, 0.9 },
    Opacity = 0.65,
    File = speck .. "/pn.speck",
    Texture = textures .. "/point4.png",
    PolygonSides = 3,
    LabelFile = speck .. "/pn.label",
    Unit = "pc",
    TextColor = { 0.25, 0.25, 0.65 },
    ScaleFactor = 425.0,
    TextSize = 16.24,
    TextMinMaxSize = { 4, 25 },
    BillboardMinMaxSize = { 0.0, 500.0 },
    EnablePixelSizeControl = true
  },
```
into
```lua
  Renderable = {
    Type = "RenderableBillboardsCloud",
    Enabled = false,
    Color = { 0.4, 0.4, 0.9 },
    Opacity = 0.65,
    File = speck .. "/pn.speck",
    Texture = textures .. "/point4.png",
    PolygonSides = 3,
    Unit = "pc",
    Labels = {
      File = speck .. "/pn.label",
      Color = { 0.25, 0.25, 0.65 },
      Size = 16.24,
      MinMaxSize = { 4, 25 },
    },
    ScaleFactor = 425.0,
    BillboardMinMaxSize = { 0.0, 500.0 },
    EnablePixelSizeControl = true
  },
```

Maybe not worth it, but worth considering at least

data/assets/scene/solarsystem/planets/saturn/hyperion/hyperion.asset
```diff
@@ -26,18 +26,14 @@ local Hyperion = {
     SegmentsPerPatch = 64,
     Layers = {},
     Labels = {
-      Enable = false,
+      -- @TODO (2021-05-21, emmbr26) These labels do not seem to be visible
+      Enabled = false,
       FileName = labelsPath .. "/hyperion.labels",
       LabelAlignmentOption = "Horizontally", -- or Circularly
```
alexanderbock, line 32: Since the label variables are in the `Labels` table, I think we would drop the "Label" prefix from them?
`LabelAlignmentOption` -> `AlignmentOption`
`LabelsFontSize` -> `FontSize`

etc

alexanderbock (commented):
- Not sure if it is related to this PR, but the "1 AU Size Comparison Grid" doesn't show up when enabled.  Tried with "11 Com" and "GJ 1265".  Enabling the property in the menu didn't make the grid show up.  Tried manipulating its other properties to force a recalculation, but no cigar
 - Could Minmax slider also be used for the "Kopparapu TEFF" value in the habitable zone?
 - Please don't kill me for this;  but maybe we should not highlight the properties that are set as "ReadOnly"?  I tried to figure out why I couldn't change the RenderableHabitableZone.size and width properties and there is no indicator that they are read-only in the UI

data/assets/scene/digitaluniverse/starorbits.asset
```diff
@@ -20,10 +20,8 @@ local sunOrbit = {
     File = speck .. "/starorbits-Sun.speck",
     MeshColor = {{ 1.0, 0.65, 0.0 }},
     --LabelFile = speck .. "/1ld.label",
-    TextColor = { 0.0, 0.2, 0.5 },
-    TextSize = 10.3,
-    TextMinSize = 0.5,
-    TextMaxSize = 30.0,
+    -- TextColor = { 0.0, 0.2, 0.5 },
```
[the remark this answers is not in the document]
WeirdRubberDuck: Commented them out just to clarify that they are actually not used. The label related properties are only read if the `LabelFile` is set, which it isn't here

data/assets/scene/solarsystem/planets/saturn/hyperion/hyperion.asset
WeirdRubberDuck: I agree

data/assets/scene/digitaluniverse/sdss.asset
```diff
@@ -30,16 +30,12 @@ local object = {
     ColorRange = { { 0.0, 0.075 }, { 1.0, 50.0 } },
     Texture = textures .. "/point3A.png",
     Unit = "Mpc",
-    -- Fade in value in the same unit as "Unit"
-    FadeInDistances = { 220.0, 650.0 },
-    BillboardMaxSize = 50.0,
-    BillboardMinSize = 0.0,
+    FadeInDistances = { 220.0, 650.0 }, -- Fade in value in the same unit as "Unit"
+    BillboardMinMaxSize = { 0.0, 5.5 },
```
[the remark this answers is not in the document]
WeirdRubberDuck: Yep. The Max size was actually set twice in this asset and 5.5 comes from what was previously line 42

WeirdRubberDuck:
> 1. Not sure if it is related to this PR, but the "1 AU Size Comparison Grid" doesn't show up when enabled.  Tried with "11 Com" and "GJ 1265".  Enabling the property in the menu didn't make the grid show up.  Tried manipulating its other properties to force a recalculation, but no cigar
> 2. Could Minmax slider also be used for the "Kopparapu TEFF" value in the habitable zone?
> 3. Please don't kill me for this;  but maybe we should not highlight the properties that are set as "ReadOnly"?  I tried to figure out why I couldn't change the RenderableHabitableZone.size and width properties and there is no indicator that they are read-only in the UI

Thanks for doing such a thorough review! And thank you for remembering to check the exoplanets related things. I did an extremely thorough go-through of all the assets, but apparently forgot the exoplanet systems... 

1. That is surely related to this PR, as the inner and outer radii properties used for the "1 AU Size Comparison Circle" has been renamed and hence is no longer set correctly. Will fix

2. Yes, it could. I was unsure whether it would actually be useful, as it is not changeable and it's main use is to just communicate those two numbers, so left it as a vec2 view. But when trying it out now it does make sense

3. I absolutely agree! The read-only are often pretty confusing. I think we could make them feel less "interactible" and also maybe put "read only" or "non-changeable" in text somewhere, either in the label or in the beginning of the info tooltip. Should be easy, as the UI knows when a property is read-only. This is done on the documentation side for the globe-browsing current LOD, but oculd be done on the UI side for all read-only properties:
image

modules/base/rendering/renderablelabels.cpp
```diff
@@ -75,25 +75,25 @@ namespace {
...
         "Label Color",
-        "The label color for the astronomical object."
+        "The label text color."
     };
 
     constexpr openspace::properties::Property::PropertyInfo FontSizeInfo = {
         "FontSize",
         "Font Size",
-        "The font size for the astronomical object labels."
+        "The font size (in pt) for the label."
```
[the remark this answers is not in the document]
WeirdRubberDuck: 👍 sounds good. So that people don't mistake it for some obscure space related distance unit 😄

data/assets/scene/solarsystem/dwarf_planets/pluto/charon/charon.asset
```diff
@@ -27,19 +27,14 @@ local Charon = {
     SegmentsPerPatch = 64,
     Layers = {},
     Labels = {
-      Enable = false,
+      Enabled = false,
       FileName = labelsPath .. "/charon.labels",
       LabelAlignmentOption = "Horizontally", -- or Circularly
       LabelsFontSize = 40.0,
       LabelsSize = 8.0,
```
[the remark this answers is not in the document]
WeirdRubberDuck: No, no need. The globe labels do not vary as much in size as the other label types, so here it wakes sense to just use the default values. I even removed the Min Max Size for all the globes that had them assigned, since they were simply just set to the default values for all globes. 

Also reduced the default max value from 1500 px to 1000 px. Could not find a use case where this made a difference and it made the slider more interactible (in most cases we don't get a visible change unless setting the max to something below 150-ish, at least on monitors). Realizing not that it might make a difference on more high-res displays, such as the dome? If that is the case I should probably change it back to 1500

modules/digitaluniverse/rendering/renderablebillboardscloud.cpp
WeirdRubberDuck: It absolutely makes sense to group them together, and makes the asset clearer to read. Didn't put too much work into things like this to avoid breaking changes for more properties than needed, but maybe it's best to just do them all at once?

Also I thought we plan on refactoring the label code down the line to make them more generic anyways? For example, similarly to how the globes do it, we could have a "LabelsComponent" that encapsulates information like this. SO didn't know how much work I should actually put into refactoring

data/assets/scene/digitaluniverse/abell.asset
```diff
@@ -27,18 +27,18 @@ local object = {
     File = speck .. "/abell.speck",
     Texture = textures .. "/point3A.png",
     LabelFile = speck .. "/abell.label",
+    Unit = "Mpc",
     TextColor = { 0.0, 0.8, 0.0 },
     TextSize = 22,
-    TextMinSize = 10.0,
-    Unit = "Mpc",
+    TextMinMaxSize = { 10, 15 },
```
WeirdRubberDuck, line 33: OBS! Here I did a stylistic choice to change the max from the previous default (20 px) to 15 pixels, to make the labels not overlap as much. Realizing that this might actually also affect the appearance on really high-res display system, with a risk of being to small. 

Change back to 20?

data/assets/scene/digitaluniverse/starlabels.asset
```diff
@@ -17,12 +17,11 @@ local object = {
...
     Opacity = 0.65,
     LabelFile = speck .. "/stars.label",
+    Unit = "pc",
     TextColor = { 0.4, 0.4, 0.4 },
     DrawLabels = true,
     TextSize = 14.7,
-    TextMinSize = 6.0,
-    TextMaxSize = 50.0,
-    Unit = "pc"
+    TextMinMaxSize = { 6, 30 }
```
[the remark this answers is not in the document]
WeirdRubberDuck: I actually don't remember if there was a motivation behind this, so probably best to change it back to 50. Might be a remnant of me trying out different sizes to figure out what looks good. Good catch

modules/base/rendering/grids/renderableradialgrid.cpp
```diff
@@ -238,12 +216,17 @@ void RenderableRadialGrid::update(const UpdateData&) {
...
     }
 
+    const float innerRadius = _radii.value().x;
+    const float outerRadius = _radii.value().y;
+
     // Circles
     const int nRadialSegments = _gridSegments.value()[0];
     const float fnCircles = static_cast<float>(nRadialSegments);
-    const float deltaRadius = (_maxRadius - _minRadius) / fnCircles;
+    //const float deltaRadius = (_maxRadius - _minRadius) / fnCircles;
```
WeirdRubberDuck, line 225: remove this comment!

data/assets/scene/digitaluniverse/abell.asset
alexanderbock: I think it's fine;  15 or 20 don't make too much of a difference..  Was looking at it on a 4K display and the text didn't stand out as too small (didn't even notice the difference, in fact)

WeirdRubberDuck: Not changing back, as the change makes them look better on most displays.

data/assets/scene/solarsystem/planets/saturn/hyperion/hyperion.asset
WeirdRubberDuck: Done, for `RenderableLabels` and `GlobeLabelsComponent`

Issue #1206, opened 2020-06-22
Error with labelling on Mercury


alexanderbock:
Report from Alex @ NCMNS

•	On Mercury, labelling of surface features shows up, changing font size gives an error, changing label size works well until an upper limit, changing label minimum height causes an error.

GPayne:
I don't see these problems anymore. I believe these problems were fixed with issue #1209.

WeirdRubberDuck:
Reopening, as I am still observing this issue. Steps to reproduce: 
1, Go to Mercury and enable the labels 
2. Make sure they are visible, then try to increase the label font size to something pretty large, such as 100
3. See your log being filled with errors

![image](https://user-images.githubusercontent.com/8808894/111120527-bcc5b300-856b-11eb-948e-db2c63403a72.png)

Also, after the error has happened it seem like the scriptlog gets broken somehow. For example, trying to use the arrow buttons in the time UI to increase the delta time step size results in the error "Could not open file 'D:\dev\OpenSpace\logs\ScriptLog.txt' for logging scripts"

alexanderbock:
Last time I saw this, there was an exception in the font rendering, leaving the font file opened.   Windows has a limited number of files it can keep opened so when OpenSpace then tries to write something to the scriptlog, that file opening fails since the maximum number of opened files has been reached.   Error should be in the font.cpp in Ghoul where the glyphs are loaded.  I did some change there recently.  I'm assuming someone isn't checking whether a specific glyph exists before rendering it

WeirdRubberDuck:
A hotfix for this issue was added in a40e110fe9f662613dd2b4f53033a65d243f13f3, but it does not actually solve the problem with the font loading. It just prevents the problem from occurring during runtime 

I suggest we close this issue and create another that is focused on the font loading error, as it is not specific for Mercury. (Updating the minimum height and text size does not lead to any errors)

PR #1611, opened 2021-05-27
Feature/jwst

Roxeena:


data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
+local assetHelper = asset.require('util/asset_helper')
+local sunTransforms = asset.require('scene/solarsystem/sun/transforms')
+
+asset.require("spice/base")
```
alexanderbock, line 4: -> '....'  to be consistent with the other `asset.require`s

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+        Type = "RenderableSphere",
+        Texture = band .. "/JWST-band.png",
+        Size = 9.2E15,
+        Segments = 50,
+        DisableFadeInOut = true,
+        Orientation = "Inside",
+        Opacity = 0.05,
+        UseAdditiveBlending = true,
+    },
+    Tag = { "fov" },
```
alexanderbock, line 97: A more wordy tag name would be good here, too, to prevent accidental doubling up

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+    GUI = {
+        Name = "JWST View Band",
+        Path = "/Solar System/Missions/JWST",
+        Hidden = true,
```
alexanderbock, line 101: Is this intended?  If this menu entry is hidden, it will be impossible to enable it

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+local JWSTModel = {
+    Identifier = "JWSTModel",
+    Parent = JWSTRotation.Identifier,
+    TimeFrame = {
+        Type = "TimeFrameInterval",
+        Start = "2020 JAN 01",
+        End = "2024 JAN 01"
+    },
+    Transform = {
+        Rotation = {
```
alexanderbock, line 114: I don't think this is necessary;  each scenegraphnode should get a StaticRotation by default which also defaults to 0,0,0

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+    Identifier = "JWSTFov",
+    Parent = JWSTModel.Identifier,
+    Renderable = {
+        Enabled = false,
+        Type = "RenderablePrism",
+        Segments = 6,
+        Lines = 3,
+        Radius = 3.25,
+        LineWidth = 1.0,
+        Color = {
```
alexanderbock, line 199: `Color = { 1.0, 1.0, 1.0 }` to make it a bit more condensed

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+            1.0, 1.0, 1.0
+        },
+        Length = 9.2E15
+    },
+    Transform = {
+        Rotation = {
+            Type = "StaticRotation",
+            Rotation = {
+                0, 0, 0.523598776 -- 30 deg around Z-axis in radians
```
alexanderbock, line 208: Would suggest:
`0, 0, 30 * 3.1415926 / 180.0` 
Then there is no need for the comment and we can later introduce.

I think this should wokr as well:
`0, 0, math.rad(30)`

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+            }
+        },
+    },
+    Tag = { "fov" },
```
alexanderbock, line 212: ^^

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+        Hidden = true,
+    }
+}
+
+local JWSTLaunchModel = {
+    Identifier = "JWSTLaunchModel",
+    Parent = JWSTLaunchPosition.Identifier,
+    TimeFrame = {
+        Type = "TimeFrameInterval",
+        Start = "2018 OCT 01 13:18",
```
alexanderbock, line 241: To not make it look weird:
"2018 OCT 01 13:18:00"

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+            }
+        }
+    },
+    Renderable = {
+        Type = "RenderableModel",
+        GeometryFile = model .. "/JWSTFBX.osmodel",
+        ModelScale = "Foot",
+        InvertModelScale = true,
+        EnableAnimation = true,
+        AnimationStartTime = "2018 10 01 15:00:00", --TODO: Update theese when the new animation is finished
```
alexanderbock, line 127: I'd suggest to put the TODO comment on the line before, rather than after.

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+            YAxis = sunTransforms.SolarSystemBarycenter.Identifier
+        }
+    },
+    Renderable = {
+        Type = "RenderableModel",
+        GeometryFile = model .. "/JWSTFBX.osmodel",
+        ModelScale = "Foot",
+        InvertModelScale = true,
+        EnableAnimation = true,
+        AnimationStartTime = "2018 10 01 15:00:00", --TODO: Update theese when the new animation is finished
```
alexanderbock, line 260: ^^

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+        Path = "/Solar System/Missions/JWST",
+    }
+}
+
+local JWSTLaunchTrail = {
+    Identifier = "JWSTLaunchTrail",
+    Parent = transforms.L2.Identifier,
+    TimeFrame = {
+        Type = "TimeFrameInterval",
+        Start = "2018 OCT 01 13:18",
```
alexanderbock, line 284: ^^

data/assets/scene/solarsystem/missions/jwst/jwst.asset
```diff
@@ -0,0 +1,315 @@
...
+        End = "2019 OCT 01"
+    },
+    Renderable = {
+        Type = "RenderableTrailTrajectory",
+        Translation = {
+            Type = "HorizonsTranslation",
+            HorizonsTextFile = JWSTHorizons .. "/horizons_jwst_launch.dat",
+        },
+        Color = { 0.9, 0.9, 0.0 },
+        StartTime = "2018 OCT 01 13:18",
```
alexanderbock, line 294: ^^

data/assets/util/circle.asset
```diff
@@ -0,0 +1,8 @@
+local circle = asset.syncedResource({
```
alexanderbock, line 1: We changed the asset files to be 2 spaces in indentation

modules/base/rendering/renderablemodel.cpp
```diff
@@ -318,17 +334,44 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
                 case Parameters::ScaleUnit::Kilometer:
                     distanceUnit = DistanceUnit::Kilometer;
                     break;
+
+                // Weired units
```
alexanderbock, line 338: -> Weird

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+ *                                                                                       *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,   *
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#define _USE_MATH_DEFINES
```
alexanderbock, line 25: We'd put that define just before the `#include <math.h>` instead.   Defining this *before* a bunch of other includes makes them potentially behave differently which can lead to very nasty bug finding missions

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo LengthInfo = {
+        "Length",
+        "Length",
+        "The length of the prism in meters."
+    };
+
+    // Generate vertices around the unit circle on the XY-plane
+    void getUnitCircleVertices(std::vector<float>& vertices, int sectorCount) {
```
alexanderbock, line 85: I'd suggest to change this to:  `std::vector<float> unitCircleVertices(int sectorCount)`.   Making the out-parameter a return value more clearly signifies that intent

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+        float sectorStep = 2 * M_PI / sectorCount;
```
alexanderbock, line 86: `float sectorStep = glm::two_pi<float>() / sectorCount;`

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+        float sectorAngle;  // in radians
+
+        for (int i = 0; i < sectorCount; ++i) {
+            sectorAngle = i * sectorStep;
```
alexanderbock, line 90: You can define the variable inside the loop. `float sectorAngle = i * sectorStep;` 
Predefining variables before a look doesn't really do anything anymore optimization-wise as the optimizer will >99% do that for you and it leads to code that is easier to read

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+    , _nShapeSegments(SegmentsInfo, 6, 3, 32)
+    , _nLines(LinesInfo, 6, 0, 32)
+    , _radius(RadiusInfo, 10.f, 0.f, 3.0e12f)
+    , _lineWidth(LineWidthInfo, 1.f, 1.f, 20.f)
+    , _lineColor(LineColorInfo, glm::vec3(1.f), glm::vec3(0.f), glm::vec3(1.f))
+    , _length(LengthInfo, 20.f, 1.f, 3.0e12f)
+{
+    const Parameters p = codegen::bake<Parameters>(dictionary);
+
+    _radius.setViewOption(properties::Property::ViewOptions::Logarithmic);
```
alexanderbock, line 136: @WeirdRubberDuck 

I don't think this is necessary anymore (or even compiles anymore) in the new `master` as you'd set the exponent of the slider now instead

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+    _length.onChange([&]() { _prismIsDirty = true; });
+
+    _nShapeSegments = p.segments;
+    _nLines = p.lines.value_or(_nShapeSegments);
+    _radius = p.radius.value_or(_radius);
+    _lineWidth = p.lineWidth.value_or(_lineWidth);
+    _lineColor = p.color.value_or(_lineColor);
+    _length = p.length.value_or(_length);
+
+    addProperty(_nShapeSegments);
```
alexanderbock, line 152: Personally, I prefer to group the functions by property, rather than by function:
So
```cpp
_radius.onChange([&]() { _prismIsDirty = true; });
_radius = p.radius.value_or(_radius);
addProperty(_radius);

_lineColor.setViewOption(properties::Property::ViewOptions::Color);
addProperty(_lineColor);
```
etc.

But personal preference

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+bool RenderablePrism::isReady() const {
+    return _shader != nullptr;
+}
+
+void RenderablePrism::initialize() {
+    updateVertexData();
+}
+
+void RenderablePrism::initializeGL() {
+    initializeShader();
```
alexanderbock, line 170: Since you are not reusing this function, it'd be cleaner to just paste the contents of the `initializeShader` function here.   That would save us one declared function and an indirection that someone would have to understand

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+    glDeleteBuffers(1, &_iboId);
+    _iboId = 0;
+}
+
+void RenderablePrism::updateVertexData() {
+    _vertexArray.clear();
+    _indexArray.clear();
+
+    // Get unit circle vertices on the XY-plane
+    std::vector<float> unitVertices;
```
alexanderbock, line 205: With the change above this would become:
```cpp
std::vector<float> unitVertices = unitCircleVertices(_nShapeSegments);
std::vector<float> unitVerticesLines = unitCircleVertices(_nLines);
```

And the `reserve` functions can move into the `unitCircleVertices` function itself

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+    unitVerticesLines.reserve(2 * _nLines);
+    getUnitCircleVertices(unitVertices, _nShapeSegments);
+    getUnitCircleVertices(unitVerticesLines, _nLines);
+
+    // Put base and top shape vertices into array
+    for (int i = 0; i < 2; ++i) {
+        float h = i * _length; // z value, 0 to _length
+
+        for (int j = 0, k = 0; j < _nShapeSegments && k < unitVertices.size(); ++j) {
+            float ux = unitVertices[k++];
```
alexanderbock, line 217: Not a great fan of actually using the preincrement and postincrement operators for their intended purpose as they obfuscate the code a bit.  My suggestion would be to have the `++k` inside the third part of the for loop and then:
```cpp
float ux = unitVertices[k];
float ux = unitVertices[k + 1];
```
or 
```cpp
float ux = unitVertices[k + 1];
float ux = unitVertices[k + 2];
```
whichever it is [this confusion is kind of the point of the comment, thinking about it 😆 ]

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+            float uy = unitVertices[k++];
+
+            _vertexArray.push_back(ux * _radius); // x
+            _vertexArray.push_back(uy * _radius); // y
+            _vertexArray.push_back(h);            // z
+        }
+    }
+
+    // Put the vertices for the connecting lines  into array
```
alexanderbock, line 226: Extra space

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+        _vertexArray.push_back(0.f);
+
+        // Center for top shape
+        _vertexArray.push_back(0.f);
+        _vertexArray.push_back(0.f);
+        _vertexArray.push_back(_length);
+    }
+    else {
+        for (int j = 0, k = 0; j < _nLines && k < unitVerticesLines.size(); ++j) {
+            float ux = unitVerticesLines[k++];
```
alexanderbock, line 241: Same here

modules/base/rendering/renderableprism.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+            // Top
+            _vertexArray.push_back(ux * _radius); // x
+            _vertexArray.push_back(uy * _radius); // y
+            _vertexArray.push_back(_length);      // z
+        }
+    }
+
+    // Indices for Base shape
+    for (uint8_t i = 0; i < _nShapeSegments; ++i) {
```
alexanderbock, line 257: I know you're setting th emax value of  `_nShapeSegments` to 32 but just a `ghoul_assert(_nShapeSegments.value <= std::numeric_limit<uint8_t>::max(), "Too many shape segments")` might ease some future pain.   If _nShapeSegments > 255, who knows what will happen /shrug

modules/base/rendering/renderableprism.h
```diff
@@ -0,0 +1,85 @@
...
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_MODULE_BASE___RENDERABLEPRISM___H__
+#define __OPENSPACE_MODULE_BASE___RENDERABLEPRISM___H__
+
+#include <openspace/properties/scalar/floatproperty.h>
+#include <openspace/properties/scalar/intproperty.h>
+#include <openspace/properties/vector/vec3property.h>
+#include <openspace/rendering/renderable.h>
```
alexanderbock, line 31: Because `RenderablePrism` derives from `Renderable`, the renderable include would be the first in the list and separated by an empty line to make it easier to see what the base class is

modules/base/rendering/renderableprism.h
```diff
@@ -0,0 +1,85 @@
...
+    properties::IntProperty _nShapeSegments;
+    properties::IntProperty _nLines;
+    properties::FloatProperty _radius;
+    properties::FloatProperty _lineWidth;
+    properties::Vec3Property _lineColor;
+    properties::FloatProperty _length;
+    UniformCache(modelViewProjection, color) _uniformCache;
+
+    std::unique_ptr<ghoul::opengl::ProgramObject> _shader;
+    const GLuint _locVertex = 0;
```
alexanderbock, line 73: I'd argue for not having this variable at all and just using `0` in the glVertexAttribArray` function instead

alexanderbock: Having it defined as a variable implies more flexibility than is actually present

modules/base/shaders/prism_fs.glsl
```diff
@@ -0,0 +1,36 @@
...
+ ****************************************************************************************/
+
+#include "fragment.glsl"
+
+in vec4 vs_positionScreenSpace;
+
+uniform vec4 vs_color;
+
+Fragment getFragment() {
+    Fragment frag;
```
alexanderbock, line 32: I recently went through the shader codes are placed the 4-space indentation with 2-space intentation to make them stand out more from C++ code

modules/base/shaders/prism_fs.glsl
```diff
@@ -0,0 +1,36 @@
...
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include "fragment.glsl"
+
+in vec4 vs_positionScreenSpace;
```
alexanderbock, line 27: Instead of passing the whole vec4 from which you are only using the `w` component, you can just do
`in float vs_depth` instead

modules/base/shaders/prism_vs.glsl
```diff
@@ -0,0 +1,40 @@
...
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#version __CONTEXT__
+
+#include "PowerScaling/powerScaling_vs.hglsl"
+
+layout (location = 0) in vec3 in_position;
+
+out vec4 vs_positionScreenSpace;
```
alexanderbock, line 31: See above

modules/base/shaders/prism_vs.glsl
```diff
@@ -0,0 +1,40 @@
...
+
+uniform mat4 modelViewProjectionTransform;
+
+void main() {
+    vec4 position = vec4(in_position, 1);
+    vec4 positionClipSpace = modelViewProjectionTransform * position;
+    vs_positionScreenSpace = z_normalization(positionClipSpace);
```
alexanderbock, line 38: And then just pass `vs_depth` here.

modules/base/shaders/prism_vs.glsl
```diff
@@ -0,0 +1,40 @@
...
+#include "PowerScaling/powerScaling_vs.hglsl"
+
+layout (location = 0) in vec3 in_position;
+
+out vec4 vs_positionScreenSpace;
+
+uniform mat4 modelViewProjectionTransform;
+
+void main() {
+    vec4 position = vec4(in_position, 1);
```
alexanderbock, line 36: since it's a floating number, prefer `1.0` over `1`

alexanderbock (commented):
Looking good :)

Todo for *after* the merge.  @micahnyc  also has functions for returning the position/rotation for a scene graph node in his branch, but defined in a different place.  I'd suggest to merge both of them and then coalesce them afterwards

data/assets/scene/solarsystem/missions/jwst/L1.asset
```diff
@@ -0,0 +1,59 @@
...
+local L1Label = {
+    Identifier = "L1Label",
+    Parent = L1.Identifier,
+    Renderable = {
+        Enabled = false,
+        Type = "RenderableLabels",
+        LabelText = "L1",
+        FontSize = 20,
+        LabelSize = 7.5,
+        LabelMaxSize = 100.0,
```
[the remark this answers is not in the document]
WeirdRubberDuck: Exactly, but also, the properties of the `RenderableLabels` have been renamed without the "Labels" in them. So:

```
Type = "RenderableLabels",        
Text = "L1",        
FontSize = 20,        
Size = 7.5,       
MinMaxSize = { 1.0 100.0 },
OrientationOption = "Camera View Direction",
....
```

should do it

modules/base/rendering/renderableprism.cpp
WeirdRubberDuck: Yep, that is true. To get the same behavior as with the previously `Logarithmic` viewoption you can set the exponent to the tenth logarithm of the properties max value (i.e. here: `_radius.setExponent(12.f)`). However, I would suggest setting it to a slightly smaller one. Best is to set a value and try it out to see if it behaves the way you'd want it

modules/base/rendering/renderableprism.cpp
WeirdRubberDuck: I have the same preference

data/assets/scene/solarsystem/missions/jwst/L2.asset
```diff
@@ -0,0 +1,128 @@
...
+}
+
+local L2Label = {
+    Identifier = "L2Label",
+    Parent = L2.Identifier,
+    Renderable = {
+        Enabled = true,
+        Type = "RenderableLabels",
+        LabelText = "L2",
+        FontSize = 14.0,
```
WeirdRubberDuck, line 66: Not super important, but would be nice if this label could also have `FontSize = 20.0`. Trying to reduce the number of different fonts sizes we use to get around a problem of the font atlas being filled when we load in too many labels (as part of a hitfix for #1206 ). 
The label size is the exponent of a 2^x scaling of the label, so setting it to just a little higher, maybe `6.9` or so should give about the same size

That said, this one will only load 2 symbols of that size, which is why it's not super important

src/scene/scene.cpp
```diff
@@ -717,6 +717,20 @@ scripting::LuaLibrary Scene::luaLibrary() {
                 "Adds an interesting time to the current scene. The first argument is "
                 "the name of the time and the second argument is the time itself in the "
                 "format YYYY-MM-DDThh:mm:ss.uuu"
+            },
+            {
+                "worldPosition",
+                &luascriptfunctions::worldPosition,
+                {},
+                "string",
+                "Returns the worldPosition of the scene graph node with the given string as identifier"
```
WeirdRubberDuck, line 726: would separate `worldPosition` into two words here

src/scene/scene.cpp
```diff
@@ -717,6 +717,20 @@ scripting::LuaLibrary Scene::luaLibrary() {
...
+            },
+            {
+                "worldRotation",
+                & luascriptfunctions::worldRotation,
+                {},
+                "string",
+                "Returns the worldRotation of the scene graph node with the given string as identifier"
```
WeirdRubberDuck, line 733: Same here with `worldRotation`, and also mention that it is actually a matrix as rotations can be represented in different ways. So "world rotation matrix"

src/scene/scene_lua.inl
```diff
@@ -908,4 +908,28 @@ int addInterestingTime(lua_State* L) {
     return 0;
 }
 
+int worldPosition(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, 1, "lua::worldPosition");
+
+    std::string identifier = ghoul::lua::value<std::string>(L, 1, ghoul::lua::PopValue::Yes);
+
+    glm::dvec3 pos = global::renderEngine->scene()->sceneGraphNode(identifier)->worldPosition();
```
WeirdRubberDuck, line 916: If the scene graph node doesn't exists, this will call the `worldPosition()` function on a nullptr and lead to an exception. To handle this, first get the pointer and then have a check that returns a lua error. For example: 

```
const SceneGraphNode* node = sceneGraphNode(identifier);
if (!node) {
    return ghoul::lua::luaError(L, "fmt::format("Did not find a match for identifier: {} ", identifier));
}
 ```
 
 (the `sceneGraphNode(string)` function, without having to access the scene pointer, is a helper that exists in `openspace/query/query.h`)

File: src/scene/scene_lua.inl [hunk 39]
```diff
@@ -908,4 +908,28 @@ int addInterestingTime(lua_State* L) {
...
+    ghoul_assert(lua_gettop(L) == 1, "Incorrect number of items left on stack");
+    return 1;
+}
+
+int worldRotation(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, 1, "lua::worldRotation");
+
+    std::string identifier = ghoul::lua::value<std::string>(L, 1, ghoul::lua::PopValue::Yes);
+
+    glm::dmat3 rot = global::renderEngine->scene()->sceneGraphNode(identifier)->worldRotationMatrix();
```
WeirdRubberDuck, line 928: see above

src/scene/scene_lua.inl
WeirdRubberDuck: Or looking at the previous functions in this file, maybe handle errors the same way as done there: 
```
    if (!node) {
        LERRORC(
            "worldPosition",
            fmt::format("Did not find a match for identifier: {} ", identifier)
        );
        return 0;
    }
```

Do we have any preference here, @alexanderbock ?

data/assets/scene/solarsystem/missions/jwst/L2.asset
WeirdRubberDuck: The label size for `RendeableLabels` is the exponent of a 10^x scaling of the label...

PR #1616, opened 2021-05-31
Reduce risk of font rendering errors from user interaction (#1206 hotfix)

WeirdRubberDuck:
* Allways load all glyphs immediately for GlobeLabelsComponent, to avoid errors on font loading
* Unify used font sizes, to avoid filling font atlas (especially for globes, as we now load all glyphs immediately)
* Make font size properties read-only, to prevent users from triggering these errors during runtime

+ fix typo for Pluto asset and reduce fade in distance of Titan labels


data/assets/scene/solarsystem/dwarf_planets/pluto/pluto.asset
```diff
@@ -51,9 +51,9 @@ local PlutoLabel = {
   Renderable = {
     Enabled = false,
     Type = "RenderableLabels",
-    Size = "Pluto",
```
alexanderbock, line 54: Oopsiedupsie

alexanderbock (approved):

data/assets/scene/solarsystem/dwarf_planets/pluto/pluto.asset
WeirdRubberDuck: 😅 Good that I had to go through all the labels once more

PR #1634, opened 2021-06-07
Adding hi-res 75cm Bennu model from asteroidmission.org

GPayne:
Adds the hi-res Bennu model from here:
https://www.asteroidmission.org/updated-bennu-shape-model-3d-files/
to the osirisrex mission. The profile has the **BennuProjection** scenegraph node, which is the original low-res Bennu model that gets images projected onto it. This new **Bennu** scenegraph node is disabled by default, but looks much better than the original.
image

The Utah data server has these 2 new files:
http://openspace.sci.utah.edu/files/solarsystem/missions/osirisrex/bennu/models/v1_v2/Bennu_v20_200k_an.obj
http://openspace.sci.utah.edu/files/solarsystem/missions/osirisrex/bennu/models/v1_v2/Bennu_v20_200k_an.mtl
which will need to be added to the data.openspaceproject.com server.

### DIFF
--- a/data/assets/scene/solarsystem/missions/osirisrex/bennu.asset
+++ b/data/assets/scene/solarsystem/missions/osirisrex/bennu.asset
@@ -5,6 +5,15 @@ local models = asset.require('./models').models
 
 local BENNU_BODY = "2101955"
 
+local LightSources = {
+  {
+    Type = "SceneGraphLightSource",
+    Identifier = "Sun",
+    Node = sunTransforms.SolarSystemBarycenter.Identifier,
+    Intensity = 1.0
+  },
+}
+
 local Bennu = {
   Identifier = "Bennu",
   Parent = transforms.BennuBarycenter.Identifier,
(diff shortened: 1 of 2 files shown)


data/assets/scene/solarsystem/missions/osirisrex/bennu.asset
```diff
@@ -16,39 +25,20 @@ local Bennu = {
     },
     Scale = {
       Type = "StaticScale",
-      Scale = 1000.0
+      Scale = 1.0
```
Roxeena, line 28: If the scale is set to 1.0 then couldn’t it just be removed to make the code cleaner?

GPayne: Agreed, since I no longer needed to scale the model.

alexanderbock:
Files are updated
http://liu-se.cdn.openspaceproject.com/files/solarsystem/missions/osirisrex/bennu/models/v1_v2/Bennu_v20_200k_an.obj
http://liu-se.cdn.openspaceproject.com/files/solarsystem/missions/osirisrex/bennu/models/v1_v2/Bennu_v20_200k_an.mtl

and available under bennu_models version 2

PR #1638, opened 2021-06-11
issue/1626

Roxeena:
Change all Spice usage of "SUN" to "SSB" and make trails and planets be placed in their correct place.


alexanderbock (approved):
Looks good to me!

What is the reason for the extra *center scene graph nodes? As far as I can see, they are only used by the globes directly and the *center translations could be included in there directly, right?   For Earth, we have a center as that is the reference frame for the TLE files that we've been using, which we don't have for other planets

alexanderbock (approved):

PR #1665, opened 2021-06-23
Feature/render at distance

micahnyc:


alexanderbock (approved):

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -179,6 +179,14 @@ namespace {
         "Enables shadow mapping algorithm. Used by renderable rings too."
     };
 
+    constexpr openspace::properties::Property::PropertyInfo RenderAtDistanceInfo = {
+        "RenderAtDistance",
+        "Render at Distance",
+        "Tells the rendering engine not to perform distance based performance culling"
```
WeirdRubberDuck, line 185: Missing whitespace after "culling" and a misspelled property. Suggest to change to:

```
        "Tells the rendering engine not to perform distance based performance culling "
        "for this globe. Turning this property on will let the globe to be seen at far "
        "away distances when normally it would be hidden."
```

WeirdRubberDuck (approved):

PR #1667, opened 2021-06-24
Thesis/2019/camera paths

WeirdRubberDuck:
Refactored and cleaned up the camera paths code and moved it from a separate module into core. The integration into core also included some refactoring of the existing navigation code. For example, the camera and navigation stuff has now got their own folders and the NavigationState has been moved out of the NavigationHandler into its own file. 

There is still some work left to do to make it all work nicely, but it would be great to get this first version into master and then continue working on it. I also want to let you guys have a chance to look at it and provide input before I do. Hence the PR 🙂 

Some known problems:
- Instability when traveling really vast distances
- The paths and speed does not work super well close to surfaces
- "Pointy" paths lead to rapid changes in perceived speed
- Including roll can lead to very fast rotations
- other things that I don't remember now

One major thing left to do is the final design of the actual camera paths. The two existing ones address a couple of different scenarios, but neither works well in all cases

Happy Midsommar y'all!


GPayne (commented):
Code looks great overall, my comments are suggestions rather than potential problems in the code. Camera paths is a great feature to add to OpenSpace!

include/openspace/navigation/path.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace openspace { 
+    struct CameraPose; 
+} // namespace openspace
+
+namespace openspace::interaction {
+
+class Path {
+public:
+    enum CurveType {
```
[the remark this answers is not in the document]
WeirdRubberDuck: Tried doing this when I implemented it, but the strongly-types enums sadly don't play nicely with our `OptionProperty`. The reason is that they are not implicitly typed to ints.

We use the plain enums for all other OptionProperties (at least all I've seen), so I suggest to keep the plain enum here as well

include/openspace/navigation/pathcurve.h
```diff
@@ -0,0 +1,99 @@
...
+
+    /*
+     * Compute and rteturn the position along the path at the specified relative
+     * distance. The input parameter should be in range [0, 1], where 1 correspond to
+     * the full length of the path
+     */
+    glm::dvec3 positionAt(double relativeDistance);
+
+    /*
+     * Compute curve parameter u that matches the input arc length s
```
Roxeena, line 49: This comment doesn't seem to match its function. The input is `u` but the comment says it's `s`, which is confusing.

src/navigation/navigationstate.cpp
```diff
@@ -0,0 +1,186 @@
...
+    anchor = p.anchor;
+    position = p.position;
+
+    referenceFrame = p.referenceFrame.value_or(anchor);
+    aim = p.aim.value_or(aim);
+
+    if (p.up.has_value()) {
+        up = *p.up;
+
+        yaw = p.yaw.value_or(yaw);
```
Roxeena, line 77: Should these really be inside the if statement? What if `yaw` and `pitch` is provided but not `up`? And if `up` is required in order to use `yaw` and `pitch` then it should be beter documented to avoid confusion.

src/navigation/pathnavigator.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+    addProperty(_includeRoll);
+    addProperty(_speedScale);
+    addProperty(_applyIdleBehaviorOnFinish);
+    addProperty(_minValidBoundingSphere);
+
+    _relevantNodeTags = std::vector<std::string>{
+        "planet_solarSystem",
+        "moon_solarSystem"
+    };;
```
Roxeena, line 120: double ;

src/navigation/orbitalnavigator.cpp
```diff
@@ -637,14 +761,22 @@ void OrbitalNavigator::updateCameraStateFromStates(double deltaTime) {
...
+    // Disable idle behavior if camera interaction happened
+    if (_idleBehavior.apply && _idleBehavior.abortOnCameraInteraction) {
+        _idleBehavior.apply = false;
+        // Prevent interpolating stop, to avoid weirdness when changing anchor, etc
+        _idleBehaviorDampenInterpolator.setInterpolationTime(0.f);
+    }
+}
+
 glm::dquat OrbitalNavigator::composeCameraRotation(
     const CameraRotationDecomposition& decomposition)
```
alexanderbock, line 774: Argument should be aligned with the opening `(`

src/navigation/path.cpp
```diff
@@ -0,0 +1,472 @@
...
+            break;
+        case CurveType::ZoomOutOverview:
+            _curve = std::make_unique<ZoomOutOverviewCurve>(_start, _end);
+            break;
+        default:
+            LERROR("Could not create curve. Type does not exist!");
+            throw ghoul::MissingCaseException();
+    }
+
+    const auto defaultDuration = [](double pathlength) {
```
alexanderbock, line 109: If this is only used once on line 113, maybe it's better to just used the `log` function directly rather than going through a lambda?

src/navigation/path.cpp
```diff
@@ -0,0 +1,472 @@
...
+    glm::dvec3 endUp = _end.rotation() * glm::dvec3(0.0, 1.0, 0.0);
+
+    double tUp = helpers::shiftAndScale(t, t1, t2);
+    tUp = ghoul::sineEaseInOut(tUp);
+    glm::dvec3 up = ghoul::interpolateLinear(tUp, startUp, endUp);
+
+    return ghoul::lookAtQuaternion(_curve->positionAt(t), lookAtPos, up);
+}
+
+double Path::speedAlongPath(double traveledDistance) {
```
alexanderbock, line 234: This function could be `const`, right?

src/navigation/path.cpp
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    return nullptr;
+}
+
+// Compute a target position close to the specified target node, using knowledge of
+// the start point and a desired distance from the node's center
+glm::dvec3 computeGoodStepDirection(const SceneGraphNode* targetNode,
+                                     const Waypoint& startPoint)
+{
```
alexanderbock, line 325: `const`

src/navigation/path.cpp
```diff
@@ -0,0 +1,472 @@
...
+        const glm::dvec3 targetToSun = sunPos - nodePos;
+
+        constexpr const float defaultPositionOffsetAngle = -30.f; // degrees
+        constexpr const float angle = glm::radians(defaultPositionOffsetAngle);
+        const glm::dvec3 axis = glm::normalize(glm::cross(targetToPrev, targetToSun));
+        const glm::dquat offsetRotation = angleAxis(static_cast<double>(angle), axis);
+
+        return glm::normalize(offsetRotation * targetToSun);
+    }
+};
```
alexanderbock, line 364: No need for the `;` at the end

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+const double PathCurve::length() const {
+    return _totalLength;
+}
+
+glm::dvec3 PathCurve::positionAt(double relativeDistance) {
+    const double u = curveParameter(relativeDistance * _totalLength);
+    return interpolate(u);
+}
+
+std::vector<glm::dvec3> PathCurve::points() {
```
alexanderbock, line 55: This can be `const`

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+    _parameterSamples.push_back({ 1.0, _totalLength });
+}
+
+// Compute the curve parameter from an arc length value, using a combination of
+// Newton's method and bisection. Source:
+// https://www.geometrictools.com/Documentation/MovingAlongCurveSpecifiedSpeed.pdf
+// Input s is a length value, in the range [0, _totalLength]
+// Returns curve parameter in range [0, 1]
+double PathCurve::curveParameter(double s) {
```
alexanderbock, line 108: This could be `const`?

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+namespace openspace::interaction {
+
+PathCurve::~PathCurve() {}
+
+const double PathCurve::length() const {
+    return _totalLength;
+}
+
+glm::dvec3 PathCurve::positionAt(double relativeDistance) {
```
alexanderbock, line 50: `const` if possible?

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+    const double uMin = _curveParameterSteps[segmentIndex - 1];
+    const double uMax = _curveParameterSteps[segmentIndex];
+
+    // Use samples to find an initial guess for Newton's method
+    // Find first sample with s larger than input s
+    auto sampleIterator = std::upper_bound(
+        _parameterSamples.begin() + startIndex,
+        _parameterSamples.begin() + endIndex,
+        ParameterPair{ 0.0 , s }, // 0.0 is a dummy value for u
+        [](ParameterPair lhs, ParameterPair rhs) {
```
alexanderbock, line 130: Should be `const ParameterPair&` to prevent a copy

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+            u = (uCandidate >= upper) ? (upper + lower) / 2.0 : uCandidate;
+        }
+    }
+
+    // No root was found based on the number of iterations and tolerance. However, it is
+    // safe to report the last computed u value, since it is within the segment interval
+    return u;
+}
+
+double PathCurve::approximatedDerivative(double u, double h) {
```
alexanderbock, line 177: `const`?

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+    if (u <= h) {
+        return (1.0 / h) * glm::length(interpolate(0.0 + h) - interpolate(0.0));
+    }
+    if (u >= 1.0 - h) {
+        return (1.0 / h) * glm::length(interpolate(1.0) - interpolate(1.0 - h));
+    }
+    return (0.5 / h) * glm::length(interpolate(u + h) - interpolate(u - h));
+}
+
+double PathCurve::arcLength(double limit) {
```
alexanderbock, line 187: `const`?

src/navigation/pathcurve.cpp
```diff
@@ -0,0 +1,237 @@
...
+    return arcLength(0.0, limit);
+}
+
+double PathCurve::arcLength(double lowerLimit, double upperLimit) {
```
alexanderbock, line 191: `const`?

src/navigation/pathcurves/avoidcollisioncurve.cpp
```diff
@@ -0,0 +1,196 @@
...
+            glm::dvec3 intersectionPointModelCoords;
+            bool collision = collision::lineSphereIntersection(
+                p1,
+                p2,
+                center,
+                radius,
+                intersectionPointModelCoords
+            );
+
+            if (collision) {
```
alexanderbock, line 155: Can do:
```
if (!collision) {
  continue;
}
```
to reduce the total sum of indentation

src/navigation/pathnavigator.cpp
```diff
@@ -0,0 +1,439 @@
...
+}
+
+bool PathNavigator::isPlayingPath() const {
+    return hasCurrentPath() && _isPlaying;
+}
+
+void PathNavigator::updateCamera(double deltaTime) {
+    ghoul_assert(camera() != nullptr, "Camera must not be nullptr");
+
+    if (!hasCurrentPath()) {
```
alexanderbock, line 161: Can merge the second `if` statement together with this:
```
if (!hasCurrentPath() || !_isPlaying) {
  return;
}
```

src/navigation/pathnavigator.cpp
```diff
@@ -0,0 +1,439 @@
...
+    catch (const ghoul::RuntimeError& e) {
+        LERROR(fmt::format("Could not create path. Reason: ", e.message));
+        return;
+    }
+
+    LINFO("Successfully generated camera path");
+}
+
+void PathNavigator::clearPath() {
+    LINFO("Clearing path...");
```
alexanderbock, line 244: I tried to get rid of the ellipses in messages as it looks a bit weird in the log files

src/navigation/pathnavigator.cpp
```diff
@@ -0,0 +1,439 @@
...
+    }
+    _isPlaying = false;
+    clearPath(); // TODO: instead of clearing this could be handled better
+
+    LINFO("Aborted camera path");
+}
+
+void PathNavigator::pausePath() {
+    if (hasFinished()) {
+        LERROR("No path to pause (path is empty or has finished).");
```
alexanderbock, line 279: We try not to use punctuation at the end of log messages

src/navigation/pathnavigator.cpp
```diff
@@ -0,0 +1,439 @@
...
+        return;
+    }
+
+    LINFO("Path paused");
+    _isPlaying = false;
+}
+
+void PathNavigator::continuePath() {
+    if (hasFinished()) {
+        LERROR("No path to resume (path is empty or has finished).");
```
alexanderbock, line 294: Same here

src/navigation/pathnavigator.cpp
```diff
@@ -0,0 +1,439 @@
...
+        return;
+    }
+
+    if (_isPlaying) {
+        LERROR("Cannot resume a path that is already playing");
+        return;
+    }
+
+    LINFO("Continuing path...");
```
alexanderbock, line 303: Same here

src/util/collisionhelper.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+#include <openspace/util/collisionhelper.h>
+
+namespace openspace::collision {
+
+// Source: http://paulbourke.net/geometry/circlesphere/raysphere.c
+bool lineSphereIntersection(glm::dvec3 p1, glm::dvec3 p2, glm::dvec3 center,
+                            double r, glm::dvec3& intersectionPoint)
+{
+    long double a, b, c;
```
alexanderbock, line 33: Can move the variable declarations further down where they are actually used to make it less old-school C style

alexanderbock: Are the `long double` really necessary?  Long double is not standardized and might produce some problems.  Maybe `double` is enough?

alexanderbock (approved):

src/util/collisionhelper.cpp
WeirdRubberDuck: You're probably right that a double is enough. In the worst case the diffs below would be in the order of the size of the universe, i.e. 10^27. We want to square those values and sum them together, but it still shouldn't result in larger numbers than the order of 10^54 something, if I'm thinking correctly. So a double is still more than enough

WeirdRubberDuck: Note to self: The same goes for all the functions in this file

src/navigation/path.cpp
WeirdRubberDuck: This is actually not a member function, but just a helper function to the `createPathFromDictionary` function below. 

Wasn't sure how to structure this part of the code nicely, but many of these functions actually aren't relevant to the `Path` class itself or need to be exposed anywhere else. So, just kept them as independent functions in the cpp file and did not include them in the header file

Is this an ok solution? Open to suggestions on how to make it nicer

src/navigation/navigationstate.cpp
WeirdRubberDuck: I agree it's not the best design, but this is how it was done before and none of the three values makes sense without the others in the case of a NavigationState. 

A better option is to combine the three angles into an EulerAngles struct in the Parameters. That would make it easy to handle it correctly in terms of testing against the specification as well (right now, just adding one of the three would not result in an error message). However, that would be a breaking change that could potentially affect quite a few users

As it's not a change made in this code, but rather just a function that has been copied, I suggest keeping it as it is for this PR. If we want to change it, we can create an issue for it. Then the change would be properly "documented" within our GitHub history

Issue #1656, opened 2021-06-17
Exponential sliders does not handle negative values well


WeirdRubberDuck:
Problematic for for example StaticTranslations... The scale must be updated to handle both fully negative and mixed negative/positive value ranges

[Relevant stackoverflow post](https://stackoverflow.com/questions/51584888/how-to-write-a-d3-positive-and-negative-logarithmic-scale)

WeirdRubberDuck:
Temporary fix in #1672, but it does not actually solve the problem. Remaining work is to design how these kind of sliders should actually behave and come up with a somewhat neat implementation

WeirdRubberDuck:
> Where did this end up in the end regarding the new user interface, [@WeirdRubberDuck](https://github.com/WeirdRubberDuck) ?

Still an issue - no change has been done to the scale of the exponential sliders

PR #1672, opened 2021-06-30
Avoid non-supported ranges for exponential slider

WeirdRubberDuck:
Temporary fix to address #1656. Actually fixing the sliders requires some more careful work and design decisions


alexanderbock (approved):
Looks tony-the-tiger great!

PR #1675, opened 2021-07-01
Reduce fontsize of shutdown warning and center it instead with a dimming of the rendering

alexanderbock:
.

### DIFF
--- a/src/rendering/renderengine.cpp
+++ b/src/rendering/renderengine.cpp
@@ -57,6 +57,7 @@
 #include <ghoul/io/model/modelreaderassimp.h>
 #include <ghoul/io/model/modelreaderbinary.h>
 #include <ghoul/logging/logmanager.h>
+#include <ghoul/misc/easing.h>
 #include <ghoul/misc/profiling.h>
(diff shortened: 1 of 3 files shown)


src/rendering/renderengine.cpp
```diff
@@ -919,13 +937,14 @@ void RenderEngine::renderShutdownInformation(float timer, float fullTime) {
         ghoul::fontrendering::CrDirection::Down
     );
 
+    const glm::vec2 size2 = _fontShutdown->boundingBox("Press ESC again to abort");
```
WeirdRubberDuck, line 940: Tiny thing, but noticed this exact sting appears twice. So maybe create a variable for it to avoid problems if someone decides to change it later?

WeirdRubberDuck (approved):
Looks great. I love it 😍 

A tiny improvement suggestion: Maybe darken the rendering a little immediately when ESC is pressed? That way it's easier to notice when it happens by accident. 

Now it takes almost a second before the visual change is really noticeable and I found it a little stressful to click ESC again if I wanted to prevent the shutdown from happening

PR #1676, opened 2021-07-06
Fix problem when pause in session recording and outputting frames (#1669)

### LINKED ISSUE #1669: Session Recording When in 'Output Frames' Playback Mode
Observed Behavior:
When playing back a recording file in save-screenshots mode ("Output Frames" in the playback GUI or `openspace.sessionRecording.enableTakeScreenShotDuringPlayback` in console), if the recording contains a pause then playback gets stuck at that pause and never un-pauses. This does not happen in normal playback when screenshots aren't being saved.
So far this has been tested with recording file in ascii mode, but it most likely is independent of the file format.

Expected Behavior:
The playback will pause and then un-pause at the same times as when the recording was made.

How to Reproduce:
Start a recording, move camera around a bit, hit spacebar to pause, move camera some more, hit spacebar again to un-pause, then stop recording.
Play back in normal mode should work as expected. Playing back with the above console command should cause playback to hang at the first pause.

WeirdRubberDuck:
Seems to me like this should fix #1669, but maybe @GPayne could verify?

Tried it out a bit and the session recording and output frames behave how I would expect it with this change. However, I noticed that the screenshots are triggered a few frames before the actual session recording starts playing back. So, we get a few extra screenshots before the session recording, and also some before the playback is resumed after it has been paused (when the GUI bar is orange)

Anyhow, it works much better than the pause preventing the session recording from being played back at all 😃

### DIFF
--- a/src/interaction/sessionrecording.cpp
+++ b/src/interaction/sessionrecording.cpp
@@ -1770,8 +1770,8 @@ bool SessionRecording::addKeyframe(Timestamps t3stamps,
 void SessionRecording::moveAheadInTime() {
     using namespace std::chrono;
 
-    bool paused = global::timeManager->isPaused();
-    if (_state == SessionState::PlaybackPaused) {
+    bool playbackPaused = (_state == SessionState::PlaybackPaused);
+    if (playbackPaused) {
         _playbackPauseOffset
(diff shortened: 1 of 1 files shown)


WeirdRubberDuck:
Closing without merge, as duplicate of #1678

PR #1690, opened 2021-07-20
Issue/1669

GPayne:
Fixes the hard-to-debug timing problems with session recording playback-with-frames. The original time manager timeline code has some built-in expectations for parts of the code happening at slightly different application times. In playback-with-frames, those times are exactly the same because the application time is set to discrete values according to the rendering framerate. This caused problems with time interpolation at the beginning and ending of delta time (and pause) interpolations. These problems included trouble with pausing, and time skips and/or duplicate rendered frames at interpolation beginning & ending.


include/openspace/util/timeline.h
```diff
@@ -95,6 +95,16 @@ bool compareTimeWithKeyframeTime(double a, const KeyframeBase& b);
 */
 bool compareKeyframeTimeWithTime(const KeyframeBase& a, double b);
 
+/**
+* Return true if the timestamp of a is smaller than or equal to b.
```
alexanderbock, line 99: The commented out lines should be aligned one character in to make the * of each line align

src/interaction/sessionrecording.cpp
```diff
@@ -1181,6 +1183,11 @@ SessionRecording::currentPlaybackInterpolationTime() const
     return _saveRenderingCurrentRecordedTime_interpolation;
 }
 
+double SessionRecording::currentApplicationInterpolationTime() const
```
alexanderbock, line 1186: The `{` should be on the same line as the function

src/scene/scene.cpp
```diff
@@ -474,6 +474,16 @@ SceneGraphNode* Scene::loadNode(const ghoul::Dictionary& nodeDictionary) {
     return rawNodePointer;
 }
 
+std::chrono::steady_clock::time_point Scene::currentTimeForInterpolation()
```
alexanderbock, line 477: Same here

src/util/timemanager.cpp
```diff
@@ -819,11 +833,37 @@ void TimeManager::interpolatePause(bool pause, double interpolationDuration) {
...
+        now = previousApplicationTimeForInterpolation();
+    }
     clearKeyframes();
     if (interpolationDuration > 0) {
         addKeyframe(now, currentKeyframe);
     }
     addKeyframe(now + interpolationDuration, futureKeyframe);
 }
 
+double TimeManager::currentApplicationTimeForInterpolation() const
```
alexanderbock, line 846: Same here

src/util/timemanager.cpp
```diff
@@ -819,11 +833,37 @@ void TimeManager::interpolatePause(bool pause, double interpolationDuration) {
...
+{
+    if (global::sessionRecording->isSavingFramesDuringPlayback()) {
+        return global::sessionRecording->currentApplicationInterpolationTime();
+    }
+    else {
+        return global::windowDelegate->applicationTime();
+    }
+}
+
+double TimeManager::previousApplicationTimeForInterpolation() const
```
alexanderbock, line 856: And here

alexanderbock (approved):

PR #1705, opened 2021-08-10
Feature/state machine

WeirdRubberDuck:
The state machine module, that is useful for creating interactive installations. 

Also renamed the old "state machine", that can only transition forwards and backwards in a list, to the "lua state machine". A better name suggestion would be appreciated.


modules/statemachine/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+##########################################################################################
+#                                                                                        #
+# OpenSpace                                                                              #
+#                                                                                        #
+# Copyright (c) 2014-2020                                                                #
```
WeirdRubberDuck, line 5: update the year in all these fiels

modules/statemachine/src/statemachine.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    currentState()->exit();
+    _transitions[transitionIndex].performAction();
+    _currentStateIndex = stateIndex;
+    currentState()->enter();
+}
+
+bool StateMachine::canTransitionTo(const std::string state) const {
+    int transitionIndex = findTransitionTo(state);
+    return (transitionIndex != -1) ? true : false;
```
Roxeena, line 152: This could be simplified to `return transitionIndex != -1`

modules/statemachine/statemachinemodule.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <modules/statemachine/include/transition.h>
+#include <openspace/documentation/documentation.h>
+#include <openspace/scripting/lualibrary.h>
+#include <ghoul/logging/logmanager.h>
+
+#include "statemachinemodule_lua.inl"
+
+namespace {
+    constexpr const char* _loggerCat = "StateMachine";
+}
```
Roxeena, line 38: } // namespace

modules/statemachine/statemachinemodule.cpp
```diff
@@ -0,0 +1,194 @@
...
+            "and Transition documentation for details. The optional thrid argument is "
+            "the identifier of the desired initial state. If left out, the first state "
+            "in the list will be used."
+        },
+        {
+            "goTo",
+            &luascriptfunctions::goTo,
+            {},
+            "string",
+            "Triggers a transition from the current state to th state with the given "
```
Roxeena, line 139: th -> the

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -0,0 +1,195 @@
...
+    ghoul::lua::checkArgumentsAndThrow(L, 1, "lua::goTo");
+    const bool isString = (lua_isstring(L, 1) != 0);
+
+    if (!isString) {
+        lua_settop(L, 0);
+        const char* msg = lua_pushfstring(
+            L,
+            "%s expected, got %s",
+            lua_typename(L, LUA_TSTRING),
+            luaL_typename(L, -1)
```
Roxeena, line 88: Shouldn't this be connected to what type the actual argument was? It would make the error message clearer for users and us in the future.

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -0,0 +1,195 @@
...
+        );
+        return luaL_error(L, "bad argument #%d (%s)", 2, msg);
```
Roxeena, line 90: I think the 2 here should be a 1 to indicate the first argument

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -0,0 +1,195 @@
...
+    ghoul::lua::checkArgumentsAndThrow(L, 1, "lua::setStartState");
+    const bool isString = (lua_isstring(L, 1) != 0);
+
+    if (!isString) {
+        lua_settop(L, 0);
+        const char* msg = lua_pushfstring(
+            L,
+            "%s expected, got %s",
+            lua_typename(L, LUA_TSTRING),
+            luaL_typename(L, -1)
```
Roxeena, line 113: See above

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -0,0 +1,195 @@
...
+        );
+        return luaL_error(L, "bad argument #%d (%s)", 2, msg);
```
Roxeena, line 115: See above

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -0,0 +1,195 @@
...
+    ghoul::lua::checkArgumentsAndThrow(L, 1, "lua::canGoTo");
+    const bool isString = (lua_isstring(L, 1) != 0);
+
+    if (!isString) {
+        lua_settop(L, 0);
+        const char* msg = lua_pushfstring(
+            L,
+            "%s expected, got %s",
+            lua_typename(L, LUA_TSTRING),
+            luaL_typename(L, -1)
```
Roxeena, line 160: See above

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -0,0 +1,195 @@
...
+        );
+        return luaL_error(L, "bad argument #%d (%s)", 2, msg);
```
Roxeena, line 162: See above

Roxeena (commented):
Looks good! :+1: Just a few nitpicky comments.

modules/statemachine/statemachinemodule_lua.inl
WeirdRubberDuck: That is what this does 😄 the `luaL_typename` gets the type name of the variable at that corresponding place in the stack. However, the index is incorrect (similar to the copy paste error you noticed, where the 2s should be 1s)

WeirdRubberDuck:
@micah and I discussed adding another optional parameter to the States, to make it easier to create transitions between states when no action is required. The state could then look something like:
```
{
    Identifier = "A",
    Enter = "",
    Exit = "", 
    Transitions = {"B", "C", "D"}
}
```
Or even `Transitions = "*"` or something, to automatically create transitions to all other states. 

This doesn't break any of the current implementation though, so I suggest we merge this PR first and create a separate issue for it. It would also add some complexity to the implementation, so could be good to discuss it a bit more as well

modules/statemachine/include/statemachine.h
```diff
@@ -0,0 +1,64 @@
...
+namespace openspace {
+
+namespace documentation { struct Documentation; }
+
+class StateMachine {
+public:
+    StateMachine(const ghoul::Dictionary& dictionary);
+    ~StateMachine() = default;
+
+    void setInitialState(const std::string initialState);
```
alexanderbock, line 41: Do you mean `const std::string&`?

modules/statemachine/include/statemachine.h
```diff
@@ -0,0 +1,64 @@
...
+    const State* currentState() const;
+    void transitionTo(const std::string newState);
```
alexanderbock, line 43: Same here

modules/statemachine/include/statemachine.h
```diff
@@ -0,0 +1,64 @@
...
+    bool canTransitionTo(const std::string state) const;
```
alexanderbock, line 44: and here

modules/statemachine/include/statemachine.h
```diff
@@ -0,0 +1,64 @@
...
+    std::vector<std::string> possibleTransitions() const;
+
+    static documentation::Documentation Documentation();
+
+private:
+    int _currentStateIndex = -1;
+    std::vector<State> _states;
+    std::vector<Transition> _transitions;
+
+    int findTransitionTo(const std::string state) const;
```
alexanderbock, line 58: 1. In general, I prefer to define the functions before defining the member variables
2. `const std::string&`

modules/statemachine/include/transition.h
```diff
@@ -0,0 +1,54 @@
...
+#include <ghoul/misc/dictionary.h>
+#include <string>
+
+namespace openspace {
+
+namespace documentation { struct Documentation; }
+
+class Transition {
+public:
+    Transition(const ghoul::Dictionary& dictionary);
```
alexanderbock, line 37: `explicit`

modules/statemachine/include/transition.h
```diff
@@ -0,0 +1,54 @@
...
+    ~Transition() = default;
+
+    std::string from() const;
```
alexanderbock, line 40: This could be returned as a `const std::string&`?

modules/statemachine/src/statemachine.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Search if the transition from _currentState to newState exists.
+// If yes then return the index to the transition, otherwise return -1
+int StateMachine::findTransitionTo(const std::string state) const {
+    if (!currentState()) {
+        return -1;
+    }
+
+    for (unsigned int i = 0; i < _transitions.size(); ++i) {
```
alexanderbock, line 161: Should be `size_t` instead of `unsigned int`

modules/statemachine/src/statemachine.cpp
```diff
@@ -0,0 +1,198 @@
...
+            return i;
+        }
+    }
+    return -1;
+}
+
+// Search if the state exist.
+// If yes then return the index to the state, otherwise return -1
+int StateMachine::findState(const std::string state) const {
+    for (unsigned int i = 0; i < _states.size(); ++i) {
```
alexanderbock, line 174: Same here

modules/statemachine/src/statemachine.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+std::vector<std::string> StateMachine::possibleTransitions() const {
+    std::vector<std::string> res;
+
+    if (!currentState()) {
+        return res;
+    }
+
+    res.reserve(_transitions.size());
+    for (unsigned int i = 0; i < _transitions.size(); ++i) {
```
alexanderbock, line 190: and here

modules/statemachine/statemachinemodule.cpp
```diff
@@ -0,0 +1,194 @@
...
+    try {
+        _machine = std::make_unique<StateMachine>(dictionary);
+        LINFO(fmt::format(
+            "State machine was created with start state: {}", currentState()
+        ));
+    }
+    catch (const documentation::SpecificationError& e) {
+        LERROR(ghoul::to_string(e.result));
+        LERROR(fmt::format("Error loading state machine: {}", e.what()));
+        return;
```
alexanderbock, line 67: This return is not necessary as the function ends directly after either way

alexanderbock (approved):

modules/statemachine/include/statemachine.h
WeirdRubberDuck: Oh, I agree!

data/assets/examples/luastatemachine.asset
```diff
@@ -0,0 +1,47 @@
+local stateMachineHelper = asset.require('util/lua_state_machine_helper')
+
+states = {
```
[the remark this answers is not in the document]
WeirdRubberDuck: Did not encounter any problems, but yeah, it should be declared as local

PR #1706, opened 2021-08-11
Update .osmodels

Roxeena:


alexanderbock:
Models are uploaded to the CDN

PR #1708, opened 2021-08-12
Feature/keybindings

alexanderbock:


apps/OpenSpace/ext/launcher/src/profile/actiondialog.cpp
```diff
@@ -0,0 +1,612 @@
...
+    //  |                      | Add action    | [oooooooooooo] |    Row 11
+    //  |                      |               |                |    Row 12
+    //  *----------------------*---------------*----------------*
+    //  | [+] [-]              |               | [Save] [Cancel]|    Row 13
+    //  *----------------------*---------------*----------------*
+    //  |=======================================================|    Row 14
+    //  *----------------------*---------------*----------------*
+    //  |                                      | [Save] [Cancel]|    Row 15
+    //  *----------------------*---------------*----------------*
+
```
GPayne, line 98: This is a good idea for showing what the intended layout is

apps/OpenSpace/ext/launcher/src/profile/actiondialog.cpp
```diff
@@ -0,0 +1,612 @@
...
+    Profile::Action* action = selectedAction();
+    std::string oldIdentifier = action->identifier;
+    if (oldIdentifier != newIdentifier) {
+        // The identifier is a bit special as we need to make sure that we didn't
+        // accidentally create a duplicate.  Also if we did create a duplicate we need to
+        // make sure that we update all keybinds that referenced the old value are updated
+        // to the new one
+
+        const auto it = std::find_if(
```
GPayne, line 427: The comment above makes me wonder if it would be possible to get more than one duplicate action? If so then this would only find the first. Maybe this should be inside of a loop that ends only when iterator == _actions.data.end()

apps/OpenSpace/ext/launcher/src/profile/timedialog.cpp
```diff
@@ -144,13 +144,13 @@ void TimeDialog::approved() {
...
         if (_relativeEdit->text().isEmpty()) {
-            _profile.clearTime();
+            _profile.time = std::nullopt;
         }
         else {
             Profile::Time t;
             t.type = Profile::Time::Type::Relative;
             t.value = _relativeEdit->text().toStdString();
-            _profile.setTime(t);
+            _profile.time = std::nullopt;
```
GPayne, line 153: Why clear the time here? Is this a copy/paste error?

apps/OpenSpace/ext/launcher/src/profile/timedialog.cpp
```diff
@@ -161,7 +161,7 @@ void TimeDialog::approved() {
             _absoluteEdit->date().toString("yyyy-MM-dd").toStdString(),
             _absoluteEdit->time().toString().toStdString()
         );
-        _profile.setTime(t);
+        _profile.time = std::nullopt;
```
GPayne, line 164: Same comment as above

include/openspace/interaction/action.h
```diff
@@ -22,58 +22,26 @@
...
-int clearShortcuts(lua_State* L) {
-    ghoul::lua::checkArgumentsAndThrow(L, 0, "lua::clearShortcuts");
-    global::shortcutManager->resetShortcuts();
-    return 0;
-}
+    std::string command;
+    std::string name;
+    std::string documentation;
+    std::string guiPath;
+    IsSynchronized synchronization = IsSynchronized::Yes;
```
GPayne, line 42: Header file comments would be useful here in order to explain what is synchronized.

include/openspace/scene/profile.h
```diff
@@ -130,72 +133,41 @@ class Profile {
...
-    std::optional<Time> _time;
-    std::vector<double> _deltaTimes;
-    std::optional<CameraType> _camera;
-    std::vector<std::string> _markNodes;
-    std::vector<std::string> _additionalScripts;
-
-    bool _ignoreUpdates = false;
 };
 
+std::string convertToScene(const Profile& profile);
```
GPayne, line 168: Some comment lines for this function would help to make the conversion more understandable, since this is an important step. Mention that the source file will contain different namespace(s) for previous version(s), etc.

GPayne (approved):

apps/OpenSpace/ext/launcher/src/profile/actiondialog.cpp
alexanderbock: Done

apps/OpenSpace/ext/launcher/src/profile/timedialog.cpp
alexanderbock: Oops.. it was indeed

include/openspace/interaction/action.h
alexanderbock: Comments added

include/openspace/scene/profile.h
alexanderbock: Added comments

apps/OpenSpace/ext/launcher/src/profile/actiondialog.cpp
alexanderbock: I rewrote the comment to make it clear that the identifier of an action has to be unique, so there is no need for a loop here as there can only be 0 or 1 hit on the check

PR #1719, opened 2021-08-18
Feature/lua function cleanup

alexanderbock:
A pull request that addresses a large amount of cleanup of the way we are handling Lua functions, extract values out of Lua states etc

### DIFF
--- a/include/openspace/interaction/joystickinputstate.h
+++ b/include/openspace/interaction/joystickinputstate.h
@@ -28,6 +28,7 @@
 #include <ghoul/misc/assert.h>
 #include <ghoul/misc/exception.h>
 #include <ghoul/misc/stringconversion.h>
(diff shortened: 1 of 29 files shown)


WeirdRubberDuck (approved):
Looks good to me :) 

The same cleanup could now be done for `pathnavigator_lua.inl` and `statemachinemodule_lua.inl`

WeirdRubberDuck:
Also, nice that each function does not have to have its own error handling for checking if the input argument was of a correct type! 

However, the resulting error messages could be a bit more user-friendly... This is what we get when a string was expected, but a table was provided
image

Would be nice if it just said "String" instead `std::basic_string<...etc`

Also, I would really want to see which function call lead to which error. In case there are many in a row that lead to errors

PR #1721, opened 2021-08-20
Add function to print out state machine to a dot file

WeirdRubberDuck:


modules/statemachine/statemachinemodule.cpp
```diff
@@ -116,6 +118,23 @@ bool StateMachineModule::canGoToState(const std::string& state) const {
     return _machine->canTransitionTo(state);
 }
 
+void StateMachineModule::saveToFile(const std::string& filename,
+                                    std::optional<std::string> directory) const
```
alexanderbock, line 122: In this case it might tbe better to say:
```
void StateMachineModule::saveToFile(const std::string& filename, std::string directory = "${TEMPORARY}/")
```
in the header

Then you don't need to worry about the whole optional thingy

modules/statemachine/statemachinemodule.cpp
```diff
@@ -116,6 +118,23 @@ bool StateMachineModule::canGoToState(const std::string& state) const {
...
+{
+    if (!_machine) {
+        LWARNING("Attempting to use uninitialized state machine");
```
alexanderbock, line 125: I think this should be an error?   Warning seems more like:  You are doing something strange that might not work.  Error is:  You are doing something strange that will not work

modules/statemachine/statemachinemodule_lua.inl
```diff
@@ -114,4 +114,15 @@ int printCurrentStateInfo(lua_State* L) {
     return 1;
 }
 
+int saveToDotFile(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, { 1, 2 }, "lua::saveToDotFile");
+
```
alexanderbock, line 119: In the rewrite I didn't have any empty lines between the check arguments and the extraction line, but I'm good eitherway

alexanderbock (approved):

modules/statemachine/statemachinemodule.cpp
WeirdRubberDuck: That's a very good point you make there ,sir 😁

PR #1737, opened 2021-08-31
Issue/1527 - GlobeRotation

WeirdRubberDuck:


modules/globebrowsing/src/globerotation.cpp
```diff
@@ -0,0 +1,236 @@
...
+    : _globe(GlobeInfo)
+    , _latitude(LatitudeInfo, 0.0, -90.0, 90.0)
+    , _longitude(LongitudeInfo, 0.0, -180.0, 180.0)
+    , _angle(AngleInfo, 0.0, 0.0, 360.0)
+    , _useHeightmap(UseHeightmapInfo, false)
+{
+    const Parameters p = codegen::bake<Parameters>(dictionary);
+
+    _globe = p.globe;
+    _globe.onChange([this]() {
```
Roxeena, line 113: Is it supposed to be possible to change the globe during runtime? If not, then the `onChange` isn't really needed right? If yes, then I think the property should be added to the GUI, which it isn't at the moment.

Roxeena (commented):
Looks good 👍
It would have been nice with something that connects the lat and long sliders for the translation and the rotation in the future. Was a little bit annoying to set both to the same value when tinkering around with it.

modules/globebrowsing/src/globerotation.cpp
WeirdRubberDuck: You can't change it in the GUI, because the property isn't added. But anyone could change it during runtime using the set property value functions

alexanderbock: Yea, I'd also say it's better not to allow re-globing.. since you, at the same time, probably also would want to reparent this object to a different scenegraph node as well, which currently does not work anyway

modules/globebrowsing/src/globerotation.cpp
```diff
@@ -0,0 +1,236 @@
...
+            _globe = _globeNode->identifier();
+        }
+    }
+}
+
+glm::vec3 GlobeRotation::computeSurfacePosition(double latitude, double longitude) const
+{
+    ghoul_assert(_globeNode, "Globe cannot be nullptr");
+
+    GlobeBrowsingModule& mod = *(global::moduleEngine->module<GlobeBrowsingModule>());
```
alexanderbock, line 156: Might be a bit more readable if:
`GlobeBrowsingModule* mod = global::module...`
and then using `->` instead of `.` below

modules/globebrowsing/src/globerotation.cpp
```diff
@@ -0,0 +1,236 @@
...
+    else {
+        float latitudeRad = glm::radians(static_cast<float>(_latitude));
+        float longitudeRad = glm::radians(static_cast<float>(_longitude));
+        yAxis = _globeNode->ellipsoid().geodeticSurfaceNormal(
+            { latitudeRad, longitudeRad }
+        );
+    }
+    yAxis = glm::normalize(yAxis);
+
+    const glm::dvec3 n = glm::vec3(0.0, 0.0, 1.0);
```
alexanderbock, line 216: Can be `constexpr`

modules/globebrowsing/src/globerotation.cpp
```diff
@@ -0,0 +1,236 @@
...
+    glm::dvec3 zAxis = glm::dvec3(0.0);
```
alexanderbock, line 217:
```
glm::dvec3 zAxis = glm::dvec3(
    yAxis.y * n.z - yAxis.z * n.y,
    yAxis.z * n.x - yAxis.x * n.z,
    yAxis.x * n.y - yAxis.y * n.x
);
```

alexanderbock (approved):

modules/globebrowsing/src/globerotation.cpp
WeirdRubberDuck: Good point. Then I'll do the same change for the GlobeTranslation

PR #1741, opened 2021-09-09
Event System

alexanderbock:
This pull requests brings in the event system.  There is an example of an event in the asset/events/toggle_trail which will automatically toggle the trail associated with a scene graph node, if it exists.  The file include/openspace/events/event.h contains a list of all events that can currently be created.  Three of which (TimeOfInterestReached, MissionEventReached, and PlanetEclipsed) are currently not emitted in the code.  The openspace.cfg file contains a new switch `PrintEvents` which will print all events that are generated to the log file for debugging.

A second option to register specific events are the new keys `OnApproach`, `OnReach`, `OnMoveAway` and `OnExit` that can either name a single or multiple actions that are invoked when the camera moves relative to the focus node at a specific distance which can be controlled on a per-scenegraph node bases using the `ApproachFactor` (default 5.0) and `ReachFactor` (default 1.25). These two factors define two shells at 5.0 * interactionsphereradius and 1.25 * interactionsphereradius and the four event types are emitted when the camera crosses these shells.

Whole system probably of particular interest to @Mik-pe  and @liuloppan aside from core team


alexanderbock:
Quick sheets - page 2

src/navigation/orbitalnavigator.cpp
```diff
@@ -720,6 +729,103 @@ void OrbitalNavigator::updateCameraStateFromStates(double deltaTime) {
     // Update the camera state
     _camera->setPositionVec3(pose.position);
     _camera->setRotation(composeCameraRotation(camRot));
+
+    const double currDistance = glm::distance(anchorPos, pose.position);
+    const double d = _anchorNode->interactionSphere();
+    const double af = _anchorNode->approachFactor();
+    const double rf = _anchorNode->reachFactor();
+
```
GPayne, line 737: I created a comment block for this section based on your hand drawing:

```
/*
 Diagram of events for a camera moving from right-to-left.
 Interaction sphere is 'O' in middle, and ')' are spherical boundaries.
 
 |<------------------->|  Approach factor * Interaction sphere
              |<------>|  Reach Factor * Interaction sphere
       
 (            (        O        )            )
 ^            ^                 ^            ^
 OnExit       OnMoveAway        OnReach      OnApproach
*/
```

GPayne (changes requested):

alexanderbock:
@GPayne   I changed the handling of the events in the session recording a bit, see if you still approve
1. Only a single event that stores the playback state as an enum
2. Added events being fired for pausing and resuming a session recording

alexanderbock:
Added a description for the events to the wiki, too:  http://wiki.openspaceproject.com/docs/builders/events/index

src/events/eventengine.cpp
```diff
@@ -0,0 +1,134 @@
...
+scripting::LuaLibrary EventEngine::luaLibrary() {
+    scripting::LuaLibrary res;
+    res.name = "event";
+    res.functions.push_back({
+        "registerEventAction",
+        &luascriptfunctions::registerEventAction,
+        {},
+        "string, string [, table]",
+        "Registers an action (second parameter) to be executed whenever an event (first "
+        "parameter is encountered. If the optional third parameter is provided, it "
```
WeirdRubberDuck, line 118: I think there should be an a closing parenthesis after "parameter" here

src/events/eventengine.cpp
```diff
@@ -0,0 +1,134 @@
...
+        &luascriptfunctions::unregisterEventAction,
+        {},
+        "string, string [, table]",
+        "Unregisters a specific combination of event (first parameter), action (second "
+        "parameter), and potentially a filter (optional third argument)"
+    });
+    return res;
+}
+
+
```
WeirdRubberDuck, line 133: super nitpicky, but extra line

src/navigation/orbitalnavigator.cpp
```diff
@@ -720,6 +729,121 @@ void OrbitalNavigator::updateCameraStateFromStates(double deltaTime) {
...
     _camera->setRotation(composeCameraRotation(camRot));
+
+    // The rest of this function is concerned with managing transitions of the camera
+    // between different distances of interest relative to the focus node. For each
+    // transition two scenarios are handled;  SceneGraphNodes can have attached actions
+    // for each transition, which are automatically triggered. Additionally, an
+    // EventCameraTransition event is fired that contains information about the focus node
+    // and the transition state that caused the vent to fire.
+
+    // Diagram of events for a camera moving from right-to-left.
```
WeirdRubberDuck, line 740: Nice diagram 👍

src/scene/scene.cpp
```diff
@@ -580,6 +582,10 @@ void Scene::updateInterpolations() {
         i.prop->interpolateValue(t, i.easingFunction);
 
         i.isExpired = (t == 1.f);
+
+        if (i.isExpired) {
+            global::eventEngine->publishEvent<events::EventInterpolationFinished>(i.prop);
```
WeirdRubberDuck, line 587: Excited for this!

src/scene/scenegraphnode.cpp
```diff
@@ -208,6 +219,36 @@ namespace {
...
+        // larger than the reachFactor or unexpected things might happen
+        std::optional<double> approachFactor [[codegen::greaterequal(0.0)]];
+
+        // This value is a multiplication factor for the interaction sphere that
+        // determines when the camera has 'reached' the scene graph node. If this value is
+        // not specified, a default value of 1.25 is used instead. This value must be
+        // smaller than the approachFactor or unexpected things might happen
+        std::optional<double> reachFactor [[codegen::greaterequal(0.0)]];
+
+        // One or multiple actions that are executed whenever the camera is focussed on
```
WeirdRubberDuck, line 234: "focused"

src/scene/scenegraphnode.cpp
```diff
@@ -208,6 +219,36 @@ namespace {
...
+        // this scene graph node and if it enters the interaction sphere of the node
+        std::optional<std::variant<std::string, std::vector<std::string>>> onApproach;
+
+        // One or multiple actions that are executed whenever the camera is focussed on
```
WeirdRubberDuck, line 238: "focused"

WeirdRubberDuck: and for `onMoveAway` and "onExit"

data/assets/events/toggle_trail.asset
```diff
@@ -0,0 +1,27 @@
+local action = asset.require('actions/toggle_trail')
+
+asset.onInitialize(function()
+  openspace.event.registerEventAction(
+    "CameraTransition",
+    action.show_trail,
+    { Before = "ApproachSphere", After = "Outside"}
```
WeirdRubberDuck, line 7: So, what this really means is "Going out of the approaching sphere" (i.e. Leaving an object?)? And the next one is "approaching from the outside" ()

Wondering if it would maybe be more clear to have one identifier per scenario, and remove the Before & After combination? Maybe "Approaching", "Reaching" and some good opposite words that make sense?

It may lead to more code, but I think it would make it easier to use for the general user.  Right now there are also some potential combinations that are ambiguous or could be confusing. For example, would `{ Before = "Outside", After = "Outside"}` be allowed? Or what does `{ Before = "Outside", After = "ReachedSphere"}` mean? (because the "ApproachSphere" is in between, right?).

Anyhow, l think removing the flexibility here and specify some clear options could be beneficial for the ease of use of this event type. But of course willing to discuss :)

src/navigation/orbitalnavigator.cpp
```diff
@@ -720,6 +729,121 @@ void OrbitalNavigator::updateCameraStateFromStates(double deltaTime) {
     // Update the camera state
     _camera->setPositionVec3(pose.position);
     _camera->setRotation(composeCameraRotation(camRot));
+
```
WeirdRubberDuck, line 732: This part should be moved to somewhere where it works for the path navigator and the future free flight navigator. Move to NavigationHandler?

include/openspace/events/event.h
```diff
@@ -0,0 +1,285 @@
...
+    //  6. Add the new enum entry into the `toString` and `fromString` methods
+    enum class Type {
+        SceneGraphNodeAdded,
+        SceneGraphNodeRemoved,
+        ParallelConnection,
+        ProfileLoadingFinished,
+        ApplicationShutdown,
+        ScreenSpaceRenderableAdded,
+        ScreenSpaceRenderableRemoved,
+        CameraTransition,
```
WeirdRubberDuck, line 67: rename to something that includes "Anchor" or "Focus"

include/openspace/events/event.h
```diff
@@ -0,0 +1,285 @@
...
+        Resumed,
+        Finished
+    };
+
+    EventSessionRecordingPlayback(State state_);
+    const State state;
+};
+
+
+struct CustomEvent : public Event {
```
WeirdRubberDuck, line 274: Add a comment about the intention about this event type and its intended use case (and especially, when it should not be used)

src/navigation/orbitalnavigator.cpp
alexanderbock: Moved:
image

Issue #1418, opened 2020-12-02
Pausing time through additional scripts in profile not working


WeirdRubberDuck:
Adding `openspace.time.setPause(true)` to additional scripts in the profile leads to Earth being off-focus (not visible) at start-up, and time being set to Sat, 01 Jan 2000.

micahnyc:
I was thinking today that in addition to fixing this, as it should work, it would be nice to have a feature that allows profiles to start paused. I'm thinking this could be a setting for users in the time window of the profile editor.

GPayne:
This problem happens with any time command (setPause, setTime, setDeltaTime) contained in the additional scripts profile section. This is because the additional scripts are executed before the timeManger is properly initialized.
I created issue #1727 to change the way that profiles are initialized, in order to solve this and other problems.

Am currently working on this in branch _feature/profileOverScene_

GPayne:
This is fixed by pull request #1742 which was just merged.

PR #1742, opened 2021-09-13
Feature/profile over scene

GPayne:
Changes the way that a profile is initialized. Profiles are no longer converted into an asset which is then imported. The individual sections of a profile are now handled independently. An input profile is now required at startup.
This also fixes issue #1418 by handling additional scripts after other parts of OpenSpace have been initialized.


alexanderbock:
Not sure if it is related, but I'm getting this after starting the default profile?

image

include/openspace/engine/configuration.h
```diff
@@ -45,6 +45,7 @@ struct Configuration {
 
     std::string windowConfiguration = "${CONFIG}/single.xml";
     std::string asset;
+    std::string profileOutPrefixName;
```
alexanderbock, line 48: This variable is only set once in OpenSpaceEngine::initialize and then read in OpenSpaceEngine::loadInitAssetSection.   The second function is never called, meaning that the stored value is never read, meaning that the variable can be removed entirely

alexanderbock: Also the Configuration::usingProfile variable is not necessary any more either

include/openspace/engine/openspaceengine.h
```diff
@@ -101,20 +102,78 @@ class OpenSpaceEngine {
     void writeStaticDocumentation();
     void createUserDirectoriesIfNecessary();
 
+    /**
+     * Sets the camera position using the time contents of a profile. The function will
+     * set an absolute position or a go-to-geolocation command using the globebrowsing
+     * module.
+     * \param p The Profile to be read.
+     */
+    void setFromProfile_camera(const Profile& p);
```
alexanderbock, line 111: We're not using `_` anywhere in functions really, so maybe:  `setCameraFromProfile` would be better?  Same for the other functions below

alexanderbock: Most of the functions also don't do anything with the OpenSpaceEngine, so they don't have to be functions of that class at all and can just be defined locally in the cpp file, too

include/openspace/engine/openspaceengine.h
```diff
@@ -101,20 +102,78 @@ class OpenSpaceEngine {
...
-    void loadSingleAsset(const std::string& assetPath);
+    void loadAsset_init(const std::string assetName);
+    void loadAsset_postInit(const std::string assetName);
     void loadFonts();
 
     void runGlobalCustomizationScripts();
     void configureLogging();
     std::string generateFilePath(std::string openspaceRelativePath);
     void resetPropertyChangeFlagsOfSubowners(openspace::properties::PropertyOwner* po);
+    void loadInitAssetSection(std::string& initAssetOutput,
```
alexanderbock, line 175: See above;  this function is not called anymore and can die, I think

include/openspace/scene/profile.h
```diff
@@ -164,20 +163,19 @@ class Profile {
      * \return The Lua library that contains all Lua functions available for profiles
      */
     static scripting::LuaLibrary luaLibrary();
+
+    std::string file_subName_assets = "";
```
alexanderbock, line 167: This variable is not used anymore and can die

include/openspace/scene/profile.h
```diff
@@ -164,20 +163,19 @@ class Profile {
...
+    std::string assetFileExtension = ".asset";
```
alexanderbock, line 168: This variable is only used in the OpenSpaceEngine::loadInitAssetSection and as described above, the function is (rightfully) not called anymore, so this variable can die, too

include/openspace/scene/profile.h
```diff
@@ -164,20 +163,19 @@ class Profile {
...
- * \return The string representation of the provided profile, ready to be loaded as an
- *         asset
+ * Function to retrieve the individual additional script lines included in the profile
+ *
+ * \param p The profile that should be processed
+ *
+ * \return vector of strings containing the individual script lines
  */
-std::string convertToScene(const Profile& profile);
+std::vector<std::string>& additionalScripts(Profile& p);
```
alexanderbock, line 178: This function is (correctly) not called anywhere and can be removed entirely

include/openspace/scene/scene.h
```diff
@@ -233,6 +235,26 @@ class Scene : public properties::PropertyOwner {
...
 
+    /**
+     * Sets a property using the 'properties' contents of a profile. The function will
+     * loop through each setProperty command. A property may be set to a bool, float,
+     * or string value (which must be converted because a Profile stores all values
+     * as strings)
+     *
+     * \param p The Profile to be read.
+     */
+    void setFromProfile_properties(const Profile& p);
```
alexanderbock, line 246: -> `setPropertiesFromProfile`

include/openspace/scene/scene.h
```diff
@@ -233,6 +235,26 @@ class Scene : public properties::PropertyOwner {
...
+
+    /**
+     * Accepts string version of a property value from a profile, converts it to the
+     * appropriate type, and then pushes the value onto the lua state.
+     *
+     * \param L the lua state to push value to
+     * \param value string representation of the value with which to set property
+     */
+    void property_pushValueFromProfileToLuaState(ghoul::lua::LuaState& L,
```
alexanderbock, line 255: This function doesn't access anything from the Scene class and just be defined as a local function in the cpp file instead

include/openspace/util/time.h
```diff
@@ -141,6 +141,20 @@ class Time {
      */
     double advanceTime(double tickTime);
 
+    /**
+     * Sets a relative time from profile.
+     * \param setTime a string containing time adjustment as described in documentation
+     * for luascriptfunctions::time_advancedTime
+     */
+    void setFromProfile_timeRelative(const std::string& setTime);
```
alexanderbock, line 149: -> setRelativeTimeFromProfile  ?

Likewise for the function below

include/openspace/util/timemanager.h
```diff
@@ -75,6 +78,15 @@ class TimeManager : public properties::PropertyOwner {
...
      */
     double deltaTime() const;
+
+    /**
+     * Sets the simulation time using the time contents of a profile. The function will
+     * set either a relative or absolute time.
+     *
+     * \param p The Profile to be read.
+     */
+    void setFromProfile_time(const Profile& p);
```
alexanderbock, line 88: -> setTimeFromProfile

src/engine/openspaceengine.cpp
```diff
@@ -331,16 +336,7 @@ void OpenSpaceEngine::initialize() {
...
-            // existing infrastructure
-            std::ofstream scene(outputAsset);
-            std::string sceneContent = convertToScene(*global::profile);
-            scene << sceneContent;
-
-            // Set asset name to that of the profile because a new scene file will be
-            // created with that name, and also because the profile name will override
-            // an asset name if both are provided.
-            global::configuration->asset = outputAsset;
+            global::configuration->profileOutPrefixName = outputProfilePrefix;
```
alexanderbock, line 339: <3 So nice

src/engine/openspaceengine.cpp
```diff
@@ -694,10 +689,10 @@ void OpenSpaceEngine::scheduleLoadSingleAsset(std::string assetPath) {
     _scheduledAssetPathToLoad = std::move(assetPath);
 }
 
-void OpenSpaceEngine::loadSingleAsset(const std::string& assetPath) {
+void OpenSpaceEngine::loadAsset_init(const std::string assetName) {
```
alexanderbock, line 692: The `_postInit` version of this function is not called anymore, meaning that the specification is not necessary anymore.   -> loadAsset

Also:  I'd change `const std::string` to `const std::string&` as you are only referencing the asset name in here when you are passing it on to the assetmanager

src/engine/openspaceengine.cpp
```diff
@@ -757,7 +749,12 @@ void OpenSpaceEngine::loadSingleAsset(const std::string& assetPath) {
     }
 
     _assetManager->removeAll();
-    _assetManager->add(assetPath);
+    if (!assetName.empty()) {
+        _assetManager->add(assetName);
+    }
+    for (auto a : global::profile->assets) {
```
alexanderbock, line 755: `auto` -> `const std::string&`

src/engine/openspaceengine.cpp
```diff
@@ -869,7 +866,16 @@ void OpenSpaceEngine::loadSingleAsset(const std::string& assetPath) {
 
     _writeDocumentationTask = std::async(&OpenSpaceEngine::writeSceneDocumentation, this);
 
-    LTRACE("OpenSpaceEngine::loadSingleAsset(end)");
+    LTRACE("OpenSpaceEngine::loadAsset_init(end)");
+}
+
+void OpenSpaceEngine::loadAsset_postInit(const std::string assetName) {
```
alexanderbock, line 872: This function is (rightfully?) not called anymore and can die in that case

src/engine/openspaceengine.cpp
```diff
@@ -1104,12 +1110,25 @@ void OpenSpaceEngine::preSynchronization() {
...
         LINFO(fmt::format("Loading asset: {}", absPath(_scheduledAssetPathToLoad)));
         global::profile->ignoreUpdates = true;
-        loadSingleAsset(_scheduledAssetPathToLoad);
+        loadAsset_init(_scheduledAssetPathToLoad);
         global::profile->ignoreUpdates = false;
         resetPropertyChangeFlagsOfSubowners(global::rootPropertyOwner);
         _hasScheduledAssetLoading = false;
         _scheduledAssetPathToLoad.clear();
     }
+    else if (!_hasInitializedProfile) {
```
alexanderbock, line 1119: Isn't `_hasInitializedProfile` the same as `_isFirstRenderingFirstFrame` (whose name probably should be `_isRenderingFirstFrame` instead) 🤔

src/engine/openspaceengine.cpp
```diff
@@ -1157,9 +1174,30 @@ void OpenSpaceEngine::preSynchronization() {
...
+
+    if (!_hasInitializedProfile) {
+        setFromProfile_camera(*global::profile);
+        setFromProfile_additionalScripts(*global::profile);
+        _hasInitializedProfile = true;
+    }
     LTRACE("OpenSpaceEngine::preSynchronization(end)");
 }
 
+void OpenSpaceEngine::loadInitAssetSection(std::string& initAssetOutput,
```
alexanderbock, line 1186: This function is (correctly) not called anywhere anymore and can die

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+                if (navStateProfile.yaw.has_value()) {
+                    nav.yaw = navStateProfile.yaw.value();
+                }
+                if (navStateProfile.pitch.has_value()) {
+                    nav.pitch = navStateProfile.pitch.value();
+                }
+                global::navigationHandler->setNavigationStateNextFrame(nav);
+            },
+            [](const Profile::CameraGoToGeo& geo) {
+                std::string geoScript = fmt::format("openspace.globebrowsing.goToGeo"
```
alexanderbock, line 1612: Maybe adding a comment here explaining the reason why this is done via a Lua script and that this complexity can be removed once we move the `goToGeo` function into the core

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+                    geoScript,
+                    scripting::ScriptEngine::RemoteScripting::Yes
+                );
+            }
+        },
+        p.camera.value()
+    );
+}
+
+void OpenSpaceEngine::setFromProfile_deltaTimes(const Profile& p) {
```
alexanderbock, line 1628: Since this function doesn't really do anything, I think it can be removed and the call to the timeManager can be done in the calling places directly instead

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+    global::timeManager->setDeltaTimeSteps(p.deltaTimes);
+}
+
+void OpenSpaceEngine::setFromProfile_modules(const Profile& p)
+{
```
alexanderbock, line 1633: The `{` can be on the previous line

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+    for (Profile::Module mod : p.modules) {
+        const std::vector<OpenSpaceModule*>& m = global::moduleEngine->modules();
+        const auto it = std::find_if(m.begin(), m.end(),
+            [&mod](const OpenSpaceModule* moduleSearch) {
+                return (moduleSearch->identifier().compare(mod.name) == 0);
```
alexanderbock, line 1638: -> `moduleSearch->identifier() == mod.name`

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+            });
+        if (it != m.end()) {
+            if (mod.loadedInstruction.has_value()) {
+                global::scriptEngine->queueScript(
+                    mod.loadedInstruction.value(),
```
alexanderbock, line 1643: To access the optional, you can just dereference it:
-> `*mod.loadedInstruction`

Same for the notLoadedInstruction below

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+                    mod.notLoadedInstruction.value(),
+                    scripting::ScriptEngine::RemoteScripting::Yes
+                );
+            }
+        }
+    }
+}
+
+void OpenSpaceEngine::setFromProfile_actions(const Profile& p)
+{
```
alexanderbock, line 1660: The `{` can be on the previous line

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+    for (Profile::Action a : p.actions) {
+        if (a.identifier.empty()) {
+            LERROR("Identifier must to provided to register action");
+        }
+        if (global::actionManager->hasAction(a.identifier)) {
+            LERROR(fmt::format("Action for identifier '{}' already existed & registered",
```
alexanderbock, line 1666: Nitpicky;  it'd format this as such:
```
LERROR(fmt::format(
    "Action for identifier '{}' already existed & registered", a.identifier
));
```
if the line fits, otherwise:
```
LERROR(fmt::format(
    "Action for identifier '{}' already existed & registered",
    a.identifier
));
```

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+                a.identifier));
+        }
+        if (a.script.empty()) {
+            LERROR(fmt::format("Identifier '{}' doesn't provide a Lua command to execute",
+                a.identifier));
+        }
+        interaction::Action action;
+        action.identifier = a.identifier;
+        action.command = a.script;
+        if (!a.name.empty()) {
```
alexanderbock, line 1676: I don't think there is a need to check if it is empty?  Since the action.name, documentation and guipath are empty by default, in the worst case we would overwrite an empty string with an empty string, which is going to be pretty cheap

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+        if (!a.guiPath.empty()) {
+            action.guiPath = a.guiPath;
+        }
+        action.synchronization = interaction::Action::IsSynchronized(a.isLocal);
+        global::actionManager->registerAction(std::move(action));
+    }
+}
+
+void OpenSpaceEngine::setFromProfile_keybindings(const Profile& p)
+{
```
alexanderbock, line 1691: `{` can be on the previous line

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+    for (Profile::Keybinding k : p.keybindings) {
+        if (k.action.empty()) {
+            LERROR("Action must not be empty");
+        }
+        if (!global::actionManager->hasAction(k.action)) {
+            LERROR(fmt::format("Action '{}' does not exist", k.action));
+        }
+        if (k.key.key == openspace::Key::Unknown) {
+            LERROR(fmt::format("Could not find key '{}'",
```
alexanderbock, line 1700: Same as above:
```
LERROR(fmt::format(
    "Could not find key '{}'",
    std::to_string(static_cast<uint16_t>(k.key.key))
));
```

I don't think the `std::to_string` is necessary either since fmt::format can accept uint16_t and do the string conversion itself

src/engine/openspaceengine.cpp
```diff
@@ -1546,6 +1584,144 @@ void OpenSpaceEngine::toggleShutdownMode() {
...
+                std::to_string(static_cast<uint16_t>(k.key.key))));
+        }
+        global::keybindingManager->bindKey(k.key.key, k.key.modifier, k.action);
+    }
+}
+
+void OpenSpaceEngine::setFromProfile_markInterestingNodes(const Profile& p) {
+    for (std::string nodeName : p.markNodes) {
```
alexanderbock, line 1708: -> `const std::string&`  no need for a copy here

src/scene/profile.cpp
```diff
@@ -737,191 +738,8 @@ scripting::LuaLibrary Profile::luaLibrary() {
     };
 }
 
-std::string convertToScene(const Profile& p) {
```
alexanderbock, line 740: So much less code <3

src/scene/scene.cpp
```diff
@@ -602,6 +603,48 @@ const std::vector<Scene::InterestingTime>& Scene::interestingTimes() const {
...
+            groupName,
+            ghoul::EasingFunction::Linear
+        );
+        //Clear lua state stack
+        lua_settop(L, 0);
+    }
+}
+
+void Scene::property_pushValueFromProfileToLuaState(ghoul::lua::LuaState& L,
+    const std::string& value)
```
alexanderbock, line 635: The `const` would be aligned with the `(` from the line above

src/scene/scene.cpp
```diff
@@ -602,6 +603,48 @@ const std::vector<Scene::InterestingTime>& Scene::interestingTimes() const {
...
+{
+    if (!luascriptfunctions::convertStringToLuaAndPush_bool(L, value)) {
+        if (!luascriptfunctions::convertStringToLuaAndPush_float(L, value)) {
```
alexanderbock, line 638: Since the functions in the `if` statement are not used anywhere else, I think it migh tbe better to write them out here as it makes them less magic as to what is happening here.
```
if (value == "true") {
    ghoul::lua::push(L, true);
}
else if (value == "false") {
    ghoul::lua::push(L, false);
}
else {
      try {
        float converted = std::stof(s);
        ghoul::lua::push(L, converted);
    }
    catch (...) {
        ghoul::lua::push(L, value == "nil" ? value : "[[" + value + "]]");
    }
}
```

src/scene/scene_lua.inl
```diff
@@ -896,4 +899,41 @@ int worldRotation(lua_State* L) {
...
 }
 
+/**
+ * \ingroup LuaScripts
+ * convertStringToLuaAndPush_bool(lua_State*, string):
+ * Used to convert a string value from a text file (e.g. profile) to a lua bool type.
+ * If string value is a bool, the value gets pushed to the lua_State, and returns true.
+ * If string value is not a bool, the lua_State is unchanged and returns false.
+ */
+bool convertStringToLuaAndPush_bool(lua_State* L, std::string s) {
```
alexanderbock, line 909: Not sure if they need this functions.  Personally, not a fan of functions that "check and then maybe do something" since they make the resulting code a bit harder to reason about

src/util/timemanager.cpp
```diff
@@ -868,4 +868,15 @@ double TimeManager::previousApplicationTimeForInterpolation() const {
     return _previousApplicationTime;
 }
 
+void TimeManager::setFromProfile_time(const Profile& p) {
+    Time t;
+
+    if (p.time->type == Profile::Time::Type::Relative) {
```
alexanderbock, line 874: ->
```
switch (p.time->type) {
    case Profile::Time::Type::Relative:
      ....;
    case Profile::Time::Type::Absolute:
     ....;
     default:  throw ghoul::MissingCaseException();
```

That way we'll get a hard error if we introduce a different time specification at some point down the line

alexanderbock (changes requested):

GPayne:
Yes, the ServerModule json.exception.parser_error is only happening with this branch. It is probably related to the handling of the meta contents, but haven't pinpointed it yet.

include/openspace/engine/configuration.h
GPayne: Removed `profileOutPrefixName` and `usingProfile` variables.

include/openspace/engine/openspaceengine.h
GPayne: Changed these names and made them independent functions.

include/openspace/engine/openspaceengine.h
GPayne: Gone.

include/openspace/scene/profile.h
GPayne: Gone

include/openspace/scene/profile.h
GPayne: Also gone

include/openspace/scene/profile.h
GPayne: Gone

include/openspace/scene/scene.h
GPayne: Change

include/openspace/scene/scene.h
GPayne: Changed to independent function.

include/openspace/util/time.h
GPayne: Changed

include/openspace/util/timemanager.h
GPayne: Changed

src/engine/openspaceengine.cpp
GPayne: Made these changes

src/engine/openspaceengine.cpp
GPayne: Done

src/engine/openspaceengine.cpp
GPayne: Deleted this function

src/engine/openspaceengine.cpp
GPayne: They mean different things, but true that `_isRenderingFirstFrame` will accomplish both, so made that change and deleted `_hasInitializedProfile`.

src/engine/openspaceengine.cpp
GPayne: Gone

src/engine/openspaceengine.cpp
GPayne: Added a comment

src/engine/openspaceengine.cpp
GPayne: Done

src/engine/openspaceengine.cpp
GPayne: I like it this way

src/engine/openspaceengine.cpp
GPayne: The std::to_string was necessary to make the compiler happy

src/engine/openspaceengine.cpp
GPayne: True

src/scene/scene.cpp
GPayne: Making the new functions allows me to describe what is happening in the function name itself. I think it's clearer than if I put it all together like that.

src/scene/scene_lua.inl
GPayne: My reason for making a separate function with a longer name is to reduce the time another developer spends trying to understand what the purpose is. I try to do this a lot, so if this is having the opposite effect then we should discuss it more in this thread.

src/util/timemanager.cpp
GPayne: Made this change

src/engine/openspaceengine.cpp
alexanderbock: I wasn't arguing for that they *should* be the same, just wondering if there *were* :)

src/scene/scene_lua.inl
alexanderbock: How about a name along the lines of:
`convertBoolStringToLuaAndMaybePush` or something?  Just to make sure that when someone reads the function call they know that it's not guaranteed that a value push to the stack will happen

GPayne: I changed it to checking what the type of the variable is, but removing the lua push so that there won't be uncertainty about whether or not a push to the lua stack was performed.

GPayne:
The aforementioned ServerModule json.exception.parser_error is now fixed, but there is a camera jitter problem when zooming close to the focus node (this is not happening in the master branch).

https://user-images.githubusercontent.com/3966290/133949164-f0b13f0c-a669-44de-b2c7-2ebf51d63ad0.mp4

.

GPayne:
A call to `updateScene` had been mistakenly deleted. Restoring this fixed the camera jitter problem.

alexanderbock (approved):

PR #1756, opened 2021-09-22
Improved numerical precision on scale value in sessionRecording files in ascii format (Issue#1732)

GPayne:


GPayne:
My test for the lowest possible scale value is zooming out past the cosmic background radiation and verifying that the recorded scale value isn't rounded down to zero.

alexanderbock (approved):

alexanderbock:
Another good test (and how we figured it out in the first place):
 - Create a session recording at the CMB sphere
 - Start OpenSpace with any stereo SGCT configuration
 - Playback the session recording once normally and a second time with `SessionRecording.IgnoreRecordedScale` turned on.  If the stereo value in the session recording is bad, the second playback will show a "jump" of the rendering as the new (incorrect) scale is applied

GPayne:
I ran the suggested test using single_sbs_stereo configuration, and both channels looked consistent with & without `SessionRecording.IgnoreRecordedScale` turned on.

PR #1759, opened 2021-09-29
JWST update

Roxeena:
Updates the JWST profile with the new animated model.

Also includes a "timelapse" script (using the script scheduler) that increases and decreases the simulation time to make the animation look smoother. The script scheduler can now also be turned off with a bool property, to be able to view the animation in real-time as well.


GPayne (approved):
This looks good to me, I'm excited to see the animations! However, I suggest adding some comments in the asset files that use the 2018 times in order to help in the understanding of why these are being used, and details about jumping between time frames. Maybe this will be documented elsewhere in a wiki or something, but I think it would be beneficial to have it here also.

alexanderbock (approved):

PR #1760, opened 2021-10-04
Fixes Problem with Profile Properties Not Handling Table Entries (Issue 1757)

GPayne:
Fixes issue#1757 and allows a profile property to have a table value rather than just a single value. A table of numbers (floats) or strings is supported. Currently only one level of depth is supported (no table-within-a-table).

### DIFF
--- a/src/scene/scene_lua.inl
+++ b/src/scene/scene_lua.inl
@@ -915,8 +915,14 @@ bool isBoolValue(const std::string& s) {
  */
 bool isFloatValue(const std::string& s) {
     try {
-        float converted = std::stof(s);
-        return true;
+        float converted = std::numeric_limits<float>::min();
+        converted = std::stof(s);
+        if (converted != std::numeric_limits<float>::min()) {
+            return true;
+        }
+        else {
+            return false;
+        }
     }
     catch (...) {
         return false;
--- a/src/scene/scene_lua.inl
+++ b/src/scene/scene_lua.inl
@@ -923,4 +923,23 @@ bool isFloatValue(const std::string& s) {
     }
 }
 
+/**
+ * \ingroup LuaScripts
+ * isNilValue(const std::string& s):
+ * Used to check if a string is a lua 'nil' value. Returns false if not.
+ */
+bool isNilValue(const std::string& s) {
+    return (s == "nil");
+}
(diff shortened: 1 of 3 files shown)


src/scene/scene_lua.inl
```diff
@@ -915,12 +915,37 @@ bool isBoolValue(const std::string& s) {
  */
 bool isFloatValue(const std::string& s) {
     try {
-        float converted = std::stof(s);
-        return true;
+        float converted = std::numeric_limits<float>::min();
+        converted = std::stof(s);
+        if (converted != std::numeric_limits<float>::min()) {
```
alexanderbock, line 920: -> `return (converted != std::numeric_limits<float>::min());`

src/scene/scene_lua.inl
```diff
@@ -915,12 +915,37 @@ bool isBoolValue(const std::string& s) {
...
         return false;
     }
 }
 
+/**
+ * \ingroup LuaScripts
+ * isNilValue(const std::string& s):
+ * Used to check if a string is a lua 'nil' value. Returns false if not.
+ */
+bool isNilValue(const std::string& s) {
```
alexanderbock, line 937: `const std::string&` can be a `std::string_view`

src/scene/scene_lua.inl
```diff
@@ -915,12 +915,37 @@ bool isBoolValue(const std::string& s) {
...
+    return (s == "nil");
+}
+
+/**
+ * \ingroup LuaScripts
+ * isTableValue(const std::string& s):
+ * Used to check if a string contains a lua table rather than an individual value.
+ * Returns false if not.
+ */
+bool isTableValue(const std::string& s) {
```
alexanderbock, line 947: `const std::string&` can be a `std::string_view`

src/scene/scene_lua.inl
```diff
@@ -915,12 +915,37 @@ bool isBoolValue(const std::string& s) {
  */
 bool isFloatValue(const std::string& s) {
```
alexanderbock, line 916: `const std::string&` can be a `std::string_view`

alexanderbock (changes requested):

Issue #898, opened 2019-06-03
Parallelized asset synchronization


alexanderbock:
Right now, it is not possible to use an HttpSynchronization from multiple assets at the same time as it will result in both assets trying to download the files simultaneously, causing all but one synchronizations to fail.
It would be good to make it possible to make these thread-safe so that we can remove the synchronization-only assets

alexanderbock:
Closed in https://github.com/OpenSpace/OpenSpace/pull/1784

PR #1784, opened 2021-11-15
Feature/assets

alexanderbock:
General cleanup and code simplification of the Asset loading system


alexanderbock:
- Removes the special case of a single "Asset" as the $root, which caused a lot of complications as that one was not actually backed by a file on disk
 - Simplifies memory management by making the AssetManager explicitly own all loaded Assets, rather than using opaque `shared_ptr`s everywhere
 - Removes `AssetListener` and `AssetLoader` class and incorporate those functions into the remaining `AssetManager` & `Asset` classes

alexanderbock:
Total changelog:
 - Total overhaul of the Asset, AssetManager, ResourceSynchronization, and HttpRequest classes and simplify all of them
 - Removes the special case of a single "Asset" as the $root, which caused a lot of complications as that one was not actually backed by a file on disk
 - Simplifies memory management by making the AssetManager explicitly own all loaded Assets, rather than using opaque shared_ptrs everywhere
 - Also make the AssetManager own all ResourceSynchronizations rather than each Asset owning their own (also closes #898 this way)
 - Adapt asset files by removing the assets that were purely there to synchronize files used by multiple other assets
 - Greatly simplify the HttpRequest / HttpDownload etc classes be coming the diamond-of-death multiple inheritance that was going on there
 - Removes AssetListener, AssetLoader, and SynchronizationWatcher classer and incorporate those functions into the remaining AssetManager & Asset classes
 - Show the failure of synchronization in the Loading Screen

GPayne (approved):

modules/sync/syncmodule.h
```diff
@@ -35,10 +37,7 @@ class SyncModule : public OpenSpaceModule {
 
     SyncModule();
 
-    std::string synchronizationRoot() const;
-
-    void addHttpSynchronizationRepository(std::string repository);
-    std::vector<std::string> httpSynchronizationRepositories() const;
```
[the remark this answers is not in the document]
alexanderbock: There are still more synchronization repositories;  we just don't need these functions anymore.  I thought that we wouldn't need the ability to dynamically add synchronization repositories at runtime anymore, so there is no need for these getter functions and we can just pass the list of repositories directly into the HttpSynchronization objects instead:

https://github.com/OpenSpace/OpenSpace/blob/feature/assets/modules/sync/syncmodule.cpp#L83

modules/sync/syncmodule_lua.inl
```diff
@@ -32,23 +32,20 @@ int syncResource(lua_State* L) {
...
-    );
+    std::unique_ptr<ResourceSynchronization> sync =
+        ResourceSynchronization::createFromDictionary(dict);
 
-    sync.start();
+    sync->start();
 
-    while (sync.isSyncing()) {
+    while (sync->isSyncing()) {
         std::this_thread::sleep_for(std::chrono::milliseconds(20));
```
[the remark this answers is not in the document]
alexanderbock: The only way this `while` loop is going to loop forever is if one of the sychronizations has failed, so I think we should be good without one?   This is what I mentioned on monday, that we now have a red item in the loading screen and the loading screen doesn't go away anymore if the **synchronization** fails

modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -125,174 +116,132 @@ void HttpSynchronization::start() {
...
+            std::make_unique<HttpFileDownload>(
+                line,
+                destination,
+                HttpFileDownload::Overwrite::Yes
+            );
+        HttpFileDownload* dl = download.get();
+        downloads.push_back(std::move(download));
 
-        sizeData[line] = { false, 0, 0 };
+        ghoul_assert(sizeData.find(line) == sizeData.end(), "Duplicate entry");
```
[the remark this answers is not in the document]
alexanderbock: I actually removed the assert since there is an explicit check (with log error) 14 lines higher 😅

include/openspace/scene/asset.h
```diff
@@ -26,156 +26,277 @@
...
+     * about successful or unsuccessful synchronizations.
+     *
+     * \param manager The AssetManager that is responsible for loading this Asset and all
+     *        of its dependencies
+     * \param assetPath The file path that will be used to load this Asset
+     * 
+     * \pre The \p assetPath must not be empty and must be an existing file
      */
-    Asset(AssetLoader* loader, SynchronizationWatcher* watcher);
+    Asset(AssetManager& manager, std::filesystem::path assetPath);
```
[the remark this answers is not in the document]
alexanderbock: Well;  at the moment there is only one AssetManager, but there is no reason that there only needs to be a single one.  So I'd prefer to keep the AssetManager reference to be prepared for multiple of them.

src/scene/assetmanager.cpp
```diff
@@ -24,106 +24,751 @@
...
+        // Instead of deleting the asset directly, we are moving it into a to-delete queue
+        // as this unloadAsset function is called from the asset that we are manipulating
+        // right now. And deleting the asset while we are in a function of the same asset
+        // might be painful
+        _toBeDeleted.push_back(std::move(*it));
+        _assets.erase(it);
+    }
+}
+
+void AssetManager::setUpAssetLuaTable(Asset* asset) {
```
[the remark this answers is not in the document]
alexanderbock: I think it's not too bad? Since it's all actions done in the function belong together?

include/openspace/scene/asset.h
```diff
@@ -26,156 +26,277 @@
...
-        Synchronizing,
-        SyncResolved,
-        SyncRejected,
-        Initialized,
-        InitializationFailed
-    };
-
+    /// This struct contains all the meta information about this asset
     struct MetaInformation {
+        /// The name of the asset
```
WeirdRubberDuck, line 58: Just a question about this comment style, with three `/`: What are the situations where we prefer this one over just two slashes?

modules/sync/syncs/httpsynchronization.cpp
```diff
@@ -125,174 +116,131 @@ void HttpSynchronization::start() {
...
 }
 
 bool HttpSynchronization::trySyncFromUrl(std::string listUrl) {
-    HttpRequest::RequestOptions opt = {};
-    opt.requestTimeoutSeconds = 0;
-
-    SyncHttpMemoryDownload fileListDownload(std::move(listUrl));
-    fileListDownload.onProgress([&c = _shouldCancel](HttpRequest::Progress) {
+    HttpMemoryDownload fileListDownload(std::move(listUrl));
+    fileListDownload.onProgress([&c = _shouldCancel](int64_t, std::optional<int64_t>) {
```
WeirdRubberDuck, line 124: Ooh, nice. Did not know it was possible to capture member variables like this, but makes sense :)

src/scene/asset.cpp
```diff
@@ -475,408 +218,124 @@ void Asset::unload() {
...
-                    id(), p->id(), e.component, e.message
-                ));
-                // TODO: rollback;
-            }
-        }
-    }
-    return true;
 }
 
+/*
```
WeirdRubberDuck, line 286: Is there a reason that this function is kept around even though it's commented out? If so, we need a comment explaining why

src/scene/assetmanager.cpp
```diff
@@ -24,106 +24,751 @@
...
+            return PathType::Absolute;
+        }
+        if (path.size() > 1 && (path[0] == '\\' || path[0] == '/')) {
+            return PathType::Absolute;
+        }
+        return PathType::RelativeToAssetRoot;
+    }
+
+    struct [[codegen::Dictionary(AssetMeta)]] Parameters {
+        std::optional<std::string> name;
```
WeirdRubberDuck, line 66: No docs? :) 

I guess we already have all that information in the files for the assets, but those are not actually exposed to the users anywhere. However, maybe we don't need to, as it should probably live on a wiki page somewhere

src/scene/assetmanager.cpp
```diff
@@ -24,106 +24,751 @@
...
+    lua_pushcclosure(
+        *_luaState,
+        [](lua_State* L) {
+            ZoneScoped
+            
+            Asset* asset = ghoul::lua::userData<Asset>(L, 1);
+            ghoul::lua::checkArgumentsAndThrow(L, 1, "lua::localResourceLua");
+
+            std::string name = ghoul::lua::value<std::string>(L);
+            std::string path = fmt::format(
```
WeirdRubberDuck, line 427: Can this be another good place to use the `operator/` of the `std::filesystem::path`? :) E.g: 
```
std::filesystem::path path = asset->path().parent_path() / name;
ghoul::lua::push(L, path.string());
```

include/openspace/scene/asset.h
alexanderbock: I don't think we have thought about it.  The way I'm using it is either
 1. It's a one line comment anywhere
 2. It's a one or two line comment in the `private` section of a class

But happy to talk about it and come to a consensus

WeirdRubberDuck (approved):
Looks good to me, mostly found typos. Great that you put some effort into providing in-code documentation on it all 🙏 

There are some trailing whitespaces around in several files that I didn't comment on here. I suggest going though the files with a trailing whitespace removal tool 

Also tried breaking it, but didn't manage to. However, when re-syncing all the files I get a warning from the DU obassociations colormap file

image

modules/sync/syncs/httpsynchronization.cpp
alexanderbock: If you wanna capture a std::unique_ptr, you need to do it like this:
```
std::unique_ptr<Foo> foo = ...;
auto fn = [f = std::move(foo)]() {  do(f); };
```

src/scene/asset.cpp
alexanderbock: No reason;  just forgot to actually remove it after commenting it out

src/scene/assetmanager.cpp
alexanderbock: Added documentation

src/scene/assetmanager.cpp
alexanderbock: I really don't like the `/` but I guess I lost that battle 😅

PR #1795, opened 2021-12-20
Feature/sgct json

alexanderbock:
Depends on https://github.com/sgct/sgct/pull/45


apps/OpenSpace/main.cpp
```diff
@@ -997,13 +997,31 @@ std::string selectedSgctProfileFromLauncher(LauncherWindow& lw, bool hasCliSGCTC
...
+            std::filesystem::path c = absPath(config);
+            
+            std::filesystem::path cj = c;
+            cj.replace_extension(".json");
+            
+            std::filesystem::path cx = c;
+            cx.replace_extension(".xml");
+
+            if (c.extension().empty()) {
+                if (std::filesystem::exists(cj)) {
```
WeirdRubberDuck, line 1009: This makes it impossible to load an XML file with the same name if a JSON version with that name already exists. The JSON version will always be the one that is loaded

Very confusing if you have used the conversion script to convert an old XML file to JSON, and want to try each of them individually

apps/OpenSpace/main.cpp
```diff
@@ -997,13 +997,31 @@ std::string selectedSgctProfileFromLauncher(LauncherWindow& lw, bool hasCliSGCTC
...
+                }
+                else {
+                    throw ghoul::RuntimeError(fmt::format(
+                        "Error loading configuration file {}. File could not be found",
+                        config
+                    ));
+                }
             }
             else {
-                config += xmlExt;
```
WeirdRubberDuck, line 1004: The `xmlExt` variable can be removed from the function

config/openvr_htcVive.json
```diff
@@ -0,0 +1,32 @@
...
+          "stereo": "side_by_side",
+          "size": { "x": 1332, "y": 840 },
+          "res": { "x": 3024, "y": 1680 },
+          "viewports": [
+            {
+              "pos": { "x": 0.0, "y": 0.0 },
+              "size": { "x": 1.0, "y": 1.0 },
+              "projection": {
+                "type": "ProjectionPlane",
+                "lowerleft": { "x": -1.7156, "y": -0.965, "z": 0.0 },
```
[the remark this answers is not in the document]
alexanderbock: Good point though.. I'll update SGCT to make it all lowercase.  => Less confusion if no need to wonder about capitalization

scripts/configuration_helper.lua
```diff
@@ -86,820 +72,386 @@ function sgct.config.single(arg) end
...
-    <Size x="1.0" y="1.0" />
-    <PlanarProjection>
-        <FOV down="]]..down..[[" left="]]..left..[[" right="]]..right..[[" up="]]..up..[[" />
-        <Orientation heading="0.0" pitch="0.0" roll="0.0" />
-    </PlanarProjection>
-</Viewport>
-]]
+  local result = {}
+
+  table.insert(result,   [[            {]])
```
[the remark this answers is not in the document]
alexanderbock: Good point.  I think at this point, this is the only place where we are doing it from Lua.  Mainly because the configurtion_helper is loaded before we initialize the OpenSpaceEngine which would be the place that would provide a lot of those libraries

apps/OpenSpace/main.cpp
alexanderbock: This could will only take care of people selecting the configuration from the launcher.  If you put `SGCTConfig = "${CONFIG}/foobar.xml` that XML file will be loaded instead.  And I think we only want to support the JSON files from the Launcher?

scripts/configuration_helper.lua
```diff
@@ -86,820 +72,386 @@ function sgct.config.single(arg) end
...
+    table.insert(result, [[              "tracked": ]] .. tostring(tracked) .. [[,]])
+  end
+
+  table.insert(result,   [[              "projection": {]])
+  table.insert(result,   [[                "type": "PlanarProjection",]])
+  table.insert(result,   [[                "fov": {]])
+  table.insert(result,   [[                  "left": ]] .. left .. [[,]])
+  table.insert(result,   [[                  "right": ]] .. right .. [[,]])
+  table.insert(result,   [[                  "up": ]] .. up .. [[,]])
+  table.insert(result,   [[                  "down": ]] .. down .. [[ ]])
```
Roxeena, line 102: Is the last empty whitespace ( `.. [[ ]]` ) needed?

scripts/configuration_helper.lua
```diff
@@ -86,820 +72,386 @@ function sgct.config.single(arg) end
...
-    </Projectionplane>
-</Viewport>
-]]
+function generateFisheyeViewport(fov, quality, tilt, background, crop, offset, tracked)
+  local result = {}
+
+  table.insert(result,   [[           {]])
+
+  if tracked then
+    table.insert(result, [[             "tracked": ]] .. tostring(tracked) .. ",")
```
Roxeena, line 118: For consistency it could be good to choose either `[[,]]` or`","` and stick to one of them instead of mixing them.

scripts/configuration_helper.lua
```diff
@@ -86,820 +72,386 @@ function sgct.config.single(arg) end
...
-    if arg["sgctDebug"] == nil then
-        arg["sgctDebug"] = false
-    end
 
-    if not arg["windowSize"] then
-        arg["windowSize"] = {1280, 720}
-    end
+function check(type_str, arg, param, subparam_type)
+  local t = type(arg[param])
+  assert(t == type_str or t == "nil", param .. " must be a " .. type_str .. " or nil")
```
Roxeena, line 315: I might be wrong but shouldn't the `t == type_str` actually be `t ~= type_str`? To catch when it is **not** the type you want?

scripts/configuration_helper.lua
```diff
@@ -86,820 +72,386 @@ function sgct.config.single(arg) end
...
-    arg["settings"] = generateSettings(arg)
-    arg["window"] = generateWindow(arg)
-    arg["user"] = generateUser(arg)
-    arg["capture"] = generateCapture(arg)
-
-    return generateCluster(arg)
+  if type_str == "table" and subparam_type and arg[param] then
+    for k, v in pairs(arg[param]) do
+      assert(
+        type(v == subparam_type),
```
Roxeena, line 320: Same here, shouldn't it be `~=` instead? And why isn't the `type` function used on `v` but it is used on `t`?

config/equirectangular_gui.json
```diff
@@ -0,0 +1,58 @@
...
+      ]
+    }
+  ],
+  "users": [
+    {
+      "eyeseparation": 0.065,
+      "pos": { "x": 0.0, "y": 0.0, "z": 0.0 }
+    }
+  ]
+}
```
Roxeena, line 58: Add new line at the end of the file. Same for all json files.

Roxeena (commented):
👍

scripts/configuration_helper.lua
alexanderbock: Nope

apps/OpenSpace/main.cpp
WeirdRubberDuck: You're probably right. But then the XML files shouldn't show up in the launcher at all :)

scripts/configuration_helper.lua
alexanderbock: I don't think so?   We are asserting that the type is correct or nil?

apps/OpenSpace/main.cpp
alexanderbock: :+1: I would agree;  we decided to only deprecate the support for XML formats and not remove it for the next release, so it would be weird for them to still be supported but not show up.
Maybe the solution is to add the extension to the configuration files shown in the launcher?

scripts/configuration_helper.lua
alexanderbock: Done

config/equirectangular_gui.json
alexanderbock: Done

apps/OpenSpace/main.cpp
WeirdRubberDuck: Ah, right. Yeah, it makes sense to show the extension in the launcher, and fix it so that any of the files can be read. 

Another idea is to update the visual appearance of the XML files in the launcher so that it "warns" the user that these will soon be deprecated. That way we would actually target the people who use the file type and inform them how to convert it to the new json format :) Should be relatively easy

alexanderbock: I did that in this PR now.
1. The extensions are shown in the launcher
2. All XML files are in dark yellow color 
3. If at least one XML file is in the list, the dropdown menu gets a tooltip informing that they need to convert their files by the next version

WeirdRubberDuck: Great :) Tried it out quickly and did not see the tooltip though

PR #1839, opened 2022-01-17
Add bool property to toggle follow anchor node rotation

### LINKED ISSUE #1822: Orbital Navigator - Follow node rotation toggle
User has expressed desire to have the follow node rotation act on a toggle rather then based on distance. Having a toggle is effectively the same as using the distance and setting to min or max, however the idea of toggling this behavior may be a paradigm users are used to.

WeirdRubberDuck:
And make property title formatting more consistent throughout OrbitalNavigator

closes #1822

### DIFF
--- a/src/navigation/orbitalnavigator.cpp
+++ b/src/navigation/orbitalnavigator.cpp
@@ -134,24 +134,33 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FollowAnchorNodeInfo = {
+        "FollowAnchorNodeRotation",
+        "Follow Anchor Node Rotation",
+        "If true, the camera will rotate with the current achor node if within a "
+        "certain distance from it. When this happens, the object will appear fixed in "
+        "relation to the camera. The distance at which the change happens is controlled "
+        "through another property."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo
+        FollowAnchorNodeDistanceInfo = {
         "FollowAnchorNodeRotationDistance",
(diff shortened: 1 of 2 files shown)


WeirdRubberDuck:
Very tiny for a PR, but would be good to get a thumbs up on the property documentation

src/navigation/orbitalnavigator.cpp
```diff
@@ -134,24 +134,33 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FollowAnchorNodeInfo = {
+        "FollowAnchorNodeRotation",
+        "Follow Anchor Node Rotation",
+        "If true, the camera will rotate with the current achor node if within a "
+        "certain distance from it. When this happens, the object will appear fixed in "
+        "relation to the camera. The distance at which the change happens is controlled "
+        "through another property."
```
Roxeena, line 142: Maybe refer to the other property by name? Then it is easier for the user to search for it if they want to change it too.

Roxeena (commented):
Looks good to me 👍

src/navigation/orbitalnavigator.cpp
WeirdRubberDuck: Yeah, I had the same idea. But the name of the property is super long (`FollowAnchorNodeRotationDistance`), and it will always be located right next to where this documentation is. So wasn't sure if it was necessary to include the full name 🤔

alexanderbock (approved):

Issue #1842, opened 2022-01-18
RenderableSphere / Atmosphere interaction


micahnyc:
As seen here, the atmosphere and renderable sphere don't interact well.  

I will submit a PR to adjust the Renderbin
![openspace2](https://user-images.githubusercontent.com/708535/150009396-8871bcba-e9b4-4e39-ba6f-894de9b5449e.jpg)

micahnyc:
test and working after #1843

PR #1843, opened 2022-01-18
unified renderbin to renderable; adapted orbitalkepler to codegen; fi…

micahnyc:
…xes 1842


micahnyc:
A few different Renderables had a RenderableType or RenderBinMode or RenderType which all represented the same thing. This unifies the code to the base Renderable class. 

While there is some loss of documentation/naming I think its even better because now the same words always represent the same thing for all renderables.

In addition, classes like RenderableSphere which had UseAdditiveBlending or BackgroundRendering which conflicted with each other now just have the one variable.

micahnyc:
Closes #1842

alexanderbock:
Looks good to me

modules/base/rendering/renderablesphere.cpp
```diff
@@ -386,14 +354,14 @@ void RenderableSphere::render(const RenderData& data, RendererTasks&) {
         glDisable(GL_CULL_FACE);
     }
 
-    if (_useAdditiveBlending) {
+    if (renderBin() == Renderable::RenderBin::PreDeferredTransparent) {
```
alexanderbock, line 357: More readable if accessing the `_renderBin` directly rather than going through the getting mething

modules/base/rendering/renderablesphere.cpp
```diff
@@ -386,14 +354,14 @@ void RenderableSphere::render(const RenderData& data, RendererTasks&) {
...
         glBlendFunc(GL_SRC_ALPHA, GL_ONE);
         glDepthMask(false);
     }
 
     _sphere->render();
 
-    if (_useAdditiveBlending) {
+    if (renderBin() == Renderable::RenderBin::PreDeferredTransparent) {
```
alexanderbock, line 364: Same here

modules/base/rendering/renderabletimevaryingsphere.cpp
```diff
@@ -383,14 +349,14 @@ void RenderableTimeVaryingSphere::render(const RenderData& data, RendererTasks&)
         glDisable(GL_CULL_FACE);
     }
 
-    if (_useAdditiveBlending) {
+    if (renderBin() == Renderable::RenderBin::PreDeferredTransparent) {
```
alexanderbock, line 352: ^^

modules/base/rendering/renderabletimevaryingsphere.cpp
```diff
@@ -383,14 +349,14 @@ void RenderableTimeVaryingSphere::render(const RenderData& data, RendererTasks&)
...
         glBlendFunc(GL_SRC_ALPHA, GL_ONE);
         glDepthMask(false);
     }
 
     _sphere->render();
 
-    if (_useAdditiveBlending) {
+    if (renderBin() == Renderable::RenderBin::PreDeferredTransparent) {
```
alexanderbock, line 359: ^^

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -391,62 +368,25 @@ RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
...
-    if (dict.hasValue<std::string>(RenderBinModeInfo.identifier)) {
-        Renderable::RenderBin cfgRenderBin = RenderBinConversion.at(
-            dict.value<std::string>(RenderBinModeInfo.identifier)
-        );
-        setRenderBin(cfgRenderBin);
-    }
-    else {
-        setRenderBin(Renderable::RenderBin::PostDeferredTransparent);
-    }
+    _startRenderIdx = p.startRenderIdx.value_or(0u);
```
alexanderbock, line 386: Just `0` should work?

Same below

src/rendering/renderable.cpp
```diff
@@ -58,6 +58,14 @@ namespace {
         openspace::properties::Property::Visibility::Hidden
     };
 
+    constexpr openspace::properties::Property::PropertyInfo RenderableRenderBinModeInfo = {
+        "RenderBinMode",
+        "Render Bin Mode",
+        "This value specifies if the plane should be rendered in the Background,"
+        "Opaque, Pre/PostDeferredTransparency, or Overlay rendering step..",
```
alexanderbock, line 65: Extra `.` at the end

src/rendering/renderable.cpp
```diff
@@ -71,6 +79,20 @@ namespace {
...
+        enum class [[codegen::map(openspace::Renderable::RenderBin)]] RenderBinMode {
+            Background,
+            Opaque,
+            PreDeferredTransparent,
+            PostDeferredTransparent,
+            Overlay
+        };
+
+        // [[codegen::verbatim(RenderableRenderBinModeInfo.description)]]
+        std::optional<RenderBinMode> renderBinMode [[codegen::key("RenderBinMode")]];
```
alexanderbock, line 93: This `key` definition is superfluous since the first letter of the variable name is capitalized automatically.

alexanderbock (changes requested):

src/rendering/renderable.cpp
```diff
@@ -58,6 +58,14 @@ namespace {
         openspace::properties::Property::Visibility::Hidden
     };
 
+    constexpr openspace::properties::Property::PropertyInfo RenderableRenderBinModeInfo = {
+        "RenderBinMode",
+        "Render Bin Mode",
+        "This value specifies if the plane should be rendered in the Background,"
```
WeirdRubberDuck, line 64: `plane` should be `renderable`?

src/rendering/renderable.cpp
```diff
@@ -71,6 +79,20 @@ namespace {
...
+            Background,
+            Opaque,
+            PreDeferredTransparent,
+            PostDeferredTransparent,
+            Overlay
+        };
+
+        // [[codegen::verbatim(RenderableRenderBinModeInfo.description)]]
+        std::optional<RenderBinMode> renderBinMode [[codegen::key("RenderBinMode")]];
+
```
WeirdRubberDuck, line 94: nitpicky, but can remove these 2 extra lines

src/rendering/renderable.cpp
```diff
@@ -140,6 +163,12 @@ Renderable::Renderable(const ghoul::Dictionary& dictionary)
     // set type for UI
     _renderableType = p.type.value_or(_renderableType);
     addProperty(_renderableType);
+
+    //only used by a few classes such as RenderableTrail and RenderableSphere
+    if (p.renderBinMode.has_value()) {
+        setRenderBin(codegen::map<Renderable::RenderBin>(*p.renderBinMode));
+    }
+
```
WeirdRubberDuck, line 171: again, nitpicky, but I would remove this extra line here

(and add a space after the `//` in the comment on line 167)

src/rendering/renderable.cpp
```diff
@@ -220,7 +249,9 @@ void Renderable::onEnabledChange(std::function<void(bool)> callback) {
 }
 
 void Renderable::setRenderBinFromOpacity() {
-    if (_renderBin != Renderable::RenderBin::PostDeferredTransparent) {
+    if ((_renderBin != Renderable::RenderBin::PostDeferredTransparent) &&
+        (_renderBin != Renderable::RenderBin::Overlay))
```
WeirdRubberDuck, line 253: What's the motivation for adding the `Overlay` here? (Wondering why it wasn't there before)

Also, with this change, this code is exactly the same as the onChange being registered in `registerUpdateRenderBinFromOpacity`, so that onChange should probably just call this function.

WeirdRubberDuck (changes requested):
Needs a few small changes but looks good to me. 

The feature is a bit difficult to use since the property is not added to the UI and cannot be updated during runtime. Would it maybe be more desirable to have it as a property with visibility "Developer" rather than "Hidden"? I could see the use of being able to change it during runtime, for debug purposes

Issue #1854, opened 2022-01-25
Update navigation UI to support camera path navigation


WeirdRubberDuck:
The time picker UI should also be updated to prevent interaction with time control while a camera path is active

PR #1858, opened 2022-01-27
Engine modes for session recording and camera path playback

### LINKED ISSUE #1845: Simulation time can be changed while a camera path is active
It is currently possible to set the simulation time (either through the GUI or through script) while a camera path is active

This should be prevented, as changing the time while flying leads to objects moving in space and the camera will not end up where expected.

WeirdRubberDuck:
A first (small) step towards better handling of "modes" like session recording, camera path playback, and so on, in terms of handling how time and navigation should be controlled. Ideally, we want to create some abstraction, so that this is centralized in a better way and also includes things like the ParallelConnection and Recording of a SessionRecording

Some noteworthy updates: 
* Handle pausing and time changes during camera paths (Fixes #1845 )
* Prevent invalid switching between Session Recording Playback and Camera Paths
* Some cleanup, like removing the basically unused ExternInteraction class

The modes can later be used by the UI, but I plan on looking more into this as part of the task on UI for camera paths (#1854 )


include/openspace/network/messagestructureshelper.h
```diff
@@ -0,0 +1,57 @@
...
+
+#ifndef __OPENSPACE_CORE___MESSAGESTRUCTURESHELPER___H__
+#define __OPENSPACE_CORE___MESSAGESTRUCTURESHELPER___H__
+
+#include <openspace/network/messagestructures.h>
+
+namespace openspace::datamessagestructures {
+
+/**
+ * Method that accepts a reference to a CameraKeyframe object, and populates
```
alexanderbock, line 33: The documentation is not accurate anymore since these three functions don't take a reference but return a new instance

alexanderbock (commented):
Looks good to me codewise.  Have to run it still

alexanderbock (approved):
Looks good to me

src/util/timemanager.cpp
```diff
@@ -403,10 +421,14 @@ void TimeManager::clearKeyframes() {
...
 void TimeManager::setTimeNextFrame(Time t) {
+    OpenSpaceEngine::Mode m = global::openSpaceEngine->currentMode();
+    if (global::openSpaceEngine->currentMode() == OpenSpaceEngine::Mode::CameraPath) {
+        LERROR("Cannot change simulation time during camera path");
+        return;
+    }
     _shouldSetTime = true;
     _timeNextFrame = std::move(t);
     clearKeyframes();
-    global::scriptScheduler->setCurrentTime(t.j2000Seconds());
```
GPayne, line 409: I see that you moved this to `progressTime` function. Seems like it should be fine, but I would give it a second look to make sure it wouldn't introduce any timing issues.

src/interaction/sessionrecording.cpp
```diff
@@ -465,10 +469,10 @@ bool SessionRecording::startPlayback(std::string& filename,
     global::eventEngine->publishEvent<events::EventSessionRecordingPlayback>(
         events::EventSessionRecordingPlayback::State::Started
     );
-    initializePlayback_triggerStart();
-
     global::navigationHandler->orbitalNavigator().updateOnCameraInteraction();
 
+    initializePlayback_triggerStart();
```
GPayne, line 474: Why move this from before to after the `updateOnCameraInteraction` call?

src/navigation/navigationhandler.cpp
```diff
@@ -328,23 +326,11 @@ void NavigationHandler::updateCameraTransitions() {
...
-    _useKeyFrameInteraction = false;
-}
-
-void NavigationHandler::triggerPlaybackStart() {
-    _playbackModeEnabled = true;
-}
-
-void NavigationHandler::stopPlayback() {
+void NavigationHandler::handleSessionRecordingPlaybackEnd() {
+    // When the OrbitalNavigator takes over after ha session recording has finished
```
GPayne, line 330: typo in comment

src/navigation/navigationhandler.cpp
```diff
@@ -328,23 +326,11 @@ void NavigationHandler::updateCameraTransitions() {
...
+    // playing, we need to make sure that it's update variables are updated correctly
     _orbitalNavigator.resetVelocities();
-    _orbitalNavigator.resetNodeMovements();
-    _playbackModeEnabled = false;
+    _orbitalNavigator.updatePreviousStateVariables();
```
GPayne, line 333: Just a suggestion: the 2 lines in this function are grouped together in `applyNavigationState` also. You could rename this function and call it from there. Renaming it would make sense because even though it's called from session recording, it isn't specific to that.

include/openspace/engine/openspaceengine.h
```diff
@@ -65,6 +65,15 @@ struct CommandlineArguments {
 
 class OpenSpaceEngine {
 public:
+    // A mode that specifies which part of the system is currently in control.
+    // The mode can be used to limit certain features, like setting time, navigation
+    // or triggering scripts
+    enum class Mode {
```
GPayne, line 71: I like these different modes, but I think it should be named `InteractionMode` or something. From elsewhere in the code, the mode of the OpenSpaceEngine would seem broad from a readability standpoint.

src/navigation/orbitalnavigator.cpp
```diff
@@ -897,17 +885,29 @@ void OrbitalNavigator::setAimNode(const std::string& aimNode) {
...
-void OrbitalNavigator::resetNodeMovements() {
+void OrbitalNavigator::updatePreviousAnchorState() {
     if (_anchorNode) {
         _previousAnchorNodePosition = _anchorNode->worldPosition();
         _previousAnchorNodeRotation = glm::quat_cast(_anchorNode->worldRotationMatrix());
     }
     else {
-        _previousAnchorNodePosition = glm::dvec3(0.0);
-        _previousAnchorNodeRotation = glm::dquat();
+        _previousAnchorNodePosition.reset();
```
GPayne, line 894: I like your renaming of `resetNodeMovements` to `updatePreviousAnchorState`. I prefer not using reset() here and just sticking with the original code that sets the position to zero because it's clear what is happening. A "position reset" could mean a set back to zero, or it could mean a set back to some previous value...

src/navigation/pathnavigator.cpp
```diff
@@ -263,11 +257,20 @@ void PathNavigator::startPath() {
...
+    if (!global::openSpaceEngine->setMode(OpenSpaceEngine::Mode::CameraPath)) {
+        LERROR("Could not start camera path");
+        return; // couldn't switch to camera path mode
+    }
+
     // Always pause the simulation time when flying, to aovid problem with objects
     // moving. However, keep track of whether the time was running before the path
     // was started, so we can reset it on finish
     if (!global::timeManager->isPaused()) {
-        global::timeManager->setPause(true);
```
GPayne, line 270: Why switch to the script queue for the pause? Is this for parallel peer? (which I guess if so would make sense)

GPayne:
I went through the session recording tests and all passed except one: If a playback attempt fails, it logs a "cannot switch to the currently active mode" error and locks-out the mouse control navigation, requiring a restart of OpenSpace. To reproduce this, enter `openspace.sessionRecording.startPlayback('badfilename')` in the console (also happens if a playback file has a format error).

src/util/timemanager.cpp
WeirdRubberDuck: I asked Malin (who added this line previously) to verify that no problems occurred by moving this line and according to her tests it did not introduce any problems

src/navigation/pathnavigator.cpp
WeirdRubberDuck: Mainly so that the pause is recorded in the session recording, and synced out to all the machines when for example in a dome setup. But yeah, I guess parallel peer is another use case

include/openspace/engine/openspaceengine.h
WeirdRubberDuck: I sort of agree, and like the name `ÌnteractionMode`. Good suggestion

Discussed this with @alexanderbock and he preferred a more generic name. But `InteractionMode` would be more descriptive and potentially reduce confusion. Thoughts @alexanderbock ?

src/interaction/sessionrecording.cpp
WeirdRubberDuck: No real reason, apart from that it was moved and grouped together with some other function calls somewhere in the middle of the process of this PR. I'll undo this change

src/navigation/navigationhandler.cpp
WeirdRubberDuck: I'm aware, and I agree. Do you maybe have a good suggestion for a name ? 😄 Because my only idea so far is something like `resetVelocitiesAndUpdateStateVariables` and that just seemed silly.

src/navigation/orbitalnavigator.cpp
WeirdRubberDuck: I get your point, and frankly, I think the implementation of these previous positions is sort of confusing and ideally I would want to reimplement them so that it's more apparent what is going on (I started looking into that for this PR but I need to truly understand the whole update method to make sure I don't break anything, so let it go for now. But that's why there are so many changes in OrbitalNavigator in this PR)

The `_previousAnchorNodePosition` is actually a `std::optional` and is used in the updateCamera method to compute the difference between the previous position and the current one, if it has a value. So I see a danger of setting it to zero, as this could lead to big undesired differences that in turn affect the camera movement. Note that I also split up the anchor and aim reset functions and replaced code in a couple of other places in the orbitalnavigator, that used the reset version. So I wouldn't want to change it. 

However, what the `reset()` really does is just setting the value to `std::nullopt`. Maybe changing the line to 
```
_previousAnchorNodePosition = std::nullopt;
```
would be clearer?

include/openspace/engine/openspaceengine.h
alexanderbock: Yea, I think `InteractionMode` would be too restrictive, since Time for example is gated through this, but that's not really anything Interaction-like.    I could go with `Controller`?

alexanderbock:
Good catch!

WeirdRubberDuck:
> I went through the session recording tests and all passed except one: If a playback attempt fails, it logs a "cannot switch to the currently active mode" error and locks-out the mouse control navigation, requiring a restart of OpenSpace. To reproduce this, enter `openspace.sessionRecording.startPlayback('badfilename')` in the console (also happens if a playback file has a format error).

Great, thanks for checking and good catch!

WeirdRubberDuck:
@GPayne I didn't get exactly the problem as you described it, but there was a problem with that the mode was successfully set even though the session recording failed to open. As the reset of the state was only done on the session recording ended call, it got stuck in the session recording mode. Now the mode is changed as the very last step when playing back a SR, which fixes that problem

A drawback with this is that the work of loading the session recording file is always done, even if it's not currently possible to switch to session recording playback. However, at least the user is informed of any problem in the file he/she tries to load even if the mode cannot be set. So this is probably not a problem at all

Sorry, rambling... 🦆  But yeah it should be fixed now

src/navigation/navigationhandler.cpp
GPayne: I don't have an idea, but I don't mind longer, descriptive names, so your suggestion doesn't sound as silly to me :)

include/openspace/engine/openspaceengine.h
GPayne: The motivation behind my comment is that the OpenSpaceEngine does a lot of different things, and so the 'mode' of the engine could apply to any of them. I'm thinking from the point of view of a developer who is studying the engine code and doesn't know much about camera paths/session recording. Another way to do this would be to keep the `Mode` name but have different names like `UserControl` rather than `Normal`.

src/navigation/navigationhandler.cpp
WeirdRubberDuck: Haha, I see. I'll give it another thought and otherwise go with it :) It would make it clearar that this is not something we only want to do when the session recording ends but in other cases too

include/openspace/engine/openspaceengine.h
alexanderbock: :+1: For `UserControl`

WeirdRubberDuck: 👍

WeirdRubberDuck:
Just waiting for a thumbs up from @GPayne on this, and then I'll consider rit ready for merge :)

PR #1863, opened 2022-01-28
Feature/Science on the Sphere

alexanderbock:
Adds all publicly available Science on a Sphere datasets from NOAA's FTP server to OpenSpace.  All of these are under the data/asset/scene/solarsystem/earth/noaa-sos folder.  Each layer will download the necessary files from the server at startup and thus require some rather large downloads.

In terms of features, this PR adds the ability to provide a Vec4 as a BackgroundColor to ScreenSpaceRenderables which is alpha-blended with the incoming pixels.  This was necessary to support SOS legends that are fully transparent and use a dark text color.  Otherwise those legends would be impossible to read


data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2000_carbon_monoxide.asset
```diff
@@ -0,0 +1,58 @@
+local globeIdentifier = asset.require("../../earth").Earth.Identifier
+
+
+local Name = "Carbon Monoxide - 2000"
+local Identifier = "noaa-sos-atmosphere-carbon_monoxide"
+local Description = [[ Carbon monoxide is known as "The Silent Killer" because it is a colorless, odorless, tasteless gas that is poisonous to humans and other oxygen breathing organisms. It is also naturally occurring in the atmosphere. Only in high concentrations is carbon monoxide deadly. At the Earth's surface the concentration of carbon monoxide is 100 ppb (parts per billion), however, in urban areas it can get up to 10 ppm (parts per million) or 100 times higher. One of the main anthropogenic sources of carbon monoxide in the atmosphere is emissions from automobiles. In areas with heavy traffic the carbon monoxide can be measured at 50 ppm. The natural sources of carbon monoxide include volcanoes and brush burning. Between the anthropogenic and natural sources, scientists estimate that the annual production of carbon monoxide is 2-5 gigatons.  ]]
+local URL = "https://sos.noaa.gov/catalog/datasets/carbon-monoxide-2000/"
+
+local syncedDirectory = asset.syncedResource({
+	Name = Name,
```
WeirdRubberDuck, line 10: Note that these initial spaces within the `syncedResource` are tabs, instead of whitespaces. Should be changed in all files

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2000_carbon_monoxide.asset
```diff
@@ -0,0 +1,58 @@
+local globeIdentifier = asset.require("../../earth").Earth.Identifier
+
+
+local Name = "Carbon Monoxide - 2000"
+local Identifier = "noaa-sos-atmosphere-carbon_monoxide"
+local Description = [[ Carbon monoxide is known as "The Silent Killer" because it is a colorless, odorless, tasteless gas that is poisonous to humans and other oxygen breathing organisms. It is also naturally occurring in the atmosphere. Only in high concentrations is carbon monoxide deadly. At the Earth's surface the concentration of carbon monoxide is 100 ppb (parts per billion), however, in urban areas it can get up to 10 ppm (parts per million) or 100 times higher. One of the main anthropogenic sources of carbon monoxide in the atmosphere is emissions from automobiles. In areas with heavy traffic the carbon monoxide can be measured at 50 ppm. The natural sources of carbon monoxide include volcanoes and brush burning. Between the anthropogenic and natural sources, scientists estimate that the annual production of carbon monoxide is 2-5 gigatons.  ]]
```
WeirdRubberDuck, line 6: Maybe line break the description, so it's readable? 

It's useful to be able to read it when looking at the asset and trying to figure out what the layer contains and whether one wants to add it

WeirdRubberDuck: Same goes for all files ofc

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2000_carbon_monoxide.asset
```diff
@@ -0,0 +1,58 @@
...
+  openspace.addScreenSpaceRenderable(colorbar);  
+end)
+
+asset.onDeinitialize(function()
+  openspace.globebrowsing.deleteLayer(globeIdentifier, "ColorLayers", layer.Identifier)
+  openspace.removeScreenSpaceRenderable(colorbar.Identifier)
+end)
+
+
+asset.meta = {
```
WeirdRubberDuck, line 51: For some reason, the meta info for the added NOAA layers are not shown in the GUI. Should be investigated

WeirdRubberDuck (changes requested):

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2000_carbon_monoxide.asset
alexanderbock: Done for all files

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2000_carbon_monoxide.asset
alexanderbock: Done for all files

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2000_carbon_monoxide.asset
alexanderbock: The `asset.export` was missing. So the meta information was missing.  The description is there but layers in general don't show the meta information it seems

data/assets/scene/solarsystem/planets/earth/noaa-sos/overlays/cables.asset
```diff
@@ -0,0 +1,40 @@
...
+
+local layer = {
+  Identifier = Identifier,
+  Name = Name,
+  FilePath = syncedDirectory .. "4096.png",
+  Description = Description
+}
+
+asset.onInitialize(function() 
+  openspace.globebrowsing.addLayer(globeIdentifier, "ColorLayers", layer)
```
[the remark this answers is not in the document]
alexanderbock: Done

WeirdRubberDuck:
Tried out some more layers and get a crash in the `TemporalTileProvider` (Folder Mode) when adding the `models/rcp-gaXX` layers. The exception happens when enabling the layer from the UI

Seems like the synced files needs to be unzipped onInitialize, similarly  to for example: `\models\ipcc_temp-ccsm-a1b.asset` ?

But ideally, the exception that causes the crash should also be caught and handled :)

WeirdRubberDuck:
Also, with the update to the most recent master, none of the layers' meta info shows up in the UI anymore (previously it was only the new NOAA layers) 🤔 But I'll create a separate issue for that as the bug is now in the master branch as well

image

modules/globebrowsing/src/tileprovider/temporaltileprovider.cpp
```diff
@@ -257,15 +257,33 @@ TemporalTileProvider::TemporalTileProvider(const ghoul::Dictionary& dictionary)
...
+                        "{}-{}T{}:{}:{}",
+                        tm.tm_year + 1900,
+                        tm.tm_yday,
+                        tm.tm_hour,
+                        tm.tm_min,
+                        tm.tm_sec
+                    );
+                }
+                else {
+                    std::string date = fmt::format(
```
GPayne, line 277: No need for the `std::string` here

GPayne (approved):

alexanderbock:
9 layers contain data files that are currently too large (>4 GiB) to be extracted by OpenSpace.  These are:
 - atmosphere/nccs_models-chem
 - atmosphere/2005_hurricane-grayir
 - atmosphere/co_gmd
 - atmosphere/2012_hurricane
 - atmosphere/carbontracker_2000_2100-fixed_scale
 - atmosphere/hurricane_season_2017
 - atmosphere/carbontracker_2000_2100-sliding_scale
 - atmosphere/nccs_models-winds
 - land/day_night-full_year

Right now, these layers throw an error on loading instructing the user to extract the zip files manually before loading the layer, similarly to how we handle the additional globebrowsing layers.  The proper solution would require fixing #1877 which is the root cause of the files not being extractable, but I deem this out-of-scope for this PR

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/2012_hurricane.asset
```diff
@@ -0,0 +1,84 @@
...
+  Identifier = Identifier .. "-names",
+  Name = Name .. "(Names)",
+  Type = "ImageSequenceTileLayer",
+  FolderPath = namesDestination,
+  Description = Description
+}
+
+local layer_sat = {
+  Identifier = Identifier .. "-sat",
+  Name = Name .. "(Sat)",
```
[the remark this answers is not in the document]
alexanderbock: Fixed

data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/carbontracker_2000_2100-fixed_scale.asset
```diff
@@ -0,0 +1,110 @@
...
+
+
+local syncedDirectory = asset.syncedResource({
+  Name = Name,
+  Type = "HttpSynchronization",
+  Identifier = Identifier,
+  Version = 1
+})
+
+local fieldDestination = syncedDirectory .. "field/"
```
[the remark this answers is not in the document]
alexanderbock: I'd disagree since the meaning behind them is different (see https://gcc.godbolt.org/z/3P8xsPvv5)
the version without the "/" at the end is pointing *at* the directory, the one with the trailing / points *into* the directory.  In all of these asset cases we want to refer to the folder itself, rather than treating it as the first argument to a path substitution

data/assets/scene/solarsystem/planets/earth/noaa-sos/land/forests.asset
```diff
@@ -0,0 +1,114 @@
...
+  Type = "ScreenSpaceImageLocal",
+  TexturePath = syncedDirectory .. "forest_label.png",
+  FaceCamera = false,
+  Scale = 0.13,
+  CartesianPosition = { 0.5, 0.0, -2.0 }
+}
+
+local label1 = {
+  Identifier = Identifier .. "-1",
+  Name = Name .. "(1)",
```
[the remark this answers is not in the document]
alexanderbock: Renamed it anyway since I agree that it's not a good name:
```cpp
local label1 = {
  Identifier = Identifier .. "-alt",
  Name = Name .. " (alt)",
```

data/assets/scene/solarsystem/planets/earth/noaa-sos/land/land_cover-slideshow.asset
```diff
@@ -0,0 +1,267 @@
...
+local label_16 = {
+  Identifier = Identifier .. "-16",
+  Name = Name .. "(16)",
+  Type = "ScreenSpaceImageLocal",
+  TexturePath = labelsDestination .. "lcc_16.jpg",
+  Enabled = false,
+  FaceCamera = false,
+  CartesianPosition = { 0.75, 0.0, -2.0 }
+}
+
```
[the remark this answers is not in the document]
alexanderbock: Agreed;  fixed

PR #1864, opened 2022-01-31
Feature/sgct gui

GPayne:
Adds a GUI for window/monitor configuration, available from the launcher.

Gives GUI control of basic to intermediate level SGCT configuration, using the new json format. Up to 3 windows are available with a single monitor, or 4 windows for a dual monitor setup. Available projection options are Planar, Fisheye, Spherical Mirror, Cylindrical, Equirectangular, and Spout (Fisheye or Equirectangular options).
The monitor(s) are shown at the top of the GUI, arranged according to the operating system's display configuration. Window(s) are shown in their selected position(s) and size(s).

At runtime, a new configuration can be created, saved, and then run from the launcher. However, the GUI cannot open an existing configuration file since it does not support all SGCT settings.


alexanderbock:
Just some general comments first:

1. Would be nice to align the Horizontal FOV / Vertical FOV text boxes
image
image

2. In front of the "Monitor 1" dropdown, an additional text would be nice.  Something like: "Monitor"  or "Location"
3. Adding "pixels" after the "Size" and "Offset" lines might give good information to the user on the units
4. The Monitor numbering is on my screen a bit too high and clips the border
image
5. Not sure how much work it would be, but it would be nice to be able to change the Size and Offset with the mouse wheel.  If you scoll it would tick the value up or down by 1 depending on the direction of the mouse wheel.
6. I'd probably think that the "Add Window" and "Remove Window" should be below the window columns
7. We should add tool tips to the individual elements to explain what they mean.  Can just copy-paste from the SGCT config wiki for that

apps/OpenSpace/ext/launcher/include/sgctedit/monitorbox.h
```diff
@@ -0,0 +1,95 @@
...
+    void computeScaledResolution_portrait(const float aspectRatio, const float maxHeight);
+    unsigned int _maxNumMonitors = 2;
+    QRectF _monitorWidgetSize;
+    QRectF _monitorBoundaryRect;
+    unsigned int _nMonitors = 1;
+    std::vector<QRect> _monitorResolution;
+    std::vector<QRectF> _monitorDimensionsScaled;
+    std::vector<QRectF> _windowResolutions;
+    std::vector<QRectF> _windowRendering = {
+        {0.0, 0.0, 0.0, 0.0},
```
alexanderbock, line 79: For most here,  ->  0.f  rather than 0.0 since they are floats, not doubles

apps/OpenSpace/ext/launcher/include/sgctedit/monitorbox.h
```diff
@@ -0,0 +1,95 @@
...
+        {0.0, 0.0, 0.0, 0.0},
+        {0.0, 0.0, 0.0, 0.0},
+        {0.0, 0.0, 0.0, 0.0}
+    };
+    unsigned int _nWindows = 1;
+    QString* _colorsForWindows;
```
alexanderbock, line 85: This one should be stored directly as a `QColor` so that it can't be manipulated from the outside without going through a setter method

apps/OpenSpace/ext/launcher/include/sgctedit/monitorbox.h
```diff
@@ -0,0 +1,95 @@
...
+    int _alphaWindowOpacity = 170;
+    float _monitorScaleFactor = 1.0;
+    float _offset[2] = {10.0, 10.0};
```
alexanderbock, line 88: This variable is not used anywhere and can die

apps/OpenSpace/ext/launcher/include/sgctedit/monitorbox.h
```diff
@@ -0,0 +1,95 @@
...
+    bool _showLabel = false;
+    float _marginFractionOfWidgetSize = 0.025;
```
alexanderbock, line 90: This is not actually a variable and should be moved into the `cpp` file and declared as a `constexpr const MarginFractionOfWidgetSize` since it is a compile-time setting instead

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include "sgctedit/monitorbox.h"
+
+MonitorBox::MonitorBox(QRect widgetDims, std::vector<QRect> monitorResolution,
+                         unsigned int nWindows, QString* winColors)
```
alexanderbock, line 28: Nitpicky;  `unsigned int` does not align with the `QRect` above

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    , _monitorResolution(monitorResolution)
+    , _nWindows(nWindows)
+    , _colorsForWindows(winColors)
+{
+    _nMonitors = monitorResolution.size();
+    _showLabel = (_nMonitors > 1);
+    mapMonitorResolutionToWidgetCoordinates();
+}
+
+MonitorBox::~MonitorBox()
```
alexanderbock, line 39: Since the destructor is empty, it shouldn't be defined and declared at all

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+}
+
+void MonitorBox::paintEvent(QPaintEvent *event)
+{
```
alexanderbock, line 44: Opening `{` goes on the same line as the function

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    _nMonitors = monitorResolution.size();
+    _showLabel = (_nMonitors > 1);
+    mapMonitorResolutionToWidgetCoordinates();
+}
+
+MonitorBox::~MonitorBox()
+{
+}
+
+void MonitorBox::paintEvent(QPaintEvent *event)
```
alexanderbock, line 43: -> `QPaintEvent* event`   same for the header file

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    Q_UNUSED(event);
```
alexanderbock, line 45: Don't use this, instead just don't give the function parameter a name to suppress the warning

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+    //Draw & fill monitors over the out-of-bounds regions
+    paintMonitorBackgrounds(painter);
+    //Draw window(s) over both out-of-bounds and monitors
+    for (unsigned int i = 0; i < _nWindows; ++i) {
+        paintWindow(painter, i);
+    }
+}
+
+void MonitorBox::paintWidgetBorder(QPainter& painter, const int width, const int height) {
```
alexanderbock, line 62: No `const` parameters

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    painter.setPen(QPen(Qt::gray, 4));
```
alexanderbock, line 63: Magic constants;  would be better to declare them locally instead to increase readability:
```cpp
    constexpr const int LineWidth = 4;
    constexpr const int Radius = 10;
    painter.setPen(QPen(Qt::gray, LineWidth));
    painter.drawRoundedRect(0, 0, width - 1, height - 1, Radius, Radius);
```

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    painter.drawRoundedRect(0, 0, width - 1, height - 1, 10, 10);
+}
+
+void MonitorBox::paintMonitorBackgrounds(QPainter& painter) {
+    painter.setPen(QPen(Qt::black, 2));
```
alexanderbock, line 68: Same here

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    painter.setFont(QFont("Arial", 14));
+    for (unsigned int i = 0; i < _nMonitors; ++i) {
+        if (i <= _monitorDimensionsScaled.size()) {
+            painter.drawRect(_monitorDimensionsScaled[i]);
+            QColor fillColor("#DDDDDD");
```
alexanderbock, line 73: `QColor` has a  constructor that takes the colors directly,  no need for string parsing:
`QColor fillColor(221, 221, 221);`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+            QBrush brush(fillColor);
+            brush.setStyle(Qt::SolidPattern);
+            painter.fillRect(_monitorDimensionsScaled[i], brush);
+            if (_showLabel) {
+                QPointF textPos = QPointF(_monitorDimensionsScaled[i].left() + 5,
```
alexanderbock, line 78: alignment:
```cpp
QPointF textPos = QPointF(
    _monitorDimensionsScaled[i].left() + 5,
    _monitorDimensionsScaled[i].top() + 18
);
```

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+                    _monitorDimensionsScaled[i].top() + 18);
+                painter.drawText(textPos, QString::fromStdString(std::to_string(i + 1)));
```
alexanderbock, line 80: -> `QString::number(i+1)`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+            }
+        }
+    }
+}
+
+void MonitorBox::paintWindowBeyondBounds(QPainter& painter, const unsigned int winIdx) {
```
alexanderbock, line 86: No const parameter for basic types

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    painter.setBrush(Qt::BDiagPattern);
+    setPenSpecificToWindow(painter, winIdx, false);
+    if (winIdx <= _windowRendering.size()) {
+        painter.drawRect(_windowRendering[winIdx]);
+    }
+    setPenSpecificToWindow(painter, winIdx, true);
+    painter.setBrush(Qt::NoBrush);
+}
+
+void MonitorBox::paintWindow(QPainter& painter, const unsigned int winIdx) {
```
alexanderbock, line 96: No const parameter for basic types

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    setPenSpecificToWindow(painter, winIdx, true);
+    if (winIdx <= _windowRendering.size()) {
```
alexanderbock, line 98: `winIdx` should be `size_t` or else it throws a warning here

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+void MonitorBox::windowDimensionsChanged(unsigned int monitorIdx, unsigned int windowIdx,
+                                                              const QRectF& newDimensions)
+{
+    mapWindowResolutionToWidgetCoordinates(monitorIdx, windowIdx, newDimensions);
+}
+
+void MonitorBox::mapMonitorResolutionToWidgetCoordinates() {
+    QSize virtualDesktopResolution;
+    float maxWidth = 0.0;
```
alexanderbock, line 132: 0.f

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    float maxHeight = 0.0;
+    for (auto m : _monitorResolution) {
```
alexanderbock, line 134: -> `const QRect&` instead of `auto`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+    for (size_t m = 0; m < _monitorResolution.size(); ++m) {
+        _monitorDimensionsScaled.push_back({
+            _monitorOffsets[m].width(),
+            _monitorOffsets[m].height(),
+            _monitorResolution[m].width() * _monitorScaleFactor,
+            _monitorResolution[m].height() * _monitorScaleFactor,
+        });
+    }
+    this->update();
```
alexanderbock, line 157: no need for the `this` here

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+void MonitorBox::computeScaledResolution_landscape(const float aspectRatio,
```
alexanderbock, line 160: `const`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+                + _monitorResolution[m].x() * _monitorScaleFactor,
+            _marginWidget + _monitorResolution[m].y() * _monitorScaleFactor
+        });
+    }
+}
+
+void MonitorBox::setNumWindowsDisplayed(unsigned int nWindows) {
+    if (_nWindows != nWindows) {
+        _nWindows = nWindows;
+        this->update();
```
alexanderbock, line 197: No need for `this`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
```diff
@@ -0,0 +1,216 @@
...
+        return;
+    }
+    QRectF wF = w;
+    _windowRendering[wIdx] = {
+        _monitorDimensionsScaled[mIdx].x() + wF.left() * _monitorScaleFactor,
+        _monitorDimensionsScaled[mIdx].y() + wF.top() * _monitorScaleFactor,
+        wF.width() * _monitorScaleFactor,
+        wF.height() * _monitorScaleFactor
+    };
+    this->update();
```
alexanderbock, line 215: Same here

apps/OpenSpace/ext/launcher/include/sgctedit/orientation.h
```diff
@@ -0,0 +1,68 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_UI_LAUNCHER___ORIENTATION___H__
+#define __OPENSPACE_UI_LAUNCHER___ORIENTATION___H__
+
+#include <QCheckBox>
```
alexanderbock, line 28: Correct order:
```cpp

#include <QWidget>

#include <sgctedit/orientationdialog.h>
#include <sgct/config.h>
#include <sgct/math.h>
#include <QCheckBox>
#include <QFrame>
#include <QIntValidator>
#include <QLabel>
#include <QLayout>
#include <QLineEdit>
#include <QPainter>
#include <QPainterPath>
#include <QPoint>
#include <QPushButton>
#include <QVector>
#include <vector>
#include <iostream>
```

Also many of the includes are not necessary and can be removed, replaced with forward declarations, or moved into the cpp file

apps/OpenSpace/ext/launcher/include/sgctedit/orientation.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <vector>
+#include <iostream>
+#include <sgctedit/orientationdialog.h>
+#include <sgct/config.h>
+#include <sgct/math.h>
+
+
+class Orientation : public QWidget
+{
```
alexanderbock, line 49: Opening `{` goes on the same line as `class`

apps/OpenSpace/ext/launcher/include/sgctedit/orientation.h
```diff
@@ -0,0 +1,68 @@
...
+Q_OBJECT
+public:
+    explicit Orientation();
```
alexanderbock, line 52: `explicit` doesn't do anything here since the constructor doesn't have any parameters

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+ *                                                                                       *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,   *
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include "sgctedit/display.h"
```
alexanderbock, line 25:
```
#include "sgctedit/orientationdialog.h"

#include "sgctedit/display.h"
#include "sgctedit/orientation.h"
```

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "sgctedit/orientation.h"
+#include "sgctedit/orientationdialog.h"
+
+Orientation::Orientation()
+{
```
alexanderbock, line 30: Opening `{` goes on the same line as the function declaration

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+    _orientationDialog = new OrientationDialog(_orientationValue, this);
+    _layoutOrientationFull = new QHBoxLayout();
```
alexanderbock, line 32: Same argument as above:  `_layoutOrientationFull = new QHBoxLayout`

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include "sgctedit/display.h"
+#include "sgctedit/orientation.h"
+#include "sgctedit/orientationdialog.h"
+
+Orientation::Orientation()
+{
+    _orientationDialog = new OrientationDialog(_orientationValue, this);
```
alexanderbock, line 31: Since it is always created, consider defining `_orientationDialog` as `OrientationDialog` instead and create it in the constructor directly without the `new`

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+    {
+        QVBoxLayout* layoutOrientationControls = new QVBoxLayout();
+        QPushButton* orientationButton = new QPushButton("Global Orientation");
+        _checkBoxVsync = new QCheckBox("VSync All Windows", this);
+        layoutOrientationControls->addWidget(_checkBoxVsync);
+        layoutOrientationControls->addWidget(orientationButton);
+        _layoutOrientationFull->addStretch(1);
+        _layoutOrientationFull->addLayout(layoutOrientationControls);
+        _layoutOrientationFull->addStretch(1);
+        connect(orientationButton, SIGNAL(released()), this,
```
alexanderbock, line 42: ->
`connect(orientationButton, &QPushButton::released, this, &Orientation::orientationDialog);`

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void Orientation::orientationDialog() {
+    _orientationDialog->exec();
+}
+
+sgct::quat Orientation::orientationValue() {
+    return _orientationValue;
+}
+
+bool Orientation::vsyncValue() {
```
alexanderbock, line 59: This function can be made `const`

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void Orientation::addButtonToLayout(QVBoxLayout* parentLayout) {
+    parentLayout->addLayout(_layoutOrientationFull);
+}
+
+void Orientation::orientationDialog() {
+    _orientationDialog->exec();
+}
+
+sgct::quat Orientation::orientationValue() {
```
alexanderbock, line 55: This function can be made `const`

apps/OpenSpace/ext/launcher/src/sgctedit/orientation.cpp
```diff
@@ -0,0 +1,67 @@
...
+    return _orientationValue;
+}
+
+bool Orientation::vsyncValue() {
+    return (_checkBoxVsync->checkState() == Qt::Checked);
+}
+
+Orientation::~Orientation()
```
alexanderbock, line 63: If moving the `_orientationDialog` to be not a pointer, the destructor is no longer necessary since the layout will be automatically destroyed by Qt when the parent layout is destroyed

apps/OpenSpace/ext/launcher/include/sgctedit/orientationdialog.h
```diff
@@ -0,0 +1,65 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_UI_LAUNCHER___ORIENTATIONDIALOG___H__
+#define __OPENSPACE_UI_LAUNCHER___ORIENTATIONDIALOG___H__
+
+#include <QCheckBox>
```
alexanderbock, line 28: Order:
```
#include <QDialog>

#include <sgct/config.h>
#include <QCheckBox>
#include <QFrame>
#include <QIntValidator>
#include <QLabel>
#include <QLayout>
#include <QLineEdit>
#include <QPainter>
#include <QPainterPath>
#include <QPoint>
#include <QPushButton>
#include <QVector>
#include <QWidget>
#include <glm/gtc/constants.hpp>
#include <iostream>
#include <vector>
```

Also the QCheckBox, QFrame, QIntValidator, QLabel, QLayout, QPainter, QPainterPath, QPoint, QPushButton, QVector, glm, iostream, and vector includes are not necessary here and the QWidget can be replaced with forward declarations

apps/OpenSpace/ext/launcher/include/sgctedit/orientationdialog.h
```diff
@@ -0,0 +1,65 @@
...
+#include <iostream>
+#include <vector>
+#include <sgct/config.h>
+
+
+class OrientationDialog : public QDialog
+{
+Q_OBJECT
+public:
+    explicit OrientationDialog(sgct::quat& orientation, QWidget* parent);
```
alexanderbock, line 52: No need for explicit since >1 parameters

apps/OpenSpace/ext/launcher/include/sgctedit/orientationdialog.h
```diff
@@ -0,0 +1,65 @@
...
+#include <QWidget>
+
+#include <glm/gtc/constants.hpp>
+#include <iostream>
+#include <vector>
+#include <sgct/config.h>
+
+
+class OrientationDialog : public QDialog
+{
```
alexanderbock, line 49: Opening `{` goes on the same line as the class definition

apps/OpenSpace/ext/launcher/src/sgctedit/orientationdialog.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "sgctedit/orientationdialog.h"
+
+OrientationDialog::OrientationDialog(sgct::quat& orientation, QWidget* parent)
+    : QDialog(parent)
+    , _orientationValue(orientation)
+{
+    setWindowTitle("Global Orientation");
+    QVBoxLayout* layoutWindow = new QVBoxLayout(this);
+
+    _linePitch = new QLineEdit(this);
```
alexanderbock, line 35: No need to create them with `this` since they will be reparented automatically whne added to a layout, so:
```cpp
_linePitch = new QLineEdit;
_lineRoll = new QLineEdit;
_lineYaw = new QLineEdit;
```

that allows a second simplification by defining them as regular member variables and getting rid of the `new` altogether

apps/OpenSpace/ext/launcher/src/sgctedit/orientationdialog.cpp
```diff
@@ -0,0 +1,101 @@
...
+        QDoubleValidator* validatorYaw = new QDoubleValidator(-180.0, 180.0, 15);
+        validatorPitch->setNotation(QDoubleValidator::StandardNotation);
+        validatorRoll->setNotation(QDoubleValidator::StandardNotation);
+        validatorYaw->setNotation(QDoubleValidator::StandardNotation);
+        _linePitch->setValidator(validatorPitch);
+        _lineRoll->setValidator(validatorRoll);
+        _lineYaw->setValidator(validatorYaw);
+    }
+    {
+        QLabel* labelPitch = new QLabel(this);
```
alexanderbock, line 53: same here, n o need for explicitly passing this `this` pointer since they get reparented

apps/OpenSpace/ext/launcher/src/sgctedit/orientationdialog.cpp
```diff
@@ -0,0 +1,101 @@
...
+        QHBoxLayout* layoutButtonBox = new QHBoxLayout;
+        QPushButton* buttonSave = new QPushButton("OK");
+        buttonSave->setToolTip("Save global orientation changes");
+        layoutButtonBox->addStretch(1);
+        layoutButtonBox->addWidget(buttonSave);
+        QPushButton* buttonCancel = new QPushButton("Cancel");
+        buttonCancel->setToolTip("Cancel global orientation changes");
+        layoutButtonBox->addWidget(buttonCancel);
+        layoutButtonBox->addStretch(1);
+        connect(buttonSave, SIGNAL(released()), this, SLOT(ok()));
```
alexanderbock, line 85:
```cpp
        connect(buttonSave, &QPushButton::released, this, &OrientationDialog::ok);
        connect(buttonCancel, &QPushButton::released, this, &OrientationDialog::cancel);
```

apps/OpenSpace/ext/launcher/include/sgctedit/sgctedit.h
```diff
@@ -0,0 +1,82 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_UI_LAUNCHER___SGCTEDIT___H__
+#define __OPENSPACE_UI_LAUNCHER___SGCTEDIT___H__
+
+#include <QApplication>
```
alexanderbock, line 28: Order:
```
#include <QDialog>

#include <sgctedit/display.h>
#include <sgctedit/filesupport.h>
#include <sgctedit/monitorbox.h>
#include <sgctedit/orientation.h>
#include <QApplication>
#include <QColor>
#include <QLabel>
#include <QLayout>
#include <QLineEdit>
#include <QMainWindow>
#include <QComboBox>
#include <QPushButton>
#include <QScreen>
#include <QTextBrowser>
#include <QWidget>
#include <iostream>
#include <string>
#include <vector>
```

and a lot of the includes can be removed/converted into forward declarations

apps/OpenSpace/ext/launcher/include/sgctedit/sgctedit.h
```diff
@@ -0,0 +1,82 @@
...
+        sgct::config::Cluster& cluster, const QList<QScreen*> screenList);
+    ~SgctEdit();
+    void addDisplayLayout(MonitorBox* monBox, QHBoxLayout* layout);
+    void createWidgets();
+    bool wasSaved() const;
+    std::string saveFilename();
+
+private:
+    void systemMonitorConfiguration(const QList<QScreen*> screenList);
+    MonitorBox* _monBox = nullptr;
```
alexanderbock, line 62: Readability:  Extra line between functions and member variables

apps/OpenSpace/ext/launcher/include/sgctedit/sgctedit.h
```diff
@@ -0,0 +1,82 @@
...
+#include <sgctedit/filesupport.h>
+#include <sgctedit/monitorbox.h>
+#include <sgctedit/orientation.h>
+
+class SgctEdit final : public QDialog
+{
+Q_OBJECT
+public:
+    SgctEdit(QWidget* parent, std::vector<sgct::config::Window>& windowList,
+        sgct::config::Cluster& cluster, const QList<QScreen*> screenList);
```
alexanderbock, line 53: `screenList` can be declared as `const QList<QScreen*>&` instead.

Maybe it doesn't need to be passed in at all and the `systemMonitorConfiguration` can just ask the application directly?

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -0,0 +1,139 @@
...
+{
+    systemMonitorConfiguration(screenList);
+    setWindowTitle("Display/Window Configuration Editor");
+    createWidgets();
+}
+
+void SgctEdit::systemMonitorConfiguration(const QList<QScreen*> screenList) {
+    size_t nScreensManaged = std::min(screenList.length(), 2);
+    for (size_t s = 0; s < nScreensManaged; ++s) {
+        int actualWidth = std::max(screenList[s]->size().width(),
```
alexanderbock, line 41: Alignment of arguments:  http://wiki.openspaceproject.com/docs/developers/coding-style.html#handling-line-breaks

```cpp
int actualWidth = std::max(
    screenList[s]->size().width(),
    screenList[s]->availableGeometry().width()
);
```

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -0,0 +1,139 @@
...
+            screenList[s]->availableGeometry().y(),
+            actualWidth,
+            actualHeight
+        });
+    }
+    _nMaxWindows = (_monitorSizeList.size() == 1) ? 3 : 4;
+}
+
+void SgctEdit::createWidgets() {
+    QVBoxLayout* layoutMainV = new QVBoxLayout(this);
```
alexanderbock, line 56: Same as above for `this` and empty constructor braces

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -0,0 +1,139 @@
...
+                else {
+                    reject();
+                }
+            }
+        );
+    }
+}
+
+void SgctEdit::addDisplayLayout(MonitorBox* monBox, QHBoxLayout* layout)
+{
```
alexanderbock, line 100: Opening `{` goes on the same line as the function definition

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -0,0 +1,139 @@
...
+                }
+                else {
+                    reject();
+                }
+            }
+        );
+    }
+}
+
+void SgctEdit::addDisplayLayout(MonitorBox* monBox, QHBoxLayout* layout)
```
alexanderbock, line 99: This function knows about the `_monBox`, so it can just use that member variable directly, rather than needing it to be passed as a function parameter

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -0,0 +1,139 @@
...
+bool SgctEdit::wasSaved() const {
+    return _saveSelected;
+}
+
+std::string SgctEdit::saveFilename() {
+    return _fileSupportWidget->saveFilename();
+}
+
+SgctEdit::~SgctEdit() {
+    if (_orientationWidget) {
```
alexanderbock, line 124: No need for these `if` statements.  Calling `delete` on the `nullptr` is legal and is a nullop

apps/OpenSpace/ext/launcher/include/sgctedit/sgctedit.h
```diff
@@ -0,0 +1,82 @@
...
+    QVBoxLayout* _displayLayout = nullptr;
+    QFrame* _displayFrame = nullptr;
+    Display* _displayWidget = nullptr;
+    QRect _monitorWidgetSize = {0, 0, 500, 500};
+    FileSupport* _fileSupportWidget = nullptr;
+    Orientation* _orientationWidget = nullptr;
+    sgct::config::Cluster& _cluster;
+    std::vector<sgct::config::Window>& _windowList;
+    bool _saveSelected = false;
+    unsigned int _nMaxWindows = 3;
```
alexanderbock, line 73: Let's just fix it to 4 and move it into the cpp file as a compile-time setting.  No need to have it as a member variable then:
`constexpr const int MaxWindows = 4;`

apps/OpenSpace/ext/launcher/include/sgctedit/sgctedit.h
```diff
@@ -0,0 +1,82 @@
...
+    QString _colorsForWindows[4] = {
```
alexanderbock, line 74: Same here, just move it to the cpp file and define as:
```cpp
constexpr const std::array<QColor, 4> ColorsForWindows = {
   QColor(43, 158, 195),
   QColor(252, 171, 16),
   QColor(68, 175, 105),
   QColor(248, 51, 60)
};
```
or if you really wanna keep the hexvalues:
```cpp
constexpr const std::array<QColor, 4> ColorsForWindows = {
   QColor(0x2B, 0x9E, 0xC3),
   QColor(0xFC, 0xAB, 0x10),
   QColor(0x44, 0xAF, 0x69),
   QColor(0xF8, 0x33, 0x3C)
};
```
but I think the decimal value are easier to read

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_UI_LAUNCHER___WINDOWCONTROL___H__
+#define __OPENSPACE_UI_LAUNCHER___WINDOWCONTROL___H__
+
+#include <QCheckBox>
```
alexanderbock, line 28: Order:
```
#include <QWidget>

#include <sgct/config.h>
#include <QCheckBox>
#include <QColor>
#include <QComboBox>
#include <QFrame>
#include <QIntValidator>
#include <QLabel>
#include <QLayout>
#include <QLineEdit>
#include <QPainter>
#include <QPainterPath>
#include <QPoint>
#include <QPushButton>
#include <QStandardItemModel>
#include <QVector>
#include <vector>
#include <iostream>
```

Not all of the includes are necessary and some can be replaced with forward declarations

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+#include <QVector>
+#include <QWidget>
+
+#include <vector>
+#include <iostream>
+#include <sgct/config.h>
+
+
+class WindowControl : public QWidget
+{
```
alexanderbock, line 50: Opening `{` goes on the same line as the class definition

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+Q_OBJECT
+public:
+    explicit WindowControl(unsigned int nMonitors, unsigned int monitorIndex,
+        unsigned int windowIndex, std::vector<QRect>& monitorDims,
+        const QString* winColors, QWidget *parent = nullptr);
```
alexanderbock, line 55: -> `QWidget* parent`

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+    int qualitySelectedIndex();
+    int qualitySelectedValue();
+    unsigned int monitorNum();
+    float fovH();
+    float fovV();
+    float heightOffset();
+    enum ProjectionIndeces : unsigned int {
+        Planar = 0,
+        Fisheye,
+        Spherical_Mirror,
```
alexanderbock, line 88: No underscores:  -> `SphericalMirror`

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+    void enableGuiWindowSelection(bool enabled);
+    void uncheckWebGuiOption();
+    int projectionSelectedIndex();
+    int qualitySelectedIndex();
+    int qualitySelectedValue();
+    unsigned int monitorNum();
+    float fovH();
+    float fovV();
+    float heightOffset();
+    enum ProjectionIndeces : unsigned int {
```
alexanderbock, line 85: -> `enum class ProjectionIndices {`

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+        Planar = 0,
+        Fisheye,
+        Spherical_Mirror,
+        Cylindrical,
+        Equirectangular
+    };
+    std::string ProjectionTypeNames[5] = {"Planar", "Fisheye", "Spherical Mirror",
```
alexanderbock, line 92: These values are only used in the initialization of the `_projectionTypes`.  Moving the `_projectionTypes` definition into the cpp file means that we don't need to explicitly list the types here, but it can all happen in the cpp instead.  Same for the quality

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+    float _marginFractionOfWidgetSize = 0.025;
+    unsigned int _nMonitors = 1;
+    unsigned int _monIndex = 0;
+    unsigned int _index = 0;
+    std::vector<QRect>& _monitorResolutions;
+    int _maxWindowSizePixels = 10000;
+    const QString* _colorsForWindows = nullptr;
+    QVBoxLayout* _layoutFullWindow = nullptr;
+    QLabel* _labelWinNum = nullptr;
+    QLineEdit* _size_x = nullptr;
```
alexanderbock, line 154: no underscores

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+                          const unsigned int windowIndex, std::vector<QRect>& monitorDims,
+                                                const QString* winColors, QWidget *parent)
+    : QWidget(parent)
+    , _nMonitors(nMonitors)
+    , _monIndex(monitorIndex)
+    , _index(windowIndex)
+    , _monitorResolutions(monitorDims)
+    , _colorsForWindows(winColors)
+{
+    createWidgets(parent);
```
alexanderbock, line 39: Just move the contents of the `createWidgets` function into the constructor.  The complexity of a function call here doesn't do anything

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+}
+
+void WindowControl::createWidgets(QWidget* parent) {
+    _windowDims = defaultWindowSizes[_index];
+    _size_x = new QLineEdit(
```
alexanderbock, line 44: I don't even understand what happens here,  is this just:
`_size_x = new QLineEdit(QString::number(_windowDims.width(), parent);`
?

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+        QString::fromUtf8(std::to_string(int(_windowDims.width())).c_str()), parent);
+    _size_y = new QLineEdit(
+        QString::fromUtf8(std::to_string(int(_windowDims.height())).c_str()), parent);
+    _offset_x = new QLineEdit(
+        QString::fromUtf8(std::to_string(int(_windowDims.x())).c_str()), parent);
+    _offset_y = new QLineEdit(
+        QString::fromUtf8(std::to_string(int(_windowDims.y())).c_str()), parent);
+    {
+        QIntValidator* validatorSize_x = new QIntValidator(10, _maxWindowSizePixels);
```
alexanderbock, line 53: No underscores

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+        QDoubleValidator* validatorFovH = new QDoubleValidator(-180.0, 180.0, 10);
+        _lineFovH->setValidator(validatorFovH);
+        QDoubleValidator* validatorFovV = new QDoubleValidator(-90.0, 90.0, 10);
+        _lineFovV->setValidator(validatorFovV);
+        _lineHeightOffset = new QLineEdit("0.0", parent);
+        QDoubleValidator* validatorHtOff= new QDoubleValidator(-1000000.0, 1000000.0, 12);
+        _lineHeightOffset->setValidator(validatorHtOff);
+    }
+
+    connect(
```
alexanderbock, line 97: See argument above for new `connect` methods.  For overloaded functions consider:
```
    connect(
        _clustersWidget, QOverload<Cluster::ID>::of(&ClustersWidget::killProcesses),
        this, QOverload<Cluster::ID>::of(&MainWindow::killAllProcesses)
    );
```

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+}
+
+QVBoxLayout* WindowControl::initializeLayout() {
+    _layoutFullWindow = new QVBoxLayout();
+    //Window size
+    QVBoxLayout* layoutWindowCtrl = new QVBoxLayout();
+
+    _labelWinNum = new QLabel();
+    _labelWinNum->setText("Window " + QString::number(_index + 1));
+    QString colorStr = "QLabel { color : ";
```
alexanderbock, line 153: ->
```cpp
std::string colorStr = fmt::format("QLabel {{ color: {}; }}", _colorsForWindows ? _colorsForWindows[_index] : "#FFFFFF");
```

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    colorStr += _colorsForWindows ? _colorsForWindows[_index] : "#FFFFFF";
+    colorStr += "; }";
+    _labelWinNum->setStyleSheet(colorStr);
+
+    QHBoxLayout* layoutWinNum = new QHBoxLayout();
```
alexanderbock, line 158: (also for rest of this file)  see above for empty constructor braces and `this` for layouts/widgets wrt reparenting

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    _layoutFullWindow->addLayout(layoutWindowCtrl);
+
+    _comboProjection->setCurrentIndex(0);
+    onProjectionChanged(0);
+    _comboQuality->setCurrentIndex(2);
+
+    return _layoutFullWindow;
+}
+
+void WindowControl::showWindowLabel(const bool show) {
```
alexanderbock, line 288: Remove `const` of parameter

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    _labelWinNum->setVisible(show);
+}
+
+void WindowControl::onSizeXChanged(const QString& newText) {
+    std::string x = newText.toStdString();
+    if (!x.empty()) {
+        _windowDims.setWidth(std::stoi(x));
```
alexanderbock, line 295: -> `_windowDims.setWidth(newText.toInt());`

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    }
+    if (_windowChangeCallback) {
+        _windowChangeCallback(_monIndex, _index, _windowDims);
+    }
+}
+
+void WindowControl::onSizeYChanged(const QString& newText) {
+    std::string y = newText.toStdString();
+    if (!y.empty()) {
+        _windowDims.setHeight(std::stoi(y));
```
alexanderbock, line 305: Same here

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    try {
+        if (!xOffset.empty()) {
+            _windowDims.setX(std::stoi(xOffset));
+            _windowDims.setWidth(prevWidth);
+        }
+        if (_windowChangeCallback) {
+            _windowChangeCallback(_monIndex, _index, _windowDims);
+        }
+    }
+    catch (...) {
```
alexanderbock, line 324: Only catch the exception thrown by the validator here;   the `catch (...)` catches every possible exception and is too wide a net here

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+        //enters an integer after the - then the value will be updated).
+    }
+}
+
+void WindowControl::onOffsetYChanged(const QString& newText) {
+    std::string yOffset = newText.toStdString();
+    float prevHeight = _windowDims.height();
+    try {
+        if (!yOffset.empty()) {
+            _windowDims.setY(std::stoi(yOffset));
```
alexanderbock, line 337: See above

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+            _windowDims.setHeight(prevHeight);
+        }
+        if (_windowChangeCallback) {
+            _windowChangeCallback(_monIndex, _index, _windowDims);
+        }
+    }
+    catch (...) {
```
alexanderbock, line 344: See above

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    _checkBoxSpoutOutput->setVisible(newSelection == ProjectionIndeces::Fisheye
+        || newSelection == ProjectionIndeces::Equirectangular);
+}
+
+void WindowControl::setDimensions(const QRectF& dimensions) {
+    _windowDims = dimensions;
+}
+
+void WindowControl::setWindowChangeCallback(
+                        std::function<void(int, int, const QRectF&)> cb)
```
alexanderbock, line 412: Alignment

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+{
+    _windowChangeCallback = cb;
```
alexanderbock, line 414: -> `_windowChangeCallback = std::move(cb);`

same below

apps/OpenSpace/ext/launcher/include/sgctedit/windowcontrol.h
```diff
@@ -0,0 +1,176 @@
...
+        unsigned int windowIndex, std::vector<QRect>& monitorDims,
+        const QString* winColors, QWidget *parent = nullptr);
+    ~WindowControl();
+    void setDimensions(const QRectF& dimensions);
+    void setWindowChangeCallback(std::function<void(int, int, const QRectF&)> cb);
+    void setWebGuiChangeCallback(std::function<void(unsigned int)> cb);
+    void showWindowLabel(const bool show);
+    QVBoxLayout* initializeLayout();
+    QRectF& dimensions();
+    QLineEdit* lineEditSizeWidth();
```
alexanderbock, line 63: None of these functions returning the internal state are used anywhere and can be removed

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+    return _checkBoxSpoutOutput;
+}
+
+std::string WindowControl::windowName() {
+    return _windowName->text().toStdString();
+}
+
+sgct::ivec2 WindowControl::windowSize() {
+    return {
+        stoi(_size_x->text().toStdString()),
```
alexanderbock, line 465: ->  `_size_x->text().toInt()`


same below

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+}
+
+sgct::ivec2 WindowControl::windowPos() {
+    return {
+        stoi(_offset_x->text().toStdString()),
+        stoi(_offset_y->text().toStdString())
+    };
+}
+
+bool WindowControl::isDecorated() {
```
alexanderbock, line 477: if all of these functions are still necessary, they can all be marked as `const`

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+float WindowControl::heightOffset() {
+    return _lineHeightOffset->text().toFloat();
+}
+
+unsigned int WindowControl::monitorNum() {
+    return _monIndex;
+}
+
+WindowControl::~WindowControl()
```
alexanderbock, line 517: Move the destructor up to just below the constructor and `{` needs to be on the asme line as the function name

alexanderbock (changes requested):
Regarding the `const` parameters for functions.  For basic types, the constness of it doesn't matter since it is an implementation detail whether a parameter is constant inside the function.  The caller of the function has to make a copy into it *anyway* therefore for the caller it doesn't make a different.
Consider:
https://gcc.godbolt.org/z/GPKG9n6fe

`foo` is declared with a `const int` but called with an int that is not-const and everything still compiles.  So declaring non-pointer, non-reference parameters as `const` is just syntactical noise and should be avoided

alexanderbock:
Also consider changing the `released` signal from the QPushButton to `clicked`  I don't remember what the details were, but there were some subtle differences between the two

alexanderbock:
#1 is primary monitor, using negative offsets for #2 

image

apps/OpenSpace/ext/launcher/include/sgctedit/display.h
```diff
@@ -0,0 +1,90 @@
...
+#include "windowcontrol.h"
+#include "monitorbox.h"
+
+
+class Display : public QWidget
+{
+Q_OBJECT
+public:
+    explicit Display(MonitorBox* monitorRenderBox, std::vector<QRect>& monitorSizeList,
+        const unsigned int nMaxWindows, const QString* winColors);
```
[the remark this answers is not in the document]
GPayne: The color is passed-in this way because it resides in SgctEdit. Both Display and WindowControl need to get the same color. The const here prevents future editors of this file from changing the color. But if someone wanted to change it at SgctEdit, then it would be changed in both Display and WindowControl, keeping them consistent.

apps/OpenSpace/ext/launcher/include/sgctedit/display.h
```diff
@@ -0,0 +1,90 @@
...
+    unsigned int _nWindowsAllocated = 0;
+    unsigned int _nWindowsDisplayed = 0;
+    unsigned int _nMaxWindows = 3;
+    unsigned int _nMonitors = 1;
+    const QString* _winColors;
+    std::vector<WindowControl*> _windowControl;
+    QPushButton* _addWindowButton = nullptr;
+    QPushButton* _removeWindowButton = nullptr;
+    unsigned int _monitorIdx = 0;
+    QVBoxLayout* _layout = nullptr;
```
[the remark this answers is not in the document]
GPayne: I'm sure you noticed that in almost all cases, I am letting Qt delete objects. This particular case was left over from when I was getting segfaults when the launcher was closed (was sensitive to the order of deletion). I made this change it looks OK now.
I don't have the same trust in Qt to properly delete everything, however. In my earlier versions I was explicitly deleting objects, but I later switched to letting Qt do this. I used valgrind to check for memory leaks in both versions, and it is reporting memory leaks when I leave it up to Qt to delete stuff. It's also reporting other leaks elsewhere in OpenSpace, so this is something I was meaning to discuss with you.

apps/OpenSpace/ext/launcher/src/sgctedit/display.cpp
```diff
@@ -0,0 +1,166 @@
...
+        initializeWindowControl();
+    }
+    connect(_addWindowButton, SIGNAL(released()), this, SLOT(addWindow()));
+    connect(_removeWindowButton, SIGNAL(released()), this, SLOT(removeWindow()));
+    initializeLayout();
+}
+
+Display::~Display() {
+    delete _monBox;
+    for (auto w : _windowControl) {
```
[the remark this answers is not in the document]
GPayne: This is a good use for `auto`, very limited scope. The coding style says to use it sparingly but does not forbid it.

apps/OpenSpace/ext/launcher/src/sgctedit/display.cpp
```diff
@@ -0,0 +1,166 @@
...
+void Display::showWindows() {
+    for (unsigned int i = 0; i < _layoutWindowWrappers.size(); ++i) {
+        _layoutWindowWrappers[i]->setVisible(i < _nWindowsDisplayed);
+    }
+    for (unsigned int i = 0; i < _frameBorderLines.size(); ++i) {
+        _frameBorderLines[i]->setVisible(i < (_nWindowsDisplayed - 1));
+    }
+    _removeWindowButton->setEnabled(_nWindowsDisplayed > 1);
+    _addWindowButton->setEnabled(_nWindowsDisplayed != _nMaxWindows);
+    for (auto w : _windowControl) {
```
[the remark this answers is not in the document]
GPayne: Claiming my same defense of using `auto` in a limited scope.

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
GPayne: I made this change. Unlike my previous defenses for the use of `auto`, the scope here is not as limited.

apps/OpenSpace/ext/launcher/src/sgctedit/filesupport.cpp
```diff
@@ -0,0 +1,278 @@
...
+        _cluster.users.push_back(user);
+    }
+}
+
+bool FileSupport::isWindowFullscreen(unsigned int monitorIdx, sgct::ivec2 wDims) {
+    sgct::ivec2 mDims = {_monitors[monitorIdx].width(), _monitors[monitorIdx].height()};
+    return ((mDims.x == wDims.x) && (mDims.y == wDims.y));
+}
+
+bool FileSupport::findGuiWindow(unsigned int& foundWindowIndex) {
```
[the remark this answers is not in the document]
GPayne: This would be a nice way to do it, but not easier. A bool is easy and the caller is a bool with a name that explains what the bool condition means.

apps/OpenSpace/ext/launcher/include/sgctedit/monitorbox.h
GPayne: I think you mean `constexpr float`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
GPayne: I'll meet you halfway. Some of these GUI setup calls have a lot of magic constants; we have some in our launcher. A few lines down I have a call to `setFont("Arial", 14)`, where 14 is obvious enough. I'll put the radius in.
A `constexpr` variable is already `const`

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
GPayne: I'll make this change, but I'm not getting a compiler warning from this

apps/OpenSpace/ext/launcher/include/sgctedit/display.h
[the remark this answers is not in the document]
alexanderbock: In that case it would be easier to understand if it comes in as a `std::array<QColor, 4>` instead that is stored in a `const std::array<QColor, 4>`.  That signals the same (the Display class mustn't change the color) and it fixes the issue of 1) having the color represented as a `QString` and 2) having a pointer to some other person holding the memory, making the understanding of the class non-local

apps/OpenSpace/ext/launcher/include/sgctedit/display.h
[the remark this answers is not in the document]
alexanderbock: :+1:   Can't say that I 100% understand how (or when) Qt deletes things either.   Yea, let's talk about it :)

apps/OpenSpace/ext/launcher/src/sgctedit/display.cpp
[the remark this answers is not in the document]
alexanderbock: I think we might want to change the coding style then 😅      My problem with `auto` in these cases is that it is impossible to see what the type of `w` is without also needing to look at the header file

apps/OpenSpace/ext/launcher/src/sgctedit/filesupport.cpp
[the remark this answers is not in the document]
alexanderbock: Just a general dislike of `out` parameters in functions, since they don't really serve any purpose anymore.   The std-library way of having it would be returning a `std::pair<unsigned int, bool>` like (std::map::emplace) but I think std::optional<unsigned int> would signal the intent of the function that it is possible to fail much clearer than either that or using a sentiel value

apps/OpenSpace/ext/launcher/include/sgctedit/monitorbox.h
alexanderbock: Eah.. jap 😅

apps/OpenSpace/ext/launcher/src/sgctedit/monitorbox.cpp
alexanderbock: Visual studio is throwing warnings left and right about these things.  Didn't mention the others since I was planning to go in and fix those warnings myself, as it would be much easier since I'm compiling directly on it

apps/OpenSpace/ext/launcher/include/sgctedit/display.h
```diff
@@ -0,0 +1,90 @@
...
+    void addWindow();
+    void removeWindow();
+
+private:
+    void initializeWindowControl();
+    void removeWindowControl();
+    void initializeLayout();
+    void showWindows();
+    void addDisplayLayout(unsigned int column, MonitorBox* monBox, QVBoxLayout* layout);
+    MonitorBox* _monBox;
```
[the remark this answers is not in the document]
GPayne: Agreed with a smart pointer being a better choice. I switched to a `shared_ptr` since this needs to be used in other classes.
It's easier to work with raw pointers when using Qt however. A lot of examples online just use raw pointers and let Qt clean up the memory. It's not just memory management; I had to use the smart pointer's `get` when passing it to Qt, which then makes it an "in the wild" pointer again, but Qt should delete it so it should be OK. I don't know of another way to pass it to Qt.

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
GPayne: Doing this will give you a blank window.

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
GPayne: Yes, not sure why it had turned into a mess

apps/OpenSpace/ext/launcher/include/sgctedit/display.h
```diff
@@ -0,0 +1,80 @@
...
+#include <QLineEdit>
+#include <QPainter>
+#include <QPainterPath>
+#include <QPoint>
+#include <QPushButton>
+#include <QTextBrowser>
+#include <QVector>
+#include <vector>
+
+class Display : public QWidget {
```
WeirdRubberDuck, line 45: wondering if this class might deserve a more descriptive name?

apps/OpenSpace/ext/launcher/include/sgctedit/sgctedit.h
```diff
@@ -0,0 +1,79 @@
...
+#include <QLayout>
+#include <QScreen>
+#include <memory>
+#include <string>
+#include <vector>
+
+class QWidget;
+
+class SgctEdit final : public QDialog
+{
```
WeirdRubberDuck, line 45: nitpicky, but this `{` should be on the previous line

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -248,30 +274,52 @@ QWidget* LauncherWindow::createCentralWidget() {
...
+        newWindowButton, &QPushButton::released,
+        [this]() {
+            openWindowEditor();
+        }
+    );
+    newWindowButton->setObjectName("small");
+    newWindowButton->setGeometry(geometry::NewWindowButton);
+    newWindowButton->setCursor(Qt::PointingHandCursor);
+
+    //QPushButton* editWindowButton = new QPushButton("Edit", centralWidget);
```
WeirdRubberDuck, line 313: can this commented out code be removed?

alexanderbock:
Looks great ❤️

WeirdRubberDuck:
I really like how this GUI turned out! 🤩 

Got two comments, based on the image below: 
1. Would be nice if the "preview window" wasn't completely squared, but rather matched the window configuration somehow, or maybe was resizeable. My intuiting when starting the GUI was to try to re-scale the window to remove some of that "empty space". Would have been nice (see image below, my desire was to not have some of the space marked in blue)

2. **Bug**: The setup below actually don't work as I expect it to. I set both of the windows to fullscreen, and then switched one to the secondary monitor. But both of them show up on the primary moinitor. Running on Windows, and using the "apply without saving" option

image

WeirdRubberDuck:
**Another bug/annoyance**:  The "remove window" button is active when editing other inputs that are not button. I instinctively pressed enter after inputting the offset value below (note in the image that the remove button is highlighted. this happens when editing any input that's not a button), which triggered the remove window action since the button was active. Very confusing and potentially really annoying 

image

PR #1868, opened 2022-02-01
Removing the asset_helper file

alexanderbock:


alexanderbock:
# Limitations
 - Not dealing with the fact whether we need all `asset.export` statements
 - Keeping the `registerIdentifierWithMeta` calls for now and punting a replacement to https://github.com/OpenSpace/OpenSpace/issues/1869

alexanderbock:
- Remove the `asset_helper` file and call the `onInitialize` and `onDeinitialize` functions directly
 - Small compile fix on Windows
 - Removes the need for the `registerIdentifierWithMeta` function by automatically doing that for all `asset.export` statements
 - Allow the passing of either identifiers (as before) or entire tables to the `removeSceneGraphNode`, `removeScreenSpaceRenderable`, `deleteLayer`,  `removeAction`, and `export` methods. In that case, the `Identifier` key from the table is extracted and used instead

Issue #1329, opened 2020-10-14
Setting the linear flight target distance to 0 causes NaNs to appear in the camera position


alexanderbock:

alexanderbock:
Reopened for now

alexanderbock:
Undid the fix from  8cf5f96 in 778b4e61b38e0bdd0477e0cd7799bb3ceed5b4d3 as it caused issues with the anchor/aim feature

WeirdRubberDuck:
Closed in #1887

Issue #1837, opened 2022-01-14
Move the linear flight functionality into the camera paths 


WeirdRubberDuck:
We want to remove the linear flight feature completely, and just use the camera paths for similar cases (or at least the PathNavigator). 

Something to think about here is **how and if** to keep the "benefits" of the linear flight compared to the camera paths around. For example, a linear flight can currently be triggered while another camera motion of the OrbitalNavigator, such as an orbit idle behavior, is happening. It is also not as sensitive to simulation time being played.

WeirdRubberDuck:
Note that this task also includes adapting the touch Gui to this change, as it heavily relies on the linear flight feature

PR #1887, opened 2022-02-17
Address some stability problems for camera paths + remove linear flight

WeirdRubberDuck:
* Change spline curve parameter interval from [0,1] to [0,nSegments], for slightly increased precision
* Use linear paths whenever precision isn't enough/paths are too long (and traverse linear paths differently, to make them less depending on the distance)
* Remove OrbitalNavigator.LinearFlight and instead add "Zoom to" helper functions (addresses #1837 )
* Add checks to prevent triggering awkward short paths, for example when flying to the current target or to the same geo position twice
* Refactor speed code and handle speed when path is shorter than start+closeup distance
* Add topic for engine mode (to be used for UI)
* Cleaner log when creating paths (some previous info messages are now debug messages)

Related PR: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/70


include/openspace/navigation/path.h
```diff
@@ -92,6 +92,11 @@ class Path {
      */
     CameraPose interpolatedPose(double distance) const;
 
+    /*
```
alexanderbock, line 95: Nitpicky;  should be `/**` to be picked up by Doxygen

include/openspace/navigation/path.h
```diff
@@ -106,6 +111,12 @@ class Path {
      */
     glm::dquat easedSlerpRotation(double t) const;
 
+    /**
+     * Compute the interpolated rotation quaternion using a method that is
+     * customized to look good for linear paths
```
alexanderbock, line 116: Maybe possible to give some more information about the nature of the customization?   Should have a parameter explanation for the `t` parameter, too

include/openspace/navigation/pathcurve.h
```diff
@@ -34,21 +35,35 @@ class Waypoint;
...
+        explicit InsufficientPrecisionError(std::string msg);
+    };
+
+    struct TooShortPathError : public ghoul::RuntimeError {
+        explicit TooShortPathError(std::string msg);
+    };
+
     virtual ~PathCurve() = 0;
 
     double length() const;
```
alexanderbock, line 48: Mentioning here the unit of the length?  I'm assuming it'd be "meters" but just in case

src/navigation/path.cpp
```diff
@@ -139,12 +148,49 @@ std::vector<glm::dvec3> Path::controlPoints() const {
 CameraPose Path::traversePath(double dt, float speedScale) {
     double speed = speedAlongPath(_traveledDistance);
     speed *= static_cast<double>(speedScale);
-    const double displacement =  dt * speed;
+    double displacement =  dt * speed;
```
alexanderbock, line 151: Nitpicky;  extra space before the `dt`

src/navigation/path.cpp
```diff
@@ -139,12 +148,49 @@ std::vector<glm::dvec3> Path::controlPoints() const {
...
+        }
+
+        // Just move along the line from the current position to the target
+        newPose.position = _prevPose.position -
+            displacement * glm::normalize(prevPosToEnd);
+
+        const double relativeDistance = _traveledDistance / pathLength();
+        newPose.rotation = interpolateRotation(relativeDistance);
+    }
+    else {
```
alexanderbock, line 181: maybe better to have an `else if` here and then another `else` that throws a `ghoul::MissingCaseException` to make it easier in the future to find places where we have to add additional checks

include/openspace/navigation/path.h
```diff
@@ -131,13 +142,21 @@ class Path {
     // Playback variables
     double _traveledDistance = 0.0;
     double _progressedTime = 0.0; // Time since playback started
+    bool _forceQuit = false;
```
alexanderbock, line 145: Maybe better to rename `_shouldForceQuit` or `_isForceQuitting` to better express the usage of this?

src/navigation/path.cpp
```diff
@@ -139,12 +148,49 @@ std::vector<glm::dvec3> Path::controlPoints() const {
...
+        // Special handling of linear paths, so that it can be used when we are
+        // traversing very large distances without introducing precision problems
+        const glm::dvec3 prevPosToEnd = _prevPose.position - _end.position();
+        const double remainingDistance = glm::length(prevPosToEnd);
+
+        // Actual displacement may not be bigger than remaining distance
+        if (displacement > remainingDistance) {
+            displacement = remainingDistance;
+            _traveledDistance = pathLength();
+            _forceQuit = true;
```
alexanderbock, line 170: Does this mean that if _forceQuit is set to true, there would be one more tick of progressing along the path?  I'd intuitively assume that it would stop immediately, meaning that we would be `displacement-remainingDistance` away from the the target location when the path finishes?

src/navigation/path.cpp
```diff
@@ -183,6 +241,45 @@ glm::dquat Path::easedSlerpRotation(double t) const {
...
+        const glm::dquat q = ghoul::lookAtQuaternion(halfWayPosition, endNodePos, endUp);
+
+        const double tScaled = ghoul::sineEaseInOut(t / tHalf);
+        return glm::slerp(_start.rotation(), q, tScaled);
+    }
+
+    // This distance is guaranteed to be strictly decreasing for linear paths
+    const double distanceToEnd = glm::distance(_prevPose.position, _end.position());
+
+    double closingUpDistance = 10.0 * _end.validBoundingSphere();
```
alexanderbock, line 262: Maybe just a comment here that specifies that 10, 2. and 0.2 are just randomly picked constants

src/navigation/path.cpp
```diff
@@ -261,14 +360,20 @@ double Path::speedAlongPath(double traveledDistance) const {
...
+    double closeUpDistance = closeUpDistanceFactor * _end.node()->boundingSphere();
+    if (closeUpDistance < LengthEpsilon) { // zero bounding sphere
         closeUpDistance = glm::distance(_end.position(), endNodePos);
     }
 
     if (traveledDistance > (pathLength() - closeUpDistance)) {
-        const double remainingDistance = pathLength() - traveledDistance;
+        double remainingDistance = 0.0;
+        if (_type == Type::Linear) {
+            remainingDistance = glm::distance(_prevPose.position,_end.position());
```
alexanderbock, line 371: Nitpicky, missing space before `_end`

src/navigation/path.cpp
```diff
@@ -261,14 +360,20 @@ double Path::speedAlongPath(double traveledDistance) const {
...
+        }
+        else {
+            remainingDistance = pathLength() - traveledDistance;
+        }
+
         speed *= remainingDistance / closeUpDistance + 0.01;
```
alexanderbock, line 377: Why the 0.01 here?

src/navigation/pathcurve.cpp
```diff
@@ -57,21 +69,20 @@ std::vector<glm::dvec3> PathCurve::points() const {
...
 
+    const double max = static_cast<double>(_nSegments);
+
     // Evenly space out parameter intervals
     _curveParameterSteps.reserve(_nSegments + 1);
-    const double dt = 1.0 / _nSegments;
-    _curveParameterSteps.push_back(0.0);
-    for (unsigned int i = 1; i < _nSegments; i++) {
-        _curveParameterSteps.push_back(dt * i);
+    for (double i = 0.0; i < max; i += 1.0) {
```
alexanderbock, line 82: Why the iteration in double?
```
for (int i = 0; i < _nSegments; i++) {
    _curveParameterSteps.push_back(static_cast<double>(i));
}
```
would read less weird and no need for the `max` variable

src/navigation/pathcurve.cpp
```diff
@@ -84,49 +95,89 @@ void PathCurve::initializeParameterData() {
...
+    }
+
     // Compute a map of arc lengths s and curve parameters u, for reparameterization
-    _parameterSamples.reserve(NrSamplesPerSegment * _nSegments + 1);
-    const double uStep = 1.0 / (_nSegments * NrSamplesPerSegment);
+    constexpr const int steps = 100;
+    const double uStep = 1.0 / static_cast<double>(steps);
+    _parameterSamples.reserve(steps * _nSegments + 1);
+
+    bool problematic = false;
```
alexanderbock, line 107: -> `isProblematic`?

src/navigation/pathcurve.cpp
```diff
@@ -84,49 +95,89 @@ void PathCurve::initializeParameterData() {
...
     _totalLength = _lengthSums.back();
 
+    if (_totalLength < LengthEpsilon) {
+        throw TooShortPathError("Path too short!");
+    }
+
     // Compute a map of arc lengths s and curve parameters u, for reparameterization
-    _parameterSamples.reserve(NrSamplesPerSegment * _nSegments + 1);
-    const double uStep = 1.0 / (_nSegments * NrSamplesPerSegment);
+    constexpr const int steps = 100;
```
alexanderbock, line 103: I'd say `Steps` since it is a literal constant

src/navigation/pathcurve.cpp
```diff
@@ -84,49 +95,89 @@ void PathCurve::initializeParameterData() {
...
+            if (std::abs(s - _parameterSamples.back().s) < LengthEpsilon) {
+                problematic = true;
+                continue;
+            }
             _parameterSamples.push_back({ u, s });
         }
     }
-    _parameterSamples.push_back({ 1.0, _totalLength });
+
+    if (problematic) {
```
alexanderbock, line 126: Why the extra boolean variable?  Shouldn't you be able to just throw the exception inside the for loop directly?

src/navigation/pathcurve.cpp
```diff
@@ -84,49 +95,89 @@ void PathCurve::initializeParameterData() {
...
 }
 
 // Compute the curve parameter from an arc length value, using a combination of
 // Newton's method and bisection. Source:
 // https://www.geometrictools.com/Documentation/MovingAlongCurveSpecifiedSpeed.pdf
 // Input s is a length value, in the range [0, _totalLength]
-// Returns curve parameter in range [0, 1]
+// Returns curve parameter in range [0, _nSegments]
 double PathCurve::curveParameter(double s) const {
     if (s <= 0.0) return 0.0;
```
alexanderbock, line 146: Return statements for these two lines should be on the next line with `{` `}`  but I'm open for discussing changing this at one of the developer meetings

src/navigation/pathnavigator_lua.inl
```diff
@@ -208,6 +207,113 @@ int flyToNavigationState(lua_State* L) {
...
+
+    return 0;
+}
+
+int zoomToDistance(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, { 1, 2 }, "lua::zoomToDistance");
+    auto [distance, duration] =
+        ghoul::lua::values<double, std::optional<double>>(L);
+
+    if (distance <= 0.f) {
```
alexanderbock, line 246: -> 0.0 since it is a `double`

src/navigation/pathnavigator_lua.inl
```diff
@@ -208,6 +207,113 @@ int flyToNavigationState(lua_State* L) {
...
+
+    return 0;
+}
+
+int zoomToDistanceRelative(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, { 1, 2 }, "lua::zoomToDistanceRelative");
+    auto [distance, duration] =
+        ghoul::lua::values<double, std::optional<double>>(L);
+
+    if (distance <= 0.f) {
```
alexanderbock, line 283: Same here

src/navigation/pathnavigator.cpp
```diff
@@ -462,6 +465,33 @@ scripting::LuaLibrary PathNavigator::luaLibrary() {
                 "that roll must be included for the target up direction to be taken "
                 "into account."
             },
+            {
+                "zoomToFocus",
+                &luascriptfunctions::zoomToFocus,
+                "[duration]",
+                "Zoom linearly to the current focus node, using the default distance."
+                "The optional input parameter specifies the duration for the motion."
```
alexanderbock, line 473: Which unit is the duration in?  Assuming seconds, of course

alexanderbock (approved):
Looks good to me!
Just a single comment:
 - I'd make the "Successfully generated camera path" into a Debug message rather than Info to keep the chatter down while starting a path

In general I wasn't able to break it at all, so that's good :sweating:

src/navigation/pathnavigator.cpp
WeirdRubberDuck: Yep, seconds

src/navigation/pathcurve.cpp
WeirdRubberDuck: True. With the current implementation, it can be done directly in the loop. I'll update so that it does

It was nice to keep the entire list of samples around for debugging purposes, but there really is no reason to keep it now that linear paths are used for this scenario

src/navigation/path.cpp
WeirdRubberDuck: No, this is the final tick along the path and should result in the target position being returned (see the next line).
The check for whether a path has finished is done as the last step in the updateCamera function of `PathNavigator`. 

I get where you are coming from though. "Force" sounds very serious and immediate. Maybe just remaining the variable to `_shouldQuit` would be sufficient?

Note: The reason the variable is needed is to handle the cases where precision might be a problem. When the distance es are large, the check `(_traveledDistance / pathLength()) >= 1.0` is not sufficient to identify if a path has finished

src/navigation/path.cpp
WeirdRubberDuck: So, the condition for that `else if` would be..? A check for any of the other curve types?

If so, then a switch statement would probably be more suitable. However, ideally, I wouldn't want to have any other special cases. Right now, the linear paths are a workaround to help with precision problems but later on, I hope that these will be solved (for example by switching the direction of the interpolation, half-way, as we've talked a bit about). So, no other additional checks should be needed here IMO 🤔

src/navigation/path.cpp
WeirdRubberDuck: The reason is that we want to prevent multiplying the speed with exactly 0.0, as this would result in a zero speed and no movement/displacement. 

Refactored the speed part of the code a bit, so hopefully, now it's clearer

src/navigation/path.cpp
alexanderbock: :+1:  I saw that after I wrote the comment;  I thought it was a binary enumA vs enumB   not enumA vs all other enums

src/navigation/path.cpp
alexanderbock: :+1:  `_shouldQuit` would make that more apparent imho.  But also not a big deal

alexanderbock:
I guess this would also close #1329?

PR #1898, opened 2022-03-01
Add helper function and timer to trigger idle behavior

WeirdRubberDuck:
Adds a helper function `openspace.navigation.triggerIdleBehavior` to immediately start an idle behavior of choice, and also a timer that can be set to trigger the default idle behavior after a certain time (see new properties in OrbitalNavigator)


src/navigation/navigationhandler_lua.inl
```diff
@@ -343,4 +343,20 @@ int addGlobalRoll(lua_State* L) {
...
 }
 
+int triggerIdleBehavior(lua_State* L) {
+    ghoul::lua::checkArgumentsAndThrow(L, {0, 1}, "lua::triggerIdleBehavior");
+    std::optional<std::string> choice = ghoul::lua::value<std::optional<std::string>>(L);
+
+    try {
+        global::navigationHandler->orbitalNavigator().triggerIdleBehavior(choice);
+    }
+    catch (ghoul::MissingCaseException&) {
```
alexanderbock, line 353: The idea for the `MissingCaseException` is to be a fallback for the cases where the developer forgot to add a case to a switch statement, not as an exception that might occur in regular usage.  I'd suggest to replace it with a ghoul::RuntimeError instead

Issue #507, opened 2018-02-14
Multi-window setup incompatible with different framebuffer sizes


alexanderbock:
When running two windows, one GUI and one spherical mirror, the edges between the fisheye cube faces show up if the size of the second window is not teh same as the cube resolution

https://slack-files.com/T667GT03H-F98BP3CBD-bbdc572ac2

Roxeena:
Related to #458?

micahnyc:
bump from user "I found this very frustrating". 

I think this might be different then #458 as that is particular to the star rendering and this is for the window/rendering as a whole. Likely they both need code to fix.

alexanderbock:
I'd prefer to move this to 0.17.0 as the fix of this will require breaking changes

benpm:
I'm working on a proposal to change the OpenSpace render engine to use Stencil testing to enable postprocessing on the combined viewport, potentially fixing issues related to local pixel info being lost on the boundaries between viewports. @alexanderbock suggested it might provide a good avenue for fixing this issue, so I'm leaving a comment here as a reminder for now.

alexanderbock:
bump from the user meeting

PR #1904, opened 2022-03-09
Fixes for resizing multiple windows. 

micahnyc:
This PR allows you to resize the gui and rendering windows independently. 

NOTE: In regards to the fonts, I don't think this is the ideal solution.

I think a more ideal solution would be to have separate font renderers for the 3D scene, and for the Dashboard. Going even further, we could allow multiple dashboards, and associate them with a specific viewport.


micahnyc:
Fixes #507

PR #1906, opened 2022-03-13
Feature/codegen lua

alexanderbock:


alexanderbock:
Left to do after this PR and considered out of scope for this:
1. Make the HTML generation on the documentation index.html better and use the types, default values etc that are now passed to the Handlebars script
2. Cleanup the actual text of the documentation. Still some arguments are explained in the text rather than relying on good variable names

WeirdRubberDuck:
Seems to work great! 🙌 

Some strings that I found confusing: 
image
image


1. 
    - Thinking that maybe `String[]` would be more readable and reduce confusion compared to our previous way of saying "optional"? We didn't end up going full typescript anyways, right?
    - The "0"? Comes from the argument not having a name? But then, why does the argument not have a name in the function in the second image? confused 🤔 Either way, maybe we should come up with a more readable default, such as `arg0, arg1, arg2...` or something

2. Does the `*` mean any type? That part is fine, but these functions should also have a "dummy name" for the input variable, or even better, we should give it a meaningful name :)


Some thoughts on "optional":
image

3. How about moving the optional question mark to the argument name, rather than the type? Think that could make it clearer that the input variable is optional. so
```
downloadFile( url (String) , savePath (String) , waitForCompletion? (Boolean) )
```
?

(the corresponding typescript version would be: `waitForCompletion?:  boolean`, right? 🤔 )

alexanderbock:
1. 
- I added argument names for the `0` parameter names.
- Changed [String] -> String[]  and same for all types
2. Yes, `*` means any type and any number of arguments.  That part is not in codegen since we can't handle an `any` type so those functions are still handwritten
3. I like the question mark on the variable name, but that would have to happen on the HTML side since we are just passing the function name as is and I'd prefer not to mix name and type on the C++ side

PR #1916, opened 2022-03-21
update cef and a few calls for it

micahnyc:
There is an issue with the text fields in our gui after this update which needs to be addressed in the -WebGuiFrontend

When using the search fields, sometimes keyevents are sent through to openspace. When using a non search field (such as naming session recording, key events are never sent to openspace. 

In addition, if you check a box or use a slider, then use a search field, then the keyevents are not sent to openspace. 

I'm guessing there is a special behavior on how we have addressed the search fields in react which needs to be updated.

### DIFF
--- a/modules/webbrowser/CMakeLists.txt
+++ b/modules/webbrowser/CMakeLists.txt
@@ -168,8 +168,6 @@ if (OS_MACOSX)
   ADD_LOGICAL_TARGET("cef_sandbox_lib" "${CEF_SANDBOX_LIB_DEBUG}" "${CEF_SANDBOX_LIB_RELEASE}")
   foreach(target IN LISTS Targets)
     # Helper executable target.
-        message(STATUS "set tar : ${target}")
-
     add_executable(${target} MACOSX_BUNDLE ${WEBBROWSER_HELPER_SOURCES})
     SET_EXECUTABLE_TARGET_PROPERTIES(${target})
     # add_cef_logical_target("libcef_lib" "${CEF_LIB_DEBUG}" "${CEF_LIB_RELEASE}")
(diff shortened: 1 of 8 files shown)


alexanderbock:
Confirmed happening for me as well.  For me however, every keystroke is sent forwards to OpenSpace in the beginning and then only intermittendly later
 - Search bar in Scene and Settings
 - Search Bar in Focus menu
 - "Star name..." in Exoplanet Systems
 - Float properties when entering numbers directly
 - Seconds / second and Negative Seconds/second in Time Selection window

*not* in:
 - "Name of recording" in Record session
 - Slide name and URL in Screensapace renderables
 - Year/Month/Day/... in Time Selection
 - String Properties

alexanderbock:
Best feature about this, btw is the fact that the UI now works in Debug mode on Windows, too!

micahnyc:
tested and fixed now on macos

alexanderbock:
Not sure if it is related to the CEF but it happens on this branch:
 - The "Play" button in the Record Session panel doesn't work anymore

WeirdRubberDuck:
> Confirmed happening for me as well. For me however, every keystroke is sent forwards to OpenSpace in the beginning and then only intermittendly later
> 
> * Search bar in Scene and Settings
> * Search Bar in Focus menu
> * "Star name..." in Exoplanet Systems
> * Float properties when entering numbers directly
> * Seconds / second and Negative Seconds/second in Time Selection window
> 
> _not_ in:
> 
> * "Name of recording" in Record session
> * Slide name and URL in Screensapace renderables
> * Year/Month/Day/... in Time Selection
> * String Properties

So, this seems to be a problem for our `<Input>` components, but not `<InlineInput>` (such as the date edit the "Select Date" part). Will investigate further

WeirdRubberDuck:
> Not sure if it is related to the CEF but it happens on this branch:
> 
> * The "Play" button in the Record Session panel doesn't work anymore

@alexanderbock seems like this broke as the part of the Lua update. The "sessionrecording.startPlayback" function has been renamed to `startPlaybackDefault`?

Issue #1910, opened 2022-03-21
Camera path speed problem when inside bounding sphere of target


WeirdRubberDuck:
The speed function for the camera paths does not handle the case when the camera is within the bounding sphere of the target. Can result in a very small speed, which basically gets the computation stuck in an infinite loop...

Example to reproduce: 
- RenderableBillboardscloud have a computed bounding sphere, so try to fly to the "Tully Galaxies" node, starting from Earth

WeirdRubberDuck:
Similar problem with the linear path

The problem is that the `startUpDistance` and `closeUpDistance` gets way too large for big bounding spheres

PR #1957, opened 2022-03-23
Add bounding spheres to a bunch of renderable types

WeirdRubberDuck:
Leaving this as a PR until #1910 is fixed. 

Please comment here if you find more renderables that might need bounding spheres :) Then we can include those in this pr as well


Roxeena:
Don't we need to set the interaction sphere as well if we want events to work?

WeirdRubberDuck:
> Don't we need to set the interaction sphere as well if we want events to work?
The interaction sphere is set to the default value (zero) for all these renderables. We can't set it to the same as the bounding sphere, as we want to be able to move the camera "into" the objects

alexanderbock (approved):

PR #1961, opened 2022-03-24
added warning for gaia module on mac; not adding renderable to scene.…

### LINKED ISSUE #843: Gaia module doesn't load files properly on mac
There is an issue with the trailling / and the sync folders. Even though the .asset files says :

File = starsFolder .. "/" 

when looking at the error in the logs, it is trying to load the file SYNCDIR/1dataset/index.bin rather then SYNCDIR/1/dataset/index.bin

micahnyc:
… fixes #843

### DIFF
--- a/data/assets/scene/milkyway/gaia/gaiastars.asset
+++ b/data/assets/scene/milkyway/gaia/gaiastars.asset
@@ -1,3 +1,9 @@
+local fullOS = openspace.systemCapabilities.fullOperatingSystem()
+if string.find(fullOS, "Darwin") then
+  openspace.printWarning("Gaia module (RenderableGaiaStars) not supported on mac")
+  return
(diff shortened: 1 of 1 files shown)


data/assets/scene/milkyway/gaia/gaiastars.asset
```diff
@@ -1,3 +1,9 @@
+local fullOS = openspace.systemCapabilities.fullOperatingSystem()
+if string.find(fullOS, "Darwin") then
+  openspace.printWarning("Gaia module (RenderableGaiaStars) not supported on mac")
```
WeirdRubberDuck, line 3: Nitpicky, but add two spaces here :)

alexanderbock (approved):

Issue #1947, opened 2022-03-23
Rapid rotations during short camera paths


WeirdRubberDuck:
Since the rotation interpolation is done over the traveled distance of the camera path, the rotation can be extremely fast for short camera paths (almost infinitely fast)

For these scenarios (and actually most scenarios, if possible), we instead want a solution that interpolates based on time instead of position, and that can keep interpolating even if the camera has reached the target position. That is, the camera path should not be finished until both position and rotation interpolation are finished. 

Note that this means that we need to compute how long the path takes to travel before starting the motion of the camera

WeirdRubberDuck:
Decided to solve it for just **linear**  paths for the release, and leave the rest as future work

Issue #1919, opened 2022-03-22
Skybrowser: "Look at target" does not work


WeirdRubberDuck:
Guessing this is being worked on, but thought I'd add an issue for it for completeness, as this is a feature I really think should be working 

Clicking the "Look at target" button the UI currently does nothing

ylvaselling:
Removed lock button since with the new renderable target the lock functionality is lost

WeirdRubberDuck:
The "Look at target" och the "Lock" button is not the same though?

ylvaselling:
True - i guess unrelated this specific issue but more generally related to "buttons that currently doesnt work"

PR #1966, opened 2022-03-29
Camera path updates (#1919 and #1947)

WeirdRubberDuck:
1. Avoids problems with speed when inside big bounding spheres, 
2. updates the speed dampening when bounding spheres are huge and
3. Use time-based interpolation of rotation for linear paths, and don't finish path until both rotation and position interpolation has finished. 

2. leads to abrupt stops when traveling vast distances, to objects with big bounding spheres. There should probably be some improvement made, eventually

3. Decided to keep the old code commented out for now. It will have to be improved to work ish for navigation states eventually. Note that it also broke the debug rendering of the camera view direction, as that assumes that interpolation is done based on distance. No biggie, though, and can be fixed later.

### DIFF
--- a/src/navigation/path.cpp
+++ b/src/navigation/path.cpp
@@ -40,16 +40,13 @@
 #include <openspace/util/universalhelpers.h>
 #include <ghoul/logging/logmanager.h>
 #include <ghoul/misc/interpolator.h>
+#include <glm/ext/quaternion_relational.hpp>
 
 namespace {
     constexpr const char _loggerCat[] = "Path";
     constexpr const float LengthEpsilon = 1e-5f;
 
     constexpr const char SunIdentifier[] = "Sun";
-
-    bool isApproximateSame(const glm::dquat& q1, const glm::dquat& q2, double precision) {
-        return 1.0 - std::abs(glm::dot(q1, q2)) < precision;
-    }
 
     // TODO: where should this documentation be?
     // It's nice to have these to interpret the dictionary when creating the path, but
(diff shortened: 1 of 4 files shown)


src/navigation/path.cpp
```diff
@@ -47,6 +47,10 @@ namespace {
 
     constexpr const char SunIdentifier[] = "Sun";
 
+    bool isApproximateSame(const glm::dquat& q1, const glm::dquat& q2, double precision) {
```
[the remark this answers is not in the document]
alexanderbock: Or:
`#include <glm/ext/scalar_relational.hpp>`
and replace `isApproximateSame` with `glm::equal`

src/navigation/path.cpp
```diff
@@ -204,7 +191,13 @@ bool Path::hasReachedEnd() const {
         return true;
     }
 
-    return (_traveledDistance / pathLength()) >= 1.0;
+    bool isPositionFinished = (_traveledDistance / pathLength()) >= 1.0;
+    bool isRotationFinished = isApproximateSame(
```
alexanderbock, line 195: Just for me sanity:
```cpp
constexpr double Epsilon = 0.00001;
bool isRotationFinished = isApproximateSame(
        _prevPose.rotation,
        _end.rotation(),
        Epsilon
```

src/navigation/path.cpp
```diff
@@ -243,45 +260,59 @@ glm::dquat Path::easedSlerpRotation(double t) const {
...
-    }
-
-    // Keep looking at the end node
-    return ghoul::lookAtQuaternion(_prevPose.position, endNodePos, endUp);
+    const glm::dvec3 a = ghoul::viewDirection(_start.rotation());
+    const glm::dvec3 b = ghoul::viewDirection(_end.rotation());
+    const double angle = std::acos(glm::dot(a, b)); // assumes length 1.0 for a & b
+
+    // Seconds per pi angles. Per default, it takes 5 seconds to turn 90 degrees
+    double factor = 5.0 / (0.5 * glm::pi<double>()); // TODO: make property to adapt this
```
alexanderbock, line 268: `0.5 * glm::pi<double>()` -> `glm::half_pi<double>()`

alexanderbock (approved):

src/navigation/path.cpp
WeirdRubberDuck: And remove outdated comment

PR #1968, opened 2022-03-30
Feature/horizons framework

### LINKED ISSUE #1862: Horizons and SPICE data does not line up when it should
When using Horizons data of JWST from the 2021 launch, the trail relative to the Sun and the trail relative to L2 doesn't line up when they should. 

jwst_precision

Blue trail is of JWST relative to the Sun (SSB) and yellow trail is of JWST relative to L2. Have already checked if it could be a problem with Sun vs SSB, that is not the problem.  It is probably not the problem with loss of precision in the Horizons data either since that is on the millimeter magnitude and the gap is much bigger than that. Alex had an idea that it might be that somewhere in the pipeline is a `float` when it should be a `double` that might be causing the problem.

Here is the data used for the trails in the image
[horizons_jwst_L2.txt](https://github.com/OpenSpace/OpenSpace/files/7957489/horizons_jwst_L2.txt)
[horizons_jwst_sun.txt](https://github.com/OpenSpace/OpenSpace/files/7957490/horizons_jwst_sun.txt)

Roxeena:
A new dialog in the launcher to download a horizons file from the Horizons database. Please try it out and play around with it. If you find a case where it doesn't work please let me know and save the resulting file.

To get to the Horizons dialog: `Edit` button next to `Assets` in the Profile Editor -> `New Asset` button on the top right -> `Generate Horizons File` button.

Additional things:
1. Update all of our existing horizons data
2. Enable use of several horizons files for one `HorizonsTranslation`
3. Use Vector table data to fix #1862

If you are on a network that restricts some http traffic you will need to install OpenSSL for the http request to work. This can be done in the QT installer/MaintenanceTool as it is an optional tool. It can be found under Qt -> Developer and Designer Tools -> OpenSSL.

Edit: The horizons data is now uploaded to the server


modules/space/horizonsfile.h
```diff
@@ -0,0 +1,145 @@
...
+    };
+
+    struct HorizonsKeyframe {
+        double time;            // J2000 seconds
+        glm::dvec3 position;    // GALACTIC cartesian coordinates in meters
+    };
+
+    struct HorizonsResult {
+
+        HorizonsResult()
```
alexanderbock, line 102: If you do instead:
```cpp
struct HorizonsResult {
    Type type = Type::Invalid;
    ResultCode errorCode = ResultCode::UnknownError;
    std::vector<HorizonsKeyframe> data = std::vector<HorizonsKeyframe>();
}
```

then you don't need to define the constructor

modules/space/horizonsfile.h
```diff
@@ -0,0 +1,145 @@
...
+
+    HorizonsFile();
+    HorizonsFile(std::filesystem::path file);
+    HorizonsFile(std::filesystem::path filePath, const std::string& result);
+
+    void setFile(std::filesystem::path file);
+    const std::filesystem::path& file() const;
+    std::filesystem::path& file();
+
+    static std::string constructUrl(Type type, const std::string& target,
```
alexanderbock, line 119: This function doesn't have to be a member function of the HorizonsFile at all and can just be a stand-alone function instead

modules/space/horizonsfile.h
```diff
@@ -0,0 +1,145 @@
...
+        const std::string& observer, const std::string& startTime,
+        const std::string& stopTime, const std::string& stepSize,
+        const std::string& unit);
+    static ResultCode isValidAnswer(const json& answer);
```
alexanderbock, line 123: Same here

modules/space/translation/horizonstranslation.cpp
```diff
@@ -83,7 +79,36 @@ HorizonsTranslation::HorizonsTranslation(const ghoul::Dictionary& dictionary)
...
+
+    if (std::holds_alternative<std::string>(p.horizonsTextFile)) {
+        std::string file = std::get<std::string>(p.horizonsTextFile);
+        if (!std::filesystem::is_regular_file(absPath(file))) {
+            LWARNING(fmt::format("The Horizons text file '{}' could not be found", file));
+            return;
+        }
+
+        std::vector<std::string> files;
+        files.reserve(1);
```
alexanderbock, line 91: No need to reserve one,  doesn't optimize anything;   either the vector already has space for at least one object (-> the reserve is a no-op) or it doesn't have space (-> the `push_back` will allocate that space).
Using reserve is useful if  you know ahead of time that you want to add a *lot* of items that could cause the vector to be reseated because the preallocated memory is not enough

modules/space/translation/horizonstranslation.cpp
```diff
@@ -83,7 +79,36 @@ HorizonsTranslation::HorizonsTranslation(const ghoul::Dictionary& dictionary)
...
+
+        for (std::string file : files) {
+            if (!std::filesystem::is_regular_file(absPath(file))) {
+                LWARNING(fmt::format("The Horizons text file '{}' could not be found",
+                    file)
+                );
+                return;
+            }
+        }
+        _horizonsTextFiles.set(files);
```
alexanderbock, line 107: -> `_horizonsTextFiles = files`

modules/space/translation/horizonstranslation.cpp
```diff
@@ -116,90 +141,68 @@ glm::dvec3 HorizonsTranslation::position(const UpdateData& data) const {
...
-    }
-
-    std::filesystem::path cachedFile = FileSys.cacheManager()->cachedFilename(file);
-    bool hasCachedFile = std::filesystem::is_regular_file(cachedFile);
-    if (hasCachedFile) {
-        LINFO(fmt::format("Cached file {} used for Horizon file {}", cachedFile, file));
-
-        bool success = loadCachedFile(cachedFile);
-        if (success) {
+    for (const std::string filePath : _horizonsTextFiles.value()) {
```
alexanderbock, line 144: -> `const std::string&`

modules/space/translation/horizonstranslation.cpp
```diff
@@ -248,35 +252,41 @@ void HorizonsTranslation::saveCachedFile(const std::filesystem::path& file) cons
...
     }
+
+    // Write of entire vector will only work if the data is plain old data type,
+    // is_pod is depricated in C++20 and replaced with both is_trivial and
+    // is_standard_layout
+    assert(std::is_trivial<CacheKeyframe>::value);
+    assert(std::is_standard_layout<CacheKeyframe>::value);
+
+    // Write data
+    fileStream.write(reinterpret_cast<const char*>(
```
alexanderbock, line 287: The `reinterpret_cast` should be on the next line

alexanderbock (changes requested):

apps/OpenSpace/ext/launcher/include/profile/assetedit.h
```diff
@@ -0,0 +1,75 @@
...
+class AssetEdit : public QDialog {
+Q_OBJECT
+public:
+    /**
+     * Constructor for AssetEdit class
+     */
+    AssetEdit(QWidget* parent);
+
+private slots:
+    //void openComponent();
```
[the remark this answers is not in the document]
Roxeena: You mentioned in one of our meetings that you wanted to keep this as inspiration for how the asset builder could work in the future. So now I am confused..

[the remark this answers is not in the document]
alexanderbock: Ah.. yea, you're right... Nevermind 😅

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -0,0 +1,1178 @@
...
+    bool isValid = true;
+    std::string message;
+
+    // File
+    if (_fileEdit->text().isEmpty()) {
+        isValid = false;
+        message = "File path not selected";
+        styleLabel(_fileLabel, true);
+    }
+
```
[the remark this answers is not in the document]
Roxeena: I liked the idea to visually separate the validation of each field and make it look less cluttered, but I can change if if we prefer the other way of course.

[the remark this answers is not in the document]
alexanderbock: How about the restructuring the function a bit more to achieve that?
```cpp
bool HorizonsDialog::isValidInput() {
    if (_fileEdit->text().isEmpty()) {
        _errorMsg->setText("File path not selected");
        styleLabel(_fileLabel, true);
        return false;
    }

    // Target field
    if (_targetEdit->text().isEmpty() &&
            (_chooseTargetCombo->count() > 0 &&
             _chooseTargetCombo->currentIndex() == 0))
    {
        _errorMsg->setText("Target not selected");
        styleLabel(_fileLabel, true);
        return false;
    }

    if (_targetEdit->text().isEmpty() && _chooseTargetCombo->count() == 0) {
        _errorMsg->setText("Target not selected");
        styleLabel(_fileLabel, true);
        return false;
    }

   // etc

   return true;
```


This way it is not one `if` statement with a lot of `else if` attachments, but it is one `if` statement after the other so the empty lines don't matter since it is not one conceptual thing anymore

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+    constexpr std::string_view EqualsSign = "%3D";
+    constexpr std::string_view QuestionMark = "%3F";
+    constexpr std::string_view AtSymbol = "%40";
+    constexpr std::string_view LeftSquareBracket = "%5B";
+    constexpr std::string_view RightSquareBracket = "%5D";
+
+    std::string replaceAll(std::string string, const std::string& from,
+                           const std::string& to)
+    {
+        if (from.empty())
```
WeirdRubberDuck, line 74: This if-statement deserves some curly brackets :) It looks lonely

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+            return "";
+
+        size_t startPos = 0;
+        while ((startPos = string.find(from, startPos)) != std::string::npos) {
```
WeirdRubberDuck, line 78: The assignment in the while loop looks kind of nasty to me. Makes it a bit difficult to read what's going on (not obvious what is being compared to `std::string::npos`)

Should this be a do-while maybe? Or initialize the variable before the while loop and then later again?

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+            string.replace(startPos, from.length(), to);
+
+            // In case 'to' contains 'from', ex replacing 'x' with 'yx'
+            startPos += to.length();
+        }
+        return string;
+    }
+
+    std::string urlEncode(const std::string& string) {
+        std::string result =
```
WeirdRubberDuck, line 88: I would probably do: 
```
std::string result = string; 
result = replaceAll(string, " ", static_cast<std::string>(WhiteSpace));
result = replaceAll(result, "#", static_cast<std::string>(HashTag));
...
```
To make it a bit more aligned and easier to read. But probably just a personal opinion and not really a problem

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+        LWARNING("Could not find signature information");
+    }
+
+    // Errors
+    if (auto it = answer.find("error"); it != answer.end()) {
+        // There was an error
+        std::string errorMsg = *it;
+
+        // Projected output length (~X) exceeds 90024 line max -- change step-size
+        if (errorMsg.find("Projected output length") != std::string::npos) {
```
WeirdRubberDuck, line 235: Do we maybe need a small helper function for doing this check? is done a LOT :)

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+    // query that can tell us if the file is valid or not.
+    // The line $$SOE indicates start of data.
+    std::string line;
+    bool foundTarget = false;
+    std::getline(fileStream, line);
+    std::getline(fileStream, line); // First line is just stars (*) no information, skip
+
+    // Valid Target?
+    if (fileStream.good() &&
+        (line.find("Revised") != std::string::npos ||
```
WeirdRubberDuck, line 295: Misaligned if statement (and maybe also a place where we could use a helper function to clean up the check if string contains a substring?)

modules/space/horizonsfile.cpp
WeirdRubberDuck: Can we use C++23 features yet? I guess not? Will be nice when we have the `std::contains` function https://en.cppreference.com/w/cpp/string/basic_string/contains

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+        if (line.find("No site matches") != std::string::npos ||
+            line.find("Cannot find central body") != std::string::npos)
+        {
+            fileStream.close();
+            return ResultCode::ErrorNoObserver;
+        }
+
+        // Are observer and target the same?
+        if (line.find("disallowed") != std::string::npos)
+        {
```
WeirdRubberDuck, line 335: bracket should be on previous line

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+            fileStream.close();
+            return ResultCode::ErrorObserverTargetSame;
+        }
+
+        // Enough data?
+        if (line.find("Insufficient ephemeris data") != std::string::npos)
+        {
```
WeirdRubberDuck, line 342: same here

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+        if (line.find("No matches found") != std::string::npos) {
+            fileStream.close();
+            return ResultCode::ErrorNoTarget;
+        }
+
+        std::getline(fileStream, line);
+    }
+
+    if (result != ResultCode::UnknownError) {
+        fileStream.close();
```
WeirdRubberDuck, line 381: since the fileStream is closed in all the checks below, maybe it can just be done before this if-statement ?

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+    switch (code) {
+        case HorizonsFile::ResultCode::Valid:
+            return;
+        case ResultCode::Empty:
+            LERROR("The horizons file is empty");
+            break;
+        case ResultCode::InvalidFormat:
+            LERROR(fmt::format(
+                "Format of file '{}' is invalid. Horizons files must have extension '.hrz'",
+                _file)
```
WeirdRubberDuck, line 411: move closing `)` to next line

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+            );
+            break;
+        case ResultCode::ErrorSize:
+            LERROR("The selected time range with the selected step size is too big, "
```
WeirdRubberDuck, line 415: should be
```
LERROR(
    "The selected time range with the selected step size is too big, "
    "try to increase the step size and/or decrease the time range"
);
```
I believe

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+                "try to increase the step size and/or decrease the time range");
+            break;
+        case ResultCode::ErrorSpan:
+            LERROR("Step size is too big, exceeds available time span for target");
+            break;
+        case ResultCode::ErrorTimeRange: {
+            LERROR("Time range is outside the valid range for target");
+
+            std::pair<std::string, std::string> validTimeRange =
```
WeirdRubberDuck, line 424: Should be formatted as: 
```
std::pair<std::string, std::string> validTimeRange = parseValidTimeRange(
    "Trajectory files",
    "************",
    "Trajectory name"
);
```

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+            break;
+        }
+        case ResultCode::ErrorNoObserver:
+            LERROR("No match was found for the observer");
+            break;
+        case ResultCode::ErrorObserverTargetSame:
+            LERROR("The observer and target are the same");
+            break;
+        case ResultCode::ErrorNoData:
+            LERROR("There is not enough data to compute the state of the target in "
```
WeirdRubberDuck, line 446: see above

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+                "relation to the observer for the selected time range.");
+            break;
+        case ResultCode::MultipleObserverStations: {
+            LWARNING("Multiple matching observer stations were found for the "
```
WeirdRubberDuck, line 450: see above

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+        dataPoint.time = timeInJ2000;
+        dataPoint.position = gPos;
+        data.push_back(dataPoint);
+
+        std::getline(fileStream, line);
+    }
+
+    fileStream.close();
+
+    LWARNING("Observer table data from Horizons might not align with SPICE data well. "
```
WeirdRubberDuck, line 706: move string to an own new line

modules/space/horizonsfile.cpp
```diff
@@ -0,0 +1,877 @@
...
+
+        // Read and save each word.
+        std::stringstream str(line);
+        std::vector<std::string> words;
+        std::string word;
+        while (str >> word) {
+            words.push_back(word);
+        }
+
+        if (words.empty()) {
```
WeirdRubberDuck, line 853: This check does not seem needed to me. will be caught by the `if {} else if {} else` below?

WeirdRubberDuck: Same on line 814 above? Also ,the parsing up there seems very similar to how it's done here. Can it be generalized?

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -0,0 +1,1173 @@
...
+        _typeCombo = new QComboBox(this);
+        _typeCombo->setToolTip("Choose Horizons data type");
+        QStringList types = {
+           "Vector table",
+           "Observer table"
+        };
+        _typeCombo->addItems(types);
+        _typeCombo->setCurrentIndex(0);
+        connect(_typeCombo, QOverload<int>::of(&QComboBox::currentIndexChanged),
```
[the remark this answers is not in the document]
Roxeena: According to Alex the guidelines are more relaxed when it comes to the Qt connect function. He commented the exact opposite since I used to have it like you are suggesting.

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -0,0 +1,1173 @@
...
+
+    // Target field
+    if (_targetEdit->text().isEmpty() && ((_chooseTargetCombo->count() > 0 &&
+        _chooseTargetCombo->currentIndex() == 0) || _chooseTargetCombo->count() == 0))
+    {
+        _errorMsg->setText("Target not selected");
+        styleLabel(_targetLabel, true);
+        return false;
+    }
+    if (_targetEdit->text().toStdString().find_first_of("¤<>§£´¨€") != std::string::npos)
```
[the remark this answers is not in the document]
Roxeena: It is only checked two times, and I think a function wouldn't simplify this or prevent repetitive code too much

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -0,0 +1,1173 @@
...
+                "Unknown error",
+                LogLevel::Error
+            );
+            if (!_latestHorizonsError.empty()) {
+                appendLog(fmt::format("Latest Horizons error: {}", _latestHorizonsError),
+                    LogLevel::Error
+                );
+            }
+            _errorMsg->setText("An unknown error occured");
+            return false;
```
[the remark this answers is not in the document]
Roxeena: Yes, if the file is supposed to remain then it is return instead of break. That is the cases where we don't know what happened or could not find all information and give to the user directly. It is mostly a safe net for error cases that have not been discovered yet or don't know how to handle. Then the user can still go to the file and hopefully get an error message from there. Otherwise we would remove the file and the user would have no clue what went wrong.

modules/space/translation/horizonstranslation.cpp
```diff
@@ -35,22 +35,24 @@
...
 } // namespace
 
 namespace {
     constexpr openspace::properties::Property::PropertyInfo HorizonsTextFileInfo = {
         "HorizonsTextFile",
         "Horizons Text File",
-        "This value is the path to the text file generated by Horizons with observer "
-        "range and Galactiv longitude and latitude for different timestamps."
+        "This value is the path to the file or files generated by Horizons with "
+        "either a Vector table or an Observer table with the correct settings."
```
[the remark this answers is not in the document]
Roxeena: I wrote about them in the [header file](https://github.com/OpenSpace/OpenSpace/blob/af0a6d61cb4c828277f916dd9a33b6188d980724/modules/space/translation/horizonstranslation.h#L47) and will probably edit our [wiki page](http://wiki.openspaceproject.com/docs/builders/ephemeris/horizons) to use the new version as well. Since the text was so long I didn't want to repeat it here but I can link to the wiki when it has been updated.

modules/space/horizonsfile.cpp
Roxeena: If it is closed here, then the check for `good()` a few lines down is invalid. We need to check if we found the start of data before we reached the end of the file i.e. the file needs to still be `good()` for it to be valid in the end of this function.

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: OK, nvm then :)

modules/space/translation/horizonstranslation.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: 👍 Adding "(see wiki)", and not the full link, would probably be enough. Links might break in the future and would be nice to not have to maintain it

modules/space/horizonsfile.cpp
Roxeena: Marked as CPP23 future improvements

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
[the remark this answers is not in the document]
Roxeena: I tried to make this more clear in the code now so hopefully it wont be as confusing now.

modules/space/horizonsfile.cpp
Roxeena: It cant really be generalized since there are som small differences between the two cases. But I added a longer explanation before the function to make it clearer what the difference is.

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -0,0 +1,1173 @@
...
+                    QString::fromStdString(station),
+                    findId(station)
+                );
+            }
+            _chooseObserverCombo->setCurrentIndex(0);
+            _chooseObserverCombo->show();
+            break;
+        }
+        case openspace::HorizonsFile::ResultCode::MultipleObserver: {
+            appendLog(
```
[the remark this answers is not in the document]
Roxeena: I agree that it is a rather long function but I think having 14 small new functions would be more cluttered. Nofix

PR #1974, opened 2022-04-05
Fix for ASCII recording to correctly handle multi-line keybinding script

GPayne:


src/interaction/sessionrecording.cpp
```diff
@@ -927,6 +927,17 @@ void SessionRecording::saveScriptKeyframeAscii(Timestamps& times,
 {
     std::stringstream keyframeLine = std::stringstream();
     saveHeaderAscii(times, HeaderScriptAscii, keyframeLine);
+    //Erase all \r (from windows newline), and all \n from line endings and replace with
```
alexanderbock, line 930: Should have spaces after the `//`

src/interaction/sessionrecording.cpp
```diff
@@ -927,6 +927,17 @@ void SessionRecording::saveScriptKeyframeAscii(Timestamps& times,
...
+    //';' so that lua will treat them as separate lines. This is done in order to treat
+    //a multi-line script as a single line in the file.
+    size_t startPos = 0;
```
alexanderbock, line 933: How about instead:
```cpp
size_t startPos = sm._script.find("\r", startPos));
while (startPos != std::string::npos) {
    sm._script.erase(startPos, 1);
    startPos = sm._script.find("\r", startPos));
}
```

and same for below?

Removes the need for an assignment in the `while` loop

GPayne: Changed

Issue #1995, opened 2022-04-11
Volumetric rendering not showing up


micahnyc:
mercury magnetosphere and mas density do not render

alexanderbock:
Data seems to be there but not rendered correctly (see gray mist on the right side)

![OpenSpace_000000](https://user-images.githubusercontent.com/4921724/162802294-b520a4a9-4f45-4fe3-bfe3-b457e721dca9.png)

alexanderbock:
![OpenSpace_000001](https://user-images.githubusercontent.com/4921724/162802492-a04fefc5-033d-48a5-a0f1-fd057e2b5e19.png)

micahnyc:
Now we know the origin story of the gray goo...

WeirdRubberDuck:
Just FYI, if it helps with figuring out the problem: The generated cartesian and spherical example volumes do not have the same problem. They show up fine

I.e. these assets: 
https://github.com/OpenSpace/OpenSpace/tree/master/data/assets/examples/volume/generated

PR #2027, opened 2022-04-14
Fix opacity issue with volume rendering, issue 1995

ElonOlsson:
opacity changed to brightness in raycaster

Closes issue #1995


modules/volume/rendering/renderabletimevaryingvolume.cpp
```diff
@@ -452,7 +451,7 @@ void RenderableTimeVaryingVolume::update(const UpdateData&) {
             _raycaster->setVolumeTexture(nullptr);
         }
         _raycaster->setStepSize(_stepSize);
-        _raycaster->setOpacity(opacity());
+        _raycaster->setBrightness(_brightness);
```
alexanderbock, line 454: I'd suggest to
```cpp
_raycaster->setBrightness(_brightness * opacity());
```
that wait we can use the `Fade` parameter that every Renderable has as well to fade out the volume rendering

modules/volume/shaders/raycast.glsl
```diff
@@ -31,8 +31,9 @@ uniform int nClips_#{id};
 uniform vec3 clipNormals_#{id}[8];
 uniform vec2 clipOffsets_#{id}[8];
 
-uniform float opacity_#{id} = 10.0;
-
+uniform float brightness_#{id} = 1.0;
+// unitless factor that multiplies with the brightness [0,1] to achieve desired visuals.
+const float sampling_interval_reference_factor = 500.0;
```
alexanderbock, line 36: -> `SamplingIntervalReferenceFactor`

alexanderbock (approved):

PR #2039, opened 2022-04-21
Issue/1964

Roxeena:
* Fix broken asset file
* Prevent crash outside data range
* Apply color map outside data range

NOTE: If the interpolation is turned off, the color map is not applied at all. This is because it lives in the `InterpolateTileProvider`, this should probably be moved in the future but for now it is left as is.


modules/globebrowsing/src/tileprovider/temporaltileprovider.cpp
```diff
@@ -505,21 +505,23 @@ TemporalTileProvider::tileProvider<TemporalTileProvider::Mode::Folder, true>(
 
     It curr = next != _folder.files.begin() ? next - 1 : next;
     It nextNext = next != _folder.files.end() ? next + 1 : curr;
+
+    if (next == _folder.files.end()) {
+        curr = next = nextNext = _folder.files.end() - 1;
```
alexanderbock, line 510: Multiple assignments per line are always tricky to parse:
```cpp
curr = _folder.files.end() - 1;
next = _folder.files.end() - 1;
nextNext = _folder.files.end() - 1;
```

modules/globebrowsing/src/tileprovider/temporaltileprovider.cpp
```diff
@@ -505,21 +505,23 @@ TemporalTileProvider::tileProvider<TemporalTileProvider::Mode::Folder, true>(
...
-        (time.j2000Seconds() - curr->first) /
-        (next->first - curr->first)
+    float factor = static_cast<float>(
+        (time.j2000Seconds() - curr->first) / (next->first - curr->first)
     );
 
-    if (_interpolateTileProvider->factor > 1.f) {
-        _interpolateTileProvider->factor = 1.f;
-    }
+    _interpolateTileProvider->factor = std::min(std::max(factor, 0.f), 1.f);
```
alexanderbock, line 524: -> `_interpolateTileProvider->factor = std::clamp(factor, 0.f, 1.f);`

PR #2040, opened 2022-04-21
interaction sphere and brightness/opacity fixes. (Closes #1994)

### LINKED ISSUE #1994: Flying to magnetosphere brings you inside planet
Assuming that it needs a bounding and/or interaction sphere

ElonOlsson:
I tested removing brightness altogether but it's needed.
What do you think about having both properties in the GUI like this? Visually it's difficult to understand the difference.

### DIFF
--- a/modules/volume/rendering/renderabletimevaryingvolume.cpp
+++ b/modules/volume/rendering/renderabletimevaryingvolume.cpp
@@ -106,7 +106,7 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo BrightnessInfo = {
         "Brightness",
         "Brightness",
-        "The volume renderers general brightness."
+        "The volume renderer's general brightness."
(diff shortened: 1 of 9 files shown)


modules/volume/rendering/renderabletimevaryingvolume.cpp
```diff
@@ -106,7 +106,7 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo BrightnessInfo = {
         "Brightness",
         "Brightness",
-        "The volumes general brightness."
+        "The volume renderers general brightness."
```
alexanderbock, line 109: -> renderer's

alexanderbock:
It's not super nice to have both but i think for now it will do, imo

alexanderbock (approved):

PR #2047, opened 2022-04-26
Issue/1989

### LINKED ISSUE #1989: XBoxController detection is a bit flimsy
We are looking for the exact string "Xbox Controller" in the asset, which will fail to detect "Wireless Xbox Controller".  Some more stable solution would be great
Option 1:  Do a RegExp based `*Xbox*` based detection
Option 2: Add a Lua function that returns the names of all connected joysticks and then use an index

Roxeena:
Add more assets for different versions of already supported joysticks, wireless Xbox, enterprise version of SpaceMouse, wireless SpaceMouse, closes #1989.

Add `any-joystick.asset` that auto detects joysticks on start up and adds the corresponding assets. There is a setting whether all connected joysticks should be added or just one. This asset only works if the joysticks are connected at the time of start up. 

Enable more buttons and axes on each joystick.


apps/OpenSpace/main.cpp
```diff
@@ -207,6 +207,71 @@ LONG WINAPI generateMiniDump(EXCEPTION_POINTERS* exceptionPointers) {
...
+
+        const unsigned char* buttons = glfwGetJoystickButtons(i, &state.nButtons);
+        state.buttons.resize(state.nButtons);
+
+        for (int j = 0; j < state.nButtons; ++j) {
+            const bool currentlyPressed = buttons[j] == GLFW_PRESS;
+
+            if (currentlyPressed) {
+                switch (state.buttons[j]) {
+                case JoystickAction::Idle:
```
alexanderbock, line 249: `case` labels should be indented by one level relative to the `switch`

apps/OpenSpace/main.cpp
```diff
@@ -207,6 +207,71 @@ LONG WINAPI generateMiniDump(EXCEPTION_POINTERS* exceptionPointers) {
...
+                    break;
+                case JoystickAction::Press:
+                case JoystickAction::Repeat:
+                    state.buttons[j] = JoystickAction::Repeat;
+                    break;
+                }
+            }
+            else {
+                switch (state.buttons[j]) {
+                case JoystickAction::Idle:
```
alexanderbock, line 261: `case` labels should be indented by one level relative to the `switch`

data/assets/util/joysticks/any-joystick.asset
```diff
@@ -0,0 +1,80 @@
+local propertyHelper = asset.require("../property_helper")
+local joystickHelper = asset.require("./joystick_helper")
+local initializeAll = false
+
+local function numItems(iTable)
```
alexanderbock, line 5: Shouldn't `#iTable` do the same here here as this function?

modules/imgui/src/guijoystickcomponent.cpp
```diff
@@ -84,19 +86,21 @@ void GuiJoystickComponent::render() {
 
     ImGui::Text("%s", "Summed contributions");
     ImGui::Text("%s", "Axes");
-    for (int i = 0; i < JoystickInputState::MaxAxes; ++i) {
+    for (int i = 0; i < global::joystickInputStates->numAxes(""); ++i) {
```
alexanderbock, line 89: Should `numAxes` get the parameter with a default value instead so that we can write:
`global::joystickInputStates->numAxes()` here?

modules/imgui/src/guijoystickcomponent.cpp
```diff
@@ -84,19 +86,21 @@ void GuiJoystickComponent::render() {
...
-            std::to_string(i).c_str(),
+            id.c_str(),
             &f,
             -1.f,
             1.f
         );
     }
     ImGui::Text("%s", "Buttons");
-    for (int i = 0; i < JoystickInputState::MaxButtons; ++i) {
+    for (int i = 0; i < global::joystickInputStates->numButtons(""); ++i) {
```
alexanderbock, line 100: Same argument as above

WeirdRubberDuck (commented):
Code looks good to me, but I couldn't really try it (bc no joystick)

PR #2050, opened 2022-04-27
Feature/skybrowser

ylvaselling:
Fixes to improve performance, bugfixes and not clog up the lua script log:

- Sky browser topic
- Listen to callback to ensure image collection has been loaded
- Enable WebGL2
- Pruned image collection
- JWST pointing event dispatch from button in GUI
- Fix FOV mismatch between target and browser
- Some settings that has been asked for (hiding all skybrowser related renderables when minimizing GUI, hide title, scroll direction,)
- Hover circle updates
- Hide things using Fade property, instead of Opacity and/or Enabled (and solve some bugs that existed in relation to that)


WeirdRubberDuck:
Making this a draft for now, as there are more changes to come

modules/skybrowser/skybrowsermodule.cpp
```diff
@@ -126,45 +161,41 @@ SkyBrowserModule::SkyBrowserModule()
...
             incrementallyAnimateTargets();
         }
         if (_cameraRotation.isAnimating() && _allowCameraRotation) {
             incrementallyRotateCamera();
         }
+
+        // Trigger callbacks (should maybe have a check to see if update is needed)
+        using K = CallbackHandle;
+        using V = CallbackFunction;
+        for (const std::pair<K, V>& it : _preSyncCallbacks) {
```
alexanderbock, line 196: I believe this would run a bit faster if `const std::pair<const K, V>&`

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -455,9 +453,9 @@ namespace {
             "Translation = {"
                 "Type = 'StaticTranslation',"
                 "Position = {"
-                        + std::to_string(galacticTarget.x) + ", "
-                        + std::to_string(galacticTarget.y) + ", "
-                        + std::to_string(galacticTarget.z) + ", "
+                    + std::to_string(galacticTarget.x) + ", "
```
alexanderbock, line 456: The `+` would move to the end of previous line instead to make visually sure that it is a continuation:

```cpp
"Position = {" +
    std::to_string(galacticTarget.x) + ", " +
    std::to_string(galacticTarget.y) + ", " +
```
```

File: modules/skybrowser/skybrowsermodule_lua.inl [hunk 3]
```diff
@@ -652,26 +650,29 @@ namespace {
  * Takes an identifier to a sky browser and adds a rendered copy to it.
  * \param raePosition Position in radius, azimuth, elevation coordinates
  * \param nCopies Number of copies
+ * \param position Position of first copy, which depends on if RAE or Cartesian coordinates
+ * are checked
  */
 [[codegen::luawrap]] void addRenderCopy(std::string identifier,
                                         int numberOfCopies = 1,
-                     glm::vec3 radiusAzimuthElevationPosition = glm::vec3(2.1f, 0.f, 0.f))
+                     glm::vec3 position = glm::vec3(2.1f, 0.f, 0.f))
```
alexanderbock, line 658: the beginning of this line should align with the `(` of the function if it fits, otherwise right-aligned

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -711,6 +716,56 @@ namespace {
...
+    using namespace openspace;
+
+    SkyBrowserModule* module = global::moduleEngine->module<SkyBrowserModule>();
+    TargetBrowserPair* pair = module->getPair(identifier);
+    if (pair) {
+        LINFO("Image collection is loaded in Screen Space Sky Browser " + identifier);
+        pair->setImageCollectionIsLoaded(true);
+        // Add all selected images to WorldWide Telescope
+        const std::deque<int>& images = pair->selectedImages();
+        std::for_each(images.rbegin(), images.rend(), [&](int index) {
```
alexanderbock, line 733:
```cpp
std::for_each(
    images.rbegin(), images.rend(),
    [&](int index) {
        const ImageData& image = module->getWwtDataHandler()->getImage(index);
        // Index of image is used as layer ID as it is unique in the image data set
        pair->browser()->addImageLayerToWwt(image.imageUrl, index);
    }
);
```

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -711,6 +716,56 @@ namespace {
...
+ * Point spacecraft to the equatorial coordinates the target points to
+ * \param identifier Identifier of the sky browser
+ */
+[[codegen::luawrap]] void pointSpaceCraft(std::string identifier) {
+    using namespace openspace;
+    SkyBrowserModule* module = global::moduleEngine->module<SkyBrowserModule>();
+    TargetBrowserPair* pair = module->getPair(identifier);
+    glm::dvec2 equatorial = pair->targetDirectionEquatorial();
+    global::eventEngine->publishEvent<events::EventPointSpacecraft>(
+        equatorial.x, equatorial.y
```
alexanderbock, line 765:
```cpp
global::eventEngine->publishEvent<events::EventPointSpacecraft>(
    equatorial.x,
    equatorial.y
);
```

modules/skybrowser/src/screenspaceskybrowser.cpp
```diff
@@ -100,32 +100,43 @@ ScreenSpaceSkyBrowser::ScreenSpaceSkyBrowser(const ghoul::Dictionary& dictionary
...
-    glm::vec2 screenPosition = _cartesianPosition.value();
-    _cartesianPosition = glm::vec3(screenPosition, skybrowser::ScreenSpaceZ);
-
     if (global::windowDelegate->isMaster()) {
         SkyBrowserModule* module = global::moduleEngine->module<SkyBrowserModule>();
         _borderColor = randomBorderColor(module->highlight());
     }
     _scale = _size.y * 0.5f;
+
+    _useRadiusAzimuthElevation.onChange([this]() {
```
alexanderbock, line 126: `[this]() {` should be on the next line and the function indented by one level

modules/skybrowser/src/screenspaceskybrowser.cpp
```diff
@@ -100,32 +100,43 @@ ScreenSpaceSkyBrowser::ScreenSpaceSkyBrowser(const ghoul::Dictionary& dictionary
...
+        std::for_each(
+            _renderCopies.begin(),
+            _renderCopies.end(),
+            [this](std::unique_ptr<properties::Vec3Property>& copy) {
+            if (_useRadiusAzimuthElevation) {
+                *copy = sphericalToRae(cartesianToSpherical(copy.get()->value()));
```
alexanderbock, line 132: no need for the `.get()` ;  `copy->value()` should work too

modules/skybrowser/src/targetbrowserpair.cpp
```diff
@@ -250,13 +261,17 @@ void TargetBrowserPair::setBorderColor(const glm::ivec3& color) {
 
 void TargetBrowserPair::setScreenSpaceSize(const glm::vec2& dimensions) {
     _browser->setScreenSpaceSize(dimensions);
-    _targetRenderable->setDimensions(dimensions);
+    _targetRenderable->setRatio(dimensions.x/dimensions.y);
```
alexanderbock, line 264: Would say `dimensions.x / dimensions.y` is easier to read

modules/skybrowser/src/targetbrowserpair.cpp
```diff
@@ -275,10 +290,16 @@ void TargetBrowserPair::incrementallyAnimateToCoordinate() {
...
-    _fadeBrowser = skybrowser::Animation(_browser->opacity(), goal, fadeTime);
-    _fadeTarget.start();
-    _fadeBrowser.start();
+    const std::string script = fmt::format(
+        "openspace.setPropertyValueSingle('Scene.{0}.Renderable.Fade', {2}, {3});"
+        "openspace.setPropertyValueSingle('ScreenSpace.{1}.Fade', {2}, {3});",
+        _targetNode->identifier(), _browser->identifier(), goal, fadeTime
+    );
+
+    openspace::global::scriptEngine->queueScript(
```
alexanderbock, line 299: I don't think the `openspace::` here is necessary?

modules/server/src/topics/skybrowsertopic.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include <modules/skybrowser/skybrowsermodule.h>
+#include <modules/skybrowser/include/targetbrowserpair.h>
+#include <openspace/engine/moduleengine.h>
+#include <openspace/engine/globals.h>
+#include <openspace/properties/property.h>
+#include <openspace/query/query.h>
+#include <ghoul/logging/logmanager.h>
+
+namespace {
+    constexpr const char* EventKey = "event";
```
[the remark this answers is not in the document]
ylvaselling: json.at doesnt accept string_views unfortunately, although it is possible to do a conversion to const char* with string_view::data (IF the string is null-terminated so I guess that means putting a \0 at the end of all strings which probably is ugly)

modules/server/include/topics/skybrowsertopic.h
```diff
@@ -0,0 +1,54 @@
...
+class SkyBrowserTopic : public Topic {
+public:
+    SkyBrowserTopic();
+    virtual ~SkyBrowserTopic();
+
+    void handleJson(const nlohmann::json& json) override;
+    bool isDone() const override;
+
+private:
+    const int UnsetOnChangeHandle = -1;
```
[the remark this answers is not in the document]
ylvaselling: It is used to initialize the _targetDataCallback, I guess I could move the initialization to the constructor?

modules/server/src/topics/skybrowsertopic.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    // Pass data for all the browsers and the corresponding targets
+    if (module->isCameraInSolarSystem()) {
+        const std::vector<std::unique_ptr<TargetBrowserPair>>& pairs = module->getPairs();
+        ghoul::Dictionary targets;
+        for (const std::unique_ptr<TargetBrowserPair>& pair : pairs) {
+            std::string id = pair->browserId();
+            // Convert deque to vector so ghoul can read it
+            std::vector<int> selectedImagesVector;
+            const std::deque<int> selectedImages = pair->selectedImages();
```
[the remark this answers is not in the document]
ylvaselling: Is changed to different structure now:)

modules/skybrowser/src/screenspaceskybrowser.cpp
```diff
@@ -50,25 +50,25 @@ namespace {
 
     constexpr const openspace::properties::Property::PropertyInfo RenderCopyInfo = {
         "RenderCopy",
-        "RAE Position Of A Copy Of The Sky Browser",
+        "Position Of A Copy Of The Sky Browser",
```
WeirdRubberDuck, line 53: Wondering if this should be "Position of a Copy of the Sky Browser", or even better, maybe we should come up with a shorter name. "Sky Browser Copy Position"? or just "Render Copy Position", as the description says what it is 

We should probably also decide on whether it is a "RenderCopy" or "DisplayCopy" (the Gui says Display Copy
)

modules/skybrowser/src/screenspaceskybrowser.cpp
```diff
@@ -50,25 +50,25 @@ namespace {
...
 
-    constexpr const openspace::properties::Property::PropertyInfo RenderOnMasterInfo = {
-        "RenderOnlyOnMaster",
-        "Render Only On Master",
-        "Render the interactive sky browser only on the master node (this setting won't "
-        "affect the copies). This setting allows mouse interactions in a dome "
-        "environment."
+    constexpr const openspace::properties::Property::PropertyInfo IsHiddenInfo = {
+        "IsHidden",
+        "Is hidden",
```
WeirdRubberDuck, line 61: "Is Hidden"

modules/skybrowser/include/renderableskytarget.h
```diff
@@ -64,13 +60,11 @@ class RenderableSkyTarget : public RenderablePlane {
...
-    properties::DoubleProperty _stopAnimationThreshold;
-    properties::DoubleProperty _animationSpeed;
 
     double _verticalFov = 10.0;
 
     glm::ivec3 _borderColor = glm::ivec3(230);
-    glm::vec2 _dimensions = glm::vec2(1.f);
+    float _ratio = 1.f;
 };
 } // namespace openspace
```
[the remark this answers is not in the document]
WeirdRubberDuck: Between line 68 and 69 there should be an empty line :)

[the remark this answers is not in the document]
WeirdRubberDuck: My guess is that there should be an empty new line between line 86 and 69 :)

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -765,6 +765,17 @@ namespace {
         );
 }
 
+/**
+ * Stop animations
+ * \param identifier Identifier of the sky browser
```
alexanderbock, line 770: The way we are parsing these comments for the Lua file doesn't follow the Doxygen way yet. So you'll have to write the description out at the moment:
```
/**
 * Stop animations.  The parameter to this function is the identifier of the sky browser
 */
```
or something like that

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -765,6 +765,17 @@ namespace {
...
+ */
+[[codegen::luawrap]] void stopAnimations(std::string identifier) {
+    using namespace openspace;
+    SkyBrowserModule* module = global::moduleEngine->module<SkyBrowserModule>();
+    TargetBrowserPair* pair = module->getPair(identifier);
```
alexanderbock, line 775: For consistency with the naming scheme, we should rename `getPair` to just `pair`.  We're only using the `get` variant if the function takes a reference to the object being filled, which is not something that we'd want to do everywhere anyway. Instead:
```cpp
TargetBrowserPair* pair = module->pair(identifier);
```

alexanderbock (approved):

WeirdRubberDuck:
Was thinking about adding the documentation for `RenderableSkyTarget` and `ScreenSpaceSkyBrowser`. Do we have a reason behind not doing so? (Thinking about that we might not want these to be created from a user asset, etc?) Or has it just been missed? :)

ylvaselling:
> Was thinking about adding the documentation for `RenderableSkyTarget` and `ScreenSpaceSkyBrowser`. Do we have a reason behind not doing so? (Thinking about that we might not want these to be created from a user asset, etc?) Or has it just been missed? :)

Hm I think I thought I needed to add doxygen comments to all the functions in order for that to work. Is that not the case?

alexanderbock:
No, it does need doxygen for all the functions.. I think having a small description at the top of the file would make sense

Issue #1529, opened 2021-03-12
Ability to set Min and Max Property values from assets


WeirdRubberDuck:
For a more useful UI, it would be good if the max and min values for numeric properties could be set from an asset file

alexanderbock:
+ exponent

PR #2064, opened 2022-05-02
User Properties

alexanderbock:
Add the ability to add user-defined properties into the system;
`openspace.addCustomProperty` takes between 2 and 5 arguments:
1. The identifier of the new property
2. The Type of the new property (same naming as the C++ types)
3. The GUI name of the property
4. The Description of the property
5. A Lua script that is called when the property is changed.  The new value is inserted into this script as the `value` parameter;

For example:
`openspace.addCustomProperty("example", "IntProperty", "Example", "This is an example property", [[ if value > 0.0 then openspace.printInfo(value) else openspace.printError(value) end ]])`

Right now there is no way to specify the minimum/maximum values of the new Property;  that is part of #1529 to add that capability to this functionality.

### DIFF
--- a/src/engine/globals.cpp
+++ b/src/engine/globals.cpp
@@ -351,7 +351,7 @@ void create() {
     ghoul_assert(userPropertyOwner, "No userPropertyOwner");
     currentPos += sizeof(properties::PropertyOwner);
 #else // ^^^ WIN32 / !WIN32 vvv
-    userPropertyOwner = new properties::PropertyOwner({ "ScreenSpace" });
+    userPropertyOwner = new properties::PropertyOwner({ "UserProperties" });
 #endif // WIN32
 
 #ifdef WIN32
--- a/src/engine/globals.cpp
+++ b/src/engine/globals.cpp
@@ -98,6 +98,7 @@ namespace {
         sizeof(interaction::SessionRecording) +
         sizeof(properties::PropertyOwner) +
         sizeof(properties::PropertyOwner) +
+        sizeof(properties::PropertyOwner) +
         sizeof(scripting::ScriptEngine) +
         sizeof(scripting::ScriptScheduler) +
         sizeof(Profile);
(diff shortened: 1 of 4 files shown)


src/engine/globals.cpp
```diff
@@ -345,6 +346,14 @@ void create() {
     screenSpaceRootPropertyOwner = new properties::PropertyOwner({ "ScreenSpace" });
 #endif // WIN32
 
+#ifdef WIN32
+    userPropertyOwner = new (currentPos) properties::PropertyOwner({ "UserProperties" });
+    ghoul_assert(userPropertyOwner, "No userPropertyOwner");
+    currentPos += sizeof(properties::PropertyOwner);
+#else // ^^^ WIN32 / !WIN32 vvv
+    userPropertyOwner = new properties::PropertyOwner({ "ScreenSpace" });
```
WeirdRubberDuck, line 354: Should this really be "ScreenSpace"?

WeirdRubberDuck (approved):
Cool :) Played with it a bit and tried creating and setting it in different ways and it all seems to work just fine

Kind of missing a parameter for the default value in the input, but it's not too much work to just to a `openspace.setPropertyValueSingle` call right after the `createCustomProperty` call.

src/engine/globals.cpp
alexanderbock: Absolutely!

src/scene/scene_lua.inl
```diff
@@ -917,6 +947,157 @@ namespace {
...
+    // to have a factory function for this but right now this is the only place where that
+    // factory would be used.
+
+    const char* gui =
+        guiName.has_value() && !guiName->empty() ?
+        guiName->c_str() :
+        identifier.c_str();
+
+    properties::Property::PropertyInfo info = {
+            identifier.c_str(),
```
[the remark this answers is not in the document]
alexanderbock: indeed

alexanderbock:
> Cool :) Played with it a bit and tried creating and setting it in different ways and it all seems to work just fine
> 
> Kind of missing a parameter for the default value in the input (since the user has no way of finding out what the default value for the property types are), but it's not too much work to just do a `openspace.setPropertyValueSingle` call right after the `createCustomProperty` call.

Yea, the problem is that that would require support for generic types in the Lua function in order to pass the parameter, which is a bit tricky right now so I think for now we have to do the `setPropertyValueSingle` thing unfortunately :(

PR #2068, opened 2022-05-03
Issue/2055

### LINKED ISSUE #2055: Perseverance and Insight placed around Sun instead or Mars
Insight model is always placed in relation to the Sun (or SSB, didn't check which) instead or Mars.

insidgth-sun-landing
insight-sun-range

Perseverance model is placed relative to the Sun when outside data range, otherwise in relation to Mars.

perseverance-sun-outside-range

Roxeena:
closes #2055

### DIFF
--- a/data/assets/scene/solarsystem/missions/insight/edl.asset
+++ b/data/assets/scene/solarsystem/missions/insight/edl.asset
@@ -151,7 +151,7 @@ local RotationKeyframes = {
 
 local InsightParent = {
   Identifier = "InsightParent",
-  Parent = mars.Identifier,
+  Parent = mars.Mars.Identifier,
   Transform = {
     Translation ={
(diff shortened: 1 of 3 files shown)


data/assets/scene/solarsystem/missions/insight/edl.asset
```diff
@@ -151,7 +151,7 @@ local RotationKeyframes = {
 
 local InsightParent = {
   Identifier = "InsightParent",
-  Parent = mars.Identifier,
+  Parent = mars.Mars.Identifier,
```
alexanderbock, line 154: Very strange that that didn't throw a Lua error;  with the addition of the strict-mode it should have thrown a Lua error when trying to access a non-existing table field

alexanderbock (approved):

PR #2071, opened 2022-05-04
a new profile for the 2012 solar storm event, with looping hotkeys

ElonOlsson:
This creates a new public profile for the 2012 july solar storm event, with ENLIL, PFSS and BATSRUS models.
Only thing I can think of is to set camera position better. How can I find the current camera position, so that I can copy it ?

Needs to be added to list of profile in .cfg also.

### DIFF
--- a/data/profiles/solarstorm2012.profile
+++ b/data/profiles/solarstorm2012.profile
@@ -1,36 +1,36 @@
 {
   "actions": [
     {
-      "documentation": "Reset button. Sets time to start at start of dataset. No loop",
+      "documentation": "Reset button. Sets time to start of dataset. No loop",
       "gui_path": "2012July",
-      "identifier": "2012July.Reset-loop",
+      "identifier": "2012july.reset_loop",
       "is_local": false,
-      "name": "Reset button.",
+      "name": "Reset button",
       "script": "openspace.time.setTime('2012-JUL-01 07:00:00.00');\nopenspace.scriptScheduler.clear();"
     },
     {
-      "documentation": "Makes the sun turn black",
+      "documentation": "Makes the Sun turn black",
       "gui_path": "2012July",
-      "identifier": "2012July.Dark-sun",
+      "identifier": "2012july.dark_sun",
       "is_local": false,
       "name": "Dark sun",
       "script": "openspace.setPropertyValueSingle('Scene.Sun.Renderable.Layers.ColorLayers.Texture.Settings.Multiplier', 0.000000);"
     },
     {
-      "documentation": "Sets time to start of data. sets high delta time. loops from start when at end of data",
+      "documentation": "Sets time to start of data, sets high delta time and loops back from start when at end of data",
       "gui_path": "2012July",
-      "identifier": "2012July.loopEnlil",
+      "identifier": "2012july.loop_enlil",
       "is_local": false,
       "name": "Loop ENLIL",
-      "script": "openspace.scriptScheduler.clear();\nopenspace.time.setDeltaTime(43200);\nopenspace.time.setTime('2012-JUL-01 07:00:00.00');\nStarttimescriptENLILLoop = \"openspace.time.setTime('2012 JUL 01 07:00:00')\";\nopenspace.scriptScheduler.loadScheduledScript('2012 AUG 01 06:58:00', StarttimescriptENLILLoop);"
+      "script": "openspace.scriptScheduler.clear();\nopenspace.time.setDeltaTime(43200);\nopenspace.time.setTime('2012-JUL-01 07:00:00.00');\nlocal StarttimescriptENLILLoop = \"openspace.time.setTime('2012 JUL 01 07:00:00')\";\nopenspace.scriptScheduler.loadScheduledScript('2012 AUG 01 06:58:00', StarttimescriptENLILLoop);"
     },
     {
-      "documentation": "Sets time to start of data. Sets higher delta time. Loops back to start, at end of data.",
+      "documentation": "Sets time to start of data, sets higher delta time and loops back from start, when at end of data.",
       "gui_path": "2012July",
-      "identifier": "2012July.LoopBatsrus",
+      "identifier": "2012july.loop_batsrus",
       "is_local": false,
       "name": "Loop BATSRUS",
-      "script": "openspace.scriptScheduler.clear();\nopenspace.time.setDeltaTime(1400);\nopenspace.time.setTime('2012-JUL-14 06:00:00.00');\nStarttimescriptBatsrusLoop = \"openspace.time.setTime('2012 JUL 14 06:00:00')\";\nopenspace.scriptScheduler.loadScheduledScript('2012 JUL 16 07:30:00', StarttimescriptBatsrusLoop);"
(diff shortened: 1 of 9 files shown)


alexanderbock:
Feel free to add it to that list as part of this PR

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
+{
+  "actions": [
+    {
+      "documentation": "Reset button. Sets time to start at start of dataset. No loop",
```
alexanderbock, line 4: "to start at start"

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "gui_path": "2012July",
+      "identifier": "2012July.Reset-loop",
```
alexanderbock, line 6: I believe we are (and if we don't, we should) stick to all lower case for the identifiers to reduce the cognitive load to remember whether something was lower case or upper case.  By making everything lower case, it'll be easy to remember

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "is_local": false,
+      "name": "Reset button.",
```
alexanderbock, line 8: I'd not have the period at the end of the name

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "script": "openspace.time.setTime('2012-JUL-01 07:00:00.00');\nopenspace.scriptScheduler.clear();"
+    },
+    {
+      "documentation": "Makes the sun turn black",
```
alexanderbock, line 12: -> Sun

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "gui_path": "2012July",
+      "identifier": "2012July.Dark-sun",
+      "is_local": false,
+      "name": "Dark sun",
+      "script": "openspace.setPropertyValueSingle('Scene.Sun.Renderable.Layers.ColorLayers.Texture.Settings.Multiplier', 0.000000);"
+    },
+    {
+      "documentation": "Sets time to start of data. sets high delta time. loops from start when at end of data",
```
alexanderbock, line 20: -> "Sets time to start of data, sets high delta time, and loops from start when at end of data"

data/profiles/solarstorm2012.profile
alexanderbock: Same argument for only using an underscore rather than a dot;  so:
`2021july.reset_loop` instead just for consistency with how we write the identifies of our scene graph nodes

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "gui_path": "2012July",
+      "identifier": "2012July.loopEnlil",
+      "is_local": false,
+      "name": "Loop ENLIL",
+      "script": "openspace.scriptScheduler.clear();\nopenspace.time.setDeltaTime(43200);\nopenspace.time.setTime('2012-JUL-01 07:00:00.00');\nStarttimescriptENLILLoop = \"openspace.time.setTime('2012 JUL 01 07:00:00')\";\nopenspace.scriptScheduler.loadScheduledScript('2012 AUG 01 06:58:00', StarttimescriptENLILLoop);"
```
alexanderbock, line 25: To be future proof when we make the Lua state that executes this strict, you should:
`local StarttimescriptENLILLopp` instead

alexanderbock, line 25: To be future proof when we make the Lua state that executes this strict, you should:
`local StarttimescriptENLILLopp` instead

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "gui_path": "2012July",
+      "identifier": "2012July.Dark-sun",
+      "is_local": false,
+      "name": "Dark sun",
+      "script": "openspace.setPropertyValueSingle('Scene.Sun.Renderable.Layers.ColorLayers.Texture.Settings.Multiplier', 0.000000);"
+    },
+    {
+      "documentation": "Sets time to start of data. sets high delta time. loops from start when at end of data",
+      "gui_path": "2012July",
+      "identifier": "2012July.loopEnlil",
```
alexanderbock, line 22: Would use an underscore here as well instead of camelcase (which would go away anyway since we use lower case everywhere)

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "is_local": false,
+      "name": "Loop ENLIL",
+      "script": "openspace.scriptScheduler.clear();\nopenspace.time.setDeltaTime(43200);\nopenspace.time.setTime('2012-JUL-01 07:00:00.00');\nStarttimescriptENLILLoop = \"openspace.time.setTime('2012 JUL 01 07:00:00')\";\nopenspace.scriptScheduler.loadScheduledScript('2012 AUG 01 06:58:00', StarttimescriptENLILLoop);"
+    },
+    {
+      "documentation": "Sets time to start of data. Sets higher delta time. Loops back to start, at end of data.",
```
alexanderbock, line 28: -> "Sets time to start of data, sets higher delta time, and loops back to start, at end of data.

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "gui_path": "2012July",
+      "identifier": "2012July.LoopBatsrus",
```
alexanderbock, line 30: Using _

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+      "is_local": false,
+      "name": "Loop BATSRUS",
+      "script": "openspace.scriptScheduler.clear();\nopenspace.time.setDeltaTime(1400);\nopenspace.time.setTime('2012-JUL-14 06:00:00.00');\nStarttimescriptBatsrusLoop = \"openspace.time.setTime('2012 JUL 14 06:00:00')\";\nopenspace.scriptScheduler.loadScheduledScript('2012 JUL 16 07:30:00', StarttimescriptBatsrusLoop);"
```
alexanderbock, line 33: Same argument as above for using `local` for the variable definition to make the profile future proof for a strict Lua state

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+    15552000.0,
+    31536000.0,
+    63072000.0,
+    157680000.0,
+    315360000.0,
+    630720000.0
+  ],
+  "keybindings": [
+    {
+      "action": "2012July.Reset-loop",
```
alexanderbock, line 72: Don't forget to update these when you change the identifier of the action

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+  "mark_nodes": [
+    "Earth",
+    "Mars",
+    "Moon",
+    "Sun",
+    "Venus",
+    "ISS"
+  ],
+  "meta": {
+    "author": "CCMC",
```
alexanderbock, line 97: How about:
"Community Coordinated Modeling Center, NASA Goddard"?  Not everyone might know what a CCMC is ;)

data/profiles/solarstorm2012.profile
```diff
@@ -0,0 +1,124 @@
...
+    "description": "This profile is showing several coronal mass ejection (CMEs) during July 2012.",
```
alexanderbock, line 98: Could you add a bit more description here?  Saying something about the models being used, why July 2012 was interesting to model, etc

alexanderbock (commented):
Looks good!  Will try it out tomorrow

PR #2083, opened 2022-05-06
Fixes position jumps in OSIRIS-REx mission

GPayne:
Updated SPICE spk kernels resolve the position jumps in issue#1651. The other complaints listed in that issue were fixed back in 2021.
These updated SPICE kernels are currently only on the Utah data server, so to try this version, it's necessary to switch to that server in openspace.cfg.


alexanderbock:
Data is now referenced on the Sweden server as well;  Fine to merge with me :)

Issue #202, opened 2017-02-03
Simplified GUI


alexanderbock:

alexanderbock:
Related to #973   in so far as when it is possible to easily switch the endpoint that is served, we can create a second UI that is much simpler for the first-time usage, which can also include a tutorial view

alexanderbock:
GUI that is not based on showing all/filtered property list, but something more specialized

ylvaselling:
Closing since the PR mentioned in this issue is merged

PR #2094, opened 2022-05-09
Add settings in OpenSpaceEngine that sets visibility in gui

ylvaselling:
This hides properties depending on their visibility setting and scene graph nodes depending on their isHidden property. Use with WebGuiFrontend on feature/properties-visibility


alexanderbock (approved):

alexanderbock:
image

Two suggestions for the property:
 1. Instead of "Visibility" I'd say "Property Visibility" to make it more obvious
 2. The "Advanced User" and "Novice User" are missing in the dropdown list

src/engine/openspaceengine.cpp
```diff
@@ -146,6 +162,19 @@ OpenSpaceEngine::OpenSpaceEngine()
...
     SpiceManager::initialize();
     TransformationManager::initialize();
+
+    addProperty(_printEvents);
+    addProperty(_visibility);
+    addProperty(_showHiddenSceneGraphNodes);
+
+    using Visibility = openspace::properties::Property::Visibility;
+    _visibility.addOptions({
+        { static_cast<int>(Visibility::NoviceUser), "Novice User"},
```
WeirdRubberDuck, line 172: veery nitpicky, but missing space before closing brackets for all of the options

src/engine/globals.cpp
```diff
@@ -399,6 +399,7 @@ void initialize() {
     rootPropertyOwner->addPropertySubOwner(global::dashboard);
 
     rootPropertyOwner->addPropertySubOwner(global::userPropertyOwner);
+    rootPropertyOwner->addPropertySubOwner(global::openSpaceEngine);
```
WeirdRubberDuck, line 402: This seems to affect the order in which they are added in the Settings menu as well.

I suggest we put it a bit higher up in the list, maybe next to `RenderEngine`? Would make more sense, conceptually, at least to me

WeirdRubberDuck (approved):
Nice :) But don't forget to update the title of the PR before merging ;) (the setting Is no longer in the CEFWebGUiModule) 

Or even better, select the "Squash and merge" option and write a meaningful commit message

src/engine/globals.cpp
ylvaselling: @alexanderbock ?

alexanderbock: Yea, I agree with @WeirdRubberDuck    in addition it might be worth thinking about sorting the Settings list alphabetically additionally?

include/openspace/properties/property.h
```diff
@@ -62,13 +62,15 @@ class Property {
...
     * The visibility classes for Property%s. The classes are strictly ordered as
-    * All > Developer > User > Hidden
+    * Hidden > Developer > AdvancedUser > User > NoviceUser > Always
     */
     enum class Visibility {
-        Hidden = 3, ///< Never visible
-        Developer = 2, ///< Visible in Developer mode
-        User = 1, ///< Visible in User mode
-        All = 0,  ///< Visible for all types, no matter what
+        Hidden = 5, ///< Never visible
```
WeirdRubberDuck, line 68: Do we really need all of these categories? Might be tricky/annoying to maintain

We should all have a discussion as a team about this when we have the chance. (Talked about it briefly during the dev meeting on May 23)

@alexanderbock

alexanderbock: Yea, let's have a talk about it, but I think we do:
 - `Hidden` and `Always` we will need for sure as sentinel values
 - `Developer` and `User` also seems a logical choice to me.
 -  `AdvancedUser` would be the types of properties that we would want knowledeable users to be able to interact with (for example the atmosphere scattering parameters, but others too)
 - `NoviceUser` the type of properties that would result in a "simplified ui" that could work as a first version for #202

alexanderbock: and the "Advanced User" / "User" / "Novice User" has a nice symmetry to it

PR #2132, opened 2022-05-31
added events for when renderables are enabled/disabled that can be us…

micahnyc:
…ed to link renderables together

### DIFF
--- a/include/openspace/events/event.h
+++ b/include/openspace/events/event.h
@@ -35,6 +35,7 @@ namespace openspace {
     class Camera;
     class Layer;
     class Profile;
+    class Renderable;
     class SceneGraphNode;
     class ScreenSpaceRenderable;
     class Time;
@@ -74,6 +75,8 @@ struct Event {
         LayerRemoved,
         SessionRecordingPlayback,
         PointSpacecraft,
+        RenderableEnabled,
+        RenderableDisabled,
         Custom
     };
     constexpr explicit Event(Type type_) : type(type_) {}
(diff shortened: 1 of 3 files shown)


micahnyc:
[isstrail.asset.txt](https://github.com/OpenSpace/OpenSpace/files/8809761/isstrail.asset.txt)

here is an example turning off / on the iss trail when you turn off the model

include/openspace/events/event.h
```diff
@@ -397,6 +400,32 @@ struct EventPointSpacecraft : public Event {
     const double duration;
 };
 
+/**
+ * This event is created whenever a renderable is enabled.  By the
+ * time this event is signalled, the renderable has already been enabled.
+ *
+ * \param Node The identifier of the node that contains the renderable
```
alexanderbock, line 407: The \param line would need to be moved to the constructor.  Since this is a non-templated struct, it doesn't have any parameters

include/openspace/events/event.h
```diff
@@ -397,6 +400,32 @@ struct EventPointSpacecraft : public Event {
...
+
+    explicit EventRenderableEnabled(const SceneGraphNode* node_);
+    const tstring node;
+};
+
+/**
+ * This event is created whenever a renderable is disabled.  By the
+ * time this event is signalled, the renderable has already been disabled.
+ *
+ * \param Node The identifier of that node that contains the renderable
```
alexanderbock, line 420: Same here

include/openspace/events/event.h
alexanderbock: Oh yea, just seeing that now..  someone (read: me) put the comments in the wrong place. Having it like:
```cpp
/**
 * This event is created when a screenspace renderable has been removed from the system.
 * When this event is created, the screenspace renderable has already been removed and is
 * no longer available
 */
struct EventScreenSpaceRenderableRemoved : public Event {
    static const Type Type = Event::Type::ScreenSpaceRenderableRemoved;

    /**
     *\param renderable_ The identifier of the screenspace renderable that was removed
     */
    explicit EventScreenSpaceRenderableRemoved(const ScreenSpaceRenderable* renderable_);
    const tstring renderable;
};
```

would generate better Doxygen comments for the vents

PR #2139, opened 2022-06-08
Feature/gui mouseinteraction fix

WeirdRubberDuck:
Aims to fix interaction problems when hovering/clicking with the mouse in a non-GUI window, when having a setup with more than one window (e.g. the single_gui.json setup)

Previously, clicking with the mouse in the rendering window without the GUI also triggered mouse clicks in all the other windows, including the GUI windows. This was disturbing for interaction in the single_gui setup

Now, the GUI modules (ImGui and webbrowser) check to see if the clicked window is actually an interaction window

related PR in SGCT: https://github.com/sgct/sgct/pull/55 (note that you might have to check out the latest commit in that branch for it to work)

Will do some more through testing before considering a merge, but feel free to look at the implementation and try it out already

### DIFF
--- a/apps/OpenSpace/main.cpp
+++ b/apps/OpenSpace/main.cpp
@@ -276,6 +276,15 @@ void checkJoystickStatus() {
     }
 }
 
+bool isGuiWindow(sgct::Window* window) {
+    if (Engine::instance().windows().size() == 1) {
+        // If we only have one window, assume it's also the GUI window.
+        // It might not have been given the 'GUI' tag
+        return true;
+    }
+    return window->hasTag("GUI");
+}
+
 //
 //  Init function
 //
@@ -574,7 +583,7 @@ void mainPostDrawFunc() {
 
 
 void mainKeyboardCallback(sgct::Key key, sgct::Modifier modifiers, sgct::Action action,
-                          int)
+                          int, sgct::Window* window)
 {
     ZoneScoped
     LTRACE("main::mainKeyboardCallback(begin)");
@@ -582,7 +591,9 @@ void mainKeyboardCallback(sgct::Key key, sgct::Modifier modifiers, sgct::Action 
     const openspace::Key k = openspace::Key(key);
     const KeyModifier m = KeyModifier(modifiers);
     const KeyAction a = KeyAction(action);
-    global::openSpaceEngine->keyboardCallback(k, m, a);
+    const IsGuiWindow isGui = IsGuiWindow(isGuiWindow(window));
+
+    global::openSpaceEngine->keyboardCallback(k, m, a, isGui);
 
     LTRACE("main::mainKeyboardCallback(begin)");
 }
@@ -590,7 +601,7 @@ void mainKeyboardCallback(sgct::Key key, sgct::Modifier modifiers, sgct::Action 
 
 
 void mainMouseButtonCallback(sgct::MouseButton key, sgct::Modifier modifiers,
-                             sgct::Action action)
+                             sgct::Action action, sgct::Window* window)
 {
     ZoneScoped
     LTRACE("main::mainMouseButtonCallback(begin)");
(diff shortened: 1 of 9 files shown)


WeirdRubberDuck:
Updated so that the isGuiWindow check is part of all the mouse/keyboard callbacks from SGCT. 

Note that this does not include touch, since this is handled in a separate module that I'm not too familiar with right now. I suggest we fix that as a separate PR

apps/OpenSpace/main.cpp
```diff
@@ -574,60 +583,61 @@ void mainPostDrawFunc() {
 
 
 void mainKeyboardCallback(sgct::Key key, sgct::Modifier modifiers, sgct::Action action,
-                          int)
+                          int, sgct::Window* window)
```
alexanderbock, line 586: This argument would be on a new line since we have overflown the column limit one

apps/OpenSpace/main.cpp
```diff
@@ -574,60 +583,61 @@ void mainPostDrawFunc() {
...
+    global::openSpaceEngine->keyboardCallback(k, m, a, isGuiWindow(window));
 
     LTRACE("main::mainKeyboardCallback(begin)");
 }
 
 
 
 void mainMouseButtonCallback(sgct::MouseButton key, sgct::Modifier modifiers,
-                             sgct::Action action)
+                             sgct::Action action, sgct::Window* window)
```
alexanderbock, line 602: Same here

include/openspace/engine/globalscallbacks.h
```diff
@@ -42,12 +42,12 @@ inline std::vector<std::function<void()>>* postSyncPreDraw;
 inline std::vector<std::function<void()>>* render;
 inline std::vector<std::function<void()>>* draw2D;
 inline std::vector<std::function<void()>>* postDraw;
-inline std::vector<std::function<bool(Key, KeyModifier, KeyAction)>>* keyboard;
-inline std::vector<std::function<bool(unsigned int, KeyModifier)>>* character;
-inline std::vector<std::function<bool(MouseButton, MouseAction, KeyModifier)>>*
+inline std::vector<std::function<bool(Key, KeyModifier, KeyAction, bool)>>* keyboard;
```
alexanderbock, line 45: Should we declare a new `Boolean`  `GuiWindow` to use here instead?

modules/cefwebgui/src/guikeyboardhandler.cpp
```diff
@@ -32,10 +32,10 @@ GUIKeyboardHandler::GUIKeyboardHandler() {
     _keyConsumed = false;
 
     global::callback::keyboard->emplace_back(
-        [&](Key, KeyModifier, KeyAction) -> bool {
+        [&](Key, KeyModifier, KeyAction, bool isGuiWindow) -> bool {
             const bool previous = _keyConsumed;
             _keyConsumed = false;
-            return previous;
+            return !isGuiWindow ? false : previous;
```
alexanderbock, line 38: -> `isGuiWindow ? previous : false`?   One less negation to worry about

modules/imgui/imguimodule.cpp
```diff
@@ -153,31 +153,45 @@ ImGUIModule::ImGUIModule()
     });
 
     global::callback::keyboard->emplace_back(
-        [&](Key key, KeyModifier mod, KeyAction action) -> bool {
+        [&](Key key, KeyModifier mod, KeyAction action, bool isGuiWindow) -> bool {
             ZoneScopedN("ImGUI")
 
+            if (!isGuiWindow) {
```
alexanderbock, line 159: ->
```cpp
if (!isGuiWindow || !_isEnabled) {
    return false:
}
else {
    return keyCallback(key, mod, action);
}
```

modules/webbrowser/src/eventhandler.cpp
```diff
@@ -149,48 +149,51 @@ namespace {
 namespace openspace {
 
 void EventHandler::initialize() {
-    global::callback::character->emplace_back(
-        [this](unsigned int charCode, KeyModifier mod) -> bool {
-            if (_browserInstance) {
+    global::callback::character->emplace(global::callback::character->begin(),
```
alexanderbock, line 152: The begin would be on a new line;  same for the others below

apps/OpenSpace/main.cpp
WeirdRubberDuck: No? Not if I interpret our coding style document correctly. We align the arguments on the new lines wiht the `(` bracket and only add new lines when we reach the column limit. At least according to the second example here:

image

apps/OpenSpace/main.cpp
WeirdRubberDuck: see comment above

include/openspace/engine/globalscallbacks.h
WeirdRubberDuck: I like that idea. Would make the meaning of the bool more apparent 

Like this: `using IsGuiWindow = bool; ` ? In this file or somewhere else?

modules/imgui/imguimodule.cpp
WeirdRubberDuck: This might be a personal preference, but do we really need the else statement here? To me it feels implied because of the return statement in the first if

But again, personal preference and I'm fine with adding the else if we think that makes it clearer

include/openspace/engine/globalscallbacks.h
WeirdRubberDuck: I guess we might also want to use it in the cpp file, and in the 

```
using F = std::function<bool (unsigned int, KeyModifier, bool)>;
```
declarations in the openspaceengine.cpp file

apps/OpenSpace/main.cpp
alexanderbock: Oops... yes, you are right!  I think I'll have to fix some of my misindentations then!

include/openspace/engine/globalscallbacks.h
alexanderbock: In ghoul/misc/boolean.h   there is a `Boolean` define that does that type safe.  So:

`Boolean(IsGuiWindow);` in this file and then use `IsGuiWindow::Yes`  and `IsGuiWindow::No` when using it.  It autoconverts into a bool, too.

modules/imgui/imguimodule.cpp
alexanderbock: Yea, I think the else statement is totally optional;  the comment was more about moving the `_isEnabled` part into the first `if`.
Alternatively of course:
`return isGuiWindow && _isEnabled ? keyCallback(key, mod, action) : false;`

include/openspace/engine/globalscallbacks.h
WeirdRubberDuck: Cool, thanks!

alexanderbock (approved):

PR #2146, opened 2022-06-14
Remove debris assets that are no longer supported by celestrak

GPayne:


alexanderbock:
I think the data is still be available but they made the old txt version non-functional?
ASAT:  http://www.celestrak.com/NORAD/elements/gp.php?INTDES=2019-006
Breeze-M:  http://www.celestrak.com/NORAD/elements/gp.php?INTDES=2012-044

GPayne:
Yes, you're right about the new format fixing these links. Even though all the others still work, this might change. So I updated all satellite assets to use the new request format.

PR #2158, opened 2022-06-21
Add gamma correction property to Screen Space Renderables

ylvaselling:
...and set default value gamma = 2.2 for sky browser display copies

### DIFF
--- a/modules/base/shaders/screenspace_fs.glsl
+++ b/modules/base/shaders/screenspace_fs.glsl
@@ -32,6 +32,7 @@ uniform sampler2D tex;
 uniform vec3 color = vec3(1.0);
 uniform float opacity = 1.0;
 uniform vec4 backgroundColor = vec4(0.0);
+uniform float gamma = 1.0;
 
 Fragment getFragment() {
   Fragment frag;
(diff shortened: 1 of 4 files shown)


modules/base/shaders/screenspace_fs.glsl
```diff
@@ -44,5 +45,6 @@ Fragment getFragment() {
   }
 
   frag.depth = vs_depth;
+  frag.color.rgb = pow(frag.color.rgb, vec3(1.0/(gamma + 0.000001)));
```
alexanderbock, line 48: I think it might be better to put that limitation onto the gamma FloatProperty,  instead of having a minimum value of 0, you can make it have a minimum value of 0.000001 instead

alexanderbock (approved):

src/rendering/screenspacerenderable.cpp
```diff
@@ -190,6 +196,9 @@ namespace {
         // [[codegen::verbatim(ScaleInfo.description)]]
         std::optional<float> scale;
 
+        // [[codegen::verbatim(UseRadiusAzimuthElevationInfo.description)]]
```
Roxeena, line 199: This refers to the `UseRadiusAzimuthElevationInfo` description and not the `GammaInfo` description

Roxeena (commented):
👍

PR #2165, opened 2022-06-28
Add enable/disable property for TimelineRotation

### LINKED ISSUE #1607: TimelineTranslation and Rotation should have an option to disable interpolation
Right now, when using a TimelineTranslation and a TimelineRotation, the values between the keyframes are automatically interpolated.  It would be nice to be able to disable that

Roxeena:
Fixes #1607


modules/base/rotation/timelinerotation.h
```diff
@@ -25,6 +25,7 @@
 #ifndef __OPENSPACE_MODULE_BASE___TIMELINEROTATION___H__
 #define __OPENSPACE_MODULE_BASE___TIMELINEROTATION___H__
 
+#include <openspace/properties/scalar/boolproperty.h>
```
alexanderbock, line 28: Order should be:
```cpp
#include <openspace/scene/rotation.h>

#include <openspace/properties/scalar/boolproperty.h>
#include <openspace/util/timeline.h>
```

Since the rotation is the base class, it comes first in the list and separated by an empty line

alexanderbock (approved):

PR #2166, opened 2022-06-28
Fix filename when downloading multiple files in urlsync

k2d222:
Before it would download all the files to the same location based on the first file's name in the list.


alexanderbock (approved):
Looks great!  Merge whenever you like

Issue #1778, opened 2021-10-30
Implement additional TLE replacements


alexanderbock:
See https://celestrak.com/NORAD/elements/gp-index.php?FORMAT=kvn

Based on this talk:  https://www.youtube.com/watch?v=XxC3XFEYy98

alexanderbock:
http://celestrak.org/NORAD/documentation/gp-data-formats.php

PR #2185, opened 2022-07-16
Feature/satellites

alexanderbock:
Handles #1778 by implementing support for OMM KVN style general pertubation files.

In the process of unifying the loading, it was also possible to pull in the JPL SBDB CSV-based file format into the same mechansism and ultimately it turned out that the RenderableSatellites and RenderableSmallBodies were the same class after this simplification and could be rolled into the RenderableOrbitalKepler class

 - Introduces a new file `kepler.cpp` that handles the loading of kepler elements from a variety of data file formats
 - Removes RenderableSatellites and RenderableSmallbodies and move their functionality into RenderableOrbitalKepler
 - Fix some spell mistakes


micahnyc:
I have this branch from 2019 which is relevant: https://github.com/OpenSpace/OpenSpace/compare/feature/tle-bodymass

Right now this stuff is hard coded to earth, feels like this would have been the time to undo that.

alexanderbock:
Yea, we can rebase that branch on this once we merge it :+1:

modules/space/kepler.cpp
```diff
@@ -0,0 +1,637 @@
...
+        }
+        else if (parts[0] == "RA_OF_ASC_NODE") {
+            current->ascendingNode = std::stof(parts[1]);
+        }
+        else if (parts[0] == "ARG_OF_PERICENTER") {
+            current->argumentOfPeriapsis = std::stof(parts[1]);
+        }
+        else if (parts[0] == "MEAN_ANOMALY") {
+            current->meanAnomaly = std::stof(parts[1]);
+        }
```
GPayne, line 556: Do we need an else case to log an error if `parts[0]` doesn't match any of the above?

modules/space/kepler.cpp
```diff
@@ -0,0 +1,637 @@
...
+        // 1. Read the year value
+        // 2. Calculate the number of days since the beginning of the year
+        // 3. Convert the number of days to a number of seconds
+        // 4. Get the number of leap seconds since January 1st, 2000 and remove them
+        // 5. Add the hh:mm:ss component
+        // 6. Adjust for the fact the epoch starts on 1st January at 12:00:00, not
+        // midnight
+
+        // 1
+        int year = std::atoi(epochString.substr(0, 4).c_str());
```
GPayne, line 296: There should probably be some format checking of `epochString` before using `std::atoi`, since it can return a zero value (or undefined behavior possibly meaning any int value) if the string contains unexpected characters. The check would verify the YYYY-MM-DD.... expected form.
This comment applies to other uses of `std::atoi` here as well.

modules/space/kepler.cpp
```diff
@@ -0,0 +1,637 @@
...
+            ));
+        }
+
+
+        // 3
+        using namespace std::chrono;
+        const int SecondsPerDay = static_cast<int>(seconds(hours(24)).count());
+        const double nSecondsSince2000 = (daysSince2000 + nDays) * SecondsPerDay;
+
+
```
GPayne, line 325: Excess spaces

modules/space/kepler.h
```diff
@@ -0,0 +1,108 @@
...
+#ifndef __OPENSPACE_MODULE_SPACE___KEPLER___H__
+#define __OPENSPACE_MODULE_SPACE___KEPLER___H__
+
+#include <filesystem>
+#include <string>
+#include <vector>
+
+namespace openspace::kepler {
+
+struct SatelliteParameters {
```
GPayne, line 34: I like how the code is now combined for earth satellites, asteroids, comets, etc. However the `SatelliteParameters` name might be confusing. It's true that the definition of 'satellite' is not bound to earth satellites, but it's common to use that term only in earth context.

modules/space/kepler.h
```diff
@@ -0,0 +1,108 @@
...
+    std::string name;
+    // International Designator (launch year + launch number of the year + piece of launch
```
GPayne, line 36: Add to the comment that `id` is only for earth satellites.
Another possibility that relates to the previous comment about the definition of 'satellite' is to make a special case:
`struct EarthSatelliteParameters : SatelliteParameters` which contains `std::string id;`

modules/space/kepler.cpp
alexanderbock: There are a bunch of optional keys that are defined in the OMM format that we are ignoring, so I think it would get too chatty if we print a message every time

modules/space/kepler.cpp
alexanderbock: Going to change that in the next PR.  I'm flirting with https://github.com/eliaskosunen/scnlib right now and would like to use that instead of manually parsing and checking for errors

modules/space/kepler.h
alexanderbock: Renamed to just `Parameters` since it already lives in the `kepler` namespace and is referenced as `kepler::Parameters` everywhere

Issue #1279, opened 2020-08-17
Adding a parameter to the `require` parameters that toggles enabled/disabled


alexanderbock:
`asset.require("foobar", true)`
for an asset that is enabled (whatever that means for an asset)

alexanderbock:
Closed in https://github.com/OpenSpace/OpenSpace/pull/2187

PR #2187, opened 2022-07-20
Add the ability to pass a boolean value into the require function that gets passed into the loaded asset as a `enabled` property

alexanderbock:
With this PR, the `asset.require` function now takes a second optional parameter which is a boolean value that defaults to `true`.   That value is accessible as a new variable `asset.enabled` inside the asset.  Loading the same asset multiple times with different parameters will cause a warning and only the value of the first `asset.require` matters.  For example:

```lua
asset.require("foo", false)
asset.require("foo", true)
```
will lead to a warning and the `asset.enabled` in "foo" will be `false`

```lua
asset.require("foo", false)
asset.require("foo", false)
```
will not result in a warning

```lua
asset.require("foo")
asset.require("foo", true)
```
currently results in a warning even though it will behave the same way out of an abundance of caution.

But how about?
```lua
asset.require("foo", true)
asset.require("foo")
```

Deals with
https://github.com/OpenSpace/OpenSpace/issues/1279


alexanderbock:
@micahnyc 
Should update the https://github.com/OpenSpace/scripts/tree/master/treks-to-openspace
and add
```lua
    Enabled = asset.enabled
```
into the generation script

PR #2188, opened 2022-07-20
Fade out renderables in atmosphere

ylvaselling:
The fading happens
* when you enter the atmosphere
* when you are in the sunny area of a globe

Properties for each atmosphere are
* The angle where the sunset starts
* The angle where the sunset ends
* The percentage of the atmosphere height where the fading should happen.

Before merge, there is one edge case that should be accounted for. It is if the camera is located in an atmosphere and then uses the flyToGeo script to fly somewhere else, the renderables are not faded in as the camera has not transitet the "fading section". This could be resolved by setting the fading coefficient to 1 in the beginning of each render loop


include/openspace/camera/camera.h
```diff
@@ -93,6 +94,7 @@ class Camera {
     float sinMaxFov() const;
     SceneGraphNode* parent() const;
     float scaling() const;
+    float atmosphereDimming() const;
```
alexanderbock, line 97: Maybe `atmosphereDimmingFactor`?  Would be nice to have a one-line description of what it is, too

modules/atmosphere/rendering/renderableatmosphere.cpp
```diff
@@ -132,6 +136,25 @@ namespace {
         "Enable/Disables hard shadows through the atmosphere"
     };
 
+    constexpr openspace::properties::Property::PropertyInfo AtmosphereDimmingHeightInfo ={
+        "AtmosphereDimmingHeight",
+        "Atmosphere Dimming Height",
+        "Percentage of the atmosphere where other objects, such as the stars, are faded."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo SunsetAngleStartInfo = {
```
alexanderbock, line 145: Would there be any drawback from combining this into one Vec2Property with the `Property::ViewOption::MinMax` view?

modules/atmosphere/rendering/renderableatmosphere.cpp
```diff
@@ -406,6 +454,53 @@ void RenderableAtmosphere::update(const UpdateData& data) {
     glm::dmat4 modelTransform = computeModelTransformMatrix(data.modelTransform);
     _deferredcaster->setModelTransform(modelTransform);
     _deferredcaster->update(data);
+
+    // Calculate atmosphere dimming coefficient
+    // Calculate if the camera is in the atmosphere and if it is in the fading region
+    float atmosphereDimming = 1.0;
```
alexanderbock, line 460: -> 1.f  since it is a `float`

src/rendering/renderable.cpp
```diff
@@ -292,7 +308,11 @@ void Renderable::registerUpdateRenderBinFromOpacity() {
 }
 
 float Renderable::opacity() const {
-    return _opacity * _fade;
+    // Rendering should depend on if camera is in the atmosphere and if camera is at the 
+    // dark part of the globe
+    float dimming = global::navigationHandler->camera()->atmosphereDimming();
+    dimming = _dimInAtmosphere ? dimming : 1.0;
```
alexanderbock, line 314: could be:
```cpp
float dimming = _dimInAtmosphere ? global::navigationHandler->camera()->atmosphereDimming() : 1.f;
```

that way we don't have to call atmosphereDimming for most items at all.
Alternatively:
```cpp
return _dimInAtmosphere ?
    _opacity * _fade * global::navigationHandler->camera()->atmosphereDimming() :
    _opacity * _fade;
```

alexanderbock (approved):

alexanderbock:
Adding
```cpp
    _camera->setAtmosphereDimming(1.f);
```

In the Scene::update in line 338 works for me to fix the issue of going into an atmosphere and teleporting out with `goToGeo`

modules/atmosphere/rendering/renderableatmosphere.cpp
ylvaselling: No, I thought of that but didn't find the property - good call!

PR #2201, opened 2022-08-04
SkyBrowser Hash Handling

alexanderbock:
This PR does some minor cleanup of the SkyBrowser module but more importantly adds the ability to compare a local hash to a remote hash to perform automatic updating of files.  It also moves the wwtimagedata folder from the modules/skybrowser into the sync folder

### DIFF
--- a/modules/skybrowser/src/utility.cpp
+++ b/modules/skybrowser/src/utility.cpp
@@ -170,7 +170,7 @@ bool isCoordinateInView(const glm::dvec3& equatorial) {
     double r = windowRatio();
 
     bool isCoordInView =
-        abs(coordsScreen.x) < r && abs(coordsScreen.y) < 1.f && coordsScreen.z < 0;
+        abs(coordsScreen.x) < r && abs(coordsScreen.y) < 1.f && coordsScreen.z < 0.f;
     return isCoordInView;
 }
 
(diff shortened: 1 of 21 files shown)


modules/skybrowser/src/utility.cpp
```diff
@@ -150,11 +167,10 @@ bool isCoordinateInView(const glm::dvec3& equatorial) {
...
     glm::dvec3 localCamera = equatorialToLocalCamera(equatorial);
     glm::dvec3 coordsScreen = localCameraToScreenSpace3d(localCamera);
-    double r = static_cast<float>(windowRatio());
-
-    bool isCoordInView = abs(coordsScreen.x) < r && abs(coordsScreen.y) < 1.f &&
-                         coordsScreen.z < 0;
+    double r = windowRatio();
 
+    bool isCoordInView =
+        abs(coordsScreen.x) < r && abs(coordsScreen.y) < 1.f && coordsScreen.z < 0;
```
WeirdRubberDuck, line 173: `coordsScreen.z < 0.f` ? Don't remember if we do an exception for 0

WeirdRubberDuck (approved):
Looks good to me and runs without problems

Great update!

ylvaselling (approved):
Tried it, works well!

modules/skybrowser/src/utility.cpp
alexanderbock: No, we don't... you're right

Issue #2093, opened 2022-05-09
Dragging on the sky browser in the GUI works poorly with fisheye projection


ylvaselling:
The dragging on the sky browser in the gui assumes that the projection is planar. Therefore, the dragging while using fisheye projection works poorly. The image doesn't move in the direction you would expect with a "drag" interaction.

The solution should not rely on screen space coordinates for the dragging, and instead use only the GUI pixel coordinates and the Right Ascension, Declination coordinates for the conversion.

PR #2204, opened 2022-08-05
Change how the calculation of the fine tuning is done

ylvaselling:
Solves #2093 by changing the calculation of the fine tuning by not assuming that there is a flat projection.


alexanderbock (approved):

alexanderbock (approved):

WeirdRubberDuck:
Why closing without merging?

ylvaselling:
I found an issue with it that was so severe I will have to rethink the entire solution

PR #2206, opened 2022-08-12
Add rounder border to sky target and browser

ylvaselling:
Add functionality to have a rounded border on sky browsers and targets. This requires an update of the code hosted on the server and it is a breaking change. Probably the code needs to be put on a server and the links need to be updated before testing is possible.

Use with these 
* https://github.com/ylvaselling/sky_browser/pull/1
* https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/104

You need to run the sky_browser repo locally and change the URLs to which localhost you are running it on

image


modules/skybrowser/shaders/target_fs.glsl
```diff
@@ -66,6 +66,13 @@ float createCrosshair(in float linewidth, in float ratio, in vec2 coord) {
   return crosshairHorizontal + crosshairVertical;
 }
 
+float roundedRectangle(vec2 coord, vec2 size, vec4 radius) {
```
WeirdRubberDuck, line 69: This function is a bit magic to me (mainly the 4-dim radius and the fact that the function overwrites the input values that makes it difficult to understand)

Maybe add a short description of how it works or some supporting comment in the function? Maybe unnecessary, I don't know. Open to opinions

modules/skybrowser/shaders/target_fs.glsl
```diff
@@ -78,13 +85,34 @@ Fragment getFragment() {
...
-    float outerEdge = createFilledRectangle(width, height, vs_st);
-    float innerEdge = createFilledRectangle(width-lineWidthY, height-lineWidthX, vs_st);
-    rectangle = outerEdge - innerEdge;
+    vec2 size = vec2(width, height);
+
+    // The radius of the corners (in pixels) clockwise starting in the top left
+    vec4 radius = vec4(clamp(borderRadius * 0.5 * min(height, width), 0.0, 0.5 * min(height, width)));
+
+    // Calculate distance to edge
+    float distance = roundedRectangle(vs_st.xy - vec2(0.5), size / 2.0f, radius);
```
WeirdRubberDuck, line 98: `float distance = roundedRectangle(vs_st.xy - vec2(0.5), 0.5 * size, radius);` ?

modules/skybrowser/src/wwtcommunicator.cpp
```diff
@@ -119,6 +123,12 @@ void WwtCommunicator::setBorderColor(glm::ivec3 color) {
     _borderColorIsDirty = true;
 }
 
+void WwtCommunicator::setBorderRadius(double radius) {
+    _borderRadius = radius;
+    std::string scr = "setBorderRadius(" + std::to_string(radius) + ");";
```
WeirdRubberDuck, line 128: `fmt::format("setBorderRadius({});", radius)`

ylvaselling:
> Any updates on how to try this? :)

No, we need to put some code on the server first :/

modules/skybrowser/shaders/target_fs.glsl
ylvaselling: Yeah tbh I'm not really sure how it works either, I found the formula online hehe. I removed the 4 dim radius since we don't use that anyways and wrote a comment. Can't see that the function overwrites the input values though

PR #2267, opened 2022-10-12
Fix for issue of scaling doesn't update border radius

ylvaselling:
This PR adds a timer to the scale onChange function that allows the browser to load for 25 frame updates before applying the border radius change.


ylvaselling:
Yes, the performance hit has been a problem for a while and will even make OpenSpace crash sometimes. Will make an issue out if it

WeirdRubberDuck (commented):
Seems to work well IMO. I tried fiddling quite a bit with the scale slider and found it very difficult to reproduce the behavior in @Roxeena's video (not sure I even managed to). Either way, it's not a huge problem since the user can just use the slider to change the value again, according to me

Also, maybe we should set the max for the scale to 1.5 or something? As a temporary solution to avoid the freezing from happening when just using a slider. (Maybe more suitable to discuss in the other issue). THe fact that the application freezes completely is very annoying

PR #2287, opened 2022-10-30
Add info tooltips to profile selector dropdown in launcher

### LINKED ISSUE #2207: Expose profile descriptions in the Launcher GUI
It would be nice to expose the meta description of the profiles in the UI somewhere, to aid the users when they browse the list of profiles. 

Maybe just show when hovering the item in the list ?

WeirdRubberDuck:
closes #2207

Note that this is a minimal and pretty naive solution. The profile loading is done on the fly and takes a little while so the tooltips are not immediately available when opening the launcher. Open to suggestions on how to improve it.

Some styling might also be nice

### DIFF
--- a/apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
+++ b/apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
@@ -434,7 +434,7 @@ void LauncherWindow::populateProfilesList(std::string preset) {
             const std::optional<std::string>& d = (*p).meta.value().description;
             if (d.has_value()) {
                 // Tooltip has to be 'rich text' to linebreak properly
-                QString tooltip = QString::fromStdString(fmt::format("<html><p>{}</p></html>", *d));
+                QString tooltip = QString::fromStdString(fmt::format("<p>{}</p>", *d));
                 _profileBox->setItemData(idx, tooltip, Qt::ToolTipRole);
             }
         }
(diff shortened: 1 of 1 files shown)


ylvaselling:
Very nice with popups! Unfortunately, I think it might be a bit hard to understand that you can hover over an item to get the popup since you need to hold the cursor there for a little bit. Maybe some info icon or question mark could help if the loading time is hard to do something about?

WeirdRubberDuck:
> Very nice with popups! Unfortunately, I think it might be a bit hard to understand that you can hover over an item to get the popup since you need to hold the cursor there for a little bit. Maybe some info icon or question mark could help if the loading time is hard to do something about?

I had the same thought, but found no obvious way to achieve that in Qt from the limited time I spent investigating this. Could do some more digging

Roxeena:
Looks good 👍 I agree that it would be nice to have some sort of hint that you can hover over the items to see more info. However, that should not be something that stops this from being merged if it ends up being a lot of work.

WeirdRubberDuck:
Unfortunately, there does not seem to be a really good way of adding any visual hint for the tooltips, what I could find at least (investigated using rich text in the QListItem, but that's not really supported). And trying to make it work would lead to a hacky and quite ugly solution, I believe. 

Hopefully, our users manage to discover the tooltips anyway somehow

alexanderbock (approved):

PR #2296, opened 2022-11-01
Make dragging in skybrowser work in fisheye projection with new mode

ylvaselling:
This PR features a new way to calculate the dragging and a new property in the sky target which sets whether the roll should be applied automatically or not. Should be used with the branch in the GUI [issue/2093](https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/115). The property for the target should be added to the sky browser panel but that is currently waiting for the sky browser optimization PR


modules/skybrowser/src/renderableskytarget.cpp
```diff
@@ -151,6 +164,26 @@ glm::ivec3 RenderableSkyTarget::borderColor() const {
     return _borderColor;
 }
 
+glm::dvec3 RenderableSkyTarget::rightVector() {
```
alexanderbock, line 167: The function can be marked `const`

modules/skybrowser/src/renderableskytarget.cpp
```diff
@@ -151,6 +164,26 @@ glm::ivec3 RenderableSkyTarget::borderColor() const {
...
+    double scaling = (_verticalFov / 70)* static_cast<double>(_size.value());
+    return scaling * _rightVector;
+}
+
+glm::dvec3 RenderableSkyTarget::upVector() {
```
alexanderbock, line 172: The function can be marked `const`

alexanderbock (approved):

PR #2304, opened 2022-11-10
Various precision fixes for camera paths

### LINKED ISSUE #2212: Flying to "1 AU Size Comparison Circle" freezes application
To reproduce: 
1. Add exoplanet system "11 Umi"
2. Find "1 AU Size Comparison Circle" in Navigation list and click the "fly to" button

One beautiful day, the camera path system is going to be free from bugs...

I believe the actual problem might be flying using the avoid collision curve to objects without bounding spheres
(Note: the comparison circles should get a bounding sphere computed!)

WeirdRubberDuck:
A collection of fixes for precision-related issues, all discovered when trying to address #2212. This is one of the most troublesome cases for the camera paths; Flying to a small (close to zero-bounding sphere) object far out in space. The position coordinates are so large that we can't always do the computations for the paths with sufficient precision. This PR adds a few checks for that in problematic situations

Also note that the RadialGrids do not get a valid bounding sphere computed until the first time they are enabled, which was why the problem appeared for the 1 AU size rings for the exoplanet systems specifically. Everything works better if the objects have correctly sized bounding spheres. 

Closes #2212, and probably also solves the "shaky" movements that users have reported when flying to ISS sometimes

### DIFF
--- a/src/navigation/pathnavigator.cpp
+++ b/src/navigation/pathnavigator.cpp
@@ -537,18 +537,25 @@ SceneGraphNode* PathNavigator::findNodeNearTarget(const SceneGraphNode* node) {
 }
 
 void PathNavigator::removeRollRotation(CameraPose& pose, double deltaTime) {
-    const glm::dvec3 anchorPos = anchor()->worldPosition();
+    // The actual position for the camera does not really matter. Use the origin,
+    // to avoid precision problems when we have large values for the position
+    const glm::dvec3 cameraPos = glm::dvec3(0.0);
     const glm::dvec3 cameraDir = glm::normalize(
         pose.rotation * Camera::ViewDirectionCameraSpace
     );
-    const double anchorToPosDistance = glm::distance(anchorPos, pose.position);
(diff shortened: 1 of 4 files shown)


src/navigation/pathnavigator.cpp
```diff
@@ -537,18 +537,25 @@ SceneGraphNode* PathNavigator::findNodeNearTarget(const SceneGraphNode* node) {
...
+    // to avoid precision problems when we have large values for the position
+    const glm::dvec3 cameraPos = glm::dvec3(0.0);
     const glm::dvec3 cameraDir = glm::normalize(
         pose.rotation * Camera::ViewDirectionCameraSpace
     );
-    const double anchorToPosDistance = glm::distance(anchorPos, pose.position);
-    const double notTooCloseDistance = deltaTime * anchorToPosDistance;
-    glm::dvec3 lookAtPos = pose.position + notTooCloseDistance * cameraDir;
+
+    // The actual distance does not really matter either. Just have to far
```
WeirdRubberDuck, line 547: Just have to be

alexanderbock (approved):

PR #2359, opened 2022-11-30
Issue/2313: Ensure all nodes in a cluster loads the image collection in the sky browser

ylvaselling:
Should fix the issue of sky browser images not loading on the clients in a cluster.

1) Opens up the port for the javascript API on all nodes 
2) Allows lua scripts to be executed only on a client
3) Checks so that the initialization process is executed on each node separately and is not propagated forward to other nodes from master

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203483820933838


modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -40,6 +40,12 @@ namespace {
     constexpr std::string_view _loggerCat = "SkyBrowserModule";
 
 
+    bool browserBelongsToCurrentNode(std::string identifier) {
+        // Convert the last char to an int
+        int nodeId = static_cast<int>(identifier[identifier.length() - 1] - '0');
```
alexanderbock, line 45: This could be dangerous when the number of nodes is bigger than 9

How about:
```cpp
std::string str;
int nodeId;
auto res = scn::scan("{}{}", str, nodeId);
if (!res) {
  throw ghoul::RuntimeError(...);
}
else {
```

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -169,6 +175,9 @@ namespace {
 [[codegen::luawrap]] void loadImagesToWWT(std::string identifier) {
     using namespace openspace;
 
+    if (!browserBelongsToCurrentNode(identifier))
```
alexanderbock, line 178: `return` should have `{` and `}` surrounding it

modules/skybrowser/skybrowsermodule_lua.inl
ylvaselling: I agree that there should be a better solution. Unfortunately, the Identifier of the browser also ends with a number, which could be multiple digits. I suppose I could change the Identifier to something like SkyBrowser0_ and then the node number comes after that? So then I could split it where the _ is

modules/skybrowser/src/screenspaceskybrowser.cpp
```diff
@@ -183,7 +183,7 @@ bool ScreenSpaceSkyBrowser::isInitialized() const {
 }
 
 void ScreenSpaceSkyBrowser::setIdInBrowser() const {
-    WwtCommunicator::setIdInBrowser(identifier());
+    WwtCommunicator::setIdInBrowser(identifier() + std::to_string(global::windowDelegate->currentNode()));
```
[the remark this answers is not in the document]
ylvaselling: Nice, I should have read all the comments first 👍

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -35,10 +35,35 @@
...
 #include <ghoul/logging/logmanager.h>
+#include <scn/scn.h>
 
 namespace {
-    constexpr std::string_view _loggerCat = "SkyBrowserModule";
+constexpr std::string_view _loggerCat = "SkyBrowserModule";
+
+bool browserBelongsToCurrentNode(std::string identifier) {
+    std::string delimiter = "_";
+    size_t found = identifier.find(delimiter);
```
alexanderbock, line 45: Just a tiny bit faster if `size_t found = identifier.find('-');` rather than "-" since '-' is guaranteed to be just a single character

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -35,10 +35,35 @@
...
+    std::string errorMessage = "The Sky Browser encountered a problem when it tried to "
+        "initialize the browser";
+    if (found == identifier.size()) {
```
alexanderbock, line 48: Should be `found == std::string::npos` for the correct check

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -35,10 +35,35 @@
...
+        throw ghoul::RuntimeError(errorMessage);
+    }
+    else {
+        std::string res = identifier.substr(found + 1, identifier.size());
+        if (res == "") {
```
alexanderbock, line 53: -> `if (res.empty())`

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -35,10 +35,35 @@
...
+            throw ghoul::RuntimeError(errorMessage);
+        }
+        // Convert the last char to an int
+        int nodeId = std::stoi(res);
+        return nodeId == openspace::global::windowDelegate->currentNode();
+    }
+}
 
+std::string getPrunedIdentifer(std::string identifier) {
```
alexanderbock, line 62: For our coding style -> `prunedIdentifier`.  We only use the "get" prefix if the return value is passed in as a reference

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -35,10 +35,35 @@
...
+    std::string delimiter = "_";
```
alexanderbock, line 63: See above re '-'

modules/skybrowser/src/screenspaceskybrowser.cpp
```diff
@@ -183,7 +183,8 @@ bool ScreenSpaceSkyBrowser::isInitialized() const {
 }
 
 void ScreenSpaceSkyBrowser::setIdInBrowser() const {
-    WwtCommunicator::setIdInBrowser(identifier());
+    WwtCommunicator::setIdInBrowser(identifier() + "_" +
```
alexanderbock, line 186: Could be shorter and easier to read:
```
WwtCommunicator::setIdInBrowser(fmt::format("{}_{}", identifier(), global::windowDelegate->currentNode())
```

alexanderbock (commented):
Looks good to me, have to still try it on our dome

alexanderbock:
Unfortunately, this didn't fix the issue for me :(

When creating an instance and making a display copy for the dome, the dome version doesn't get initialized properly and also doesn't get any commands later.
Here are two examples;  `master` node is on the right, dome is on the left
Photo 28 01 23, 20 24 23

Photo 28 01 23, 20 24 11


Also got an error message on the master node:
image

Likewise on the client:
image

ylvaselling:
Oops, I set the delimiter to the wrong character after addressing the comments on the PR. It should work now

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -503,7 +503,7 @@ std::string prunedIdentifier(std::string identifier) {
     glm::vec3 positionTarget = glm::vec3(0.9f, 0.4f, -2.1f);
     glm::dvec3 galacticTarget = skybrowser::localCameraToGalactic(positionTarget);
     if (glm::any(glm::isnan(galacticTarget))) {
-        galacticTarget = { skybrowser::CelestialSphereRadius, 0.0, 0.0 };
+        galacticTarget = glm::dvec3 (0.0, 0.0, skybrowser::CelestialSphereRadius);
```
alexanderbock, line 506: No need for the extra space between `dvec3` and the `(`

alexanderbock (approved):

Issue #2362, opened 2022-12-01
Runtime Exception thrown after adding a key binding in the profile editor


ylvaselling:
When adding a keybind, an error popup showed.
![image](https://user-images.githubusercontent.com/24265984/205156543-200ac662-1dc3-4a7f-9afc-aef0e40a276e.png)

This is the keybind I added:
![image](https://user-images.githubusercontent.com/24265984/205160294-b9da0599-ae6d-4c96-9fbd-3b1ba6293e9c.png)

Expected behavior is that there is no error popup

WeirdRubberDuck:
Odd, I can't reproduce this 🤔 Does it happen every time you try to create a keybinding?

WeirdRubberDuck:
Odd, I can't reproduce this 🤔 Does it happen every time you try to create a keybinding? At which point in the creation process?

GPayne:
I didn't get exactly the same error message, but I made it crash also. I set everything up exactly as you did in your screenshot. I clicked the Actions Save button, but purposely skipped the Keybindings Save button and just clicked the Save for the whole dialog box (so that the action but not the key shortcut is saved).
You might have accidentally done the same, or similar, thing.

As I recall, the dialog box's Save button used to be grayed-out while things were being edited above. That functionality should be restored in order to prevent this from happening.

PR #2401, opened 2022-12-14
Fix to prevent invalid entries for action or keybinding  in profile edit

GPayne:
In order to save entries in the "Actions and Keybindings" dialog of the profile editor, both the Actions and Keybindings sections must be in a valid state. This means that they must not have an incomplete entry being changed or added (clicking Save in the section will transition to a valid state).


alexanderbock (approved):

alexanderbock:
Closes #2362

PR #2414, opened 2022-12-16
Change sky browser to use id's instead of indices to get the images

ylvaselling:
WebGui Frontend PR: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/126

The WebGUI-Frontend still uses indices for the simple reason that ghoul didn't have an option for a map. If people think this is very ugly I could implement the map in ghoul. 

Note: the order of the images has changed in the "All images". It used to be sorted by name - now it isn't anymore. Do we think that is something we want to keep?


modules/skybrowser/skybrowsermodule.cpp
```diff
@@ -318,9 +318,12 @@ void SkyBrowserModule::setHoverCircle(SceneGraphNode* circle) {
...
 
-void SkyBrowserModule::moveHoverCircle(int i, bool useScript) {
-    const ImageData& image = _dataHandler.image(i);
-
+void SkyBrowserModule::moveHoverCircle(const std::string& imageUrl, bool useScript) {
+    std::optional<const ImageData> found = _dataHandler.image(imageUrl);
+    if (!found.has_value()) {
+        return;
+    }
+    const ImageData image = found.value();
```
alexanderbock, line 326: -> `const ImageData image = *found;`   since we are using that in many other places (mostly the codegen:ed Parameter structs

modules/skybrowser/skybrowsermodule_lua.inl
```diff
@@ -296,10 +305,7 @@ namespace {
 
     // Create Lua table to send to the GUI
     ghoul::Dictionary list;
-
-    for (int i = 0; i < module->nLoadedImages(); i++) {
-        const ImageData& img = module->wwtDataHandler().image(i);
-
+    for (auto const& [id, img] : module->wwtDataHandler().images()) {
```
alexanderbock, line 308: -> `const auto&`

modules/skybrowser/src/targetbrowserpair.cpp
```diff
@@ -151,13 +151,21 @@ double TargetBrowserPair::verticalFov() const {
...
 
 ghoul::Dictionary TargetBrowserPair::dataAsDictionary() const {
     glm::dvec2 spherical = targetDirectionEquatorial();
     glm::dvec3 cartesian = skybrowser::sphericalToCartesian(spherical);
+    SkyBrowserModule* module = global::moduleEngine->module<SkyBrowserModule>();
+    std::vector<std::string> selectedImagesIndices;
+    
+    for (const std::string& imageUrl : selectedImages()) {
+        selectedImagesIndices.push_back(
+            module->wwtDataHandler().image(imageUrl).value().identifier
```
alexanderbock, line 166: Can do:
`module->wwtDataHandler().image(imageUrl)->identifier`

maybe also a:
`ghoul_assert(module->wwtDataHandler().image(imageUrl).has_value(), "...");`
above juuust to be sure that the image actually exists?

modules/skybrowser/src/wwtdatahandler.cpp
```diff
@@ -330,23 +331,23 @@ void WwtDataHandler::loadImages(const std::string& root,
         }
     }
 
-    // Sort images in alphabetical order
```
alexanderbock, line 333: Why no more sorting?  I think if it doesn't cost too much, timewise it'd be nicer to keep it

alexanderbock (approved):

modules/skybrowser/src/wwtdatahandler.cpp
ylvaselling: I can't remember 😅 I'll implement the sort

WeirdRubberDuck (approved):

PR #2426, opened 2023-01-04
Fix bug that channels in cluster don't set the field of view to the same value after animating

ylvaselling:


alexanderbock:
Looks good to me;  I didn't see the error on my end before and can't see it now

alexanderbock (approved):

PR #2455, opened 2023-01-27
Fix bug for screenspace sky browser scale crash

ylvaselling:
* Removed some functionality that was "too smart" and a bit unnecessary. The class adjusted its texture depending on window resolution etc, which is not ideal in a cluster setup
* Removed unused functions and variables
* Moved functions to classes where they made more sense


alexanderbock (approved):
:+1:

PR #2465, opened 2023-02-01
Touch module code cleanup

### LINKED ISSUE #1041: Touch interaction picking refactor
Currently the touch implementation holds a hard-coded list of objects that are pickable. Consider making this list configurable.

WeirdRubberDuck:
* Remove unused feature that allowed us to "pick" nodes using touch (it didn't really work and had some nasty hardcoded and costly implementation). Fixes #1041 
* General refactoring of code and removing redundant code
* Make touch markers prettier (change default color and smoothen edges)

* Add module property to control which renderable types are "directly touchable"
* Add SGN property to control which individual nodes are "directly touchable"
* Fix stuttering when zooming in closer than the orbitalnavigator allows


modules/touch/src/touchinteraction.cpp
```diff
@@ -390,37 +348,42 @@ void TouchInteraction::updateStateFromInput(const std::vector<TouchInputHolder>&
     }
 
     if (_tap) {
-        // check for doubletap
+        // @TODO (2023-02-01, emmbr) This if is not triggered on every touch tap.
```
alexanderbock, line 351: Could it be the `_tap = false` at the bottom of the `step` function?

modules/touch/src/touchinteraction.cpp
```diff
@@ -433,14 +396,16 @@ void TouchInteraction::updateStateFromInput(const std::vector<TouchInputHolder>&
         resetAfterInput();
     }
 
-    if (_directTouchMode && _selected.size() > 0 && numFingers == _selected.size()) {
+    if (_directTouchMode && _selectedNodeSurfacePoints.size() > 0 &&
```
alexanderbock, line 399: would even do:
`!_selectedNodeSurfacePoints.empty()`

modules/touch/src/touchinteraction.cpp
```diff
@@ -512,36 +478,10 @@ void TouchInteraction::directControl(const std::vector<TouchInputHolder>& list)
     }
 }
 
-void TouchInteraction::findSelectedNode(const std::vector<TouchInputHolder>& list) {
-    // trim list to only contain visible nodes that make sense
-    // @TODO (emmbr 2023-01-31) This hardcoded list should be removed and replaced by something
-    // else. Either a type of renderable that can always be directly manipulated, or a list
-    // that can be set in config/assets. Or both?
-    std::string selectables[31] = {
-        "Sun", "Mercury", "Venus", "Earth", "Mars", "Ceres", "Jupiter", "Saturn", "Uranus",
```
alexanderbock, line 521: 💀 🔫

modules/touch/src/touchinteraction.cpp
```diff
@@ -759,20 +632,23 @@ int TouchInteraction::interpretInteraction(const std::vector<TouchInputHolder>&
     if (_zoomOutTap) {
         return ZOOM_OUT;
     }
-    else if (_doubleTap) {
-        return PICK;
-    }
+    //else if (_doubleTap) {
```
alexanderbock, line 635: I'd remove the comment alltogether.  I think when we do the picking, we'd want to handle the taps more centrally in the code anyway

modules/touch/touchmodule.cpp
```diff
@@ -35,26 +35,37 @@
...
-    , _touchActive(TouchActiveInfo, true)
+    , _touchIsEnabled(EnableTouchInfo, true)
+    , _hasActiveTouchEvent(EventsInfo, false)
 {
     addPropertySubOwner(_touch);
     addPropertySubOwner(_markers);
-    addProperty(_touchActive);
-    _touchActive.onChange([&] {
+    addProperty(_touchIsEnabled);
+    _touchIsEnabled.onChange([&] {
```
alexanderbock, line 62: -> `[&]() {`

() was missing in the old version already

alexanderbock (approved):

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: Okay, I kept it for the sake of not removing all the intention of using the `_doubleTap` variable for something else than just tapping in the bottom corner (which is currently not working anyways though). But yeah sure I'll just remove it

modules/touch/touchmodule.cpp
WeirdRubberDuck: Good 👁️

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: Nope, that's not it. But the logic for the tap depends on a huge check when removing a finger from the touch screen, so I suspect something might go wrong there. 

https://github.com/OpenSpace/OpenSpace/blob/d7eec159fc707c577c9632113843c1fb72d879c8/modules/touch/touchmodule.cpp#L249

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: Realized that I forgot to include a "replacement" for this list (that had a second purpose of saying which nodes we can do direct manipulation on). Plan to discuss this during our meeting today

modules/touch/src/directinputsolver.cpp
```diff
@@ -212,6 +212,13 @@ bool DirectInputSolver::solve(const std::vector<TouchInputHolder>& list,
                               const std::vector<SelectedBody>& selectedBodies,
                               std::vector<double>* parameters, const Camera& camera)
 {
+    ZoneScopedN("Direct touch input solver")
+
+    ghoul_assert(
+        selectedBodies.size() >= list.size(),
```
GPayne, line 218: This name is a little confusing. As I understand, these are selected points on a surface or body, rather than multiple selected nodes. Is that correct?

GPayne (approved):
I looked through the code and don't have any individual line comments to make. It looks like some changes will have a big impact on the functionality, beyond a code cleanup. Unfortunately I no longer have a working touch table that I can test on, so I can't evaluate the touch behavior myself.

include/openspace/rendering/renderable.h
```diff
@@ -79,6 +79,8 @@ class Renderable : public properties::PropertyOwner {
     double boundingSphere() const;
     double interactionSphere() const;
 
+    std::string typeAsString() const;
```
alexanderbock, line 82: This can probably return `std::string_view` as the source will live for a long time and that way we don't need to make a copy every time we ask for it

include/openspace/scene/scenegraphnode.h
```diff
@@ -139,6 +139,8 @@ class SceneGraphNode : public properties::PropertyOwner {
     double reachFactor() const;
     double approachFactor() const;
 
+    bool isDirectlyTouchable() const;
```
alexanderbock, line 142: How about a more generic name, even though we only use it for touch right now?    Like `supportsDirectInteraction` instead?  To not tie it to the touch interaction too much

modules/touch/src/touchinteraction.cpp
```diff
@@ -249,6 +206,36 @@ namespace {
...
+        DirectManipulationThresholdInfo =
+    {
+        "DirectManipulationThreshold",
+        "Direct manipulation threshold",
+        "This threshold affects the distance from the interaction sphere at which the "
+        "direct manipulation interaction mode starts being active. The value is given "
+        "as a factor times the interaction sphere"
+    };
+
+    //constexpr openspace::properties::Property::PropertyInfo OrbitSpinningThreshold = {
```
alexanderbock, line 227: Do we still need the commented out code?

modules/touch/src/touchinteraction.cpp
```diff
@@ -347,25 +323,17 @@ TouchInteraction::TouchInteraction()
     addProperty(_centroidStillThreshold);
     addProperty(_panEnabled);
     addProperty(_interpretPan);
-    addProperty(_slerpTime);
     addProperty(_friction);
-    addProperty(_pickingRadiusMinimum);
+
+    addProperty(_enableDirectManipulation);
+    addProperty(_directTouchDistanceThreshold);
+    //addProperty(_orbitSpeedThreshold);
```
alexanderbock, line 330: Ditto

alexanderbock (approved):

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: It's related to a feature that I just commented out, but that we might want to bring back. So wasn't sure what to do with the properties

modules/touch/src/directinputsolver.cpp
WeirdRubberDuck: I agree that it's confusing, but limited myself when it came to poking around in the code for the directinputsolver. As seen in the function signature, the struct for that varaible is currently also named "SelectedBody". I have changed the name in the touchInteraction class in another commit, so that it's a bit more obvious in the code that calls this function at least. 

But yes the `SelectedBody` is a struct with a touch input id, a pointer to a scene graph node and a surface position (or rather, a position in local coordinates on the interaction sphere). If you've got a good name suggestion, I could go ahead and rename it in this PR as well.

WeirdRubberDuck:
> I looked through the code and don't have any individual line comments to make. It looks like some changes will have a big impact on the functionality, beyond a code cleanup. Unfortunately I no longer have a working touch table that I can test on, so I can't evaluate the touch behavior myself.

It removes two previously existing features: 
* "Picking" other scene graph nodes that are not the anchor. Which was not very usable and could introduce weird behavior when triggered by mistake. 
* Allowing the node to start spinning based on direct manipulation. We did not really get this to work on the touch screen I have, and when we did it was by mistake and led to undesired rotations. I kept the code for this commented out, as I think it would be a nice feature to bring back at some point. Ideally by utilizing the code in the orbital navigator a bit more, so you don't have to  keep track of two sets of friction parameters

include/openspace/scene/scenegraphnode.h
WeirdRubberDuck: Sure, but the description should still mention touch, for now, I think. As it currently IS very tied to the touch interaction, and I want that to be clear to the users. It's not going to be a relevant property for any user not using touch interaction, a of now at least

modules/touch/src/directinputsolver.cpp
GPayne: I agree with limiting the refactoring and keeping away from the directinputsolver code. The slight confusion about the name is probably not worth it.

modules/touch/include/touchinteraction.h
```diff
@@ -67,13 +66,11 @@ class TouchInteraction : public properties::PropertyOwner {
 public:
     TouchInteraction();
 
-    // for interpretInteraction()
-    enum Type {
+    enum InteractionType {
```
Roxeena, line 69: In our [coding style](https://github.com/OpenSpace/OpenSpace/wiki/Coding-Style#8-enumerations-must-not-be-in-upper-case-and-should-be-strongly-typed), it says that enums should be "_strongly typed_" so this would become `enum class InteractionType`. But I'm not sure how strongly we keep to this guideline.

modules/touch/include/touchinteraction.h
```diff
@@ -184,11 +177,14 @@ class TouchInteraction : public properties::PropertyOwner {
...
     properties::BoolProperty  _panEnabled;
     properties::FloatProperty _interpretPan;
-    properties::FloatProperty _slerpTime;
     properties::Vec4Property _friction;
-    properties::FloatProperty _pickingRadiusMinimum;
     properties::FloatProperty _constTimeDecay_secs;
 
+    properties::BoolProperty _enableDirectManipulation;
+    properties::FloatProperty _directTouchDistanceThreshold;
+    //properties::FloatProperty _orbitSpeedThreshold;
```
Roxeena, line 185: Is this future work to fix? If so should we add a comment about why they are commented out?

modules/touch/touchmodule.cpp
```diff
@@ -28,44 +28,98 @@
...
+    addProperty(_hasActiveTouchEvent);
+
+    _defaultDirectTouchRenderableTypes.onChange([&]() {
+        _sortedDefaultRenderableTypes.clear();
+        for (const std::string& s : _defaultDirectTouchRenderableTypes.value()) {
+            ghoul::TemplateFactory<Renderable>* fRenderable =
+                FactoryManager::ref().factory<Renderable>();
+
+            if (!fRenderable->hasClass(s)) {
+                LWARNING(fmt::format(
```
Roxeena, line 96: Should we maybe lower the level of this logging? Thinking that if you are close to a node of unsupported type this could clutter the log. Or maybe it will only warn once?

src/navigation/orbitalnavigator.cpp
```diff
@@ -539,6 +539,20 @@ glm::quat OrbitalNavigator::anchorNodeToCameraRotation() const {
...
+glm::dvec3 OrbitalNavigator::pushToSurfaceOfAnchor(
+                                                  const glm::dvec3& cameraPosition) const
+{
+    SurfacePositionHandle posHandle =
+        calculateSurfacePositionHandle(*_anchorNode, cameraPosition);
+
+     return pushToSurface(
+         cameraPosition,
+         _anchorNode->worldPosition(),
+        posHandle
```
Roxeena, line 552: Is it me or is there a misalignment here with 1 whitespace?

modules/touch/src/touchinteraction.cpp
Roxeena: My suggestion is to add a comment above that explains this and then keep it commented out.

Roxeena (approved):
👍

modules/touch/touchmodule.cpp
WeirdRubberDuck: It only warns when setting the property value, which people would not really do that often during runtime :)

include/openspace/scene/scenegraphnode.h
alexanderbock: Agree that the description should mention touch;  that we can change later once the value is applied to other interaction forms, too

modules/touch/include/touchinteraction.h
alexanderbock: :+1: for strongly typing it.  It prevents a bunch of potential errors and confusions

src/navigation/orbitalnavigator.cpp
alexanderbock: Yep, `posHandle` only has 3 whitespace friends

modules/touch/include/touchinteraction.h
WeirdRubberDuck: 👍

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: Leaving this in there for now; future me's (or someone else's) problem

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: Ended up removing them, and all related code, completely. To limit the number of TODO comments in this class, and as we would not want to use this specific implementation anyways.

modules/touch/src/touchinteraction.cpp
WeirdRubberDuck: Removed. See previous comment

modules/touch/include/touchinteraction.h
WeirdRubberDuck: Removed. See previous comment

PR #2483, opened 2023-02-02
Add option to specify an offset distance for RenderableNodeLine

WeirdRubberDuck:
Make it possible to specify a distance form the start and/or end node at which the rendering of the line should start

Also make it possible to specify the value as a multiplier times the bounding sphere


alexanderbock (approved):

modules/base/rendering/renderablenodeline.cpp
```diff
@@ -125,7 +150,52 @@ RenderableNodeLine::RenderableNodeLine(const ghoul::Dictionary& dictionary)
...
+                "Could not find end node '{}'", _identifier, _end.value()
+            ));
+            return;
+        }
+
+        if (_useRelativeOffsets) {
+            // Recompute previous offsets to relative values
+            double startBs = startNode->boundingSphere();
+            double endBs = endNode->boundingSphere();
+            _startOffset = startBs > 0.0 ? _startOffset / startBs : 0.0;
```
Roxeena, line 190: Maybe provide some warning when it is set to zero because of no bounding sphere? I got confused the first time I tried it out with the transform nodes and not the planet nodes.

Roxeena (commented):
When the offset is set too large for the line to handle, it points in the "other" direction, see images. And when both offsets are set this large, the line flickers around like crazy. 

Also, it would be great if these new properties could be part of the parameters so you can provide them in the asset file. For example, if you know the exact measurements you want to offset, it would be nice to not have to manually set it every time you show it. 

OpenSpace_000000

OpenSpace_000001

WeirdRubberDuck:
> When the offset is set too large for the line to handle, it points in the "other" direction, see images. And when both offsets are set this large, the line flickers around like crazy.
> 
> Also, it would be great if these new properties could be part of the parameters so you can provide them in the asset file. For example, if you know the exact measurements you want to offset, it would be nice to not have to manually set it every time you show it.
> 
> OpenSpace_000000
> 
> OpenSpace_000001

The "offsets too large leading to point in the other direction" is something I chose not to care about. I think that behavior is more apparent than not seeing anything at all when editing the value using the slider. Also, these property should be "AdvancedUser" (Builder) properties and not really have to be edited once OpenSpace is running, so I don't think it's too big a problem that their behavior is a bit weird when not using "suitable" values. 

Have not seen the flickering when both offsets are large, but that is for sure a precision issue. Not sure if we can get around it, but I'll give it a look. 

And yes, of course, it should be possible to add from an asset file! I just forgot to add them to the Parameters. Thanks for noticing :)

modules/base/rendering/renderablenodeline.cpp
```diff
@@ -65,6 +67,30 @@ namespace {
         "This value specifies the line width"
     };
 
+    constexpr openspace::properties::Property::PropertyInfo StartOffsetInfo = {
```
WeirdRubberDuck, line 70: Add the new properties to the Parameters

PR #2497, opened 2023-02-08
Replace profile actions/keybind editor button with single close button

GPayne:
In the profile editor for actions & keybindings: Switched from main save & cancel buttons to a single "Close" button. This new button doesn't really do anything, because of changes to save behavior in the actions and keybindings sections. Pressing Save in either of those sections will save the corresponding data.
The new Close button is necessary to be consistent with other editor dialog windows, because otherwise the user would have to manually X-close the window.


Roxeena (commented):
👍

WeirdRubberDuck (approved):

PR #2515, opened 2023-02-17
Improve behavior for gui window and separate render window

GPayne:
Removes our usage of "monitor" key in sgct config files, but does not remove that key in the sgct repository code.
Update of config/gui_projector.json to position non-gui window in 2nd monitor (dimensions will need to be customized to fit).
If option "show only user interface on the first window" is checked, then all subsequent windows will have 2D rendering disabled.

### DIFF
--- a/apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
+++ b/apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
@@ -277,6 +277,11 @@ sgct::config::Cluster SgctEdit::generateConfiguration() const {
         window.tags.push_back("GUI");
         window.draw2D = true;
         window.draw3D = false;
+
+        //Disable 2D rendering on all non-GUI windows
+        for (unsigned int w = 1; w < node.windows.size(); ++w) {
+            node.windows[w].draw2D = false;
+        }
     }
 
     cluster.nodes.push_back(node);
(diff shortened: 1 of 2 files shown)


apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -277,6 +277,11 @@ sgct::config::Cluster SgctEdit::generateConfiguration() const {
         window.tags.push_back("GUI");
         window.draw2D = true;
         window.draw3D = false;
+
+        //Disable 2D rendering on all non-GUI windows
```
alexanderbock, line 281: Missing space after `//`

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -277,6 +277,11 @@ sgct::config::Cluster SgctEdit::generateConfiguration() const {
...
+        for (unsigned int w = 1; w < node.windows.size(); ++w) {
```
alexanderbock, line 282: More modern version of this:
```cpp
for (sgct::Window& w : node.windows) {
    w.draw2D = false;
}
```

GPayne: I also like using that version, but I did the for loop because I need to skip past the first element of `node.windows` since it has the GUI.

alexanderbock: Do'h!  Yea, I missed that.   In that case it should be a `size_t` instead of an `unsigned int` anyway since that's the return type of the `size` function

Issue #1617, opened 2021-05-31
Atmosphere causes issues with globelabels rendering on AMD graphics cards


WeirdRubberDuck:
When rendering `GlobeLabelsComponents` on an AMD card, the color of the label gets blended-ish with the atmosphere. For example: first image shows labels on an AMD card VS in the second image on an Nvidia card (set with fade in min distance to 1 meter, so that no fading happens). 

The same behavior has been observed on multiple globes that has atmospheres.

![image](https://user-images.githubusercontent.com/8808894/120182448-b52eb580-c20e-11eb-8d5a-da3fa9f0664e.png)

![image](https://user-images.githubusercontent.com/8808894/120182545-d42d4780-c20e-11eb-9205-00b51b9c7697.png)

Issue #1610, opened 2021-05-26
Globelabels are making the stars shine through the planet


alexanderbock:
When enabling the globe labels, the background RenderableStars are rendered on top of the labels, maybe an issue with writing the position in the g buffer?

https://user-images.githubusercontent.com/4921724/119635630-2d166d80-be14-11eb-966f-8cd604ffc0c6.mp4

WeirdRubberDuck:
Saw this problem too. Note that it only happens when the opacity of the labels is less than 1.0, in which case we normally (for renderables, that is) switch the render bin from `Opaque` to `PreDeferredTransparent`. This is not being done for the Globe labels, as the class `GlobeLabelsComponent` is not a Renderable, and the labels are always being rendered in the `Opaque` bin

Maybe switching the render bin would solve the problem?

alexanderbock:
Yea, I think it would solve the problem

PR #2520, opened 2023-02-20
Double renderbins for renderables

WeirdRubberDuck:
Fixes https://github.com/OpenSpace/OpenSpace/issues/1610 and fixes https://github.com/OpenSpace/OpenSpace/issues/1617 by rendering the globelabels after the atmosphere

Also fixes a problem with opacity not really working for globelabels on celestial bodies with an atmosphere

And allows for rendering the upcoming GeoJson component as part of the renderable globe

### DIFF
--- a/include/openspace/rendering/renderable.h
+++ b/include/openspace/rendering/renderable.h
@@ -99,6 +99,8 @@ public:
     void setRenderBin(RenderBin bin);
     bool matchesRenderBinMask(int binMask);
 
+    bool matchesSecondaryRenderBin(int binMask);
+
     void setFade(float fade);
 
     bool isVisible() const;
@@ -128,6 +130,10 @@ protected:
     SceneGraphNode* _parent = nullptr;
     bool _shouldUpdateIfDisabled = false;
     RenderBin _renderBin = RenderBin::Opaque;
+
+    // An optional renderbin that renderables can use for certain components, in cases
+    // where all parts of the renderable should not be rendered in the same bin
+    std::optional<RenderBin>_secondaryRenderBin = std::nullopt;
 
 private:
     // We only want the SceneGraphNode to be able manipulate the parent, so we don't want
--- a/include/openspace/rendering/renderable.h
+++ b/include/openspace/rendering/renderable.h
@@ -82,8 +82,9 @@ public:
 
     std::string_view typeAsString() const;
 
-    virtual void render(const RenderData& data, RendererTasks& rendererTask);
     virtual void update(const UpdateData& data);
+    virtual void render(const RenderData& data, RendererTasks& rendererTask);
+    virtual void renderSecondary(const RenderData& data, RendererTasks& rendererTask);
 
     // The 'surface' in this case is the interaction sphere of this renderable. In some
     // cases (i.e., planets) this corresponds directly to the physical surface, but in
--- a/include/openspace/rendering/renderable.h
+++ b/include/openspace/rendering/renderable.h
@@ -98,9 +98,9 @@ public:
 
     RenderBin renderBin() const;
     void setRenderBin(RenderBin bin);
-    bool matchesRenderBinMask(int binMask);
+    bool matchesRenderBinMask(int binMask) const noexcept;
 
-    bool matchesSecondaryRenderBin(int binMask);
+    bool matchesSecondaryRenderBin(int binMask) const noexcept;
 
     void setFade(float fade);
 
@@ -134,7 +134,7 @@ protected:
 
     // An optional renderbin that renderables can use for certain components, in cases
     // where all parts of the renderable should not be rendered in the same bin
-    std::optional<RenderBin>_secondaryRenderBin = std::nullopt;
+    std::optional<RenderBin> _secondaryRenderBin;
 
 private:
     // We only want the SceneGraphNode to be able manipulate the parent, so we don't want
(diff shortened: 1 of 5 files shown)


src/rendering/renderable.cpp
```diff
@@ -268,7 +270,14 @@ void Renderable::setRenderBin(RenderBin bin) {
 }
 
 bool Renderable::matchesRenderBinMask(int binMask) {
```
alexanderbock, line 272: Function can be defined as `const` and `noexcept`

src/rendering/renderable.cpp
```diff
@@ -268,7 +270,14 @@ void Renderable::setRenderBin(RenderBin bin) {
...
-    return binMask & static_cast<int>(renderBin());
+    return binMask & static_cast<int>(_renderBin);
+}
+
+bool Renderable::matchesSecondaryRenderBin(int binMask) {
```
alexanderbock, line 276: Likewise here

include/openspace/rendering/renderable.h
```diff
@@ -129,6 +132,10 @@ class Renderable : public properties::PropertyOwner {
     bool _shouldUpdateIfDisabled = false;
     RenderBin _renderBin = RenderBin::Opaque;
 
+    // An optional renderbin that renderables can use for certain components, in cases
+    // where all parts of the renderable should not be rendered in the same bin
+    std::optional<RenderBin>_secondaryRenderBin = std::nullopt;
```
alexanderbock, line 137: The default constructor for `std::optional` initializes it with a `std::nullopt`, so no need for the assignment here

alexanderbock: Also missing space between `>` and `_`

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -806,6 +807,10 @@ void RenderableGlobe::render(const RenderData& data, RendererTasks& rendererTask
     _lastChangedLayer = nullptr;
 }
 
+void RenderableGlobe::renderSecondary(const RenderData& data, RendererTasks&) {
+    _globeLabelsComponent.draw(data);
```
alexanderbock, line 811: The old `draw` was in a `try`/`catch` block;  are we potentially letting an exception escape here that might lead to a crash?

src/scene/scenegraphnode.cpp
```diff
@@ -747,21 +747,31 @@ void SceneGraphNode::render(const RenderData& data, RendererTasks& tasks) {
...
+        .renderBinMask = data.renderBinMask,
+        .modelTransform = {
+            .translation = _worldPositionCached,
+            .rotation = _worldRotationCached,
+            .scale = _worldScaleCached
+        }
+    };
+
+    if (_renderable->matchesSecondaryRenderBin(data.renderBinMask)) {
+        TracyGpuZone("Render Secondary")
```
alexanderbock, line 762: -> `Render Secondary Bin` ?

modules/globebrowsing/src/renderableglobe.cpp
WeirdRubberDuck: Good catch ;)  I thought the try-catch was only related to the layer rendering, but the font rendering can also throw a texture unit error. 

Looking at the code in the existing catch block, I don't think we would have to do anything of this?  

image

But instead maybe just print the message? 
image

src/scene/scenegraphnode.cpp
WeirdRubberDuck: Sure

ylvaselling:
Not sure if this was introduced in this PR but I found this when I tested it:

https://user-images.githubusercontent.com/24265984/221919802-62c62df9-f40f-4858-abbe-39b10918f872.mp4

alexanderbock:
You mean the flickering?  That has been there for a long while.  The size reference layer is using an internal FBO that screws up the OpenSpace rendering for one frame while a new tile is rendered.  We should probably have a look at that soon

ylvaselling:
Yes, exactly. I see! Then this PR seems fine by me:)

include/openspace/rendering/renderable.h
```diff
@@ -97,7 +98,9 @@ class Renderable : public properties::PropertyOwner {
 
     RenderBin renderBin() const;
     void setRenderBin(RenderBin bin);
-    bool matchesRenderBinMask(int binMask);
+    bool matchesRenderBinMask(int binMask) const noexcept;
+
+    bool matchesSecondaryRenderBin(int binMask) const noexcept;
```
ylvaselling, line 103: Just curious about what the noexcept does?

ylvaselling (approved):

include/openspace/rendering/renderable.h
WeirdRubberDuck: It's a specifier that essentially tells the compiler that this function won't throw any exceptions, allowing it to do further optimizations :)

WeirdRubberDuck: Both these checks have to be done 1-2 times every frame, for each renderable. So any way it can be made faster, the better!

Or at least that's my thought on why you suggested adding them @alexanderbock ?

alexanderbock: Yea, exactly.  If we start using it a bit more often, we might save a few CPU cycles here and there.  And good to start wherever we can.

For more information: https://eel.is/c++draft/except.spec#1
:P

PR #2525, opened 2023-02-22
Optimization update for RenderableTrailTrajectory

Arohdin:
### Key changes:
- Implemented iterative approach for calculating all vertex positions in the trail. Instead of doing all calculations at once, which takes time and locks the thread, we instead calculate a smaller portion (chunk) of the vertices each frame. Once all vertices are calculated we commit it for rendering.
- Moved the calculations for the bounding sphere into the "vertex calculation" loop in order to minimize unnecessary looping. Before we also calculated the bounding sphere each frame, but this is now changed to only be done after each full sweep.

### These changes have the following effect:
- Iterative approach: The program now runs without locking up when first loading large RenderableTrailTrajectories into the scene, albeit at a lower framerate.
- Bounding Sphere: Since bounding sphere calculations are only done when needed. 
Example, when using scene with both Voyager 1 and 2 trails, my setup went from 80-100 fps to 200-400 fps.

### Before:

https://user-images.githubusercontent.com/9076357/220638897-90bfc491-c3b2-4d94-a9d1-71a346a8637a.mp4

### After:

https://user-images.githubusercontent.com/9076357/220638961-d4176cd8-3db0-4f5e-ad2b-b7f0071834df.mp4

### DIFF
--- a/modules/base/rendering/renderabletrailtrajectory.cpp
+++ b/modules/base/rendering/renderabletrailtrajectory.cpp
@@ -83,6 +83,13 @@ namespace {
         "'false', only the trail until the current time in the application will be shown"
     };
 
+    constexpr openspace::properties::Property::PropertyInfo SweepChunkSizeInfo = {
+        "SweepChunkSize",
+        "Sweep Chunk Size",
+        "The number of vertices that will be calculated each frame whenever the trail needs to be recalculated. "
+        "A greater value will result in more calculations per frame."
+    };
+
     struct [[codegen::Dictionary(RenderableTrailTrajectory)]] Parameters {
         // [[codegen::verbatim(StartTimeInfo.description)]]
         std::string startTime [[codegen::annotation("A valid date in ISO 8601 format")]];
@@ -98,6 +105,9 @@ namespace {
 
         // [[codegen::verbatim(RenderFullPathInfo.description)]]
         std::optional<bool> showFullTrail;
+
+        // [[codegen::verbatim(SweepChunkSizeInfo.description)]]
+        std::optional<int> sweepChunkSize;
     };
 #include "renderabletrailtrajectory_codegen.cpp"
 } // namespace
@@ -145,6 +155,8 @@ RenderableTrailTrajectory::RenderableTrailTrajectory(const ghoul::Dictionary& di
     _renderFullTrail = p.showFullTrail.value_or(_renderFullTrail);
     addProperty(_renderFullTrail);
 
+    _sweepChunkSize = p.sweepChunkSize.value_or(_sweepChunkSize);
+
     // We store the vertices with ascending temporal order
     _primaryRenderInformation.sorting = RenderInformation::VertexSorting::OldestFirst;
 }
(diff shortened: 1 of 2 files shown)


Arohdin:
Found a bug - let's hold for now.

Arohdin:
> Found a bug - let's hold for now.

Bug fixed in [4e519f7](https://github.com/OpenSpace/OpenSpace/commit/4e519f7a1dde0df08425ec51173264bd0d1880dc)
This should be fine to review and merge now.

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -173,30 +183,55 @@ void RenderableTrailTrajectory::deinitializeGL() {
...
-        for (int i = 0; i < nValues; ++i) {
+            // Make space for the vertices
+            _vertexArray.clear();
+            _vertexArray.resize(_nValues);
+        }
+        
+        // Calculate sweeping range for this iteration
+        int startIndex = _sweepIteration * _sweepChunk;
+        int nextIndex = (_sweepIteration + 1) * _sweepChunk;
+        int stopIndex = (nextIndex < _nValues) ? nextIndex : _nValues;
```
alexanderbock, line 203: This is just: `int stopIndex = std::min(nextIndex, _nValues);`

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -173,30 +183,55 @@ void RenderableTrailTrajectory::deinitializeGL() {
...
                 {},
-                Time(_start + i * totalSampleInterval),
+                Time(_start + i * _totalSampleInterval),
                 Time(0.0)
-            });
+                });
             _vertexArray[i] = { p.x, p.y, p.z };
+
+            // Set max and min vertex for bounding sphere calculations
+            _maxVertex.x = std::max(_maxVertex.x, p.x);
```
alexanderbock, line 215: I believe this would be the same:
```cpp
_minVertex = glm::min(_minVertex, p);
_maxVertex = glm::max(_maxVertex, p);
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -173,30 +183,55 @@ void RenderableTrailTrajectory::deinitializeGL() {
...
+
+        // Calculate all vertex positions
+        for (int i = startIndex; i < stopIndex; ++i) {
             const glm::vec3 p = _translation->position({
                 {},
-                Time(_start + i * totalSampleInterval),
+                Time(_start + i * _totalSampleInterval),
                 Time(0.0)
-            });
+                });
```
alexanderbock, line 211: One indentation too much

modules/base/rendering/renderabletrailtrajectory.cpp
Arohdin: Indeed it is 👍 
_note to self: use std when possible_

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -173,30 +183,55 @@ void RenderableTrailTrajectory::deinitializeGL() {
...
+        if (_sweepIteration == 0) {
+            // Convert the start and end time from string representations to J2000 seconds
+            _start = SpiceManager::ref().ephemerisTimeFromDate(_startTime);
+            _end = SpiceManager::ref().ephemerisTimeFromDate(_endTime);
 
-        // Make space for the vertices
-        _vertexArray.clear();
-        _vertexArray.resize(nValues);
+            _totalSampleInterval = _sampleInterval / _timeStampSubsamplingFactor;
+            _nValues = static_cast<int>((_end - _start) / _totalSampleInterval);
```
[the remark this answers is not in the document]
Arohdin: In the cpp doc it says that uint64_t is "provided if and only if the implementation directly supports the type" - does this matter in our case?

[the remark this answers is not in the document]
alexanderbock: Nah, we can just assume that they exist.  We stopped supporting compiling on 32 bit machines a _long_ time ago and I don't think we need to get that support back

alexanderbock (approved):

Issue #2513, opened 2023-02-17
Apollo Terrain Visibility Issue on Linux


GPayne:
When running the Apollo profile in linux, and then loading Apollo 11 or 17 missions (e.g. by running the `os.missions.apollo17.setup.landingsite` action), the lunar terrain is hidden. This includes the mission-specific lunar lander areas with annotations, etc:
![linux_with_both_WAC_enabled](https://user-images.githubusercontent.com/3966290/219532481-bb818646-fc9b-4550-be95-aea6ca78eaab.png)

This is not a problem on windows:
![linux_with_both_WAC_disabled](https://user-images.githubusercontent.com/3966290/219532507-6559618b-e714-4b84-a2c6-5848dc95be94.png)

The problem relates to the way the layers are ordered. In the GUI under **Solar System -> Planets -> Earth -> Moon -> Moon -> Layers -> ColorLayers** the layers listed are sourced by the following assets:
1. Moon terrain layers from scene/solarsystemplanets/earth/moon/default_layers.asset
2. A11* and A17* layers from lem_flipbook.asset and other asset files

If the WAC layers are below the A11* & A17* layers in the GUI listing, then the former hides the latter. In linux, the layers from # 2 appear above those from # 1 in the listing, which causes A11* & A17* layers to be hidden. In windows, the order is reversed so the problem doesn't happen.
This can be reproduced on windows by:
1) focusing on Apollo 17 LEM and zooming into it.
2) fast-forward the time if the LEM is not in the sunlight.
3) enable the "A17 LEM" moon color layer
4) hover over the WAC [Utah] or WAC [Sweden] layer to get the drag icon, then drag it to a position below "A17 LEM", and the annotated terrain around the LEM will disappear. If the moved WAC layer is disabled, then the LEM terrain will reappear again.

GPayne:
The AssetManager currently uses a `std::unordered_set` container for assets that are added or removed. This will consistently give the same order of objects, even though that order isn't specified. The problem here is mostly likely that the gcc compiler uses a different hash to determine container order than MSVC uses, causing a different order on linux.
This is an example where the order of how assets are loaded matters, so branch **issue/2513** switched to using a `std::list` container for asset add/remove, so there is a predictable "first in, first out" behavior.

PR #2543, opened 2023-03-11
Switch to std::list for adding & removing assets in deterministic order

GPayne:
See issue #2513 for the problem of non-deterministic asset add/removal, with an example of a problem caused by it. This change of containers for assets to be added/removed seems to avoid the problem.
This change was run on the linux image test server. Results were positive but with a few possible artifacts due to this change. Needs some additional testing.

### DIFF
--- a/src/scene/assetmanager.cpp
+++ b/src/scene/assetmanager.cpp
@@ -300,24 +300,23 @@ void AssetManager::update() {
 void AssetManager::add(const std::string& path) {
     ghoul_precondition(!path.empty(), "Path must not be empty");
     // First check if the path is already in the remove queue. If so, remove it from there
-    const auto it = _assetRemoveQueue.find(path);
-    if (it != _assetRemoveQueue.end()) {
-        _assetRemoveQueue.erase(it);
-    }
-
-    _assetAddQueue.insert(path);
+    _assetRemoveQueue.remove_if(
+        [&path](const std::string& value) {
+            return (path == value);
+        }
+    );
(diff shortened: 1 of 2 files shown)


src/scene/assetmanager.cpp
```diff
@@ -300,24 +300,23 @@ void AssetManager::update() {
 void AssetManager::add(const std::string& path) {
     ghoul_precondition(!path.empty(), "Path must not be empty");
     // First check if the path is already in the remove queue. If so, remove it from there
-    const auto it = _assetRemoveQueue.find(path);
-    if (it != _assetRemoveQueue.end()) {
-        _assetRemoveQueue.erase(it);
-    }
-
-    _assetAddQueue.insert(path);
+    _assetRemoveQueue.remove_if(
```
alexanderbock, line 303: This can be just `_assetRemoveQueue.remove(path)`

src/scene/assetmanager.cpp
```diff
@@ -300,24 +300,23 @@ void AssetManager::update() {
...
     ghoul_precondition(!path.empty(), "Path must not be empty");
-
     // First check if the path is already in the add queue. If so, remove it from there
-    const auto it = _assetAddQueue.find(path);
-    if (it != _assetAddQueue.end()) {
-        _assetAddQueue.erase(it);
-    }
-
-    _assetRemoveQueue.insert(path);
+    _assetAddQueue.remove_if(
```
alexanderbock, line 314: Same here

Issue #2546, opened 2023-03-14
SGCT editor creates stretched image for portrait windows


ylvaselling:
I tried to create a window fitted for 1080 x 1920 pixels. That created a stretched image in OpenSpace. I had expected a regular image

![image](https://user-images.githubusercontent.com/24265984/225118879-ab2988ea-30b5-4138-a93d-071201f9d813.png)
![image](https://user-images.githubusercontent.com/24265984/225118927-4b27ca25-1adc-4510-bbf7-61459e12d1eb.png)

GPayne:
The window GUI's FOV calculations for window size need to be fixed for handling portrait windows like this.
For now, you can get it to look right by unlocking the FOV values and entering your own. Keeping the horizontal FOV of 80 and setting the vertical to 110 degrees looks about right.

GPayne:
Closing this upon merging PR#2554

PR #2554, opened 2023-03-20
Use correct FOV calculations to handle portrait or landscape windows

GPayne:
Fix for issue #2546. Provides a correct FOV for both portrait or landscape window configuration. Also improves on the previous scaling code which didn't work well for elongated windows.

### DIFF
--- a/apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
+++ b/apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
@@ -816,14 +816,22 @@ void WindowControl::onFovLockClicked() {
 }
 
 void WindowControl::updatePlanarLockedFov() {
-    const float aspectRatio = _windowDimensions.width() / _windowDimensions.height();
-    const float ratio = aspectRatio / IdealAspectRatio;
-    if (ratio >= 1.f) {
-        _planar.fovH->setValue(std::min(DefaultFovH * ratio, 180.f));
-        _planar.fovV->setValue(DefaultFovV);
+    bool landscapeOrientation = (_windowDimensions.width() >= _windowDimensions.height());
+    float aspectRatio;
+    if (landscapeOrientation) {
+        aspectRatio = _windowDimensions.width() / _windowDimensions.height();
     }
     else {
-        _planar.fovH->setValue(DefaultFovH);
-        _planar.fovV->setValue(std::min(DefaultFovV / ratio, 180.f));
+        aspectRatio = _windowDimensions.height() / _windowDimensions.width();
     }
+
+    const float lockedFov = DefaultFovV;
+    float adjustedFov =
+        2.f * atan(aspectRatio * tan(lockedFov * std::numbers::pi / 180.f / 2.f));
+    // Convert to degrees and limit to 180°
+    adjustedFov *= 180.f / std::numbers::pi;
+    adjustedFov = std::min(adjustedFov, 180.f);
+
+    _planar.fovH->setValue(landscapeOrientation ? adjustedFov : lockedFov);
+    _planar.fovV->setValue(landscapeOrientation ? lockedFov : adjustedFov);
 }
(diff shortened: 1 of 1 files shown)


apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -816,14 +816,22 @@ void WindowControl::onFovLockClicked() {
...
-        _planar.fovH->setValue(DefaultFovH);
-        _planar.fovV->setValue(std::min(DefaultFovV / ratio, 180.f));
+        aspectRatio = _windowDimensions.height() / _windowDimensions.width();
     }
+
+    const float lockedFov = DefaultFovV;
+    float adjustedFov =
+        2.f * atan(aspectRatio * tan(lockedFov * std::numbers::pi / 180.f / 2.f));
+    // Convert to degrees and limit to 180°
+    adjustedFov *= 180.f / std::numbers::pi;
```
alexanderbock, line 832: Same here

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -816,14 +816,22 @@ void WindowControl::onFovLockClicked() {
...
+    if (landscapeOrientation) {
+        aspectRatio = _windowDimensions.width() / _windowDimensions.height();
     }
     else {
-        _planar.fovH->setValue(DefaultFovH);
-        _planar.fovV->setValue(std::min(DefaultFovV / ratio, 180.f));
+        aspectRatio = _windowDimensions.height() / _windowDimensions.width();
     }
+
+    const float lockedFov = DefaultFovV;
```
alexanderbock, line 828: Do we need the `lockedFov` variable?  Why not use the `DefaultFovV` in the terniary below directly?

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -816,14 +816,22 @@ void WindowControl::onFovLockClicked() {
...
+    float adjustedFov =
+        2.f * atan(aspectRatio * tan(lockedFov * std::numbers::pi / 180.f / 2.f));
```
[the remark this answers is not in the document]
GPayne: Included ` <numbers>`

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
GPayne: Switched to the float version of `pi` in both locations.

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
GPayne: Reason is for code readability. The `DefaultFovV` is used elsewhere mainly for initializing values, for the assumed orientation (which is landscape). But if a portrait orientation is selected, then this value will be used as the horizontal fov.
Here, I chose to call the angle that we don't change `lockedFov`, and the one that we change in order to make the view look correct is `adjustedFov`.

WeirdRubberDuck (approved):
Works well and looks good to me

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
WeirdRubberDuck: I get the reasoning. The name "DeafultFovV" already implies to me that this is a value that is not changed, however :)

Using it directly in the ternary would IMO also make the ternary a bit clearer to me as well, since it then is clearer that this FOV is the default **vertical** fov that is used. Hope you get what I mean

GPayne: The readability concern was with the bottom 2 lines, where the `_planar.fovH` might be set to `DefaultFovV` depending on the orientation.
However, to address the concerns with this, I got rid of the added `lockedFov` variable and renamed `DefaultFovV` & `DefaultFovH` to `DefaultFovShortEdge` and `DefaultFovLongEdge`. These names make sense with either orientation.

Issue #2450, opened 2023-01-26
Update static data


micahnyc:
For v19 we should update the data that we have pulled from other sources:

- [x] exoplanet data (@WeirdRubberDuck)
- [x] sssb data (@GPayne)
- [x] generic kernels (@alexanderbock)

----- a note from DWeigal, we should consider updating out wwt-webclient code base
- [ ] wwt-webclient

micahnyc:
From dan: 

there's already and updated kernel for the Jupiter system with the 12 new moons: jup345 as well as some moons in the jup344 kernel that aren't currently in OpenSpace https://ssd.jpl.nasa.gov/sats/ephem/ (current kernels are 310 and 341). Looks like an easy update for the next version and we could share our updated minor moons asset

alexanderbock:
Exoplanets should be updated again

PR #2556, opened 2023-03-21
Updating SPICE Kernels and adding new Moons to Outer Planets

alexanderbock:
- Updated the kernels for Jupiter, Saturn, Neptune, and Uranus
 - Added new moons from https://ssd.jpl.nasa.gov/sats/ephem

Addresses the kernel part of #2450


data/assets/scene/solarsystem/planets/saturn/minor/norse_group.asset
```diff
@@ -433,25 +433,24 @@ local norseGroup = {
...
-  --     Path = "/Solar System/Planets/Saturn/Moons/Norse Group"
-  --   },
-  --   TrailColor = trailColor,
-  --   OrbitPeriod = 1242.36,
-  --   Kernels = kernels368
-  -- },
+    Kernels = kernel452
+  },
+  {
+    Identifier = "Surtur", -- Rising
```
WeirdRubberDuck, line 439: Is this comment literally just here as a reference to an Amon Amaranth album? ;)

WeirdRubberDuck (changes requested):
Only checked the Neptune moons but hm, I don't think Nereid should look quite like this: 

image

The rendering of the moon trails should also be verified. A lot of the moons end up in the origin of the planet. 
Illustrated by labels here (also, I think the labels should probably live under the title for their respective moons, which they currently seem to not do)
image

alexanderbock:
Addressed the comments;  some of them were copy paste errors, some others were errors on the JPL page that have been reported upstream

alexanderbock:
OBS:
The kernels were changed without updating the version number, so if you have tried it before, you need to remove sync/http/neptune_kernels/2 and redownload.  Or manually add http://liu-se.cdn.openspaceproject.com/files/solarsystem/planets/neptune/kernels/v1_v2/nep095.bsp into that folder


Also for future self:  Whenever we add SPICE items, we should test with the SPICE exceptions enabled as it would have detected those errors

PR #2557, opened 2023-03-21
Fading of propertyowners through GUI, and more owners with fade property

### LINKED ISSUE #1827: GUI feature to fade in object when enabled and out when disabled
Often when enabling things in the scene, we also want them to fade in nicely. The way this is done now is by 
1. Set opacity to 0
2. Enable renderable
3. Change opacity from 0 to 1 over a number of seconds

It would be nice if this could be done automatically by just one click in the GUI. For example by holding CTRL when clicking the checkbox to enable the renderable

The GUI could check if the SGN has an opacity property and handle the logic for the fading

### LINKED ISSUE #2170: Improve connection between fade and enabled/disabled
Discussion to be had, but in general, if the object is not visible due to a parameter that is not visible to the user (e.g. using "Fade" instead of "Opacity"), we should consider it as being disabled. 

Work is being done on making the GUI's"Enabled checkbox" take this into account, but it could be a concept that we build into the c++ parts of the code as well. For example, not running the rendering code if Fade == 0

### LINKED ISSUE #2421: Add Fade property to more things
Such as globe layers, components, labels, etc. All of which we want to be able to programmatically fade in/out. 

And while looking at this, maybe also add a `openspace.fadeIn(IdentifierOfRenderedThingy)` function (and fade out)

WeirdRubberDuck:
Adds a fadeDuration property in the OpenSpace engine, that can be accessed by the UI to do automatic fading when clicking the enabled checkboxes. See separate PR in GUI for more details about that: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/141
Fixes #1827

This PR also adds the `Fade` property to more object, such as Globelabels, layers, labelscomponents and (fixes #2421). This is achieved through the `Fadeable` parent class, that encapsulates the opacity and fade properties, as well as provides the option to connect a Fadeable to another Fadeable (e.g. a parent propertowner of a subpropertyowner) so that it fades out together with the parent.

Part of this involved rewriting the Layers so that the opacity property moved out of the settings, and the LabelsComponent so that it was given its own enabled property (replaces the previous "DrawLabels" checks). **There are both breaking changes**
 
 
 The UI part of this PR also addresses / fixes #2170, by making faded out items appear disabled in the UI. Also, the isVisible function now takes fade into account


WeirdRubberDuck:
OBS! The "fade in atmosphere" feature is currently not at all part of this. Maybe that feature should just set the fade property? That way it would be reflected in the UI when it's faded out

WeirdRubberDuck:
Ready for merge? @alexanderbock @Roxeena ?

WeirdRubberDuck:
@alexanderbock @ylvaselling (and @Roxeena ?) let me know if you thing these PRs are ready for merge. I consider them done now

Roxeena:
I tried it out and looked at it and I think it is ready to be merged 👍

WeirdRubberDuck:
I'll go ahead and merge it; need the Fadeable code for the GeoJson work

PR #2560, opened 2023-03-22
added action for global lighting of all globes; closes #2494; also fi…

### LINKED ISSUE #2494: Global lighting option
Two users asked about the Global/Standard illumination options that were added. They want an option/action to set this globally instead of having to do it for each planet.

This "flashlight" mode could also apply to 3d models, adding a camera light source if one is not present.

micahnyc:
…x casing

### DIFF
--- a/data/assets/actions/planets/planet_lighting.asset
+++ b/data/assets/actions/planets/planet_lighting.asset
@@ -45,11 +45,11 @@ end
 
 local function getIlluminationAction(node, node_lower, global)
   local actionString = "_standard_illumination"
-  local actionName = "Standard Illumination"
+  local actionName = "standard illumination"
   local actionCommand = getIlluminationCommand(node, global)
   if (global) then
     actionString = "_global_illumination"
-    actionName = "Global Illumination"
+    actionName = "global illumination"
     actionCommand = getIlluminationCommand(node, global)
   end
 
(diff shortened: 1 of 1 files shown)


data/assets/actions/planets/planet_lighting.asset
```diff
@@ -95,6 +95,41 @@ local saturn_standard = getIlluminationAction("Saturn", "saturn", false)
 asset.export("titan_standard", saturn_standard.Identifier)
 
 
+--handling Earth special
+--if we have #2258 we can do all globes with night layers instead of just earth
+--if we have #1826 we can do the default night layer instead of the first
```
alexanderbock, line 100: Missing space between the `--` and the first character of the comment

data/assets/actions/planets/planet_lighting.asset
```diff
@@ -95,6 +95,41 @@ local saturn_standard = getIlluminationAction("Saturn", "saturn", false)
...
+local all_globes_global = {
```
alexanderbock, line 101: Maybe a better name required?  `all_globes_global_illumination` for example?

data/assets/actions/planets/planet_lighting.asset
```diff
@@ -95,6 +95,41 @@ local saturn_standard = getIlluminationAction("Saturn", "saturn", false)
...
+  Identifier = "os.all_globes_global_lighting",
+  Name = "All globes global illumination",
+  Command = [[
+    openspace.setPropertyValue("Scene.*.Renderable.PerformShading", false)
```
alexanderbock, line 105: This will also disable the shading on all RenderableModels, too

Issue #929, opened 2019-07-05
Using caching MRF files for globebrowsing performance


alexanderbock:

alexanderbock:
Preliminary test concluded that the performance was not improved.  Will reinvestigate when the HDR branch is in `master`

PR #2565, opened 2023-03-28
Add ability to cache globe data using MRFs

jockekilby:
If activated, data is cached using [caching mrf's](https://github.com/nasa-gibs/mrf/blob/master/doc/MUG.md#caching-mrf) as it is loaded from the original dataset. Subsequent reads of tiles that have been cached do not query the original dataset but picks it up from the mrf data.

Supports anything that can be used a dataset for a gdal raster but raw images such as JPGs are explicitly disallowed as they lack necessary geotransform data.

Solves #929 

* Activated on a global scale by setting the following keys in openspace.cfg (ModuleConfigurations/GlobeBrowsing)
```
MRFCacheEnabled = true,
MRFCacheLocation = "<somepath>",
```

* Settings can be overriden on a per-layer basis by adding the following entries to the layer entity of an asset
```lua
CacheSettings = { Enabled = true, Compression = "LERC", BlockSize = 512, Quality=25 },
```
The following per-layer settings are available and override the global module settings:
* Enabled : enable/disable caching for this layer. 
* Compression: The compression algorithm to use for tile cache storage. (JPEG, PNG, or LERC)
* BlockSize: Size of tiles in cache, should be a multiple of 2 although this is not required.
* Quality: The quality setting of the JPEG compression (when used)

Note that heightlayers must use LERC compresison and this is the default unless overridden by a layer.


alexanderbock:
Had a crash when starting up for the first time:
```
        MRFCacheEnabled = true,
        MRFCacheLocation = "${BASE}/cache_mrf"
```

Exception thrown in temporaltileprovider.cpp
image

image

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -519,6 +522,87 @@ void RawTileDataReader::initialize() {
         }
     }
 
+    if (_cacheProperties.enabled) {
+        ZoneScopedN("MRF Caching");
+
+        std::string datasetIdentifier = std::to_string(std::hash<std::string>{}(_datasetFilePath));
+        std::stringstream path;
+        path << module.mrfCacheLocation() << "/" << _cacheProperties.path << "/" << datasetIdentifier << "/";
```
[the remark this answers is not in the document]
alexanderbock: Can be replaced with:
```
std::string path = fmt::format("{}/{}/{}/", module.mrfCacheLocation(), _cacheProperties.path, datasetIdentifier);
```

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -519,6 +522,87 @@ void RawTileDataReader::initialize() {
...
+        std::string root = absPath(path.str()).string();
+        std::string mrf = root + datasetIdentifier + ".mrf";
+        std::string cache = root + datasetIdentifier + ".mrfcache";
```
alexanderbock, line 533: Or even here

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -519,6 +522,87 @@ void RawTileDataReader::initialize() {
...
+                // Already existing directories causes a 'failure' but no error
+                if (ec) {
+                    throw ghoul::RuntimeError(fmt::format(
+                        "Failed to create directories for cache at: {}. Error Code: {}, message: {}",
+                        root, std::to_string(ec.value()), ec.message()
+                    ));
+                }
+            }
+
+            auto* driver = GetGDALDriverManager()->GetDriverByName("MRF");
```
alexanderbock, line 547: We refrain from using `auto` unless it is to shorten long names or when we'd otherwise duplicate the class name to increase readability;
so  -> `GDALDriver* driver = ...`

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -519,6 +522,87 @@ void RawTileDataReader::initialize() {
...
+                        "Failed to load dataset: {}. GDAL Error: {}",
+                        _datasetFilePath, CPLGetLastErrorMsg()
+                    ));
+                }
+
+                {
+                    // Check if there is a geotransform present, if not assume its bounds are
+                    // [-180, 180] W<->E and [-90, 90] N<->S and North-up
+                    double geoTransform[6];
+                    auto err = src->GetGeoTransform(geoTransform);
```
alexanderbock, line 561: Same here with `CPLErr`

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -519,6 +522,87 @@ void RawTileDataReader::initialize() {
...
+                        if (err != CPLErr::CE_None) {
+                            throw ghoul::RuntimeError(fmt::format(
+                                "Failed to set default Geo Transform of dataset: {}. GDAL Error: {}",
+                                _datasetFilePath, CPLGetLastErrorMsg()
+                            ));
+                        }
+                    }
+                }
+
+                char** createOpts = NULL;
```
alexanderbock, line 579: -> `nullptr`

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -519,6 +522,87 @@ void RawTileDataReader::initialize() {
...
+                auto dst = static_cast<GDALDataset*>(driver->CreateCopy(mrf.c_str(), src, FALSE, createOpts, NULL, NULL));
+                if (!dst) {
+                    throw ghoul::RuntimeError(fmt::format(
+                        "Failed to create MRF Caching dataset dataset: {}. GDAL Error: {}",
+                        mrf, CPLGetLastErrorMsg()
+                    ));
+                }
+                GDALClose(dst);
+                GDALClose(src);
+            } else {
```
alexanderbock, line 598: We place the `else` on a new line after the closing `}`

modules/globebrowsing/src/rawtiledatareader.h
```diff
@@ -53,7 +54,7 @@ class RawTileDataReader {
      * \param baseDirectory, the base directory to use in future loading operations
      */
     RawTileDataReader(std::string filePath, TileTextureInitData initData,
-        PerformPreprocessing preprocess = PerformPreprocessing::No);
+        TileCacheProperties cacheProperties, PerformPreprocessing preprocess = PerformPreprocessing::No);
```
alexanderbock, line 57: Exceeding line length?

modules/globebrowsing/src/tilecacheproperties.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifndef __OPENSPACE_MODULE_GLOBEBROWSING___TILE_CACHE_PROPERTIES___H__
+#define __OPENSPACE_MODULE_GLOBEBROWSING___TILE_CACHE_PROPERTIES___H__
+
+#include <string>
+
+namespace openspace::globebrowsing {
+
+struct TileCacheProperties {
+    bool enabled;
```
alexanderbock, line 33: If there are good default values, initializing the member variables to some known-good state?
`bool enabled = false;`  etc to prevent eventual reading of uninitialized memory

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -105,24 +127,56 @@ DefaultTileProvider::DefaultTileProvider(const ghoul::Dictionary& dictionary)
...
 
+    // Get the name of the layergroup to which this layer belongs
+    auto it = std::find_if(
+        layers::Groups.begin(),
+        layers::Groups.end(),
+        [id = _layerGroupID](const layers::Group& gi) {
+            return gi.id == id;
+        }
+    );
+    auto layerGroup = it != layers::Groups.end() ? it->name : std::to_string(static_cast<int>(_layerGroupID));
```
alexanderbock, line 138: Line length;

`auto` is okay is here since it is an iterator that would be long to write otherwise

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -105,24 +127,56 @@ DefaultTileProvider::DefaultTileProvider(const ghoul::Dictionary& dictionary)
...
+
+    auto identifier = p.identifier.value_or("unspecified");
```
alexanderbock, line 140: Better to use actual type here

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -105,24 +127,56 @@ DefaultTileProvider::DefaultTileProvider(const ghoul::Dictionary& dictionary)
...
+    auto enclosing = p.globeName.value_or("unspecified");
+
+    std::stringstream path;
+    path << "/" << enclosing << "/" << layerGroup << "/" << identifier << "/";
```
alexanderbock, line 144: Same with `fmt::format` as above as it is faster and easier to read (and C++ standard with std::format which not all compilers support yet)

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -105,24 +127,56 @@ DefaultTileProvider::DefaultTileProvider(const ghoul::Dictionary& dictionary)
...
+
+    GlobeBrowsingModule& module = *global::moduleEngine->module<GlobeBrowsingModule>();
+    auto enabled = module.isMRFCachingEnabled();
+
+    std::string compression = _layerGroupID == layers::Group::ID::HeightLayers ? "LERC" : "JPEG";
```
alexanderbock, line 149: Line length

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -105,24 +127,56 @@ DefaultTileProvider::DefaultTileProvider(const ghoul::Dictionary& dictionary)
...
+    auto layerGroup = it != layers::Groups.end() ? it->name : std::to_string(static_cast<int>(_layerGroupID));
+
+    auto identifier = p.identifier.value_or("unspecified");
+    auto enclosing = p.globeName.value_or("unspecified");
+
+    std::stringstream path;
+    path << "/" << enclosing << "/" << layerGroup << "/" << identifier << "/";
+
+    GlobeBrowsingModule& module = *global::moduleEngine->module<GlobeBrowsingModule>();
+    auto enabled = module.isMRFCachingEnabled();
```
alexanderbock, line 147: -> `bool`

modules/globebrowsing/src/tileprovider/defaulttileprovider.h
```diff
@@ -57,6 +57,7 @@ class DefaultTileProvider : public TileProvider {
     layers::Group::ID _layerGroupID = layers::Group::ID::Unknown;
     bool _performPreProcessing = false;
     bool _padTiles = true;
+    TileCacheProperties _cacheProperties{ false };
```
alexanderbock, line 60: We prefer to use the `= ` notation since to me this looks more like a function definition.   If the `enabled = false` is set as a default argument in the `struct`, then here we only need `TileCacheProperties _cacheProperties;` anyway

jockekilby:
I had made an oopsie in default value initialisation which caused blockSize and quality settings to have unreasonable values 😄 

See [here](https://github.com/OpenSpace/OpenSpace/pull/2565/commits/0dbc1fb6111af4ece419b7ca9d1b9751a9e317da)

ylvaselling:
Maybe the crash wasn't fixed - in that case please ignore this: got a crash on start  
image

Set the config to `ModuleConfigurations = {
    GlobeBrowsing = {
        MRFCacheEnabled = true,
        MRFCacheLocation = "${BASE}/mrf_cache"
        }
        `
        `(D) Asset                Initializing asset "C:\\Users\\ylvaselling\\Documents\\Work\\OpenSpace\\data\\assets\\scene\\solarsystem\\sun\\layers\\colorlayers\\sun_texture.asset"`

alexanderbock:
It's a different crash;  the `getMRFCache` function is not returning anything if `std::filesystem::exists(mrf)`

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -439,6 +442,94 @@ RawTileDataReader::~RawTileDataReader() {
     }
 }
 
+std::optional<std::string> RawTileDataReader::getMRFCache() {
```
alexanderbock, line 445: Should be `mrfCache` since we only use `get` for functions that take in a reference or pointer, not returning the value in the return value

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -439,6 +442,94 @@ RawTileDataReader::~RawTileDataReader() {
...
+    GlobeBrowsingModule& module = *global::moduleEngine->module<GlobeBrowsingModule>();
+
+    std::string datasetIdentifier =
+        std::to_string(std::hash<std::string>{}(_datasetFilePath));
+    std::string path = fmt::format("{}/{}/{}/",
+        module.mrfCacheLocation(), _cacheProperties.path, datasetIdentifier);
+    std::string root = absPath(path).string();
```
alexanderbock, line 452: Can do `std::filesystem::path root = absPath(path);`

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -439,6 +442,94 @@ RawTileDataReader::~RawTileDataReader() {
...
+            createOpts = CSLSetNameValue(createOpts, "CACHEDSOURCE", _datasetFilePath.c_str());
+            createOpts = CSLSetNameValue(createOpts, "NOCOPY", "true");
+            createOpts = CSLSetNameValue(createOpts, "uniform_scale", "2");
+            createOpts = CSLSetNameValue(createOpts, "compress", _cacheProperties.compression.c_str());
+            createOpts = CSLSetNameValue(createOpts, "quality", std::to_string(_cacheProperties.quality).c_str());
+            createOpts = CSLSetNameValue(createOpts, "blocksize", std::to_string(_cacheProperties.blockSize).c_str());
+            createOpts = CSLSetNameValue(createOpts, "indexname", cache.c_str());
+            createOpts = CSLSetNameValue(createOpts, "DATANAME", cache.c_str());
+
+            auto dst = static_cast<GDALDataset*>(driver->CreateCopy(mrf.c_str(), src, FALSE, createOpts, NULL, NULL));
```
alexanderbock, line 513: `FALSE` -> `false`
`NULL` -> nullptr

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -439,6 +442,94 @@ RawTileDataReader::~RawTileDataReader() {
...
+                    root, std::to_string(ec.value()), ec.message()
+                ));
+                return std::nullopt;
+            }
+        }
+
+        GDALDriver* driver = GetGDALDriverManager()->GetDriverByName("MRF");
+        if (driver != nullptr) {
+            auto src = static_cast<GDALDataset*>(GDALOpen(_datasetFilePath.c_str(), GA_ReadOnly));
+            if (!src) {
```
alexanderbock, line 472: Can do `defer { GDALClose(src); }` to ensure the dataset is closed even in the cases that an `std::nullopt` is returned

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -70,6 +73,25 @@ namespace {
         // Determines if the tiles should be preprocessed before uploading to the GPU
         std::optional<bool> performPreProcessing;
 
+        struct CacheSettings {
+            // Specifies whether to use caching or not
+            std::optional<bool> enabled;
+
+            // The compression algorithm to use for cached tiles
+            std::optional<std::string> compression;
```
alexanderbock, line 81: Can do:
```cpp
enum class Compression {
   PNG,
   JPEG,
   LERC
};

std::optional<Compression> compression;
```

And people can pass in a string that is checked against one of the enum values

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -70,6 +73,25 @@ namespace {
...
+
+            // The quality setting of the compression alogrithm, only valid for JPEG
+            std::optional<int> quality;
```
alexanderbock, line 84: `std::optional<int> quality [[codegen::inrange(0, 100)]];

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
```diff
@@ -70,6 +73,25 @@ namespace {
...
+
+            // The block-size of the MRF cache
+            std::optional<int> blockSize;
```
alexanderbock, line 87: std::optional<int> blockSize [[codegen::greater(0)]];

modules/globebrowsing/src/rawtiledatareader.cpp
jockekilby: Nifty!

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
jockekilby: That's not the only types of compression though. I opted for a simple string property to avoid having to explicitly add each option, but we can if you think it makes sense.

https://github.com/nasa-gibs/mrf/blob/master/doc/MUG.md#types-of-tile-compressions-supported-by-mrf

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
jockekilby: 👍

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
jockekilby: 👍

jockekilby:
crash (re)fixed 😄 
rebased & (force) pushed
updated for all comments aside from compression comment by @alexanderbock

openspace.cfg
```diff
@@ -137,12 +137,9 @@ Paths = {
...
 ModuleConfigurations = {
     GlobeBrowsing = {
-        WMSCacheEnabled = false,
-        -- OfflineMode = true,
-        -- NoWarning = true,
-        WMSCacheLocation = "${BASE}/cache_gdal",
-        WMSCacheSize = 1024, -- in megabytes PER DATASET
-        TileCacheSize = 2048 -- for all globes (CPU and GPU memory)
+        TileCacheSize = 2048, -- for all globes (CPU and GPU memory)
+        MRFCacheEnabled = true,
```
alexanderbock, line 141: I don't think we want it to be `true` by default.  Would be good to have the key in here, but defaulted to `false` instead

alexanderbock:
Ah, that should say `constexpr std::array<std::string_view, 11> unsupported = {`

modules/globebrowsing/src/tileprovider/defaulttileprovider.cpp
jockekilby: Fixed as we discussed

jockekilby:
Undrafted & ready for review & tests

alexanderbock:
https://tenor.com/view/do-it-star-wars-gif-4928619

PR #2570, opened 2023-03-30
Limit min/max distance

Roxeena:
Adds the possibility to limit the orbital navigator and the touch interaction to not zoom closer or further than a set limit. It also adds Lua functions for the orbital navigator to set these distances based on the interaction sphere of the anchor.


alexanderbock:
I'm wondering if it would make sense to move the minimum allowed distance and maximum allowed distance together into the new PropertyOwner rather than keeping them separate?  It would break a few people, but logically it would make more sense to me

include/openspace/navigation/orbitalnavigator.h
```diff
@@ -112,6 +114,9 @@ class OrbitalNavigator : public properties::PropertyOwner {
     void setRetargetInterpolationTime(float durationInSeconds);
     void updatePreviousStateVariables();
 
+    void setMinimumAllowedDistance(float distance);
+    void setMaximumAllowedDistance(double distance);
```
alexanderbock, line 118: Why is one of them a `float` and the other a `double`?  I'd assume that both should be doubles?

include/openspace/navigation/orbitalnavigator.h
```diff
@@ -186,8 +197,18 @@ class OrbitalNavigator : public properties::PropertyOwner {
...
 
+    struct LimitZoomOut : public properties::PropertyOwner {
+        LimitZoomOut();
+
+        properties::BoolProperty enabled;
+        properties::FloatProperty maximumAllowedDistance;
+    };
+
+    LimitZoomOut _limitZoomOut;
+
```
alexanderbock, line 211: See comment regarding moving the minimum limits into this PropertyOwner

include/openspace/navigation/orbitalnavigator.h
```diff
@@ -186,8 +197,18 @@ class OrbitalNavigator : public properties::PropertyOwner {
...
     properties::BoolProperty _followAnchorNodeRotation;
     properties::FloatProperty _followAnchorNodeRotationDistance;
+    properties::BoolProperty _enableMinimumAllowedDistanceLimit;
     properties::FloatProperty _minimumAllowedDistance;
 
+    struct LimitZoomOut : public properties::PropertyOwner {
+        LimitZoomOut();
+
+        properties::BoolProperty enabled;
+        properties::FloatProperty maximumAllowedDistance;
```
alexanderbock, line 207: Currently, the maximum allowed distance is given in meters and the minimum allowed distance is given in x * interaction sphere.
It would be nice to harmonize that to use either for both of them.  Personally I think I would tend to using the interaction sphere rather than meters?

modules/touch/src/touchinteraction.cpp
```diff
@@ -147,13 +147,20 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo
-    ZoomBoundarySphereMultiplierInfo = {
-        "ZoomBoundarySphereMultiplier",
-        "Multiplies a node's boundary sphere by this in order to limit zoom & prevent "
+    ZoomInBoundarySphereMultiplierInfo = {
```
alexanderbock, line 150: The `{` would be on a new line

modules/touch/src/touchinteraction.cpp
```diff
@@ -147,13 +147,20 @@ namespace {
...
+        "ZoomInBoundarySphereMultiplier",
+        "Multiplies a node's boundary sphere by this in order to limit zoom in & prevent "
         "surface collision",
         "" // @TODO Missing documentation
     };
 
+    constexpr openspace::properties::Property::PropertyInfo
+    ZoomOutBoundarySphereMultiplierInfo = {
```
alexanderbock, line 158: Same here and `ZoomOutBoundarySphereMultiplierInfo` would be indented by one level

modules/touch/src/touchinteraction.cpp
```diff
@@ -300,7 +313,8 @@ TouchInteraction::TouchInteraction()
     addProperty(_zoomSensitivityExponential);
     addProperty(_zoomSensitivityProportionalDist);
     addProperty(_zoomSensitivityDistanceThreshold);
-    addProperty(_zoomBoundarySphereMultiplier);
+    addProperty(_zoomInBoundarySphereMultiplier);
+    addProperty(_zoomOutBoundarySphereMultiplier);
```
alexanderbock, line 317: Not sure about this;

would it make sense to make this a Vec2Property instead with the "Range" view attached?

src/navigation/orbitalnavigator.cpp
```diff
@@ -146,13 +148,35 @@ namespace {
         "this factor with the approximate radius of the node"
     };
 
+    constexpr openspace::properties::Property::PropertyInfo
+        EnabledMinimumAllowedDistanceLimitInfo = {
```
alexanderbock, line 152: `{` on a new line

src/navigation/orbitalnavigator.cpp
```diff
@@ -923,6 +966,29 @@ void OrbitalNavigator::updatePreviousStateVariables() {
...
+void OrbitalNavigator::setMinimumAllowedDistance(float distance) {
+    if (_limitZoomOut.enabled && distance > _limitZoomOut.maximumAllowedDistance) {
+        LWARNING("Setting minumum allowed distance larger than maximum allowed distance");
+    }
+
+    _minimumAllowedDistance = distance;
+}
+
+void OrbitalNavigator::setMaximumAllowedDistance(double distance) {
+    if (distance < 50.f) {
```
alexanderbock, line 978: -> `50.0` since it is a `double`

src/navigation/orbitalnavigator_lua.inl
```diff
@@ -0,0 +1,61 @@
...
+[[codegen::luawrap]] void setRelativeMinDistance(float multiplier) {
+    using namespace openspace;
+    const SceneGraphNode* node = global::navigationHandler->anchorNode();
+    if (!node) {
+        throw ghoul::lua::LuaError("Could not determine current focus node");
+    }
+
+    double is = node->interactionSphere();
+    global::navigationHandler->orbitalNavigator().setMinimumAllowedDistance(
+        is * multiplier
```
alexanderbock, line 39: I don't think this works as intended since currently the minimum allowed distance is a factor in itself, so this should just be `setMinimumAllowedDistance(multiplier)`

src/navigation/orbitalnavigator_lua.inl
```diff
@@ -0,0 +1,61 @@
...
+    );
+}
+
+/**
+* Set maximum allowed distance to a multiplier of the interaction sphere of the focus node
+*/
+[[codegen::luawrap]] void setRelativeMaxDistance(float multiplier) {
```
alexanderbock, line 46: -> `double`

include/openspace/navigation/orbitalnavigator.h
Roxeena: The properties are floats so I'm changing both of these to use floats instead

src/navigation/orbitalnavigator.cpp
```diff
@@ -1543,8 +1619,19 @@ glm::dvec3 OrbitalNavigator::pushToSurface(const glm::dvec3& cameraPosition,
     const double surfaceToCameraSigned = glm::length(actualSurfaceToCamera) *
         glm::sign(dot(actualSurfaceToCamera, referenceSurfaceOutDirection));
 
-    return cameraPosition + referenceSurfaceOutDirection *
-        glm::max(_minimumAllowedDistance - surfaceToCameraSigned, 0.0);
+    // Adjustment for if the camera is inside the min distance
+    double adjustment = 0.0;
+    if (std::abs(minHeight) > 0.0) {
```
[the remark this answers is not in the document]
Roxeena: Wouldn't that suffer from the same floating issue as checking `minHeight == 0.0`?

src/navigation/orbitalnavigator_lua.inl
Roxeena: I think we discussed this in a previous dev meeting.. Anyway this is not an issue since the property in the orbital navigator is a min/max distance in meters and the Lua function does the multiplier -> meters conversion using the bounding sphere

include/openspace/navigation/orbitalnavigator.h
alexanderbock: No, floats is fine;  was just assuming the wider type

modules/touch/src/touchinteraction.cpp
Roxeena: wont fix

PR #2574, opened 2023-04-03
Allow Window Config GUI to Edit Files

GPayne:
These changes allow the window configuration GUI to edit files that have been previously created by the GUI. This does not include valid SGCT config files that have functionality outside of the window GUI's subset defined by **sgcteditor.schema.json**.
Adds JSON schema validation support to SGCT, with tests in both SGCT and OpenSpace.


alexanderbock:
I can't read a configuration file that I created previously:
```
{
  "generator": {
    "major": 0,
    "minor": 19,
    "name": "OpenSpace"
  },
  "masteraddress": "localhost",
  "nodes": [
    {
      "address": "localhost",
      "port": 20401,
      "windows": [
        {
          "border": true,
          "id": 0,
          "pos": {
            "x": 50,
            "y": 50
          },
          "size": {
            "x": 1920,
            "y": 1080
          },
          "viewports": [
            {
              "pos": {
                "x": 0.0,
                "y": 0.0
              },
              "projection": {
                "fov": {
                  "down": 25.264999389648438,
                  "left": 40.0,
                  "right": 40.0,
                  "up": 25.264999389648438
                },
                "type": "PlanarProjection"
              },
              "size": {
                "x": 1.0,
                "y": 1.0
              },
              "tracked": true
            }
          ]
        }
      ]
    }
  ],
  "scene": {
    "orientation": {
      "w": 0.0,
      "x": 0.0,
      "y": 0.0,
      "z": 0.0
    }
  },
  "users": [
    {
      "eyeseparation": 0.06499999761581421,
      "pos": {
        "x": 0.0,
        "y": 0.0,
        "z": 4.0
      }
    }
  ],
  "version": 1
}
```

With message
image

I think the `monitor` tag should be optional?

alexanderbock:
If we are editing a file the dialog should switch from "Save As" to "Save" instead
image

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -637,20 +678,117 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
...
 }
 
-void LauncherWindow::openWindowEditor() {
-    SgctEdit editor(this, _userConfigPath);
-    int ret = editor.exec();
-    if (ret == QDialog::DialogCode::Accepted) {
-        sgct::config::Cluster cluster = editor.cluster();
+void LauncherWindow::editRefusalDialog(const std::string title, const std::string msg) {
+    QMessageBox msgBox;
+    msgBox.setText(QString::fromUtf8(msg.c_str()));
```
alexanderbock, line 683: Better to:
`QString::fromStdString(msg);` and `QString::fromStdString(title)`.  Our `title` and `msg` are not Utf8 anyway

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -637,20 +678,117 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
...
+            handleReturnFromWindowEditor(
+                editor.cluster(),
+                editor.saveFilename(),
+                saveWindowCfgPath
+            );
+        }
+    }
+    else {
+        try {
+            sgct::config::GeneratorVersion previewGenVersion =
```
alexanderbock, line 706: Could do a `using namespace sgct` here to make the rest of the `else` case shorter

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -637,20 +678,117 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
...
+                "This configuration file is valid for generating a display, but "
+                "its format does not match the window editor requirements and "
+                "cannot be opened in the editor"
+            );
+            if (previewGenVersion.versionCheck(minimumVersion)) {
+                try {
+                    preview = sgct::readConfig(
+                        winCfg,
+                        "This configuration file is unable to generate a proper display "
+                        "due to a problem detected in the readConfig function"
```
alexanderbock, line 725: -> `'readConfig` function?

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -637,20 +678,117 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
...
+                    );
+                }
+                catch (std::runtime_error& e) {
+                    //Re-throw an SGCT error exception with the runtime exception message
+                    throw sgct::Error(
```
alexanderbock, line 730: I don't think we want to throw `sgct::Error`s from outside the SGCT application.  Would be better to throw some other type of error.   The error codes are not meant to be user-extendable

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -637,20 +678,117 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
...
+                    );
+                }
+            }
+            else {
+                editRefusalDialog(
+                    "File Format Version Error",
+                    fmt::format(
+                        "File '{}' does not meet the minimum required version of {}.",
+                        winCfg,
+                        minimumVersion.versionString()
```
alexanderbock, line 762: This line can be on the previous one without a line break needed (https://github.com/OpenSpace/OpenSpace/wiki/Coding-Style#handling-line-breaks)

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -65,15 +64,154 @@ namespace {
...
+}
+
+void SgctEdit::setupProjectionTypeInGui(sgct::config::Viewport& vPort,
+                                        WindowControl* wCtrl)
+{
+    std::visit(overloaded{
+        [&](sgct::config::CylindricalProjection p) {
+            if (p.quality && p.heightOffset) {
+                wCtrl->setProjectionCylindrical(
+                    static_cast<int>(p.quality.value()),
```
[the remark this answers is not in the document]
alexanderbock: Also can just write `*p.quality` instead of `p.quality.value()`

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -65,15 +64,154 @@ namespace {
...
+                    wCtrl->setProjectionEquirectangular(
+                        static_cast<int>(p.quality.value()),
+                        true
+                    );
+                }
+                else if (p.mapping ==
+                         sgct::config::SpoutOutputProjection::Mapping::Fisheye)
+                {
+                    wCtrl->setProjectionFisheye(
+                        static_cast<int>(p.quality.value()),
```
alexanderbock, line 201: same here

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -65,15 +64,154 @@ namespace {
...
+                        true
+                    );
+                }
+            }
+        },
+        [&](sgct::config::NoProjection p) { (void)p; },
```
alexanderbock, line 207: better to just ` [&](sgct::config::NoProjection) {},`

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -235,63 +376,101 @@ void SgctEdit::apply() {
...
-sgct::config::Cluster SgctEdit::generateConfiguration() const {
-    sgct::config::Cluster cluster;
-
-    sgct::config::Scene scene;
-    scene.orientation = _settingsWidget->orientation();
-    cluster.scene = std::move(scene);
+void SgctEdit::generateConfiguration() {
+    _cluster.scene = sgct::config::Scene();
+    _cluster.scene->orientation = _settingsWidget->orientation();
+    if (_cluster.nodes.size() == 0) {
```
alexanderbock, line 385: -> `_cluster.nodes.empty()`

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -235,63 +376,101 @@ void SgctEdit::apply() {
...
+void SgctEdit::generateConfigSetupVsync() {
     if (_settingsWidget->vsync()) {
-        sgct::config::Settings::Display display;
-        display.swapInterval = 1;
-        
-        sgct::config::Settings settings;
-        settings.display = display;
-
-        cluster.settings = settings;
+        if (!_cluster.settings || !_cluster.settings.value().display ||
```
alexanderbock, line 399: -> `_cluster.settings->display`

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -235,63 +376,101 @@ void SgctEdit::apply() {
...
+            !_cluster.settings.value().display.value().swapInterval)
```
alexanderbock, line 400: -> `_cluster.settings->display->swapInterval`

also need to check that `display` actually has a value first or else it will crash

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -235,63 +376,101 @@ void SgctEdit::apply() {
...
+        {
+            sgct::config::Settings::Display display;
```
alexanderbock, line 402: ->
```cpp
sgct::config::Settings settings = {
    .display = sgct::config::Settings::Display {
       .swapInterval = 1
    }
};
```

alternatively

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -235,63 +376,101 @@ void SgctEdit::apply() {
...
-    return cluster;
+void SgctEdit::generateConfigIndividualWindowSettings(sgct::config::Node& node) {
+    for (size_t i = 0; i < node.windows.size(); ++i) {
+        // First apply default settings to each window...
+        node.windows[i].id = i;
+        node.windows[i].draw2D = true;
+        node.windows[i].draw3D = true;
+        node.windows[i].viewports.back().isTracked = true;
+        node.windows[i].tags.erase(
+            std::remove_if(
```
alexanderbock, line 456: -> `std::remove(node.windows[i].tags.begin(), node.windows[i].tags.end(), "GUI")`?

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -43,23 +43,30 @@ namespace {
         "Primary", "Secondary", "Tertiary", "Quaternary"
     };
 
+    const int nQualityTypes = 10;
```
alexanderbock, line 46: `constexpr int nQualityTypes = 10;`

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -293,11 +302,11 @@ void WindowControl::createWidgets(const QColor& windowColor) {
...
-            "Planar Projection",
-            "Fisheye",
-            "Spherical Mirror Projection",
-            "Cylindrical Projection",
-            "Equirectangular Projection"
+            ProjectionTypes[0],
+            ProjectionTypes[1],
+            ProjectionTypes[2],
+            ProjectionTypes[3],
+            ProjectionTypes[4],
```
alexanderbock, line 309: No need for the extra `,` at the end

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -625,10 +634,30 @@ void WindowControl::resetToDefaults() {
...
+
+void WindowControl::setMonitorSelection(int monitorIndex) {
+    _monitor->setCurrentIndex(monitorIndex);
+}
+
 void WindowControl::showWindowLabel(bool show) {
     _windowNumber->setVisible(show);
 }
 
+void WindowControl::setWindowName(std::string& windowName) {
```
alexanderbock, line 653: Can be `const std::string&` instead

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -720,17 +749,55 @@ sgct::config::Window WindowControl::generateWindowInformation() const {
...
     if (!_windowName->text().isEmpty()) {
         window.name = _windowName->text().toStdString();
     }
-    return window;
+}
+
+void WindowControl::setProjectionPlanar(float hfov, float vfov) {
+    _planar.fovH->setValue(hfov);
+    _planar.fovV->setValue(vfov);
+    _projectionType->setCurrentIndex(static_cast<int>(ProjectionIndices::Planar));
```
alexanderbock, line 764: If this use is the only for the `ProjectIndices`, they shouldn't be an `enum class`, but defined with separate `constexpr int` instead, in my view

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
```diff
@@ -720,17 +749,55 @@ sgct::config::Window WindowControl::generateWindowInformation() const {
...
+void WindowControl::setProjectionEquirectangular(int quality, bool spoutOutput) {
+    setQualityComboBoxFromLinesResolution(quality, _equirectangular.quality);
+    _equirectangular.spoutOutput->setChecked(spoutOutput);
+    _projectionType->setCurrentIndex(
+        static_cast<int>(ProjectionIndices::Equirectangular)
+    );
+}
+
+void WindowControl::setQualityComboBoxFromLinesResolution(int lines, QComboBox* combo) {
+    for (unsigned int v = 0; v < nQualityTypes; ++v) {
```
alexanderbock, line 795: Would put in a:
`ghoul_assert(combo, "Invalid pointer")`
in the beginning to be sure

ylvaselling:
I'm not sure if we already discussed this but I tried the branch and just thought this error message is a bit confusing (tried to edit the single.json):

image

apps/OpenSpace/ext/launcher/src/sgctedit/windowcontrol.cpp
GPayne: The `ProjectionIndices` struct is used in other places as well. I agree that the casts are a bit verbose, but having these defined in an `enum` feels more readable and expected that individually assigning int values.

alexanderbock:
Same file from above I now get
image

Would be good to either get a more actionable error message or (better) if we could make it work anyway since I definitely created that file with the editor

apps/OpenSpace/ext/launcher/include/launcherwindow.h
```diff
@@ -79,30 +81,46 @@ Q_OBJECT
...
+    bool isUserConfigSelected() const;
+
 private:
     QWidget* createCentralWidget();
     void setBackgroundImage(const std::string& syncPath);
 
     void openProfileEditor(const std::string& profile, bool isUserProfile);
-    void openWindowEditor();
+    void openWindowEditor(const std::string& winCfg, bool isUserWinCfg);
+    void editRefusalDialog(const std::string title, const std::string msg);
```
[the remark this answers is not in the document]
alexanderbock: Still the case for `editRefusalDialog` function

alexanderbock:
Regarding Ylva's comment, we decided to grey out the edit button in that case

apps/OpenSpace/ext/launcher/include/launcherwindow.h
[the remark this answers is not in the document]
GPayne: Change

GPayne:
I added exceptions for "OpenSpace" 0.18 and 0.19 to allow these to be edited (they have to pass the schema validation though).
Also new: if file is invalid version or non-user config, then edit button is disabled with hint text visible when hovering over the button, but not grayed-out.

PR #2578, opened 2023-04-04
Adding static Settings to Renderable

alexanderbock:
Adds a new `struct` to the `Renderable` class for class-fixed settings that are passed into the constructor.  Right now, it supports setting:
 - `automaticallyUpdateRenderBin`:  Whether the render bin of the `Renderable` should be changed automatically based on the opacity and fade properties
 - `shouldUpdateIfDisabled`:  If it is true, the `Renderable` will have it's `update` method called even if it is disabled

There are some more various changes in the PR as well to cleanup the Renderable as well that should not change the behavior


include/openspace/rendering/renderable.h
```diff
@@ -54,6 +54,11 @@ class Camera;
 
 class Renderable : public properties::PropertyOwner {
 public:
+    struct Settings {
+        bool automaticallyUpdateRenderBin = true;
```
Roxeena, line 58: Is this value stored in the renderable somewhere? It looks like the `Settings` object is destroyed after the construction of the `Renderable` and cannot be accessed after that. For the models, I would like to somehow find out if the user specifically specified a `RenderBin` in the asset file so I won't do the automatic `Renderbin` switching there either

src/rendering/renderable.cpp
```diff
@@ -153,21 +153,27 @@ ghoul::mm_unique_ptr<Renderable> Renderable::createFromDictionary(
...
+    , _shouldUpdateIfDisabled(settings.shouldUpdateIfDisabled)
 {
     ZoneScoped;
 
-    // I can't come up with a good reason not to do this for all renderables
-    registerUpdateRenderBinFromOpacity();
-
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
+    if (p.renderBinMode.has_value()) {
```
Roxeena, line 169: This check is also done on line 213, maybe they can be combined?

PR #2590, opened 2023-04-10
disabled assets for profile start

micahnyc:
(used profile editor to make changes)

### DIFF
--- a/data/profiles/asteroids.profile
+++ b/data/profiles/asteroids.profile
@@ -65,12 +65,105 @@
     "url": "https://www.openspaceproject.com",
     "version": "1.0"
   },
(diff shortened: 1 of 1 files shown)


data/profiles/asteroids.profile
```diff
@@ -65,12 +65,105 @@
     "url": "https://www.openspaceproject.com",
     "version": "1.0"
   },
+  "properties": [
+    {
+      "name": "Scene.sssb_transneptunian_object_asteroid.Renderable.Fade",
```
alexanderbock, line 70: Their `enabled` state should probably also be set to `false`

PR #2608, opened 2023-04-13
Video module

Roxeena:
Adds a new module that can decode videos (using libmpv) and has a few renderables that can have a video attached to them, globes, spheres, planes, screenspace planes, etc. There is an example asset in the examples folder that you can use to test it out. 

There is a known render bug that will be fixed later. If you have a video plane or a video sphere and place it in front of an atmosphere, the atmosphere is still rendered on top when it shouldn't. This is an issue that we already had with our other planes and spheres too.

@ylvaselling do you have anything you want to add to this description?


modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+#include <ghoul/opengl/openglstatecache.h>
+
+namespace {
+    constexpr std::string_view _loggerCat = "VideoPlayer";
+
+    constexpr openspace::properties::Property::PropertyInfo VideoInfo = {
+        "Video",
+        "Video",
+        "This should point to the video that should be played. It can "
+        "be either a file path or a url to a youtube video."
```
alexanderbock, line 46: I don't think it can be a YouTube video anymore?

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+        return false;
+    }
+    return true;
+}
+
+void* getOpenGLProcAddress(void*, const char* name) {
+    return reinterpret_cast<void*>(
+        global::windowDelegate->openGLProcedureAddress(name)
+    );
+}
```
alexanderbock, line 139: These functions can be defined in the anonymous namespace above which makes them less visible to the outside

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+    mpv_observe_property(
+        _mpvHandle,
+        static_cast<uint64_t>(key),
+        keys[key],
+        formats[key]
+    );
+}
+
+void VideoPlayer::setPropertyStringMpv(const char* name, const char* value)
+{
```
alexanderbock, line 159: This `{` can be on the previous line

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+        switch (*p.playbackMode) {
+        case Parameters::PlaybackMode::RealTimeLoop:
+            _playbackMode = PlaybackMode::RealTimeLoop;
+            break;
+        case Parameters::PlaybackMode::MapToSimulationTime:
+            _playbackMode = PlaybackMode::MapToSimulationTime;
+            break;
+        default:
+            LERROR("Missing playback mode in VideoTileProvider");
+            throw ghoul::MissingCaseException();
```
alexanderbock, line 282: Everything here would be indented one level higher

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+        }
+    }
+
+    if (_playbackMode == PlaybackMode::MapToSimulationTime) {
+        if (!p.startTime.has_value() || !p.endTime.has_value()) {
+            LERROR("Video tile layer tried to map to simulation time but lacked start or"
+                " end time");
```
alexanderbock, line 289: The `);` goes on the next line

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+
+    // Starting MPV in a paused state seems to reduce problems with initialization
+    setPropertyStringMpv("pause", "");
+
+    //setPropertyStringMpv("load-stats-overlay", "");
+
+    //mpv_set_property_string(_mpvHandle, "script-opts", "autoload-disabled=yes");
+
+    // Verbose mode
+    mpv_set_property_string(_mpvHandle, "msg-level", "all=v");
```
alexanderbock, line 408: Can use the `setPropertyStringMpv` function here as well?

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+                updateFrameDuration();
+            }
+
+            LINFO(fmt::format("Detected fps: {}", *fps));
+            _seekThreshold = 2.0 * (1.0 / _fps);
+            break;
+        }
+        case MpvKey::Pause: {
+            int* videoIsPaused = reinterpret_cast<int*>(prop->data);
+            _isPaused = *videoIsPaused == 1;
```
alexanderbock, line 733: maybe `_isPaused = (*videoIsPaused == 1);`  to make it more obvious

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+
+                if (w == -1 || h == -1) {
+                    LERROR("Invalid width or height params");
+                    return;
+                }
+                resizeFBO(w, h);
+            }
+            break;
+        }
+
```
alexanderbock, line 798: Extra empty line

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+        default: {
+            throw ghoul::MissingCaseException();
+            break;
```
alexanderbock, line 801: No need for the `break` here, since the `throw` in the line above will yeet us out of this function anyway

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+    _videoResolution = glm::ivec2(width, height);
+
+    // Delete old FBO and texture
+    glDeleteFramebuffers(1, &_fbo);
+    _frameTexture.reset(nullptr);
+
+    createFBO(width, height);
+}
+
+VideoPlayer::~VideoPlayer() {}
```
alexanderbock, line 947: I'd put the destructor just after the constructor since they belong logically together

modules/video/src/videotileprovider.cpp
```diff
@@ -0,0 +1,160 @@
+/*****************************************************************************************
+ *                                                                                       *
+ * OpenSpace                                                                             *
+ *                                                                                       *
+ * Copyright (c) 2014-2022                                                               *
```
alexanderbock, line 5: -> 2023

modules/video/src/videotileprovider.cpp
```diff
@@ -0,0 +1,160 @@
...
+    _videoPlayer.initialize();
+}
+
+VideoTileProvider::~VideoTileProvider() {}
+
+void VideoTileProvider::internalDeinitialize() {
+    _videoPlayer.destroy();
+}
+
+} // namespace openspace::video
```
alexanderbock, line 160: -> `// namespace openspace`

modules/video/video.asset
```diff
@@ -0,0 +1,30 @@
+local layer = {
+  Identifier = "TestVideo",
+  Video = "C:/Users/ylvaselling/Documents/Work/Testmovies/result.mp4",
```
alexanderbock, line 3: Same as above;  if we need another test video, you can send it to me;  otherwise we can reuse one of the other test videos

alexanderbock: File should be moved to data/assets/examples either way if you want to keep it

modules/video/video.asset
```diff
@@ -0,0 +1,30 @@
+local layer = {
```
alexanderbock, line 1: This asset should require the earth asset since we depend on it

modules/video/video.asset
```diff
@@ -0,0 +1,30 @@
...
+
+asset.export("layer", layer)
+
+asset.meta = {
+  Name = "Aqua Modis (Temporal)",
+  Version = "1.1",
+  Description = "GIBS hosted layer",
+  Author = "NASA EOSDIS Global Imagery Browse Services",
+  URL = "https://earthdata.nasa.gov/eosdis/science-system-description/eosdis-components/gibs",
+  License = "NASA"
```
alexanderbock, line 29: Meta information needs to be updated

alexanderbock (changes requested):

modules/globebrowsing/CMakeLists.txt
```diff
@@ -146,18 +146,19 @@ target_precompile_headers(${globebrowsing_module} PRIVATE
   <map>
 )
 
+# Gdal
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/gdal_data DESTINATION modules/globebrowsing)
 
-if (WIN32)
+if(WIN32)
```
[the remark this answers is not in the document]
ylvaselling: My VSCode edits Cmakelists files just when I open them :'(

modules/video/src/videoplayer.cpp
```diff
@@ -0,0 +1,949 @@
...
+        case MpvKey::IsSeeking: {
+            bool* isSeekingBool = reinterpret_cast<bool*>(prop->data);
+            _isSeeking = *isSeekingBool;
+            break;
+        }
+        case MpvKey::Fps: {
+            double* fps = reinterpret_cast<double*>(prop->data);
+            if (*fps < glm::epsilon<double>()) {
+                _fps = 24.0;
+                LWARNING("Detected fps was 0. Falling back on 24 fps");
```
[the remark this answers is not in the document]
ylvaselling: Fps can be negative?

PR #2614, opened 2023-04-14
Updating assets that use updated Horizons data files

GPayne:
All assets that use Horizons ephemeris were checked recently for updated positions. For all files where the ephemeris had been updated, the new file was downloaded from JPL and added to the Utah data server with an updated version number.
For testing, it will be necessary to switch the openspace.cfg file's sync entry to:
`HttpSynchronizationRepositories = { "http://openspace.sci.utah.edu/request" }`
since data.openspaceproject.com does not yet have this data.


alexanderbock (approved):
Data is on the data.openspaceproject.com server as well now

PR #2622, opened 2023-04-17
macos build fixes

micahnyc:


modules/webbrowser/cmake/webbrowser_helpers.cmake
```diff
@@ -160,7 +160,9 @@ function(run_cef_linux_config CEF_TARGET CEF_ROOT)
 endfunction ()
 
 function(set_modules_dependency_on_cef_libraries LIB_DEPENDENT)
-  target_link_libraries(${LIB_DEPENDENT} INTERFACE libcef_lib)
+  if (WIN32 OR OS_LINUX)
+    target_link_libraries(${LIB_DEPENDENT} INTERFACE libcef_lib)
+  endif ()
```
alexanderbock, line 165: This seems like it is just patching another bug;  even on MacOS we ought to link against CEF?

modules/webbrowser/src/webbrowserapp.cpp
```diff
@@ -49,6 +49,8 @@ void WebBrowserApp::OnBeforeCommandLineProcessing(const CefString&,
     commandLine->AppendSwitch("use-mock-keychain");
     commandLine->AppendSwitch("enable-begin-frame-scheduling");
     commandLine->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");
+    commandLine->AppendSwitch("--disable-gpu-sandbox");
+    commandLine->AppendSwitch("--no-sandbox");
```
alexanderbock, line 53: Didn't we disable this some time ago @ylvaselling?   Maybe setting these flags is only necessary on Apple?

src/scene/scene.cpp
```diff
@@ -858,7 +858,11 @@ std::string makeIdentifier(std::string s) {
     // marks. Hence, we first convert '_' to whitespaces to avoid them being replaced
     // in the puncutation check
     std::replace(s.begin(), s.end(), '_', ' ');
-    std::replace_if(s.begin(), s.end(), std::ptr_fun<int, int>(&std::ispunct), '-');
+    std::replace_if(
+                    s.begin(),
+                    s.end(),
+                    [](char c) { return std::ispunct(c) == 0; },
+                    '-');
```
alexanderbock, line 865:
```suggestion
    std::replace_if(
        s.begin(),
        s.end(),
        [](char c) { return std::ispunct(c) == 0; },
        '-'
    );
```

modules/webbrowser/src/webbrowserapp.cpp
micahnyc: I think we do only need this on apple, it just felt cleaner to not have the ifdef apple in there, happy to add it thou

modules/webbrowser/cmake/webbrowser_helpers.cmake
micahnyc: I agree, it felt weird taking it out, but I was getting linker errors with it in, and working w/o out it....something to investigate

modules/webbrowser/src/webbrowserapp.cpp
ylvaselling: No, I haven't done anything with the sandbox flags. [Here](https://team-openspace.slack.com/archives/CBLM14A87/p1651014724397579) is the message which describes my findings from investigating the CEF flags

modules/webbrowser/cmake/webbrowser_helpers.cmake
jockekilby: Thought I could be of more assistance, sorry

alexanderbock:
Not sure what happened, but the list of files now have a whole lot of changes that don't have anything to do with the pull request itself.  That should be sorted out or else it will make the Git log harder to read

jockekilby:
I think the commit by @JohnstoneJ could simply be reverted to fix the extra code-changes @alexanderbock 
I might be missing something but it looks like the commit is a squashed series of duplicate commits..?

JohnstoneJ:
Yes please revert it, it was a mistake! Sorry about that

PR #2643, opened 2023-04-25
handled underscore and period special in c++ for cef bug; also remove…

### LINKED ISSUE #853: Disable CEF internal Keyboard shortcuts
Currently the CEF has a few internal keyboard commands like F5 for refresh (and maybe CTRL+T) we don't want these things to happen as they can be considered "unexpected"

Perhaps some commands like refresh can be exposed as a lua command, but not *necessary* for this issue.

### LINKED ISSUE #896: Windows CEF misinterprets keyCodes
In our current windows cef gui, you cannot type . ' _ < " and a few other things. Right now this is mostly a problem for . however I see it becoming more of a problem moving forward. For example, when typing underscore _ it actually pates from the clipboard into the input field.

After many hours of debugging, googling, hacking, googling, etc I believe that it's somehow and issue related to cef and glfw, and may or may not be specific to using cef in osr mode.

The only refernces I could find to these issues we have are here: https://github.com/BalazsJako/ImGuiColorTextEdit/issues/18
https://magpcss.org/ceforum/viewtopic.php?f=6&t=11874
https://bitbucket.org/chromiumembedded/cef/issues/2229
https://bitbucket.org/chromiumembedded/cef/issues/2228/cefclient-incorrectly-translates-key-codes
https://bitbucket.org/chromiumembedded/cef/issues/2139/cefclient-osr-delivers-key-events-with

The first one is actually an issue in an ImGui module, and our ImGui works fine, and we still get this error in CEF without the imgui module enabled in cmake. But it mentions glfw.

The second one has some things similar to what we are having too, 

The 3rd one seems related. 

The 4th seems to be the issue, and mentions a fix to cef. which says is fixed in the last link, but then apparently is not.

For now, I have made a incomplete fix in our gui front end to at least all a . to be put in so that decimal values can be entered.

micahnyc:
…d F5 special code and changed to keybinding; closes #853; closes #896

### DIFF
--- a/modules/webbrowser/src/eventhandler.cpp
+++ b/modules/webbrowser/src/eventhandler.cpp
@@ -447,17 +447,16 @@ bool EventHandler::specialKeyEvent(Key key, KeyModifier mod, KeyAction action) {
                 return true;
             }
             return false;
-        case Key::Period:
-            {
-                CefKeyEvent keyEvent;
-                keyEvent.windows_key_code = mapFromGlfwToWindows(Key::KeypadDecimal);
-                keyEvent.character = mapFromGlfwToCharacter(Key::KeypadDecimal);
-                keyEvent.native_key_code = mapFromGlfwToNative(Key::KeypadDecimal);
-                keyEvent.modifiers = static_cast<uint32>(mod);
-                keyEvent.type = keyEventType(action);
-                _browserInstance->sendKeyEvent(keyEvent);
-                return true;
-            }
+        case Key::Period: {
+            CefKeyEvent keyEvent;
+            keyEvent.windows_key_code = mapFromGlfwToWindows(Key::KeypadDecimal);
+            keyEvent.character = mapFromGlfwToCharacter(Key::KeypadDecimal);
+            keyEvent.native_key_code = mapFromGlfwToNative(Key::KeypadDecimal);
+            keyEvent.modifiers = static_cast<uint32>(mod);
+            keyEvent.type = keyEventType(action);
+            _browserInstance->sendKeyEvent(keyEvent);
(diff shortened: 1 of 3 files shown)


modules/webbrowser/src/eventhandler.cpp
```diff
@@ -427,17 +427,37 @@ bool EventHandler::keyboardCallback(Key key, KeyModifier modifier, KeyAction act
...
+                keyEvent.character = mapFromGlfwToCharacter(Key(45));
+                keyEvent.native_key_code = mapFromGlfwToNative(Key(45));
+                keyEvent.modifiers = static_cast<uint32>(mod);
+                keyEvent.type = keyEventType(action);
+                _browserInstance->sendKeyEvent(keyEvent);
+                return true;
+            }
+            return false;
+        case Key::Period:
+            {
```
WeirdRubberDuck, line 451: Should be indented one step less and cna move start bracket to same line as case: 
```
case Key::Period: {
    CefKeyEvent keyEvent;
    keyEvent.windows_key_code = mapFromGlfwToWindows(Key::KeypadDecimal);
    keyEvent.character = mapFromGlfwToCharacter(Key::KeypadDecimal);
    keyEvent.native_key_code = mapFromGlfwToNative(Key::KeypadDecimal);
    keyEvent.modifiers = static_cast<uint32>(mod);
    keyEvent.type = keyEventType(action);
    _browserInstance->sendKeyEvent(keyEvent);
    return true;
}
```

WeirdRubberDuck (approved):
Wow, this makes me so happy 🤩

alexanderbock (approved):

PR #2658, opened 2023-04-27
added multiple and offset to blue marble height to make it similar lo…

micahnyc:
…oking to esri terrain

### DIFF
--- a/data/assets/scene/solarsystem/planets/earth/layers/heightlayers/blue_marble_height.asset
+++ b/data/assets/scene/solarsystem/planets/earth/layers/heightlayers/blue_marble_height.asset
@@ -12,7 +12,11 @@ local layer = {
   Identifier = "Earth_Bluemarble_Height",
   Enabled = asset.enabled,
   FilePath = texturesPath .. "earth_bluemarble_height.jpg",
-  Description = "Topographic layer from Blue Marble Next Generation"
+  Description = "Topographic layer from Blue Marble Next Generation",
+  Settings = {
+    Multiplier = 40,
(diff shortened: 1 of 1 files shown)


alexanderbock:
Do we know _why_ the layers are misaligned?  The Earth reference ellipsoid is very well defined and it seems like we could fix this a better way than hacking a few correction values

micahnyc:
> 

Not sure, I'm guessing the layer comes from this page: https://visibleearth.nasa.gov/images/73934/topography 

It states "Data in these images were scaled 0-6400 meters. " So maybe a multiplier of 64 instead of 40?

alexanderbock:
Ouh.. that's weird.  Maybe we should reach out to them what they mean exactly to do the correct thing?  Since there are heights on earth <0 and >6400

alexanderbock:
:+1:

data/assets/scene/solarsystem/planets/earth/layers/heightlayers/blue_marble_height.asset
```diff
@@ -12,7 +12,11 @@ local layer = {
   Identifier = "Earth_Bluemarble_Height",
   Enabled = asset.enabled,
   FilePath = texturesPath .. "earth_bluemarble_height.jpg",
-  Description = "Topographic layer from Blue Marble Next Generation"
+  Description = "Topographic layer from Blue Marble Next Generation",
+  Settings = {
+    Multiplier = 40,
```
alexanderbock, line 17: Adding a comment where the numbers come from

PR #2699, opened 2023-05-16
Issue/2227 - Option to just specify a scene graph node in profile editor camera dialog

WeirdRubberDuck:
Adds a tab for specifying a scene graph node (and makes that the default one)

Also added some descriptive text to each tab. Let me know what you think of them.

Used the already existing camera path waypoint creation code to compute the camera position, but unfortunately, the actual camera pose can't be computed until we reach the "first correct frame", so the code ended up being much uglier than I wanted it to be... (the check in navigationhandler.cpp). But it works, and that's what most important by now :) 

Do we also want to allow the user to set an optional height value? Since this is supported by the data structure used for the camera position, this might make sense

### DIFF
--- a/include/openspace/navigation/waypoint.h
+++ b/include/openspace/navigation/waypoint.h
@@ -63,7 +63,7 @@ private:
  */
 Waypoint waypointFromCamera();
 
-struct NodeInfo {
+struct NodeCameraStateSpec {
     std::string identifier;
     std::optional<glm::dvec3> position;
     std::optional<double> height;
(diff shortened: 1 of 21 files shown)


WeirdRubberDuck:
Still have to create tests, but feel free to check out the implementation before that

WeirdRubberDuck:
Oh, do I need to update the profile version somehow? @GPayne @alexanderbock

alexanderbock:
1. Specifying an optional height would be nice, but not necessary in my view
2. Yea, updating the Profile version 1.3 is necessary to prevent people from creating a profile in 0.19.0 and trying to load it in 0.18 and crashing hard

To do that:
1. Duplicate the code in https://github.com/OpenSpace/OpenSpace/blob/issue/2227/src/scene/profile.cpp#L596-L608
2. `namespace` should be `version12`
3. Function name should be:
```cpp
void convertVersion12to13(nlohmann::json& profile) {
    // Version 1.3 introduced to GoToNode camera initial position
    profile["version"] = Profile::Version{ 1, 3 };
}
```
4. In https://github.com/OpenSpace/OpenSpace/blob/issue/2227/src/scene/profile.cpp#L740-L743 add another check afterwards if the major version is 1.2 and if so, call the convertVersion12to13 function

include/openspace/navigation/waypoint.h
```diff
@@ -51,10 +52,42 @@ class Waypoint {
...
+ *             information is the identifier of the node, but a position or height
+ *             above the bounding sphere may also be given.
+ * \param startPoint an optional previous waypoint. If not specified, the current camera
+ *                   position will be used.
+ * \param userLinear if true, the new waypoint will be computed along a straight line
+ *                   from the start waypoint to the scene graph node or position.
+ * \return the computed WayPoint
+ */
+Waypoint computeWaypointFromNodeInfo(const NodeCameraStateSpec& spec,
+    std::optional<const Waypoint> startPoint = std::nullopt, bool useLinear = false);
```
alexanderbock, line 89: -> `std::optional<Waypoint>` would be enough since it is a parameter to the function

alexanderbock (approved):

WeirdRubberDuck:
Done :)

PR #2719, opened 2023-05-27
Issue/2531

GPayne:
In a multi-window configuration, this provides the ability to select different render options for window 1 if it is selected to be the only GUI window. Options are no rendering (only GUI), copying the projection config from another window, or its own unique settings.
This will also slightly change the behavior of older config files. In a multi-window config setup where the "GUI" tag is in window 1, rendering will now be done on window 1 unless `draw3d` is set to false.


apps/OpenSpace/ext/launcher/include/sgctedit/settingswidget.h
```diff
@@ -72,17 +74,81 @@ Q_OBJECT
...
+     * return a value even if the checkbox is not enabled.
+     * 
+     * \return -1 if in a disabled state (e.g. when showUiOnFirstWindow() returns false)
+     *          0 if no window is being mirrored (only the UI will appear in first
+     *            window, with no 3D rendering)
+     *          1 if the second window's viewport is mirrored in the first window
+     *          2 if the third window's viewport is mirrored in the first window
+     *          3 if the fourth window's viewport is mirrored in the first window
+     */
+    int graphicsSelectionForShowUiOnFirstWindow() const;
```
alexanderbock, line 96: I think it might be easier to use if this function would return an `enum class` that has a value for each of the numeric return values

apps/OpenSpace/ext/launcher/include/sgctedit/settingswidget.h
```diff
@@ -72,17 +74,81 @@ Q_OBJECT
...
+     * Sets value of the graphics selection combo box for which other window that the
+     * first window will mirror.
+     * 
+     * \param selection int value for the combo box selection.
+     *                  0 if no window is being mirrored (only the UI will appear)
+     *                  1 if the second window's viewport is mirrored in the first window
+     *                  2 if the third window's viewport is mirrored in the first window
+     *                  3 if the fourth window's viewport is mirrored in the first window
+     */
+    void setGraphicsSelectionForShowUiOnFirstWindow(int selection);
```
alexanderbock, line 108: Same here

apps/OpenSpace/ext/launcher/include/sgctedit/settingswidget.h
```diff
@@ -72,17 +74,81 @@ Q_OBJECT
...
     sgct::quat _orientationValue = sgct::quat(0.f, 0.f, 0.f, 0.f);
     QCheckBox* _checkBoxVsync = nullptr;
     QCheckBox* _showUiOnFirstWindow = nullptr;
+    QComboBox* _firstWindowGraphicsSelection = nullptr;
+    QBoxLayout* _firstWindowSelectionLayout = nullptr;
+    int _stateOfUiOnFirstWindowPreviousCount = 1;
+    bool _stateOfUiOnFirstWindowWhenDisabled = false;
+
+    void showUiOnFirstWindowClicked(bool checked);
+    void firstWindowGraphicsSelectionChanged(const QString &text);
```
alexanderbock, line 151: We place the function declarations as the first thing in the visibility block, so these two functions would come directly after the `private` and before the member variables

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -657,45 +657,49 @@ void LauncherWindow::populateWindowConfigsList(std::string preset) {
...
+        this,
+        &LauncherWindow::onNewWindowConfigSelection
+    );
+    // Call combobox selected callback to refresh the file status of the current selection
+    onNewWindowConfigSelection(_windowConfigBox->currentIndex());
+}
+
+void LauncherWindow::onNewWindowConfigSelection(int newIndex) {
+    std::filesystem::path pathSelected = absPath(selectedWindowConfig());
+    std::string fileSelected = pathSelected.generic_string();
```
alexanderbock, line 669:
```suggestion
    std::string fileSelected = pathSelected.string();
```

should work?

apps/OpenSpace/ext/launcher/src/sgctedit/settingswidget.cpp
```diff
@@ -78,13 +96,75 @@ bool SettingsWidget::vsync() const {
...
+    _firstWindowGraphicsSelection->setCurrentIndex(selection);
+}
+
 void SettingsWidget::setVsync(bool enableVsync) {
     _checkBoxVsync->setChecked(enableVsync);
 }
+
+void SettingsWidget::nWindowsDisplayedChanged(int newCount) {
+    constexpr int countOneWindow = 1;
+    constexpr int countTwoWindows = 2;
```
alexanderbock, line 125:
```suggestion
    constexpr int CountOneWindow = 1;
    constexpr int CountTwoWindows = 2;

```

Since they are constants

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -130,22 +130,71 @@ SgctEdit::SgctEdit(sgct::config::Cluster& cluster, const std::string& configName
...
+    for (size_t i = 0; i < nWindows; ++i) {
+        sgct::config::Window& w = _cluster.nodes.front().windows[i];
+        //First window needs to have "GUI" tag if this mode is set
+        if (i == 0) {
+            if (std::find(w.tags.begin(), w.tags.end(), "GUI") != w.tags.end()) {
+                firstWindowGuiIsEnabled = true;
+            }
+            else {
+                firstWindowGuiIsEnabled = false;
+            }
```
alexanderbock, line 158:
```suggestion
            auto it = std::find(w.tags.begin(), w.tags.end(), "GUI");
            firstWindowGuiIsEnabled = it != w.tags.end();
```

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -465,27 +541,84 @@ void SgctEdit::generateConfigIndividualWindowSettings(sgct::config::Node& node)
...
 
+void SgctEdit::deleteFromTags(sgct::config::Window& window) {
+    for (std::string tag : {"GUI", "GUI_No_Render", "GUI_Render_Win1",
+                            "GUI_Render_Win2", "GUI_Render_Win3", "GUI_Render_Win4"})
+    {
+        window.tags.erase(
+            std::remove(window.tags.begin(), window.tags.end(), tag),
+            window.tags.end()
+        );
+    }
```
alexanderbock, line 585:
```suggestion
    constexpr std::array<std::string_view, 6> Tags = {
        "GUI",
        "GUI_No_Render",
        "GUI_Render_Win1",
        "GUI_Render_Win2",
        "GUI_Render_Win3",
        "GUI_Render_Win4"
    };
    for (std::string_view tag : Tags) {
        window.tags.erase(
            std::remove(window.tags.begin(), window.tags.end(), tag),
            window.tags.end()
        );
    }
```

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -465,27 +541,84 @@ void SgctEdit::generateConfigIndividualWindowSettings(sgct::config::Node& node)
...
+}
+
 sgct::config::Cluster SgctEdit::cluster() const {
     return _cluster;
 }
+
+void SgctEdit::firstWindowGraphicsSelectionChanged(const QString &text) {
+    (void)text;
```
alexanderbock, line 593:
```suggestion
void SgctEdit::firstWindowGraphicsSelectionChanged(const QString&) {
```

apps/OpenSpace/ext/launcher/src/sgctedit/sgctedit.cpp
```diff
@@ -465,27 +541,84 @@ void SgctEdit::generateConfigIndividualWindowSettings(sgct::config::Node& node)
...
+    if (_settingsWidget->showUiOnFirstWindow()) {
+        _displayWidget->activeWindowControls()[0]->setVisibilityOfProjectionGui(
+            newSetting == 1
+        );
+    }
+}
+
+void SgctEdit::nWindowsDisplayedChanged(int newCount) {
+    if (!_settingsWidget)
+        return;
```
alexanderbock, line 608:
```suggestion
    if (!_settingsWidget) {
        return;
    }
```

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
GPayne: Yes, it works

apps/OpenSpace/ext/launcher/include/sgctedit/settingswidget.h
GPayne: I had forgotten to change the comment header which was outdated. I think it's less confusing now because the value matches the 1-based window number to use. If set to zero then no window is used (no graphics).
I prefer leaving it as an integer since that is what QComboBox uses, and the values for this and its set function are only used with that object.

apps/OpenSpace/ext/launcher/include/sgctedit/settingswidget.h
GPayne: Same response as for the getter function above

PR #2725, opened 2023-05-30
Switched multiple OpenGL calls to a single call for better efficiency

Arohdin:
Switched multiple OpenGL calls to single a call for better efficiency from `glDrawArrays` to `glMultiDrawArrays` .
This increases performance when rendering Main Asteroid Belt from 20 fps to 85 fps on my machine.

## Before
mainbelt_before

## After
mainbelt_after

### DIFF
--- a/modules/space/rendering/renderableorbitalkepler.cpp
+++ b/modules/space/rendering/renderableorbitalkepler.cpp
@@ -290,14 +290,11 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
 
     glLineWidth(_appearance.lineWidth);
 
-    const size_t nrOrbits = _segmentSize.size();
-    gl::GLint vertices = 0;
-
     //glDepthMask(false);
     //glBlendFunc(GL_SRC_ALPHA, GL_ONE)
 
-    GLint* _si = &_startIndex[0];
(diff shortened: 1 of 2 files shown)


modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -296,11 +296,11 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
     //glDepthMask(false);
```
alexanderbock, line 296: Just above this, the two lines declaring `nrOrbits` and `vertices` are no longer needed

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -296,11 +296,11 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
...
     //glBlendFunc(GL_SRC_ALPHA, GL_ONE)
 
+    GLint* _si = &_startIndex[0];
+    GLint* _ss = &_segmentSize[0];
```
alexanderbock, line 300: More elegant of writing this:
```suggestion
    GLint* _si = _startIndex.data();
    GLint* _ss = _segmentSize.data();
```

alexanderbock (approved):
Awesome!

PR #2728, opened 2023-05-31
Fix for date format in sssb v2 asteroids files (issue 2649)

GPayne:
The JPL SSSB site changed their date format from `YYYYMMDD` to `YYYY-MM-DD`. Our v1 asteroids files have the first date format, and our v2 files have the second.
This version will accept either of these two format, so that the older v1 files will still work.

**Important Note for Testing this PR:**
The crash mentioned in issue 2649 only occurs if OpenSpace was built in Debug mode.
In order to verify this fix, the following are required:

- Build & run OpenSpace in Debug mode
- Delete all cache folders that begin with _sssb_data__
- Delete all folders in sync/http that begin with _sssb_data__
- In openspace.cfg entry for `Sync{HttpSynchronizationRepositories}`, comment-out all other servers except for `http://openspace.sci.utah.edu/request`.

### DIFF
--- a/modules/space/kepler.cpp
+++ b/modules/space/kepler.cpp
@@ -262,12 +262,9 @@ namespace {
         size_t nDashes = std::count_if(
             epoch.begin(),
             epoch.end(),
-            []( char c ) {return c =='-';}
+            [](char c) { return c == '-'; }
         );
-        std::string formatString = "{:4}{:2}{:2}{}";
-        if (nDashes == 2) {
-            formatString = "{:4}-{:2}-{:2}{}";
-        }
+        std::string formatString = (nDashes == 2) ? "{:4}-{:2}-{:2}{}" : "{:4}{:2}{:2}{}";
         auto [res, year, monthNum, dayOfMonthNum, fractionOfDay] =
             scn::scan_tuple<int, int, int, double>(e, formatString);
         if (!res) {
--- a/modules/space/kepler.cpp
+++ b/modules/space/kepler.cpp
@@ -239,8 +239,9 @@ namespace {
     }
 
     double epochFromYMDdSubstring(const std::string& epoch) {
-        // The epochString is in the form:
+        // The epochString can be in one of two forms:
         // YYYYMMDD.ddddddd
+        // YYYY-MM-DD.ddddddd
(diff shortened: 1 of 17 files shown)


modules/space/kepler.cpp
```diff
@@ -258,8 +259,17 @@ namespace {
             e += ".0";
         }
         // 1, 2
+        size_t nDashes = std::count_if(
+            epoch.begin(),
+            epoch.end(),
+            []( char c ) {return c =='-';}
```
alexanderbock, line 265:
```suggestion
            [](char c) { return c == '-'; }
```

modules/space/kepler.cpp
```diff
@@ -258,8 +259,17 @@ namespace {
...
+        );
+        std::string formatString = "{:4}{:2}{:2}{}";
+        if (nDashes == 2) {
+            formatString = "{:4}-{:2}-{:2}{}";
+        }
```
alexanderbock, line 270: Consider:
```suggestion
        std::string formatString =
            nDashes == 2 ?
            "{:4}{:2}{:2}{}" :
            "{:4}-{:2}-{:2}{}";
```

modules/space/kepler.cpp
```diff
@@ -598,10 +608,12 @@ std::vector<Parameters> readSbdbFile(std::filesystem::path file) {
 
     std::string line;
     std::getline(f, line);
+    // Newer versions downloaded from the JPL SBDB website have " around variables
+    line.erase(remove(line.begin(), line.end(), '\"'), line.end());
     if (line != ExpectedHeader) {
         throw ghoul::RuntimeError(fmt::format(
             "Expected JPL SBDB file to start with '{}' but found '{}' instead",
-            ExpectedHeader, line
+            ExpectedHeader, line.substr(0, 100)
```
alexanderbock, line 616:
```suggestion
            ExpectedHeader, line.substr(0, std::min(line.size(), 100))
```
In case the line is shorter than 100 characters

alexanderbock (approved):

alexanderbock (approved):

alexanderbock:
Files are updated on the server

PR #2735, opened 2023-06-02
Fix for preventing adding property from scriptlog if already in add mode

GPayne:
The profile editor's property dialog now has two ways of adding a property: manual or fill-from-scriptlog. This change prevents user from being able to select the scriptlog fill option if the editor is currently in edit mode (adding or changing a property).


alexanderbock (approved):
Yey

PR #2736, opened 2023-06-02
Portrait window config for issue#2383

GPayne:
Adds a new window config with a 9:16 portrait render window. Uses a render-less GUI window to the side in landscape orientation, which is necessary in order to access the full GUI menu.


alexanderbock (approved):
Looks neat!  Still not used to vertical video after all these years ;)

PR #2747, opened 2023-06-07
Fixes line ending problems with horizons vector files on linux

GPayne:
This fix allows importing and rendering from horizons files in vector format by removing the former dependency on windows vs linux line endings.

### DIFF
--- a/modules/space/horizonsfile.cpp
+++ b/modules/space/horizonsfile.cpp
@@ -504,8 +504,9 @@ HorizonsResult readHorizonsFile(std::filesystem::path file) {
     // " Before data starts, Observer table doesn't
     std::string line;
     std::getline(fileStream, line);
+    const std::string vectorTabHeader = "JDTDB";
     while (line[0] != '$') {
-        if (line == "JDTDB") {
+        if (line.substr(0, vectorTabHeader.size()) == vectorTabHeader) {
             fileStream.close();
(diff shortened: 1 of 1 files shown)


modules/space/horizonsfile.cpp
```diff
@@ -504,8 +504,9 @@ HorizonsResult readHorizonsFile(std::filesystem::path file) {
     // " Before data starts, Observer table doesn't
     std::string line;
     std::getline(fileStream, line);
+    const std::string vectorTabHeader = "JDTDB";
```
alexanderbock, line 507:
```suggestion
    constexpr std::string_view VectorTabHeader = "JDTDB";
```

modules/space/horizonsfile.cpp
```diff
@@ -504,8 +504,9 @@ HorizonsResult readHorizonsFile(std::filesystem::path file) {
...
     while (line[0] != '$') {
-        if (line == "JDTDB") {
+        if (line.substr(0, vectorTabHeader.size()) == vectorTabHeader) {
```
alexanderbock, line 509:
```suggestion
        if (line.starts_with(VectorTabHeader)) {
```

Shiny new C++20

or
`if (line.starts_with("JDTBD"))`

GPayne: Switched to the cool new way

PR #2748, opened 2023-06-07
Fixes #2540 by correctly showing all points, including the proper end point

### LINKED ISSUE #2540: RenderableTrailTrajectory renders one point too few
### RenderableTrailTrajectory does not include the last point in the trail, instead it always renders one point too few.

rtt-too-few-points

### This may be two different issues in one.

1. It always renders one point too few. As seen in the images below, even though the number of vertices is 3, it only renders 2.
rtt-3pt

2. We're calculating one points too few. Example, let's say we have 100m trail we want to render and we want to do it with 10m interval - using the same math as we do in the code we get that we would need 10 points ( 100 / 10 ).
However, if you want to draw lines that covers from 0 to 100m then you'd need 11 points.
```
 1       2      3      4      5      6      7      8      9      10     11 points
  | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
 0       10     20     30     40     50     60     70     80     90    100 m
```

Arohdin:
#### Before:
image

#### After:
image

____________________________________

#### Before:
image

##### After:
**Note**: the end point is added afterwards and is not included in this variable
image

### DIFF
--- a/modules/base/rendering/renderabletrailtrajectory.cpp
+++ b/modules/base/rendering/renderabletrailtrajectory.cpp
@@ -253,8 +253,9 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
         }
         ++_sweepIteration;
 
+        // Full sweep is complete here.
         // Adds the last point in time to the _vertexArray so that we
-        // ensure that points for _start and _end always exists.
+        // ensure that points for _start and _end always exists
         if (stopIndex == _numberOfVertices) {
             const glm::vec3 p = _translation->position({
                 {},
@@ -262,9 +263,7 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
                 Time(0.0)
             });
             _vertexArray[stopIndex] = { p.x, p.y, p.z };
-        }
 
-        if (stopIndex == _numberOfVertices) {
             _sweepIteration = 0;
             setBoundingSphere(glm::distance(_maxVertex, _minVertex) / 2.f);
         }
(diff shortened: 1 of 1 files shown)


modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -299,10 +309,13 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
             0.0,
             (data.time.j2000Seconds() - _start) / (_end - _start)
         );
-        _primaryRenderInformation.count = std::min(
-            static_cast<GLsizei>(ceil(_vertexArray.size() * t)),
-            static_cast<GLsizei>(_vertexArray.size() - 1)
-        );
+        if (data.time.j2000Seconds() < _end) {
+            _primaryRenderInformation.count = static_cast<GLsizei>(std::max(1.0,_vertexArray.size() * t));
```
alexanderbock, line 313:
```suggestion
            _primaryRenderInformation.count = static_cast<GLsizei>(std::max(1.0, _vertexArray.size() * t));
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -218,19 +218,19 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
...
                 maxNumberOfVertices
             );
 
             // We need to recalcuate the _totalSampleInterval if _numberOfVertices eqals
             // maxNumberOfVertices. If we don't do this the position for each vertex
             // will not be correct for the number of vertices we are doing along the trail
-            _totalSampleInterval = (_numberOfVertices == maxNumberOfVertices) ?
-                (timespan / _numberOfVertices) : _totalSampleInterval;
+            _totalSampleInterval = std::max(1.0, (_numberOfVertices == maxNumberOfVertices) ?
+                (timespan / _numberOfVertices) : _totalSampleInterval);
```
alexanderbock, line 229:
```suggestion
            _totalSampleInterval = (_numberOfVertices == maxNumberOfVertices) ?
                (timespan / _numberOfVertices) : _totalSampleInterval;
            _totalSampleInterval = std::max(1.0, _totalSampleInterval);
```

Suggestion to make it easier to understand

alexanderbock (approved):

PR #2753, opened 2023-06-08
Don't call update function unless layer has been initialized

ylvaselling:
Added a simple check to see if the layer is initialized before calling the update function.


alexanderbock (approved):

Roxeena (commented):
👍

Issue #2696, opened 2023-05-15
Easy to go inside Earth


ylvaselling:
I think something must have happened to the bounding sphere of Earth because it is very easy to go inside Earth now. Also happens to the Moonzzzzzz

alexanderbock:
The min-distance boolean gating might be causing this

alexanderbock:
Didn't happen Venus and Mars

ylvaselling:
![image](https://github.com/OpenSpace/OpenSpace/assets/24265984/f804587e-692f-46ad-b122-0a89547d57bf)

To reproduce:

Open default profile
Go to the Moon as far as you can. I get the above image

ylvaselling:
Also happens to some places on Mars - it seems to depend on the height map

![image](https://github.com/OpenSpace/OpenSpace/assets/24265984/aa845a5a-e27d-4d62-9830-3b1f6323834f)

ylvaselling:
![image](https://github.com/OpenSpace/OpenSpace/assets/24265984/347d4d09-8a4e-49c8-8979-661c9413e644)
 Earth - go to Kazakhstan and zoom in as far as possible

WeirdRubberDuck:
I'm really struggling to reproduce this on master, even with your exact coordinates. Verified that the "min distance" check works as it should and it seems to do so, so my guess is that there is some problem with the height layer loading. 

Does the problem persist even if you let the height map load for a while? I.e if you go back to the same position again, does it still allow you to go through the surface? 

And are you using the new MRF cache?  I'm not, currently, so maybe that could be related somehow?

WeirdRubberDuck:
Aha! Nvm! Managed to get the problem on the Moon 🌚

WeirdRubberDuck:
This is actually an older problem. I went back, and it occurs easily in 0.18.0 and 0.17.0 too

The problem seems to be a mismatch between the rendered height and the computed height in the C++ code (the one we get for SurfacePositionHandles, RenderableGlobes' `getHeight`, etc.). Haven't found the cause of the error yet though... As I was expecting it to be a new problem, I need to change my mindset a bit when continuing the debugging

The fact that it's not a new problem makes it a bit less critical IMO, and I'll leave it off the 0.19.0 release in order to prioritize other things

(Considered the solution of just temporarily increasing the minimum distance, but that would be a pain when looking at small things in the software, such as spacecraft. So decided to leave it as it is for now)

micahnyc:
[...]

PR #2756, opened 2023-06-09
Offset sample coordinate by 0.5 when estimating height values

WeirdRubberDuck:
This commit seems to somewhat fix the problem mentioned in #2696, by changing the coordinate of where we sample the height value, to account for a 0.5 offset. 

It's not a perfect solution (managed to go through the planets still in a few places), but it seems to greatly improve the behavior. Please give it a try and see if you agree (e.g. @micahnyc @ylvaselling )


WeirdRubberDuck:
Long term solution is to rewrite the rendering code for the globe-browsing to be a bit simpler, and then hopefully also find the true cause of the error

ylvaselling:
Tried it out and it does indeed improve it although not fix it altogether

WeirdRubberDuck:
> Tried it out and it does indeed improve it although not fix it altogether

Great.  I came to the same conclusion.

alexanderbock (approved):

PR #2766, opened 2023-06-13
Fixes #625: RenderableTrailTrajectory connects to spacecraft/object when rendering full trail

### LINKED ISSUE #625: RenderableTrailTrajectory does not go through center of spacecraft when full trail is enabled
For example when lookign at Juno and enabling the full trail for the space craft + orbit lines.  If done, they don't match up with the accurate location

Possible solution:  Having the floating position in there as well?

Arohdin:
PR solves issue #625 which is that `RenderableTrailTrajectory` does not connect to spacecrafts/objects when using rendering full trail (see final results further down).

Current implementation contains two parts that makes this solution work and look good.

1. When an object is in between two trail vertices, we render a trail from the preceding vertex to the object and from the object to the vertex that comes after. This was not done before as the trail only connected to it's own vertices during full trail rendering.

#### Before
https://github.com/OpenSpace/OpenSpace/assets/9076357/2a7ffe75-b331-46b0-ae37-38c7f6f7c856

#### After
https://github.com/OpenSpace/OpenSpace/assets/9076357/4c39969a-4cc5-4d76-a46f-614d30bb5880

2. As seen above, the first solution "solves" the issue of the trail not connected to the spacecraft. However, due to the inaccuracies of the trail vertex points (float precision issue) it still looks really bad. 
To remedy this we introduce "padding" on each sides of the spacecraft. This padding replaces X number of vertex points (default 250) **_before and after_** the spacecraft with vertex points recalculated and then rendered in model space of the spacecraft. 
However, in order for the inaccurate vertex points and the accurate vertex points to have a smooth transition at the border (were old and new vertex points meet) we move each point with a factor from 0 - 1 (or 1 - 0 depending on if the new vertex is before or after the spacecraft).
By doing this we get a smooth transition with accurate vertices close to the spacecraft and then less accurate, but properly transitioned and connected, vertices at the borders.

#### Final result
https://github.com/OpenSpace/OpenSpace/assets/9076357/7d662599-7444-47ad-94fc-db933f1316d8

### Note:
1. Performance wise my PC runs this at 370 fps with 500 new vertices per frame and roughly 330 fps with 5000 vertices each frame (Note: 500 vertices equals 250 each side of the spacecraft). Master, without this implementation, runs similar to 500 vertices.

2. The number of new vertex points that are calculated each frame can be set in GUI or via asset.
image

https://github.com/OpenSpace/OpenSpace/assets/9076357/c06b9802-8a11-4375-8a40-cb9323776e53


modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -296,26 +331,184 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
...
+                std::distance(
+                    _timeVector.begin(),
+                    std::lower_bound(_timeVector.begin(), _timeVector.end(), j2k)
+                )
+            );
+
+
+            // Second segment
+            _secondSegRenderInformation.first = _firstSegRenderInformation.count;
+
```
alexanderbock, line 352:
```suggestion
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -296,26 +331,184 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
...
+                    _numberOfReplacementPoints : _secondSegRenderInformation.count;
+            
+            int numOfPoints = prePaddingDelta + postPaddingDelta + 1;
+
+            _replacementPointsRenderInformation.first = 0;
+            _replacementPointsRenderInformation.count = numOfPoints;
+;
+
+            // All points before the object
+            glm::dvec3 v(p.x, p.y, p.z);
```
alexanderbock, line 377:
```suggestion
            glm::dvec3 v = glm::dvec3(p.x, p.y, p.z);
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -296,26 +331,184 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
...
+                glm::dvec3 newPoint = dp - v;
+
+                double mult = 0.0;
+                if (i == prePaddingDelta - 1) {
+                    if (i == 0) {
+                        mult = 1.0;
+                    }
+                    else {
+                        mult = 0.0;
+                    }
```
alexanderbock, line 403:
```suggestion
                    mult = (i == 0) ? 1.0 : 0.0;
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -296,26 +331,184 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
...
+                    _dVertexArray[floatPointIndex].x,
+                    _dVertexArray[floatPointIndex].y,
+                    _dVertexArray[floatPointIndex].z
+                );
+
+                glm::dvec3 dv = fp - dp;
+                glm::dvec3 newPoint = dp - v;
+
+                double mult = 0.0;
+                if (i == postPaddingDelta - 1) {
```
alexanderbock, line 445: Same as above

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -296,26 +331,184 @@ void RenderableTrailTrajectory::update(const UpdateData& data) {
...
+                    mult = 1.0 - 
+                        (postPaddingDelta - i) / static_cast<double>(postPaddingDelta);
+                }
+
+                newPoint = newPoint + dv * mult;
+                _replacementPoints.push_back({
+                    static_cast<float>(newPoint.x),
+                    static_cast<float>(newPoint.y),
+                    static_cast<float>(newPoint.z)
+                    });
```
alexanderbock, line 458:
```suggestion
                });
```

Arohdin:
> How about making the `TrailVBOLayout` into:
> 
> ```c++
> template <typename T>
> struct TrailVBOLayout {
>     T x, y, z;
> };
> ```
> 
> and then using `TrailVBOLayout` or `TrailVBOLayout`? Would save us from the extra struct
> 
> Regarding name, I think the current name is fine; alternative maybe: "AccurateWindowSize" or something like that?

Resolved in [5c1511d](https://github.com/OpenSpace/OpenSpace/pull/2766/commits/5c1511d65032879d1eb4051cb2ce67125a4d6e2d) .

alexanderbock:
Fixes #625

Arohdin:
Remaking PR later

Issue #2773, opened 2023-06-14
Add property to make sky browser display copy animate if target moves


ylvaselling:
There have been multiple people wanting to have the same behavior in the display copies as it is in the GUI, which is that the sky browser updates while the target is moving. This could be added as a property so that the default behavior doesn't change.

PR #2774, opened 2023-06-14
Add property to make it possible to set if the browser display copies…

ylvaselling:
… should animate while the target is moving

Closes issue #2773

### DIFF
--- a/modules/skybrowser/src/targetbrowserpair.cpp
+++ b/modules/skybrowser/src/targetbrowserpair.cpp
@@ -91,7 +91,9 @@ void TargetBrowserPair::fineTuneTarget(const glm::vec2& translation) {
 }
 
 void TargetBrowserPair::synchronizeAim() {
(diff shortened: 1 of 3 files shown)


modules/skybrowser/src/targetbrowserpair.cpp
```diff
@@ -91,7 +91,9 @@ void TargetBrowserPair::fineTuneTarget(const glm::vec2& translation) {
 }
 
 void TargetBrowserPair::synchronizeAim() {
-    if (!_targetAnimation.isAnimating() && _browser->isInitialized()) {
+    bool shouldUpdate = _browser->shouldUpdateWhileTargetAnimates() ? 
+         true : !_targetAnimation.isAnimating();
```
alexanderbock, line 95:
```suggestion
    bool shouldUpdate =
        _browser->shouldUpdateWhileTargetAnimates() ||
        !_targetAnimation.isAnimating();
```

PR #2800, opened 2023-06-27
Sphere maintenance

Roxeena:
Make `RenderableSphere` into an abstract class that now more classes derive from,


modules/base/rendering/renderablesphere.cpp
```diff
@@ -46,27 +45,25 @@ namespace {
...
-        "This value specifies an image that is loaded from disk and is used as a texture "
-        "that is applied to this sphere. This image is expected to be an equirectangular "
-        "projection",
-        // @VISIBILITY(2.33)
+    constexpr openspace::properties::Property::PropertyInfo SegmentsInfo = {
+        "Segments",
+        "Number of Segments",
+        "This value specifies the number of segments that the sphere is separated in",
+        // @VISIBILITY(2.67)
         openspace::properties::Property::Visibility::User
```
alexanderbock, line 60: I'd say that this one should become and AdvancedUser property

modules/base/rendering/renderablesphere.cpp
```diff
@@ -77,28 +74,18 @@ namespace {
...
-    constexpr openspace::properties::Property::PropertyInfo SegmentsInfo = {
-        "Segments",
-        "Number of Segments",
-        "This value specifies the number of segments that the sphere is separated in",
-        // @VISIBILITY(2.67)
+    constexpr openspace::properties::Property::PropertyInfo MirrorTextureInfo = {
+        "MirrorTexture",
+        "Mirror Texture",
+        "Mirror the texture along the x-axis",
         openspace::properties::Property::Visibility::User
```
alexanderbock, line 81: Same here

modules/base/rendering/renderablesphere.cpp
```diff
@@ -239,19 +217,18 @@ void RenderableSphere::initializeGL() {
     );
 
     ghoul::opengl::updateUniformLocations(*_shader, _uniformCache, UniformNames);
-
-    loadTexture();
 }
 
 void RenderableSphere::deinitializeGL() {
-    _texture = nullptr;
+    _sphere.release();
```
alexanderbock, line 223:
```suggestion
    _sphere = nullptr;
```

alexanderbock: Does the same thing, but is more understandable as `_sphere` being a pointer

modules/base/rendering/renderablesphereimageonline.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    _textureUrl = p.url;
+    _textureUrl.onChange([this]() {
+        _textureIsDirty = true;
+    });
+    addProperty(_textureUrl);
+}
+
+void RenderableSphereImageOnline::deinitializeGL() {
+    _texture.release();
```
alexanderbock, line 73:
```suggestion
```

No need for this if we set it to `nullptr` immediately after

alexanderbock (approved):

PR #2801, opened 2023-06-29
Allow negative values for KeplerTranslation values

### LINKED ISSUE #2784: Add Support for Negative Angles in KeplerTranslation
Angles in a **KeplerTranslation** (e.g. Inclination, AscendingNode) are usually positive in the range of 0 - 360 degrees.
If a negative angle is specified, the Renderable fails to load with an error.
A fix for this would involve adding +360° to the value if it is negative.

alexanderbock:
Closes #2784

### DIFF
--- a/modules/space/translation/keplertranslation.cpp
+++ b/modules/space/translation/keplertranslation.cpp
@@ -121,16 +121,16 @@ namespace {
         double semiMajorAxis;
 
         // [[codegen::verbatim(InclinationInfo.description)]]
-        double inclination [[codegen::inrange(0.0, 360.0)]];
+        double inclination [[codegen::inrange(-360.0, 360.0)]];
 
         // [[codegen::verbatim(AscendingNodeInfo.description)]]
-        double ascendingNode [[codegen::inrange(0.0, 360.0)]];
+        double ascendingNode [[codegen::inrange(-360.0, 360.0)]];
 
         // [[codegen::verbatim(ArgumentOfPeriapsisInfo.description)]]
-        double argumentOfPeriapsis [[codegen::inrange(0.0, 360.0)]];
+        double argumentOfPeriapsis [[codegen::inrange(-360.0, 360.0)]];
 
         // [[codegen::verbatim(MeanAnomalyAtEpochInfo.description)]]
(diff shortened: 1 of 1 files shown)


modules/space/translation/keplertranslation.cpp
```diff
@@ -198,10 +198,12 @@ KeplerTranslation::KeplerTranslation(const ghoul::Dictionary& dictionary)
     setKeplerElements(
         p.eccentricity,
         p.semiMajorAxis,
-        p.inclination,
-        p.ascendingNode,
-        p.argumentOfPeriapsis,
-        p.meanAnomaly,
+        p.inclination < 0.0 ? p.inclination + 360.0 : p.inclination,
```
GPayne, line 201: How about:
`p.inclination < 0.0 ? (fmod(p.inclination, 360.0) + 360.0) : p.inclination`
in case the negative value is less than -360, since the `codegen::inrange` won't guard against such a value being read?

alexanderbock: The codegen checks against `[[codegen::inrange(-360.0, 360.0)]];` so < -360 values can't happen

GPayne: I tested an input file with an angle greater than 360.0 and it didn't throw any errors, so that's why I think the < -360 value could happen.

alexanderbock: Oh... gotcha. Will investigate

alexanderbock: I couldn't reprocude it;  loading these asset files correctly produce errors for all the fail ones:

[2784.zip](https://github.com/OpenSpace/OpenSpace/files/11949890/2784.zip)

GPayne: Sorry for the confusion, I had been testing with the `readSbdbFile` function in kepler.cpp, where the angles outside the range of -360 to 360 are allowed, and do work. Anyway, I withdraw my suggestion.

alexanderbock: No worries!  Ended up finding a different bug while tracking this down ;)

Issue #2826, opened 2023-07-25
Absolute time from profile not applied


jockekilby:
**Describe the bug**
An entry like the following is not applied correctly when loading a profile: 
```json
"time": {
    "type": "absolute",
    "value": "2018-10-30T23:00:00.500",
    "is_paused": true
  }
```
The paused value is applied but it seems like the time value is overwritten. A workaround is to add the following to the profile:
```json
"additional_scripts": [
    "openspace.time.setTime(\"2018-10-30T23:00:00.500\")"
  ]
```
This fixes the issue but the wiki clearly states that __additional_scripts__ should not be used when there is an explicit setting in the profile for it. The example JSON above is also using the value from the wiki and would be expected to work

**To Reproduce**
1. Add the _time_ JSON content above to your profile and time is __not__.
2. Add the _additional_scripts_ content and time changes.

**Expected behavior**
Time should be applied as described in the wiki by using __time_ entry

WeirdRubberDuck:
This actually seems to be a problem with the start as paused option, not the absolute time. 

Tried setting it from the profile editor and get the same behavior for both absolute and relative time. When starting with time paused, OpenSpace starts at year 2000

I suspect this issue is related: https://github.com/OpenSpace/OpenSpace/issues/2510

WeirdRubberDuck:
Adding this to the next milestone, as I think we should fix it ASAP

Issue #2510, opened 2023-02-15
Unexpected behavior when using setPause and setTime during the same frame


Arohdin:
**Explanation**
There is an unexpected behavior when using setPause and SetTime during the same frame (example, via scripts).
Example current datedate 2023-01-01 - If you do setTime before setPause 

```
openspace.time.setTime('1990-02-14T14:14:14.645')
openspace.time.setPause(true)
```
The expected behavior would be that:
1. OpenSpace sets time to 1990-02-14 14:14:14
2. OpenSpace pauses the time
So when the scripts finishes you would be at 1990 and with the time paused.
However, the outcome is that time is paused but it's not 1990 (it's the same date as it was when you ran the script).

The expected behavior can be achieved by switching the execution order of the two commands to
```
openspace.time.setPause(true)
openspace.time.setTime('1990-02-14T14:14:14.645')
```

**Apparent cause**
The issue seems to be that setTime will clear the keyframes in OpenSpace and then set _timeNextFrame to 1990, but then setPause will get the current time (which has not changed yet to 1990 as it's still the same frame) and then create a new keyframe for pause with duration over 0 seconds. 
When OpenSpace then executes everything the next frame will be 1990 and the frame after that will be 2023 (assuming 2023 was the current year when you executed the script) again as it applies the keyframe that is in queue.

PR #2837, opened 2023-08-03
Set pause directly instead of interpolating over 0 duration for setPause

WeirdRubberDuck:
Seems to solve problems in #2826 and #2510

Would be great if some people could try it out; it almost seems a bit too simple and I'm afraid it might cause problems in other scenarios (e.g. session recording?). Someone that has more experience with the time system or more ideas on what to test, please help me verify that it does not. 

There is still some weirdness when setting the time and interpolating the pause in the same frame, but perhaps that is to be expected. It does not quite make sense to do those two actions simultaneously, IMO

### DIFF
--- a/src/util/timemanager.cpp
+++ b/src/util/timemanager.cpp
@@ -867,7 +867,7 @@ void TimeManager::interpolatePreviousDeltaTimeStep(double durationSeconds) {
 }
 
 void TimeManager::setPause(bool pause) {
-    interpolatePause(pause, 0);
+    _timePaused = pause;
 }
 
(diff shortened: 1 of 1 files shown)


WeirdRubberDuck:
I guess one thing to figure out will be: Is there any danger in setting the `_isPaused` flag in the middle of a frame?

Can it cause problems somehow ?

WeirdRubberDuck:
Another fix was done for #2826, so will close this PR. 

We should looik fopr another fix for #2510 as well (investigating the problem with the interpolatePause function)

PR #2840, opened 2023-08-03
Rename Lua function to bring them more in line with the function naming in C++

alexanderbock:
All of the Lua functions that had a `get` prefix now have the same version without the get.  The old prefixed version is still there, but prints a deprecation warning.  Since they might be used in user webpage code, it would be harsh to not give a version's worth of time to update.

### DIFF
--- a/data/assets/customization/gui.asset
+++ b/data/assets/customization/gui.asset
@@ -1,4 +1,4 @@
-asset.export("webguiDevelopmentMode", true)
+asset.export("webguiDevelopmentMode", false)
 
 -- To make changes to the webgui:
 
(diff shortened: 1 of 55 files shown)


data/assets/customization/gui.asset
```diff
@@ -1,4 +1,4 @@
-asset.export("webguiDevelopmentMode", false)
+asset.export("webguiDevelopmentMode", true)
```
WeirdRubberDuck, line 1: this should not be committed

WeirdRubberDuck (approved):
Looks good to me. Want me to make the corresponding UI changes or do you prefer doing so yourself?

data/assets/customization/gui.asset
alexanderbock: Oops, indeed

Issue #2841, opened 2023-08-04
Frustum culling can be too aggressive while close to planetary surface


jockekilby:
Performing frustum culling may sometimes cull away tiles when camera is close to planetary surface.
See screenshots below with culling on and off. Suggest to make [constexpr setting](https://github.com/OpenSpace/OpenSpace/blob/master/modules/globebrowsing/src/renderableglobe.cpp#L72) a debug property instead to allow a user to control it.

**Cull on**
<img width="1281" alt="cull_on" src="https://github.com/OpenSpace/OpenSpace/assets/454642/ed982f85-8f4f-47c2-b7fc-afc2e39cdfd9">

**Cull off**
<img width="1281" alt="cull_off" src="https://github.com/OpenSpace/OpenSpace/assets/454642/bafcadac-7d2a-47b0-b96c-6b4c49311e49">

Issue #2834, opened 2023-08-03
Tiles are split once more than they should with regards to their corresponding maxChunkLevel


jockekilby:
Globebrowsing Tiles are split once more than they should with regards to their corresponding maxChunkLevel since the function which calculates a desired level data only takes into account if the current leven is "OK" or not, not wether the next level will be "OK".

See https://github.com/OpenSpace/OpenSpace/blob/master/modules/globebrowsing/src/renderableglobe.cpp#L2360

jockekilby:
See below, with and without limiting desired level by wether or not next level is available.
(Showing UV coordinates alongside chunk edges and underlying color map)

<img width="1281" alt="WAC_max_tile_no_split" src="https://github.com/OpenSpace/OpenSpace/assets/454642/adea1624-6553-449a-a3f0-f6e3aa64054e">
<img width="1281" alt="WAC_max_tile_split" src="https://github.com/OpenSpace/OpenSpace/assets/454642/4bf211d0-19c2-4c22-ab42-dd921986624b">

jockekilby:
On my machine I get about 5-6 FPS improvement when close to the surface of a globe after having fixed this.

Issue #2472, opened 2023-02-01
Return of the Stair Stepping (of heightmap)


micahnyc:
Stair stepping of the height map has been reintroduced. Seen easiest when accurate normals is enabled. This is happening even when turning off the stereoscopic depth.

See attached screenshots of random place on the moon in v17.2 vs v18/master.

![OpenSpace_000000](https://user-images.githubusercontent.com/708535/216125990-6cef1084-4593-4be7-9cff-ecc3bb23813d.png)
![OpenSpace_000001](https://user-images.githubusercontent.com/708535/216125996-435209fd-f776-4562-914e-13f14bafe88d.png)
![OpenSpace_000000](https://user-images.githubusercontent.com/708535/216126010-105b7719-fb60-4f3a-9bc8-889b5c01950b.png)
![OpenSpace_000001](https://user-images.githubusercontent.com/708535/216126014-1eb80791-3f81-438c-beec-8db0142a1d84.png)

alexanderbock:
Fixed;  we subsampled the height map to a glorious resolution of 64x64

![OpenSpace_000000](https://user-images.githubusercontent.com/4921724/229607388-5569a18a-9b59-4e7e-b08d-5ec5f837f788.png)

micahnyc:
Still happening for LROC dems.  probably all  globebrowsing files added with tif/vrt

![image](https://user-images.githubusercontent.com/708535/235179320-ae283b1c-6522-45e5-a691-2b2218a8412a.png)

alexanderbock:
Which files are they?

micahnyc:
These are the ones from our website download for the apollo sites.

alexanderbock:
9447a351b4a3d41a432dd7dbe0edd04fa370a667 should address some of that.  I still see some stairstepping, but I believe that it is included in the data.

We might want to consider removing the info parts and loading the files from OnMoon instead?

alexanderbock:
@micahnyc  Need feedback on if it is fixed

micahnyc:
seems to still be happening on mars in WCC

![image](https://github.com/OpenSpace/OpenSpace/assets/708535/cf3375d0-37d0-4bfd-96de-b7581214442a)

alexanderbock:
Stair stepping does not seem to occur in the image data

![OpenSpace_000001](https://github.com/OpenSpace/OpenSpace/assets/4921724/971ba741-bbcc-4e64-ae43-fe76c29b5ea1)
[...]

PR #2842, opened 2023-08-04
Remove tile-padding to reduce stair-stepping issues

jockekilby:
Solves #2472 #2834 and #2841 

Tiles were previously padded both during read and rendering with 4 pixels on each side (or the corresponding amount of pixels during read). This caused issues both with color & height maps.

For colormaps the default rendering was done without padding but the rasterreads were performed with padding, this caused the entire texture to become squeezed and visually "move" towards the corners as higher-resolution tiles were loaded.

In addition; for heightmaps it caused excessive sampling in-between pixels which enhanced stair-stepping artifacts.

All of the changes in this PR, alongside with potential changes to the underlying VRT dataset as described [in this PR](https://github.com/OpenSpace/openspace.github.io/pull/30) provides significant improvements to globe-browsing when close to the surface and using high-resolution local patches.

### DIFF
--- a/modules/globebrowsing/src/renderableglobe.cpp
+++ b/modules/globebrowsing/src/renderableglobe.cpp
@@ -2365,6 +2365,8 @@ int RenderableGlobe::desiredLevelByAvailableTileData(const Chunk& chunk) const {
         const std::vector<Layer*>& lyrs = _layerManager.layerGroup(gi.id).activeLayers();
         for (Layer* layer : lyrs) {
             Tile::Status status = layer->tileStatus(chunk.tileIndex);
+            // Ensure that the current tile is OK and that the tileprovider for the current
+            // layer has enough data to support an additional level.
             if (status == Tile::Status::OK && layer->tileProvider()->maxLevel() > currLevel + 1) {
                 return UnknownDesiredLevel;
             }
--- a/modules/globebrowsing/src/renderableglobe.cpp
+++ b/modules/globebrowsing/src/renderableglobe.cpp
@@ -1606,7 +1606,7 @@ void RenderableGlobe::setCommonUniforms(ghoul::opengl::ProgramObject& programObj
         // If it does not it will still produce "correct" normals. If the resolution is
         // higher the shadows will be softer, if it is lower, pixels will be visible.
         // Since default is 64 this will most likely work fine.
-        constexpr float TileDelta = 1.f / DefaultSkirtedGridSegments;
+        constexpr float TileDelta = 1.f / 512;
         const glm::vec3 deltaTheta0 = modelViewTransformMat3 *
             (glm::vec3(corner10 - corner00) * TileDelta);
         const glm::vec3 deltaTheta1 = modelViewTransformMat3 *
--- a/modules/globebrowsing/src/renderableglobe.cpp
+++ b/modules/globebrowsing/src/renderableglobe.cpp
@@ -97,6 +97,7 @@ namespace {
     // time being.  --abock  2018-10-30
     constexpr int DefaultSkirtedGridSegments = 64;
     constexpr int UnknownDesiredLevel = -1;
+    constexpr int DefaultHeightTileResolution = 512;
 
     const openspace::globebrowsing::GeodeticPatch Coverage =
         openspace::globebrowsing::GeodeticPatch(0, 0, 90, 180);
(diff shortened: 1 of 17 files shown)


modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -69,7 +69,6 @@ namespace {
 
     // Global flags to modify the RenderableGlobe
     constexpr bool LimitLevelByAvailableData = true;
-    constexpr bool PerformFrustumCulling = true;
```
alexanderbock, line 72: Should check that this doesn't degrade performance.  The idea behind having them as `constexpr` variables was to enable some compiler optimizations.  But  maybe it doesn't matter?  ¯\_(ツ)_/¯

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1606,7 +1614,7 @@ void RenderableGlobe::setCommonUniforms(ghoul::opengl::ProgramObject& programObj
         // If it does not it will still produce "correct" normals. If the resolution is
         // higher the shadows will be softer, if it is lower, pixels will be visible.
         // Since default is 64 this will most likely work fine.
-        constexpr float TileDelta = 1.f / DefaultSkirtedGridSegments;
+        constexpr float TileDelta = 1.f / 512;
```
alexanderbock, line 1617: Should this stay unchanged?

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -2357,7 +2365,7 @@ int RenderableGlobe::desiredLevelByAvailableTileData(const Chunk& chunk) const {
         const std::vector<Layer*>& lyrs = _layerManager.layerGroup(gi.id).activeLayers();
         for (Layer* layer : lyrs) {
             Tile::Status status = layer->tileStatus(chunk.tileIndex);
-            if (status == Tile::Status::OK) {
+            if (status == Tile::Status::OK && layer->tileProvider()->maxLevel() > currLevel + 1) {
```
alexanderbock, line 2368: Would be good to have a short comment explaining the +1 in here

modules/globebrowsing/src/renderableglobe.cpp
jockekilby: Well I would imagine that if the first condition evaluates to true [here](https://github.com/OpenSpace/OpenSpace/blob/6d87eb39b1e78149e57ed968162a84d61f19eede/modules/globebrowsing/src/renderableglobe.cpp#L1900
) then the second one won't be checked. If so the performance cost of this change would be one boolean check per globe & renderpass which I should think is OK. 
I can't really find any documentation supporting my assumption about the multiple-condition IF's but a quick test shows  no difference between having it constexpr or not (on my machine)

modules/globebrowsing/src/renderableglobe.cpp
jockekilby: since [the default texture size](https://github.com/OpenSpace/OpenSpace/blob/6d87eb39b1e78149e57ed968162a84d61f19eede/modules/globebrowsing/src/tiletextureinitdata.cpp#L109) for heightmaps changed I believe it makes sense to change this value to match. It is also conceptually weird that the resolution used for normal calculation is based on how many segments the skirt of a tile has.  I can't really see any major visual difference between leaving it or changing it but for consistency I think it should be changed.
But the comment should be updated though! (if we decide to change this value that is)

alexanderbock: I'm fine with the change in resolution, but I think the `DefaultSkirtedGridSegments` variable should be defined instead.

modules/globebrowsing/src/renderableglobe.cpp
alexanderbock: Yea, there definitely is short-circuit evaluation of it;  and yea, it probably doesn't make much difference

modules/globebrowsing/src/renderableglobe.cpp
jockekilby: Makes sense, lets define a new variable then to hold this value? 
DefaultSkirtedGridSegments_ is also used to determine the resolution of the grid and it feels like an unnecessary performance degradation to increase the number of vertices on a grid by a factor 64 😄

alexanderbock: Yea, sounds good!

PR #2844, opened 2023-08-06
Add search text box and filtering capability to Launcher profile assets editor

GPayne:
It can be hard to find an asset in the profile editor since there are many assets, and the tree view has collapsed branches by default for directories that don't contain any currently-selected assets.
This adds a text box above the tree view to allow for easy searching of the available asset files.

### DIFF
--- a/apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
+++ b/apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
@@ -26,8 +26,8 @@
 #define __OPENSPACE_UI_LAUNCHER___ASSETSDIALOG___H__
 
 #include <QDialog>
-#include <QSortFilterProxyModel>
 #include <QRegularExpression>
+#include <QSortFilterProxyModel>
 #include "assettreemodel.h"
 
 class QTextEdit;
@@ -41,7 +41,7 @@ public:
      *
      * \param parent The QObject* object that is the Qt parent of this object
      */
-    SearchProxyModel(QObject* parent = nullptr) : QSortFilterProxyModel(parent) {}
+    SearchProxyModel(QObject* parent = nullptr);
 
 public slots:
     /**
@@ -55,8 +55,9 @@ protected:
     bool filterAcceptsRow(int sourceRow, const QModelIndex& sourceParent) const override;
 
 private:
-    QRegularExpression* _regExPattern = nullptr;
     bool acceptIndex(const QModelIndex& idx) const;
+
+    QRegularExpression* _regExPattern = nullptr;
 };
 
 class AssetsDialog final : public QDialog {
--- a/apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
+++ b/apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
@@ -26,11 +26,35 @@
 #define __OPENSPACE_UI_LAUNCHER___ASSETSDIALOG___H__
 
 #include <QDialog>
-
+#include <QSortFilterProxyModel>
+#include <QRegExp>
 #include "assettreemodel.h"
 
 class QTextEdit;
 class QTreeView;
+
+class SearchProxyModel : public QSortFilterProxyModel {
+public:
+    /**
+     * Constructor for SearchProxyModel class
+     *
+     * \param parent The QObject* object that is the Qt parent of this object
+     */
+    SearchProxyModel(QObject* parent = nullptr) : QSortFilterProxyModel(parent) {}
+
+    /**
+     * Sets the regular expression pattern to apply to the filter
+     *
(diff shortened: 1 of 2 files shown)


WeirdRubberDuck:
What Qt version are you using? I can't compile this code using Qt 6.4.0

When trying to compile this, I got problems with the compiler not being able to find `QRegExp`. Did some quick research and this seems to have been removed in Qt6, and should seemingly be replaced with the `QRegularExpression` class. 

https://doc.qt.io/qt-6/qtcore-changes-qt6.html
https://doc.qt.io/qt-6/qregexp.html#porting-to-qregularexpression

alexanderbock:
QRegExp is deprecated in Qt6; there is a page on how to convert the old usage:  https://doc.qt.io/qt-6/qregexp.html#porting-to-qregularexpression

GPayne:
Latest commit now uses `QRegularExpression`. The code works with both Qt6 and Qt5 now.

apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
```diff
@@ -27,13 +27,14 @@
 
 #include <QDialog>
 #include <QSortFilterProxyModel>
-#include <QRegExp>
+#include <QRegularExpression>
```
alexanderbock, line 30: Should be sorted before the `QSortFilterProxyModel`

apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
```diff
@@ -26,12 +26,39 @@
...
 
+class SearchProxyModel : public QSortFilterProxyModel {
+Q_OBJECT
+public:
+    /**
+     * Constructor for SearchProxyModel class
+     *
+     * \param parent The QObject* object that is the Qt parent of this object
+     */
+    SearchProxyModel(QObject* parent = nullptr) : QSortFilterProxyModel(parent) {}
```
alexanderbock, line 44: The definition of the constructor should move to the cpp file

apps/OpenSpace/ext/launcher/include/profile/assetsdialog.h
```diff
@@ -26,12 +26,39 @@
...
+     * \param pattern The QString reference containing the regex pattern
+     */
+    void setFilterRegularExpression(const QString& pattern);
+
+protected:
+    bool filterAcceptsRow(int sourceRow, const QModelIndex& sourceParent) const override;
+
+private:
+    QRegularExpression* _regExPattern = nullptr;
+    bool acceptIndex(const QModelIndex& idx) const;
```
alexanderbock, line 59:
```suggestion
    bool acceptIndex(const QModelIndex& idx) const;

    QRegularExpression* _regExPattern = nullptr;
```

In general, we sort functions first, then member variables

apps/OpenSpace/ext/launcher/src/profile/assetsdialog.cpp
```diff
@@ -62,7 +62,6 @@ namespace {
     void traverseToFindFilesystemMatch(AssetTreeModel& model, QModelIndex parent,
                                        int nRows, const std::string& path)
     {
-
         int startIndex = 0;
         std::string token = "${USER_ASSETS}/";
         if (path.find(token) == 0) {
```
alexanderbock, line 67: Can be replaced with `path.starts_with(token)`

apps/OpenSpace/ext/launcher/src/profile/assetsdialog.cpp
```diff
@@ -259,3 +280,57 @@ void AssetsDialog::parseSelections() {
...
+        if (matchIterator.hasNext()) {
+            return true;
+        }
+        for (int row = 0; row < idx.model()->rowCount(idx); ++row) {
+            if (acceptIndex(idx.model()->index(row, 0, idx))) {
+                return true;
+            }
+        }
+    }
+    return false;
```
alexanderbock, line 335:
```suggestion
    if (!idx.isValid() || !_regExPattern) {
        return false;
    }
    QString text = idx.data(Qt::DisplayRole).toString();
    QRegularExpressionMatchIterator matchIterator = _regExPattern->globalMatch(text);
    if (matchIterator.hasNext()) {
        return true;
    }
    for (int row = 0; row < idx.model()->rowCount(idx); ++row) {
        if (acceptIndex(idx.model()->index(row, 0, idx))) {
            return true;
        }
    }

    return false;
```

Would reduce overall indentation

alexanderbock (approved):

apps/OpenSpace/ext/launcher/src/profile/assetsdialog.cpp
GPayne: All changes made

alexanderbock (approved):

PR #2846, opened 2023-08-09
Issue/2225 - Add exoplanets from a CSV file

### LINKED ISSUE #2225: Make it easier to add an exoplanet system from data file
Currently, the user has to run the data preparation task manually to be able to load any new exoplanet system into OpenSpace

However, we could expose the loading functionality in a Lua script that takes in a path to a CSV for example and creates an exoplanet system from the data in that CSV file.

WeirdRubberDuck:
Closes #2225 

I just did minimal changes to make it work, but the code might need some restructuring and improvement now that we allow loading files that we are enot as much in control over. 

Some thoughts from my side: 
1. The data-loading logic now lives in the previous data preparation task, but maybe it's worth breaking it out into a separate handler that could be instantiated in the ExopanetsModule? To avoid using static functions in the task for reusing the CSV loading code. 

2. We may also have to validate that all columns we expect to be in the dataset are there. For instance, the inclination is not included per default if one just downloads a data table as is from the Exoplanet Archive, but as we replace it with default values, the user won't know that that data is missing. 
 
3. Currently there is no limit to how big the file to load can be. Should there be?  Or is it a "use at own risk"-feature?

 Let me know what you think

### DIFF
--- a/modules/exoplanets/exoplanetshelper.h
+++ b/modules/exoplanets/exoplanetshelper.h
@@ -48,7 +48,7 @@ struct ExoplanetDataEntry {
     float bigOmegaLower = std::numeric_limits<float>::quiet_NaN();
 
     // Star known to be binary?
-    bool binary;
+    bool binary = false;
 
     // Star B − V color
     float bmv = std::numeric_limits<float>::quiet_NaN();
--- a/modules/exoplanets/exoplanetshelper.h
+++ b/modules/exoplanets/exoplanetshelper.h
@@ -30,56 +30,104 @@
 #include <string>
 #include <vector>
 
-namespace {
-    constexpr float fNaN = std::numeric_limits<float>::quiet_NaN();
-    constexpr float dNaN = std::numeric_limits<double>::quiet_NaN();
-    constexpr float iNaN = std::numeric_limits<int>::quiet_NaN();
-} // namespace
-
 namespace openspace::exoplanets {
 
 struct ExoplanetDataEntry {
-    float a = fNaN;            // Orbital semi-major axis in AU
-    double aUpper = dNaN;      // Upper uncertainty of orbital semi-major axis
-    double aLower = dNaN;      // Lower uncertainty of orbital semi-major axis
-    float bigOmega = fNaN;     // Longitude of ascending node in degrees
-    float bigOmegaUpper = fNaN; // Upper uncertainty of longitude of ascending node
-    float bigOmegaLower = fNaN; // Lower uncertainty of longitude of ascending node
-    bool binary;                // Star known to be binary?
-    float bmv = fNaN;          // Star B − V color
-    float ecc = fNaN;          // Orbital eccentricity
-    float eccUpper = fNaN;     // Upper uncertainty of orbital eccentricity
-    float eccLower = fNaN;     // Lower uncertainty of orbital eccentricity
-    float i = fNaN;            // Orbital inclination in degrees (for transiting systems only)
-    float iUpper = fNaN;       // Upper uncertainty of orbital inclination
-    float iLower = fNaN;       // Lower uncertainty of orbital inclination
-    int nPlanets = iNaN;       // Number of known planets in the planetary system
-    int nStars = iNaN;         // Number of stars in the planetary system
-    float omega = fNaN;        // Argument of periastron in degrees
-    float omegaUpper = fNaN;   // Upper uncertainty of argument of periastron
-    float omegaLower = fNaN;   // Lower uncertainty of argument of periastron
-    double per = dNaN;         // Orbital period in days
-    float perUpper = fNaN;     // Upper uncertainty of period
-    float perLower = fNaN;     // Lower uncertainty of period
-    double r = dNaN;           // Radius of the planet in Jupiter radii
-    double rUpper = dNaN;      // Upper uncertainty of radius of the planet
-    double rLower = dNaN;      // Lower uncertainty of radius of the planet
-    float rStar = fNaN;        // Estimated radius of the star in solar radii
-    float rStarUpper = fNaN;   // Upper uncertainty of estimated star radius
-    float rStarLower = fNaN;   // Lower uncertainty of estimated star radius
-    float luminosity = fNaN;   // Star luminosity, in units of solar luminosities
-    float luminosityUpper = fNaN; // Upper uncertainty of star luminosity
-    float luminosityLower = fNaN; // Lower uncertainty of star luminosity
-    float teff = fNaN;         // Star's effective temperature in Kelvin
-    float teffUpper = fNaN;    // Upper uncertainty of effective temperature
-    float teffLower = fNaN;    // Lower uncertainty of effective temperature
-    double tt = dNaN;          // Epoch of transit center in HJD-2440000
-    float ttUpper = fNaN;      // Upper uncertainty of epoch of transit center
-    float ttLower = fNaN;      // Lower uncertainty of epoch of transit center
+    // Orbital semi-major axis in AU
+    float a = std::numeric_limits<float>::quiet_NaN();
+    // Upper uncertainty of orbital semi-major axis
+    double aUpper = std::numeric_limits<double>::quiet_NaN();
+    // Lower uncertainty of orbital semi-major axis
+    double aLower = std::numeric_limits<double>::quiet_NaN();
 
-    float positionX = fNaN;    // Star position's X-coordinate in parsec
-    float positionY = fNaN;    // Star position's Y-coordinate in parsec
-    float positionZ = fNaN;    // Star position's Z-coordinate in parsec
+    // Longitude of ascending node in degrees
+    float bigOmega = std::numeric_limits<float>::quiet_NaN();
+    // Upper uncertainty of longitude of ascending node
+    float bigOmegaUpper = std::numeric_limits<float>::quiet_NaN();
+    // Lower uncertainty of longitude of ascending node
+    float bigOmegaLower = std::numeric_limits<float>::quiet_NaN();
+
+    // Star known to be binary?
+    bool binary;
+
+    // Star B − V color
+    float bmv = std::numeric_limits<float>::quiet_NaN();
+
+    // Orbital eccentricity
+    float ecc = std::numeric_limits<float>::quiet_NaN();
+    // Upper uncertainty of orbital eccentricity
+    float eccUpper = std::numeric_limits<float>::quiet_NaN();
+    // Lower uncertainty of orbital eccentricity
+    float eccLower = std::numeric_limits<float>::quiet_NaN();
+
+    // Orbital inclination in degrees (for transiting systems only)
+    float i = std::numeric_limits<float>::quiet_NaN();
+    // Upper uncertainty of orbital inclination
+    float iUpper = std::numeric_limits<float>::quiet_NaN();
+    // Lower uncertainty of orbital inclination
+    float iLower = std::numeric_limits<float>::quiet_NaN();
+
+    // Number of known planets in the planetary system
+    int nPlanets = std::numeric_limits<int>::quiet_NaN();
+
+    // Number of stars in the planetary system
+    int nStars = std::numeric_limits<int>::quiet_NaN();
+
+    // Argument of periastron in degrees
+    float omega = std::numeric_limits<float>::quiet_NaN();
+    // Upper uncertainty of argument of periastron
+    float omegaUpper = std::numeric_limits<float>::quiet_NaN();
+    // Lower uncertainty of argument of periastron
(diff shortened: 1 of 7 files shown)


modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -674,6 +673,96 @@ listOfExoplanetsDeprecated()
...
+            system.planetNames.push_back(planetData.name);
+            system.planetsData.push_back(planetData.dataEntry);
+            updateStarDataFromNewPlanet(system.starData, planetData.dataEntry);
+
+            hostNameToSystemDataMap[planetData.host] = system;
+        }
+    }
+
+    // Add all the added exoplanet systems
+    for (auto entry : hostNameToSystemDataMap) {
```
alexanderbock, line 753:
```suggestion
    for (const std::pair<const std::string, ExoplanetSystem>& entry : hostNameToSystemDataMap) {
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -146,37 +130,91 @@ void ExoplanetsDataPreparationTask::perform(
...
+            _inputSpeckPath,
+            _teffToBvFilePath
+        );
+
+        // Create look-up table
+        long pos = static_cast<long>(binFile.tellp());
+        std::string planetName = planetData.host + " " + planetData.component;
+        lutFile << planetName << "," << pos << std::endl;
+
+        binFile.write(reinterpret_cast<char*>(&(planetData.dataEntry)), sizeof(ExoplanetDataEntry));
```
alexanderbock, line 156: This line looks too long to me?

alexanderbock, line 156:
```suggestion
        binFile.write(reinterpret_cast<char*>(&planetData.dataEntry), sizeof(ExoplanetDataEntry));
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -146,37 +130,91 @@ void ExoplanetsDataPreparationTask::perform(
...
+}
+
+std::vector<std::string>
+ExoplanetsDataPreparationTask::readFirstDataRow(std::ifstream& file)
+{
+    std::string line;
+
+    // Read past any comments and empty lines
+    while (std::getline(file, line)) {
+        bool shouldSkip = line[0] == '#' || line.empty();
```
alexanderbock, line 169:
```suggestion
        bool shouldSkip = line.empty() || line[0] == '#';
```

Otherwise we try to look at the first character of a line that is empty

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -194,202 +232,206 @@ void ExoplanetsDataPreparationTask::perform(
...
+            float dataInLogSolar = readFloatData(data);
+            p.luminosityUpper = static_cast<float>(std::pow(10, dataInLogSolar));
+        }
+        else if (column == "st_lumerr2") {
+            float dataInLogSolar = readFloatData(data);
+            p.luminosityLower = static_cast<float>(-std::pow(10, dataInLogSolar));
+        }
+        // Is the planet orbiting a binary system?
+        else if (column == "cb_flag") {
+            p.binary = static_cast<bool>(readIntegerData(data));
```
alexanderbock, line 382:
```suggestion
            p.binary = readIntegerData(data) != 0;
```

?

modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp
```diff
@@ -423,17 +465,21 @@ glm::vec3 ExoplanetsDataPreparationTask::starPosition(const std::string& starNam
...
+{
     if (std::isnan(teff)) {
         return std::numeric_limits<float>::quiet_NaN();
     }
 
-    std::ifstream teffToBvFile(_teffToBvFilePath);
+    std::ifstream teffToBvFile(conversionFile);
     if (!teffToBvFile.good()) {
-        LERROR(fmt::format("Failed to open teff_bv.txt file"));
+        LERROR(fmt::format("Failed to open file '{}'", conversionFile));
```
alexanderbock, line 477:
```suggestion
        LERROR(fmt::format("Failed to open file {}", conversionFile));
```

passing a `std::filesystem::path` will already add "..." around it

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -25,6 +25,7 @@
 #ifndef __OPENSPACE_MODULE_EXOPLANETS___EXOPLANETSDATAPREPARATIONTASK___H__
 #define __OPENSPACE_MODULE_EXOPLANETS___EXOPLANETSDATAPREPARATIONTASK___H__
 
+#include <modules/exoplanets/exoplanetshelper.h>
 #include <openspace/util/task.h>
```
alexanderbock, line 29:
```suggestion
#include <openspace/util/task.h>

#include <modules/exoplanets/exoplanetshelper.h>
```

since we derive from `Task`

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -35,22 +36,69 @@ namespace openspace::exoplanets {
...
     void perform(const Task::ProgressCallback& progressCallback) override;
     static documentation::Documentation documentation();
 
+    /**
+     * The exoplanet file includes a set of commented lines in the beginning. This
+     * funciton skips those lines, as well as any empty lines, and continues reading
+     * until it find the first valid data line (which will be the column names).
+     *
+     * \param file The inpur file stream used for reading the file
+     * \return The ocntent of the first data row (i.e. the column names).
```
alexanderbock, line 57:
```suggestion
     * \return The content of the first data row (i.e. the column names).
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -35,22 +36,69 @@ namespace openspace::exoplanets {
...
     std::string description() override;
     void perform(const Task::ProgressCallback& progressCallback) override;
     static documentation::Documentation documentation();
 
+    /**
+     * The exoplanet file includes a set of commented lines in the beginning. This
+     * funciton skips those lines, as well as any empty lines, and continues reading
+     * until it find the first valid data line (which will be the column names).
+     *
+     * \param file The inpur file stream used for reading the file
```
alexanderbock, line 56:
```suggestion
     * \param file The input file stream used for reading the file
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -35,22 +36,69 @@ namespace openspace::exoplanets {
...
+     * Parse a row in the CSV file of exoplanets. Assumes the same format and column
+     * names as provided by the NASA Exoplanet Archive.
+     * (https://exoplanetarchive.ipac.caltech.edu/)
+     *
+     * \param row The row to parse, given as a string
+     * \param columnNames The list of column names in the file, from the CSV header
+     * \param positionSourceFile A SPECK file to use for getting the position of the star.
+     *  This is used to make sur ethe position of the star matches those of other star
+     *  datasets. If not file is provided, the position from the CSV data file is read
+     *  and used instead
```
alexanderbock, line 71:
```suggestion
     * \param positionSourceFile A SPECK file to use for getting the position of the star.
     *        This is used to make sure the position of the star matches those of other star
     *        datasets. If no file is provided, the position from the CSV data file is read
     *        and used instead
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -35,22 +36,69 @@ namespace openspace::exoplanets {
...
+     * \param bvFromTeffConversionFile A text file containing a mapping between Effective
+     *  temperatature (teff) values and B-V color index values. Each line should include
+     *  two values separated by a comma: first the teff value and hten the B-V value
```
alexanderbock, line 74:
```suggestion
     * \param bvFromTeffConversionFile A text file containing a mapping between effective
     *        temperature (teff) values and B-V color index values. Each line should include
     *        two values separated by a comma: first the teff value and hten the B-V value
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -35,22 +36,69 @@ namespace openspace::exoplanets {
...
     std::filesystem::path _outputBinPath;
     std::filesystem::path _outputLutPath;
     std::filesystem::path _teffToBvFilePath;
 
-    glm::vec3 starPosition(const std::string& starName);
+    /**
+     * Try to find the star position from an input speck file. If not found, the
+     * returned position will contain NaN values.
+     *
+     * \param starName The name of the star, to look for in the file
```
alexanderbock, line 93:
```suggestion
     * \param starName The name of the star to look for in the file
```

alexanderbock:
1. I think the functionality could be extracted into a different loader class, but I don't think it would be too important to do at this point. I could see us doing it if we discover the need for a different data format that we also need to be able to load.
2. Maybe we should have a way to visually show that we are missing some data, rather than using default values first before needing to care about the loading?  Alternatively, using the `std::optional` suggestion could be used already?
3. I don't think we need to worry about that.  We got plenty of places where we don't worry about the sizes. For example 3D models, etc.

WeirdRubberDuck:
> 1. I think the functionality could be extracted into a different loader class, but I don't think it would be too important to do at this point. I could see us doing it if we discover the need for a different data format that we also need to be able to load.
> 2. Maybe we should have a way to visually show that we are missing some data, rather than using default values first before needing to care about the loading?  Alternatively, using the `std::optional` suggestion could be used already?
> 3. I don't think we need to worry about that.  We got plenty of places where we don't worry about the sizes. For example 3D models, etc.

I agree with these points. 

But yeah, this is probably a good opportunity to look at better communicating when default values are used. Some kind of visual representation would be ideal, but some textual version could maybe also work (I've been thinking of adding some generated UI descriptions for the exoplanets, and it could maybe be included there) 

Would be great if we could discuss this during one of next week's meetings :)

modules/exoplanets/exoplanetshelper.h
```diff
@@ -30,52 +30,56 @@
...
-    double tt;          // Epoch of transit center in HJD-2440000
-    float ttUpper;      // Upper uncertainty of epoch of transit center
-    float ttLower;      // Lower uncertainty of epoch of transit center
-    // Star position's X-coordinate in parsec
-    float positionX = std::numeric_limits<float>::quiet_NaN();
-    // Star position's Y-coordinate in parsec
-    float positionY = std::numeric_limits<float>::quiet_NaN();
-    // Star position's Z-coordinate in parsec
-    float positionZ = std::numeric_limits<float>::quiet_NaN();
+    float a = fNaN;            // Orbital semi-major axis in AU
```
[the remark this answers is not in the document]
WeirdRubberDuck: I did it like this at first, but found it quite difficult to read so went with the committed method to try to mediate that. 

But yeah, you're right. The documentation not showing correctly is a problem. So I'll do it this way instead 👍

modules/exoplanets/exoplanetshelper.h
```diff
@@ -30,52 +30,56 @@
...
+    double aUpper = dNaN;      // Upper uncertainty of orbital semi-major axis
+    double aLower = dNaN;      // Lower uncertainty of orbital semi-major axis
+    float bigOmega = fNaN;     // Longitude of ascending node in degrees
+    float bigOmegaUpper = fNaN; // Upper uncertainty of longitude of ascending node
+    float bigOmegaLower = fNaN; // Lower uncertainty of longitude of ascending node
+    bool binary;                // Star known to be binary?
```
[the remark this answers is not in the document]
WeirdRubberDuck: What's a good value? Just false?

modules/exoplanets/exoplanetshelper.h
```diff
@@ -33,43 +33,95 @@
...
+
+    // Longitude of ascending node in degrees
+    float bigOmega = std::numeric_limits<float>::quiet_NaN();
+    // Upper uncertainty of longitude of ascending node
+    float bigOmegaUpper = std::numeric_limits<float>::quiet_NaN();
+    // Lower uncertainty of longitude of ascending node
+    float bigOmegaLower = std::numeric_limits<float>::quiet_NaN();
+
+    // Star known to be binary?
+    bool binary;
```
WeirdRubberDuck, line 51:
```suggestion
    bool binary = false;
```

PR #2857, opened 2023-08-14
Change name of line fade in RenderableTrail from Fade to LineFade

### LINKED ISSUE #2856: RenderableTrail has a property called Fade which shadows renderable property Fade
**Describe the bug**
RenderableTrail has a property called Fade (under Appearance) which shadows renderable property Fade. This leads to undefined behavior when using the URI for fade - it is unclear which fade is accessed. To fix - rename the fade property in RenderableTrails and everywhere it is used.

ylvaselling:
Close #2856

### DIFF
--- a/modules/base/rendering/renderabletrail.cpp
+++ b/modules/base/rendering/renderabletrail.cpp
@@ -86,7 +86,7 @@ namespace {
         openspace::properties::Property::Visibility::NoviceUser
     };
 
-    constexpr openspace::properties::Property::PropertyInfo FadeInfo = {
+    constexpr openspace::properties::Property::PropertyInfo LineFadeInfo = {
         "LineFade",
         "Line fade",
         "The fading factor that is applied to the trail if the 'EnableFade' value is "
@@ -138,7 +138,7 @@ namespace {
         // [[codegen::verbatim(EnableFadeInfo.description)]]
         std::optional<bool> enableFade;
 
-        // [[codegen::verbatim(FadeInfo.description)]]
+        // [[codegen::verbatim(LineFadeInfo.description)]]
         std::optional<float> lineFade;
 
         // [[codegen::verbatim(LineWidthInfo.description)]]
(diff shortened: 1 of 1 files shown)


WeirdRubberDuck:
Not sure that this change is necessary; Shouldn't we just be able to handle the checking against the property better in the UI?  Otherwise I'm pretty sure the error will emerge again at some point. 

Worth a discussion;  I know that there is some complication in doing this since not only renderables have the Fade property.

WeirdRubberDuck:
Looking a bit more into detail of this property, renaming it to "LineFade", makes sense. I mean, it shows up as "Line Fade" in the UI anyways 

However, we should still address the underlying cause of the problem, which is the check in the UI code.  But I think it's ok to go ahead and merge this

modules/base/rendering/renderabletrail.cpp
```diff
@@ -87,7 +87,7 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FadeInfo = {
```
WeirdRubberDuck, line 89:
```suggestion
    constexpr openspace::properties::Property::PropertyInfo LineFadeInfo = {
```

modules/base/rendering/renderabletrail.cpp
```diff
@@ -139,7 +139,7 @@ namespace {
         std::optional<bool> enableFade;
 
         // [[codegen::verbatim(FadeInfo.description)]]
```
WeirdRubberDuck, line 141:
```suggestion
        // [[codegen::verbatim(LineFadeInfo.description)]]
```

modules/base/rendering/renderabletrail.cpp
WeirdRubberDuck: Also in the constructor of the Appearance property owner

alexanderbock (approved):

PR #2866, opened 2023-08-22
Add options to disable roll and zoom in OrbitalNavigator

WeirdRubberDuck:
To discuss: Do we need similar settings for other camera motions, like panning and global rotation?  Should I add these too?

We used this setting in the Stort installation, to prevent the user from moving closer to objects in certain scenarios, without having to set exact distances for limiting the vertical navigation

Added warnings that show if either property is disabled. Feels like one of those things the user could easily forget is disabled, and be confused about. The warning increases the visibility of the effect, and warns the user about the consequences

### DIFF
--- a/include/openspace/navigation/orbitalnavigator.h
+++ b/include/openspace/navigation/orbitalnavigator.h
@@ -214,6 +214,9 @@ private:
 
     LimitZoom _limitZoom;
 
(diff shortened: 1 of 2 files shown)


alexanderbock (approved):
I'd only argue for addingthe other ones for symmetry, but can't think of a good usecase for them

PR #2868, opened 2023-08-23
Add a proper blackout to the ScreenSpaceRenderables and a property and codegen option to toggle that behavior

alexanderbock:
Adds a new property to all ScreenSpaceRenderables `RenderDuringBlackout` that controls whether a ScreenSpaceRenderable should listen to the BlackoutFactor of the RenderEngine or not.  If it does listen, it will darken the image by the same value as the 3D rendering, if it does not listen, it will always render at full brightness.
Also fixed the issue where no ScreenSpaceRenderables are rendered when the blackout factor is 0


Arohdin (approved):
Checked it functionality wise and it works as expected 👍

Issue #1372, opened 2020-11-05
Option to orbit at constant latitude


micahnyc:
When Orbiting Earth for example, it would be nice if you could rotate at a constant latitude so that you could show specific bands of the planet. 

Currently this is possible if the pilot adds the correct amount of constant roll as the rotation is being applied but an option could be added to the orbitalnavigator to do this for them.

WeirdRubberDuck:
This was added with PR #1667, as an "IdleBehavior" of the orbital navigator. The `OrbitAtConstantLatitude` option orbits around the "North vector" of a node, while keeping the camera's local rotation. If the up-vector points upwards on the screen, the camera will rotate to the right

The implementation is really simple and needs some testing. For example, maybe it's always desired to rotate to the right, even if a planet's up vector is pointing downwards on the screen?

alexanderbock:
@WeirdRubberDuck  isn't this fixed with the camera paths?

WeirdRubberDuck:
@alexanderbock the behavior Micah was looking for was something different. Rather than an automatic orbit, he wanted the ability to lock part of the rotation during manual rotation, so that we can for example only move the camera over a certain latitude band 

Do I remember that correctly, @micahnyc ?

alexanderbock:
Micah says "Yes"

Issue #740, opened 2018-10-26
Add "preferred up vector" for scene graph nodes + lock up vector


emiax:
I think it would be useful to be able to constrain navigation to use a certain up vector. Use cases:
- Keeping north pointing up
- Keeping the sky up when the focus node is on a planetary surface (rover or other point of interest)

There may be other ways to achieve the same thing, but just throwing this out here.

micahnyc:
I was thinking about this today, something less automatic but to be able to lock/unlock the friction on certain axis for roll. 

It would be helpful for 'helicopter' moves on planet surfaces.

Issue #1946, opened 2022-03-23
Add camera path helper data per scene graph node


WeirdRubberDuck:
Such as:
- Arrival distance factor (Factor to be multiplied with bounding sphere)
- Up vector (to be able to fly to up)
- Front / preferred view direction (?)

To be used by camera path system when computing target position

WeirdRubberDuck:
1. Would be particularly useful for the digital universe renderables

WeirdRubberDuck:
Needs more discussion when the work is starting (thinking of events and interaction spheres and factors)

WeirdRubberDuck:
Related issue: https://github.com/OpenSpace/OpenSpace/issues/740

PR #2874, opened 2023-08-30
Feature - OrbitalNavigator Setting - Orbit Around Up

WeirdRubberDuck:
Adds a setting to let the user orbit around a certain axis (referred to as the "up") using the OrbitalNavigator. The axis can be set to either the local X Y or Z of the anchor node. 

Input in the x-direction will rotate around the provided vector, while input in the y-direction will move the camera as regular.  It works best when the "up" vector is set to point upwards in screen-space (we should maybe add a way to enforce this)

Should close https://github.com/OpenSpace/OpenSpace/issues/1372  (and partly https://github.com/OpenSpace/OpenSpace/issues/740?)

Here's a demonstration video of how it works. Feel free to try it out.  
https://github.com/OpenSpace/OpenSpace/assets/8808894/51f0cf30-1088-4f3d-a3ee-148e57796e6b

### DIFF
--- a/include/openspace/navigation/orbitalnavigator.h
+++ b/include/openspace/navigation/orbitalnavigator.h
@@ -442,6 +442,9 @@ private:
      */
     void orbitAroundAxis(const glm::dvec3 axis, double angle, glm::dvec3& position,
         glm::dquat& globalRotation);
+
+    double rotationSpeedScaleFromCameraHeight(const glm::dvec3& cameraPosition,
+        const SurfacePositionHandle& positionHandle) const;
 };
 
 } // namespace openspace::interaction
--- a/include/openspace/navigation/orbitalnavigator.h
+++ b/include/openspace/navigation/orbitalnavigator.h
@@ -332,14 +332,24 @@ private:
         double targetDistance);
 
     /**
+     * Modify the camera position and global rotation to rotate around the up vector
+     * of the current anchor based on x-wise input
+     *
+     * The up-vector to rotate around is determined by the "_upToUseForRotation" property
+     */
+    void rotateAroundAnchorUp(double deltaTime, double speedScale,
+        glm::dvec3& cameraPosition, glm::dquat& globalCameraRotation);
+
+    /**
      * Translates the horizontal direction. If far from the anchor object, this will
      * result in an orbital rotation around the object. This function does not affect the
      * rotation but only the position.
      *
      * \return a position vector adjusted in the horizontal direction.
      */
-    glm::dvec3 translateHorizontally(double deltaTime, const glm::dvec3& cameraPosition,
-        const glm::dvec3& objectPosition, const glm::dquat& globalCameraRotation,
+    glm::dvec3 translateHorizontally(double deltaTime, double speedScale,
+        const glm::dvec3& cameraPosition, const glm::dvec3& objectPosition,
+        const glm::dquat& globalCameraRotation,
         const SurfacePositionHandle& positionHandle) const;
 
     /*
--- a/include/openspace/navigation/orbitalnavigator.h
+++ b/include/openspace/navigation/orbitalnavigator.h
@@ -233,6 +233,15 @@ private:
 
     properties::BoolProperty _invertMouseButtons;
 
+    properties::BoolProperty _shouldRotateAroundUp;
+
+    enum class UpDirectionChoice {
+        XAxis = 0,
+        YAxis,
+        ZAxis
+    };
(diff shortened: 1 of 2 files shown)


WeirdRubberDuck:
Would benefit from each type of scenegraph node knowing about it's up direction, as mentioned in:

https://github.com/OpenSpace/OpenSpace/issues/1946
https://github.com/OpenSpace/OpenSpace/issues/740

WeirdRubberDuck:
Did not manage to reproduce the black screen, but added some safety checks in cases where the rotation should not be updated just in case. 

Let me know if you notice the problem, and try to see if you can reproduce it

include/openspace/navigation/orbitalnavigator.h
```diff
@@ -233,6 +233,15 @@ class OrbitalNavigator : public properties::PropertyOwner {
 
     properties::BoolProperty _invertMouseButtons;
 
+    properties::BoolProperty _shouldRotateAroundUp;
```
alexanderbock, line 236: More a general question then just for here;  should it be called "rotate around axis" rather than up?  Since we can choose what "up" is, it might not match with the persons understanding of "up" necessarily

src/navigation/orbitalnavigator.cpp
```diff
@@ -698,6 +718,24 @@ OrbitalNavigator::OrbitalNavigator()
...
+            "Local X"
+        },
+        {
+            static_cast<int>(UpDirectionChoice::YAxis),
+            "Local Y"
+        },
+        {
+            static_cast<int>(UpDirectionChoice::ZAxis),
+            "Local Z"
+        },
```
alexanderbock, line 735:
```suggestion
        { static_cast<int>(UpDirectionChoice::XAxis), "Local X" },
        { static_cast<int>(UpDirectionChoice::YAxis), "Local Y" },
        { static_cast<int>(UpDirectionChoice::ZAxis), "Local Z" }
```

is a bit more compact to me and seems readable to me

src/navigation/orbitalnavigator.cpp
```diff
@@ -1583,83 +1637,87 @@ double OrbitalNavigator::interpolateCameraToSurfaceDistance(double deltaTime,
...
 
     // Transform to world space
     const glm::dquat rotationDiffWorldSpace = globalCameraRotation *
         joystickRotationDiffCamSpace * mouseRotationDiffCamSpace *
         websocketRotationDiffCamSpace * scriptRotationDiffCamSpace *
         glm::inverse(globalCameraRotation);
 
+    const glm::dmat4 modelTransform = _anchorNode->modelTransform();
+    const glm::dvec3 outDirection = glm::normalize(glm::dmat3(modelTransform) *
+        positionHandle.referenceSurfaceOutDirection);
```
alexanderbock, line 1713:
```suggestion
    const glm::dvec3 outDirection = glm::normalize(
        glm::dmat3(modelTransform) *
        positionHandle.referenceSurfaceOutDirection
    );
```

alexanderbock (approved):

include/openspace/navigation/orbitalnavigator.h
WeirdRubberDuck: I get your point. We should probably discuss it. 

The reason I chose to use "Up" is that I consider the axis property to be a "temporary" property, until each scene graph node knows about its up direction (https://github.com/OpenSpace/OpenSpace/issues/740)

However ,I guess that could be seen as a 4th axis option and that we could still want to orbit around any of the 3 axes

alexanderbock: Ah.. in that case I withdraw my comment.  Using the Up vector of the scene graph nodes is a lot better.. found the x/y/z choices to be confusing and removing the dropdown down the line makes sense :+1:

src/navigation/orbitalnavigator.cpp
```diff
@@ -1583,83 +1637,87 @@ double OrbitalNavigator::interpolateCameraToSurfaceDistance(double deltaTime,
...
+        break;
+    case static_cast<int>(UpDirectionChoice::YAxis):
+        axis = glm::dvec3(0.0, 1.0, 0.0);
+        break;
+    case static_cast<int>(UpDirectionChoice::ZAxis):
+        axis = glm::dvec3(0.0, 0.0, 1.0);
+        break;
+    default:
+        throw ghoul::MissingCaseException();
+    }
```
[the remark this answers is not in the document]
WeirdRubberDuck: Nice, thanks! I like that

Issue #2897, opened 2023-09-22
macOS/Linux build failure


garrettsummerfi3ld:
**Describe the bug**

A compilation of OpenSpace cannot be completed on macOS/Linux platforms.

This is the output of a Ubuntu 22.04 LTS with GNU GCC11 build with `make`:

```cpp
[ 71%] Building CXX object src/CMakeFiles/openspace-core.dir/documentation/core_registration.cpp.o
In file included from /home/runner/work/OpenSpace/OpenSpace/src/documentation/core_registration.cpp:41:
/home/runner/work/OpenSpace/OpenSpace/include/openspace/rendering/renderable.h:163:53: error: could not convert ‘<brace-enclosed initializer list>()’ from ‘<brace-enclosed initializer list>’ to ‘const openspace::Renderable::AlternativeTransform&’
  163 |         const AlternativeTransform& altTransform = {}) const;
      |                                                     ^
      |                                                     |
      |                                                     <brace-enclosed initializer list>
/home/runner/work/OpenSpace/OpenSpace/include/openspace/rendering/renderable.h:207:82: error: could not convert ‘<brace-enclosed initializer list>()’ from ‘<brace-enclosed initializer list>’ to ‘const openspace::Renderable::AlternativeTransform&’
  207 |         const RenderData& data, const AlternativeTransform& altModelTransform = {}) const;
      |                                                                                  ^
      |                                                                                  |
      |                                                                                  <brace-enclosed initializer list>
cc1plus: note: unrecognized command-line option ‘-Wno-unknown-attributes’ may have been intended to silence earlier diagnostics
gmake[3]: *** [src/CMakeFiles/openspace-core.dir/build.make:125: src/CMakeFiles/openspace-core.dir/documentation/core_registration.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:2393: src/CMakeFiles/openspace-core.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:4224: apps/OpenSpace/CMakeFiles/OpenSpace.dir/rule] Error 2
gmake: *** [Makefile:1112: OpenSpace] Error 2
Error: Process completed with exit code 2.
```

**To Reproduce**
1. Clone repository
2. Run Cmake steps
3. Build
4. Build errors

**Expected behavior**

Completed build without errors

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04 LTS, macOS 12, 13, 14
 - Compilers:
[...]

PR #2898, opened 2023-09-22
Non-Windows compiler fix for empty initializer list

GPayne:
Compilers besides Windows msvc throw an error when encountering an empty initializer list for a struct, even though that struct is fully-specified with default values.
This fix explicity provides the three `std::nullopt` values for the `AlternativeTransform` struct.

### DIFF
--- a/include/openspace/rendering/renderable.h
+++ b/include/openspace/rendering/renderable.h
@@ -160,7 +160,9 @@ protected:
      * \return the resulting model transformation matrix in double precision
      */
     glm::dmat4 calcModelTransform(const RenderData& data,
-        const AlternativeTransform& altTransform = {}) const;
+        const AlternativeTransform& altTransform = {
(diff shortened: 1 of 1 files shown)


garrettsummerfi3ld:
This commit does solve issue #2897 on my builds.

alexanderbock:
Seems to be a known GCC and Clang bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88165.
The problem is that the `AlternativeTransform` struct is not considered complete until the surrounding `Renderable` class is finished.  So the options would be to move the `AlternativeTransform` out of the `Renderable` class or the functions out of the `Renderable` class.

Can't think of a better way since we:
1. Want it to be default constructed
2. Want the functions to be protected, so they can't be freestanding
3. Want the AlternativeTransform to be protected since noone else but subclasses should care about the type

PR #2916, opened 2023-10-11
Simplify Apollo LEM Model Rotation and allow using IdleBehavior to or…

WeirdRubberDuck:
…bit around up-vector

Previously, the resulting coordinate system was skewed in relation to the surface (as only a GlobeTranslation and no GlobeRotation was added)


WeirdRubberDuck:
Note that the 11 LEM model I changes is apparently not the one used in the apollo profile... Very confused about this!

data/assets/scene/solarsystem/missions/apollo/17/lem.asset
```diff
@@ -21,6 +21,12 @@ local Apollo17Lem = {
       Longitude = -329.22833,
       Latitude = 20.19092,
       UseHeightmap = true
+    },
+    Rotation = {
+      Type = "GlobeRotation",
+      Globe = moonAsset.Moon.Identifier,
+      Longitude = -329.22833,
```
WeirdRubberDuck, line 28: OBS! Apparnetly these coordinates are not correct!

data/assets/scene/solarsystem/missions/apollo/11/lem.asset
```diff
@@ -42,7 +48,7 @@ local Apollo11LemModel = {
   Renderable = {
     Type = "RenderableModel",
     GeometryFile = lem_model .. "LM-2_ver2clean.obj",
-    RotationVector = { 91.044090, 171.229706, 111.666664 },
+    RotationVector = { -90.0, 0, 0 },
```
WeirdRubberDuck, line 51: Fix rotation so that it has the correct offset and orientation as before

data/assets/scene/solarsystem/missions/apollo/17/lem.asset
```diff
@@ -43,7 +49,7 @@ local Apollo17LemModel = {
     Type = "RenderableModel",
     GeometryFile = model .. "LM-2_ver2clean.obj",
     SpecularIntensity = 0.0,
-    RotationVector = { 110.255219, 171.229706, 126.666664 },
+    RotationVector = { -90.0, 0, 0 },
```
WeirdRubberDuck, line 52: Fix rotation so that it has the correct offset and orientation as before

alexanderbock:
Would need correct numbers on rotation vectors

PR #2917, opened 2023-10-13
Layers maintenance

Roxeena:
All layers are now separated into different assets depending on our different servers to make it easier for users to choose which server they want to use. 

Something to discuss is how we would like to make it easier for users to set what server they want to use at a higher level. Should it be a .cfg variable, an environment variable, a UserProperty, etc... And how should we handle backward compatibility for existing profiles and the _base.asset_ which already includes the _default_layers.asset_?

In the future, we might want to add some way to automatically choose which server to use depending on location but this is not included in this PR.


openspace.cfg
```diff
@@ -228,10 +228,10 @@ ShutdownCountdown = 3
 ScreenshotUseDate = true
 BypassLauncher = false
 
-PreferedLayerServer = "All"
--- PreferedLayerServer = "NewYork"
--- PreferedLayerServer = "Sweden"
--- PreferedLayerServer = "Utah"
+-- LayerServer = "All"
```
alexanderbock, line 231: I think we should keep the default as "All" until we automatically detect the location

src/engine/configuration.cpp
```diff
@@ -581,4 +560,22 @@ Configuration loadConfigurationFromFile(const std::filesystem::path& configurati
...
+}
+
+std::string layerServerToString(openspace::Configuration::LayerServer server) {
+    using Server = openspace::Configuration::LayerServer;
+    switch (server) {
+        case Server::All: return "All";
+        case Server::NewYork: return "NewYork";
+        case Server::Sweden: return "Sweden";
+        case Server::Utah: return "Utah";
+    }
```
alexanderbock, line 578:
```suggestion
        default: throw ghoul::MissingCaseException();
    }
```

or else a warning is raised here because of a missing return value

src/engine/configuration.cpp
```diff
@@ -581,4 +560,22 @@ Configuration loadConfigurationFromFile(const std::filesystem::path& configurati
     return result;
 }
 
+openspace::Configuration::LayerServer stringToLayerServer(std::string_view server) {
+    using Server = openspace::Configuration::LayerServer;
+    if (server == "All") { return Server::All; }
+    else if (server == "NewYork") { return Server::NewYork; }
+    else if (server == "Sweden") { return Server::Sweden; }
+    else if (server == "Utah") { return Server::Utah; }
+}
```
alexanderbock, line 569:
```suggestion
    throw ghoul::MissingCaseException();
}
```
or else a warning is raised here because of a missing return value

src/engine/openspaceengine_lua.inl
```diff
@@ -235,4 +235,24 @@ namespace {
     openspace::setCameraFromProfile(*openspace::global::profile);
 }
 
+/**
+ * Returns the whole configuration object as a Dictionary
+ */
+[[codegen::luawrap]] ghoul::Dictionary configuration() {
+    ghoul::Dictionary res;
```
alexanderbock, line 242: Can move the `ghoul::Dictionary res;` declaration down to line 246. No need to have the value while accessing the configuration

openspace.cfg
Roxeena: Yea ofcourse, ops ^^'

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+    }
+    res.setValue("GlobalCustomizationScripts", globalCustomizationScriptsDict);
+
+    ghoul::Dictionary fontsDict;
+    for (auto it = fonts.begin(); it != fonts.end(); ++it) {
+        fontsDict.setValue(it->first, it->second);
+    }
+    res.setValue("Fonts", fontsDict);
+
+    ghoul::Dictionary fontSizeDict; {
```
alexanderbock, line 344: No need for the `{` and the extra indentation here

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+        fontSizeDict.setValue("FrameInfo", static_cast<double>(fontSize.frameInfo));
+        fontSizeDict.setValue("Shutdown", static_cast<double>(fontSize.shutdown));
+        fontSizeDict.setValue("Log", static_cast<double>(fontSize.log));
+        fontSizeDict.setValue("CameraInfo", static_cast<double>(fontSize.cameraInfo));
+        fontSizeDict.setValue("VersionInfo", static_cast<double>(fontSize.versionInfo));
+    }
+    res.setValue("FontSize", fontSizeDict);
+
+    ghoul::Dictionary loggingDict; {
+        loggingDict.setValue("Level", logging.level);
```
alexanderbock, line 354: Same here

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+        for (size_t i = 0; i < logging.logs.size(); ++i) {
+            logsDict.setValue(std::to_string(i), logging.logs[i]);
+        }
+        loggingDict.setValue("Logs", logsDict);
+    }
+    res.setValue("Logging", loggingDict);
+
+    res.setValue("ScriptLog", scriptLog);
+
+    ghoul::Dictionary documentationDict; {
```
alexanderbock, line 368: and here

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+        documentationDict.setValue("Path", documentation.path);
+    }
+    res.setValue("Documentation", documentationDict);
+
+    res.setValue("VersionCheckUrl", versionCheckUrl);
+    res.setValue("UseMultithreadedInitialization", useMultithreadedInitialization);
+
+    ghoul::Dictionary loadingScreenDict; {
```
alexanderbock, line 376: and here

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+            "IsShowingLogMessages",
+            loadingScreen.isShowingLogMessages
+        );
+    }
+    res.setValue("LoadingScreen", loadingScreenDict);
+
+    res.setValue("IsCheckingOpenGLState", isCheckingOpenGLState);
+    res.setValue("IsLoggingOpenGLCalls", isLoggingOpenGLCalls);
+    res.setValue("IsPrintingEvents", isPrintingEvents);
+    res.setValue("consoleKey", ghoul::to_string(consoleKey));
```
alexanderbock, line 389:
```suggestion
    res.setValue("ConsoleKey", ghoul::to_string(consoleKey));
```
?

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+    res.setValue("BypassLauncher", bypassLauncher);
+    res.setValue("LayerServer", layerServerToString(layerServer));
+
+    ghoul::Dictionary moduleConfigurationsDict;
+    for (auto it = moduleConfigurations.begin(); it != moduleConfigurations.end(); ++it) {
+        moduleConfigurationsDict.setValue(it->first, it->second);
+    }
+    res.setValue("ModuleConfigurations", moduleConfigurationsDict);
+
+    ghoul::Dictionary openGLDebugContextDict; {
```
alexanderbock, line 408: same here

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+        openGLDebugContextDict.setValue("IsActive", openGLDebugContext.isActive);
+        openGLDebugContextDict.setValue("PrintStacktrace", openGLDebugContext.printStacktrace);
+        openGLDebugContextDict.setValue("IsSynchronous", openGLDebugContext.isSynchronous);
+
+        ghoul::Dictionary identifierFiltersDict;
+        for (size_t i = 0; i < openGLDebugContext.severityFilters.size(); ++i) {
+            ghoul::Dictionary identifierFilterDict; {
```
alexanderbock, line 415: and here

src/engine/configuration.cpp
```diff
@@ -318,6 +318,138 @@ namespace {
...
+    for (size_t i = 0; i < openGLDebugContext.severityFilters.size(); ++i) {
+        severityFiltersDict.setValue(
+            std::to_string(i),
+            openGLDebugContext.severityFilters[i]
+        );
+    }
+    openGLDebugContextDict.setValue("SeverityFilters", severityFiltersDict);
+    res.setValue("OpenGLDebugContext", severityFiltersDict);
+
+    ghoul::Dictionary httpProxyDict; {
```
alexanderbock, line 438: and here

src/engine/configuration.cpp
Roxeena: I added it to make the code a little more clear since it is dictionaries inside dictionaries, but I can change it if we want. I just thought it made it more readable.

alexanderbock: I think we could do the same thing with empty lines instead, breaking the different Dictionaries into more visible blocks by having two empty lines before and after for example.
If you really want to keep the additional scopes, putting the `{` on a separate line would be less jarring and look less like a mistake:

```cpp
ghoul::Dictionary fontSizeDict;
{
        fontSizeDict.setValue("FrameInfo", static_cast<double>(fontSize.frameInfo));
        fontSizeDict.setValue("Shutdown", static_cast<double>(fontSize.shutdown));
        fontSizeDict.setValue("Log", static_cast<double>(fontSize.log));
        fontSizeDict.setValue("CameraInfo", static_cast<double>(fontSize.cameraInfo));
        fontSizeDict.setValue("VersionInfo", static_cast<double>(fontSize.versionInfo));
}
```

though I do think that they distract more than they help

src/engine/configuration.cpp
```diff
@@ -283,6 +304,11 @@ namespace {
         // multiprojector setups where a launcher window would be undesired
         std::optional<bool> bypassLauncher;
 
+        // Set which layer server should be preferd to be used, the options are
+        // \"All\", \"NewYork\", \"Sweden\" and \"Utah\".
+        std::optional<std::string> preferedLayerServer [[codegen::inlist("All", "NewYork",
```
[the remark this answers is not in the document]
Roxeena: Alex and I talked about this and think this can be resolved by adding the option "None". That gives the user complete freedom to add any layer they want, we will just include the layers that do not depend on any single server and they can add anything they want on top of that.

apps/OpenSpace/ext/launcher/src/settingsdialog.cpp
```diff
@@ -269,11 +270,37 @@ void SettingsDialog::createWidgets() {
...
+            "Sweden",
+            "Utah",
+            "None"
+        });
+        _layerServer->setCurrentText("All");
+        connect(
+            _layerServer,
+            &QComboBox::textActivated,
+            [this](const QString& value) {
+                _currentEdit.layerServer = openspace::stringToLayerServer(value.toStdString());
```
alexanderbock, line 297: This line looks like it is too long?

apps/OpenSpace/ext/launcher/src/settingsdialog.cpp
```diff
@@ -269,11 +270,37 @@ void SettingsDialog::createWidgets() {
...
     {
-        QLabel* label = new QLabel("MRF Caching");
+        QLabel* label = new QLabel("Layer Settings");
         label->setObjectName("heading");
         layout->addWidget(label, 13, 0, 1, 2);
 
-        _mrf.isEnabled = new QCheckBox("Enable Caching");
+        QLabel* conf = new QLabel("Layer Server");
+        conf->setToolTip(
+            "This setting sets the default server to be used for the layers"
```
alexanderbock, line 279:
```suggestion
            "This setting sets the default server to be used for the layers "
            "that are hosted by the OpenSpace team"
```

data/assets/base.asset
```diff
@@ -19,13 +19,18 @@ asset.require("scene/milkyway/constellations/constellation_keybinds")
...
+elseif (server == "NewYork") then
   asset.require("scene/solarsystem/planets/newyork_layers")
 elseif (server == "Sweden") then
   asset.require("scene/solarsystem/planets/sweden_layers")
 elseif (server == "Utah") then
   asset.require("scene/solarsystem/planets/utah_layers")
+elseif (server == "None") then
+  -- Do nothing
 else
+  openspace.printWarning("Did not recognise the specified Layer server, defaults to all")
```
alexanderbock:
```suggestion
  openspace.printWarning("Did not recognize the specified Layer server, defaults to 'All'")
```

*eagle flying above*

src/engine/settings.cpp
```diff
@@ -74,6 +74,11 @@ namespace version1 {
         }
         settings.bypassLauncher = get_to<bool>(json, "bypass");
 
+        std::optional<std::string> layerServer = get_to<std::string>(json, "layerServer");
```
alexanderbock, line 77:
```suggestion
        std::optional<std::string> layerServer = get_to<std::string>(json, "layerserver");
```

Since we do lowercase characters for the keys in JSON objects

src/engine/settings.cpp
```diff
@@ -165,6 +170,9 @@ void saveSettings(const Settings& settings, const std::filesystem::path& filenam
     if (settings.bypassLauncher.has_value()) {
         json["bypass"] = *settings.bypassLauncher;
     }
+    if (settings.layerServer.has_value()) {
+        json["layerServer"] = layerServerToString(*settings.layerServer);
```
alexanderbock, line 174:
```suggestion
        json["layerserver"] = layerServerToString(*settings.layerServer);
```

alexanderbock (approved):
:+1:

PR #2932, opened 2023-11-06
Point cloud rewrite (RIP RenderableBillboardsCloud)

### LINKED ISSUE #1746: FadeDistance appears to only respond to distances relative to the Sun/absolute origin
The FadeDistance value for objects (such as renderablestar or renderableplaneimage) only appears to respond to distance from the absolute origin of the universal coordinate space in software--that is, the Sun/Solar System barycenter. For approaching exoplanet systems it would be good to be able to set a distancefade to replicate the effect of the milky way version of the star fading out and being replaced with the local in-system star glare. FadeDistance should be relative to local distance from the origin of the specified object, not the Sun.

WeirdRubberDuck:
Rewrite point cloud renderables (RenderableBillboardsCloud) to make it more intuitive and generic. 

Some changes/news: 
- Introduces a data loader that can load CSV files in addition to Speck files (only numeric attributes work for now)
- The XYZ positions for the points can be set to any column in a CSV dataset using the `DataMapping` component
- RenderableBillboardsCloud has been split into two: A base calss `RenderablePointCloud` and sub-class `RenderablePolygonCloud`. These live in the base module, rather than digitaluniverse
- Everything related to Scale/Sizing has been combined into one property owner: SizeSettings. The settings have also been renamed to be a bit more intuitive/self-explanatory
  - There is also a new `ScaleFactor` that actually works as a factor on the visual size of the points. The previous one has been renamed to `ScaleExponent`
  - The `ScaleExponent` is now computed to a "suitable-ish" value if not specified, based on the largest coordinate in the dataset. That way, the points should always be visible even if no exponent is included in the asset
- Generic components such as LabelsComponent and DataLoader/SpeckLoader have been moved to core (from the space module)
- Adds a reusable `ColorMappingComponent`; a property owner that includes settings related to the color mapping based on (currently) cmap files.  There is a plan to keep working on this, possibly together with @ElonOlsson. In the future, we will probably have some kind of json based format for both transfer functions and colormaps
- Cmap files can now include specific colors for `aboveRange`, `belowRange` and `nan` values. Makes them a bit more verbose and easier to understand. These can also be set in the ColorMapping settings in the assets.
- A texture sprite for the point is no longer mandatory
- Removes (unused) existing RenderablePoints class
- Fix fading so that it is done based on the local coordinate system and not world space (fixes #1746), and allow inverted fading direction (see example asset)

~~Will have to update the datasets that include the colormaps. [Here is a link to a folder with updated files that can be used for testing](https://liuonline-my.sharepoint.com/:f:/g/personal/emmbr26_liu_se/EmVYTP0UPAtDup_6rJyPNBEB_zSQ9llbuDefKTbZmHmw6g?e=MwS6Qt). Just replace your files in your sync folder with these. Note that you may also have to remove the files from your cache folder~~ 
The files are now updates and synced 

OBS! Note that all the DU renderables should look exactly the same. The exception is the "Local Group" (localdwarfs.asset), as this had a color map that was not completely correctly used. We should decide what to do with this.


WeirdRubberDuck:
OBS! Will create an example asset that showcases all of the new features, includingthe CSV support, in a better way.

WeirdRubberDuck:
CSV support implemented (positions columns must be called x, y, and z for now) and example asset added

ElonOlsson:
I checked it out and tried it a little and just wanted to add my comments, 
* It looked like Earths atmosphere was shining through the circles, as well as orbit lines
* For some of them I was confused about the opacity. It said 100% but I could still see points shine through each other. Is it an adaptive blending type thing? If so, maybe it would be nice with a property to toggle that on and off?

WeirdRubberDuck:
> I checked it out and tried it a little and just wanted to add my comments,
> 
> * It looked like Earths atmosphere was shining through the circles, as well as orbit lines
> * For some of them I was confused about the opacity. It said 100% but I could still see points shine through each other. Is it an adaptive blending type thing? If so, maybe it would be nice with a property to toggle that on and off?

Thanks for testing it! 

1. Yes I have also noticed the issues with the Earth atmosphere there.  Not sure if it is worth fixing in this PR
2. Yes, that has to do with the blending. There is an option to "Enable Additive Blending" for the points. I have that setting set to false for a couple of the point clouds in the example asset, and in those cases there is no blending.

data/assets/scene/digitaluniverse/deepsky.asset
```diff
@@ -25,22 +25,16 @@ local DeepSkyObjects = {
       MinMaxSize = { 16, 20 },
       Unit = "pc"
     },
-    Color = { 1.0, 1.0, 0.0 },
```
WeirdRubberDuck, line 28: This color seem to have disappeared!

data/assets/scene/digitaluniverse/deepsky.asset
```diff
@@ -25,22 +25,16 @@ local DeepSkyObjects = {
...
-    --ColorOption = { "proximity" },
-    --ColorOption = { "prox5Mpc" },
-    --ColorRange = { { 1.0, 30.0 } },
     Unit = "pc",
     --FadeInDistances = { 0.05, 1.0 }, -- Fade in value in the same unit as "Unit"
-    BillboardMinMaxSize = { 0.0, 8.22 }, -- in pixels
-    --CorrectionSizeEndDistance = 22.0,
-    --CorrectionSizeFactor = 10.45
-    EnablePixelSizeControl = true
+    SizeSettings = {
```
WeirdRubberDuck, line 33: OBS!  Something goes wrong with the sizing.  These look huge from Earth, and they didn't use to.  Should be fixed!

WeirdRubberDuck: Need to double-check this.

include/openspace/data/dataloader.h
```diff
@@ -22,20 +22,27 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#ifndef __OPENSPACE_MODULE_SPACE___SPECKLOADER___H__
-#define __OPENSPACE_MODULE_SPACE___SPECKLOADER___H__
+#ifndef __OPENSPACE_CORE___DATALOADER___H__
```
WeirdRubberDuck, line 25: Should this file maybe actually be called "Point Data Loader"? 

Since it's so specific to point-based datasets?

WeirdRubberDuck:
Ready for review!   But I still need to update the datafiles and upload them to the server, so for now, replace the files in your sync folder with the ones in the link above when testing 

Here it is again
https://liuonline-my.sharepoint.com/:f:/g/personal/emmbr26_liu_se/EmVYTP0UPAtDup_6rJyPNBEB_zSQ9llbuDefKTbZmHmw6g?e=MwS6Qt

data/assets/examples/pointclouds/points.asset
```diff
@@ -0,0 +1,232 @@
...
+    Type = "RenderablePointCloud",
+    File = asset.resource("data/dummydata.csv"),
+    Coloring = {
+      ColorMapping = {
+        File = asset.resource("viridis.cmap")
+      }
+    },
+    SizeSettings = {
+      -- Here we set the exponent for the scale explicitly, to a value that
+      -- gives the point a suitable size based on their world-space coordinates
```
WeirdRubberDuck, line 73:
```suggestion
      -- gives the points a suitable size based on their world-space coordinates
```

data/assets/examples/pointclouds/points_datamapping.asset
```diff
@@ -0,0 +1,66 @@
...
+asset.export(Example)
+
+
+
+asset.meta = {
+  Name = "Example - Point Cloud with Custom Data Mapping",
+  Version = "1.0",
+  Description = [[Example of a point cloud where the X, Y and Z position are mapped to
+    other columns in the dataset. The data mapping also includes some settings for
+    missing data values and columns to exclude]],
```
WeirdRubberDuck, line 62:
```suggestion
    missing data values and columns to exclude when loading the dataset]],
```

data/assets/scene/digitaluniverse/abell.asset
```diff
@@ -34,16 +34,20 @@ local Object = {
       Unit = "Mpc",
       TransformationMatrix = TransformMatrix
     },
-    Color = { 1.0, 0.4, 0.2 },
     Opacity = 1.0,
-    --ColorMap = speck .. "abell.cmap",
+    Coloring = {
+      FixedColor = { 1.0, 0.4, 0.2 },
+      --ColorMap = speck .. "abell.cmap", -- TODO: Decide whether to add
```
WeirdRubberDuck, line 40: DiscussDicsuss whether this should be added or removed

include/openspace/data/dataloader.h
```diff
@@ -22,20 +22,28 @@
...
 #include <string>
 #include <vector>
 
-namespace openspace::speck {
-
-BooleanType(SkipAllZeroLines);
+namespace openspace::dataloader {
 
+/**
+ * A dataset representing object with positions and various other data columns.
```
WeirdRubberDuck, line 40:
```suggestion
 * A dataset representing objects with positions and various other data columns.
```

include/openspace/rendering/colormappingcomponent.h
```diff
@@ -0,0 +1,108 @@
...
+#include <openspace/properties/vector/vec4property.h>
+#include <ghoul/opengl/texture.h>
+
+namespace openspace {
+
+namespace documentation { struct Documentation; }
+
+/**
+ * This is a component that can be used to consistently hold parameters and properties
+ * for controlling color mapping between different renderables. This includes things
```
WeirdRubberDuck, line 45:
```suggestion
 * for controlling color mapping in different types of renderables. This includes things
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1020 @@
...
+        "This read only property includes information about how many points are being "
+        "rendered.",
+        openspace::properties::Property::Visibility::User
+    };
+
+    // A RenderablePointCloud can be used to render point-based datasets in 3D space,
+    // optionally including color mapping, a sprite texture and labels.
+    //
+    // There are a lot of properties that affect the visuals of the points, such as
+    // a scale,.... @TODO: Elaborate
```
WeirdRubberDuck, line 227: TODO: Finish!

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1020 @@
...
+        return std::vector<float>();
+    }
+
+    std::vector<float> result;
+    result.reserve(nAttributesPerPoint() * _dataset.entries.size());
+
+    // what datavar in use for the index color
+    int colorParamIndex = currentColorParameterIndex();
+
+    // what datavar in use for the size scaling (if present)z
```
WeirdRubberDuck, line 983:
```suggestion
    // What datavar is in use for the size scaling (if present)
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1020 @@
...
+    ZoneScoped;
+
+    if (_dataset.entries.empty()) {
+        return std::vector<float>();
+    }
+
+    std::vector<float> result;
+    result.reserve(nAttributesPerPoint() * _dataset.entries.size());
+
+    // what datavar in use for the index color
```
WeirdRubberDuck, line 980:
```suggestion
    // What datavar is in use for the index color
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1020 @@
...
+    int colorParamIndex = currentColorParameterIndex();
```
WeirdRubberDuck, line 981: Consider updating rendering code so that we only need to update these values when the color parameter option changes, instead of regenerating the entire data slice

(Same for the size variable below)

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,190 @@
...
+    LDEBUG("Creating Polygon Texture");
+    constexpr gl::GLsizei TexSize = 512;
+
+    glGenTextures(1, &_pTexture);
+    glBindTexture(GL_TEXTURE_2D, _pTexture);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+    // Stopped using a buffer object for GL_PIXEL_UNPACK_BUFFER
```
WeirdRubberDuck, line 105: What does this comment mean ? Remove?

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,190 @@
...
+    }
+    if (_polygonVbo == 0) {
+        glGenBuffers(1, &_polygonVbo);
+    }
+
+    glBindVertexArray(_polygonVao);
+    glBindBuffer(GL_ARRAY_BUFFER, _polygonVbo);
+
+    constexpr std::array<GLfloat, 4> VertexData = {
+        //      x      y     z     w
```
WeirdRubberDuck, line 157:
```suggestion
        //      x      y     z     w
```
```suggestion
        //  x  y  z  w
```

modules/base/rendering/pointcloud/renderablepolygoncloud.h
```diff
@@ -0,0 +1,71 @@
...
+
+namespace ghoul::opengl { class Texture; }
+
+namespace openspace {
+
+namespace documentation { struct Documentation; }
+
+/**
+ * A billboarded point cloud, but with dynamically created polygon shapes instead of a
+ * custom texture. Overwrites the sprite set in
```
WeirdRubberDuck, line 40: Fix comment (ends abruptly)

modules/base/shaders/billboardpoint_fs.glsl
```diff
@@ -24,45 +24,86 @@
...
+
 uniform bool useColorMap;
+uniform sampler1D colorMapTexture;
+uniform float cmapRangeMin;
+uniform float cmapRangeMax;
+uniform bool hideOutsideRange;
+
 uniform float fadeInValue;
 
+vec4 sampleColorMap(float dataValue) {
```
WeirdRubberDuck, line 54: This is the shader snippet that the ColorMappingComponent could generate. 

How best to do this?

modules/base/shaders/billboardpoint_gs.glsl
```diff
@@ -137,30 +112,21 @@ void main() {
     // width and height
     vec2 sizes = abs(halfViewSize * (topRight - bottomLeft));
 
-    if (enabledRectSizeControl && (length(sizes) > maxBillboardSize)) {
+    if (length(sizes) > maxBillboardSize) {
```
WeirdRubberDuck, line 115: TODO: Should verify that this works as we expect it in the dome

src/data/csvloader.cpp
```diff
@@ -0,0 +1,174 @@
...
+        float result;
+        auto [p, ec] = std::from_chars(str.data(), str.data() + str.size(), result);
+        if (ec == std::errc() && std::isfinite(result)) {
+            return result;
+        }
+        return std::numeric_limits<float>::quiet_NaN();
+#else
+        // clang is missing float support for std::from_chars
+        return !str.empty() ? std::stof(str.c_str(), nullptr) : NAN;
+        // @TODO: This won't work with non-finite or any type of invalid argument..
```
WeirdRubberDuck, line 60: Help!  How to fix this for clang?

src/data/dataloader.cpp
```diff
@@ -0,0 +1,698 @@
...
+        }
+        LINFOC("DataLoader", fmt::format("Loading file {}", filePath));
+        T dataset = loadFunction(filePath, specs);
+
+        if (!dataset.entries.empty()) {
+            LINFOC("DataLoader", "Saving cache");
+            saveCacheFunction(dataset, cached);
+        }
+
+        if (specs.has_value() && !(*specs).excludeColumns.empty()) {
```
WeirdRubberDuck, line 110: Remove this!

src/data/dataloader.cpp
```diff
@@ -0,0 +1,698 @@
...
+        // Text
+        checkSize<uint16_t>(e.text.size(), "Text too long");
+        uint16_t len = static_cast<uint16_t>(e.text.size());
+        file.write(reinterpret_cast<const char*>(&len), sizeof(uint16_t));
+        file.write(e.text.data(), len);
+    }
+}
+
+Labelset loadFileWithCache(std::filesystem::path filePath)
+{
```
WeirdRubberDuck, line 460: Can be moved to previous line

src/data/datamapping.cpp
```diff
@@ -0,0 +1,186 @@
...
+        std::optional<std::string> x;
+
+        // Specifies the column name for the y coordinate
+        std::optional<std::string> y;
+
+        // Specifies the column name for the z coordinate
+        std::optional<std::string> z;
+
+        // Specifies whether to do case sensitive checks when reading column names.
+        // Deafault is not to, so that 'X' nad 'x' are both valid column names for the
```
WeirdRubberDuck, line 92:
```suggestion
        // Default is not to, so that 'X' and 'x' are both valid column names for the
```

src/rendering/colormappingcomponent.cpp
```diff
@@ -257,8 +266,13 @@ ColorMappingComponent::ColorMappingComponent(const ghoul::Dictionary& dictionary
...
 
-    // @TODO: read valueRange from asset if specified. How to avoid overriding it
-    // in initialize?
+    if (p.parameter.has_value()) {
+        _providedParameter = *(p.parameter);
+    }
+
+    if (p.valueRange.has_value()) {
+        _providedRange = *(p.valueRange);
+    }
```
alexanderbock, line 275:
```suggestion
    _providedParameter = p.parameter.value_or(_providedParameter);
    _providedRange = p.valueRange.value_or(_providedRange);
```

?

WeirdRubberDuck: Actually, this does not work.  The `_providedParameter` and `_providedRange` variables are std::optional, and cannot be used as default values for the `value_or` function

WeirdRubberDuck: But setting them to just the value in the parameters work: 
```
    _providedParameter = p.parameter;
    _providedRange = p.valueRange;
```

data/assets/scene/digitaluniverse/exoplanets.asset
```diff
@@ -64,7 +63,7 @@ asset.export(Object)
 
 asset.meta = {
   Name = "Exoplanets",
-  Version = "1.1",
+  Version = "3.0",
```
WeirdRubberDuck, line 66:
```suggestion
  Version = "2.0",
```

Roxeena (commented):
Tested it out, looks good 👍

data/assets/scene/digitaluniverse/quasars.asset
```diff
@@ -23,21 +23,32 @@ local colormaps = asset.resource({
...
-    ScaleFactor = 540.9,
-    FadeInDistances = { 1000.0, 10000.0 }, -- Fade in value in the same unit as "Unit"
-    BillboardMinMaxSize = { 0.0, 11.1 },
-    EnablePixelSizeControl = true,
-    ColorMap = colormaps .. "viridis.cmap",
-    ColorOption = { "redshift", "Tlookback", "distMpc" },
-    ColorRange = { { 0.102, 7.085 }, { 1.4, 13.0 }, { 440.5, 8852.099609 } },
-    UseColorMap = false
+    Fading = {
+      FadeInDistances = { 1000.0, 10000.0 } -- Fade in value in the same unit as "Unit"
```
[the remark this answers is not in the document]
WeirdRubberDuck: It will be on the wiki, but should also be added to the property documentation :) Thanks for pointing it out

data/assets/scene/digitaluniverse/quasars.asset
```diff
@@ -23,21 +23,32 @@ local colormaps = asset.resource({
...
+    Coloring = {
+      FixedColor = { 1.0, 0.4, 0.2 },
+      ColorMapping = {
+        File = colormaps .. "viridis.cmap",
+        ParameterOptions = {
+          { Key = "redshift", Range = { 0.102, 7.085 } },
+          { Key = "Tlookback", Range = { 1.4, 13.0 } },
+          { Key = "distMpc", Range = { 440.5, 8852.099609 } }
+        },
+        Enabled = false
```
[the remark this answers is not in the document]
WeirdRubberDuck: Disables the color map

data/assets/scene/digitaluniverse/tully.asset
```diff
@@ -33,23 +33,28 @@ local TullyGalaxies = {
       Unit = "Mpc",
       TransformationMatrix = TransformMatrix
     },
-    Color = { 1.0, 0.4, 0.2 },
     Opacity = 0.99,
-    ScaleFactor = 504.0,
     File = speck .. "tully.speck",
     Texture = textures .. "point3A.png",
-    --ColorMap = speck .. "tully.cmap",
```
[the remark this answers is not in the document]
WeirdRubberDuck: This file actually does not exist and haven't done so in any of our 3 synced data versions for tully... So I cannot reply to that question 😅

data/assets/examples/pointclouds/points.asset
```diff
@@ -0,0 +1,236 @@
...
+    -- like for the data file above
+    Texture = openspace.absPath("${DATA}/test3.jpg"),
+    -- Disable additive blending, so that points will be rendered with their actual color
+    -- and overlapping points will be sorted by depth. This works best when the points
+    -- have an opacity of one
+    UseAdditiveBlending = false,
+    Coloring = {
+      ColorMapping = {
+        -- Disable color map per default. When enabled, the texture color will be
+        -- multiplied with the color form the color map
```
WeirdRubberDuck, line 144:
```suggestion
        -- multiplied with the color from the color map
```

data/assets/examples/pointclouds/points_colormappingsettings.asset
```diff
@@ -0,0 +1,76 @@
...
+    Type = "RenderablePointCloud",
+    File = asset.resource("data/dummydata.csv"),
+    Coloring = {
+      ColorMapping = {
+        File = asset.resource("viridis.cmap"),
+        -- Set the default choice of parameter and value range explicitly. Values
+        -- outside this range will be given special colors
+        Parameter = "normaldist_withMissing",
+        ValueRange = { -0.25, 0.25},
+        -- Show missing data values, so we can sow these in a specific colors
```
WeirdRubberDuck, line 16:
```suggestion
        -- Show missing data values, so we can show these in a specific color
```

data/assets/scene/digitaluniverse/2mass.asset
```diff
@@ -9,28 +9,34 @@ local speck = asset.resource({
...
-    ColorMap = speck .. "lss.cmap",
-    ColorOption = { "redshift", "prox5Mpc" },
-    ColorRange = { { 0.0, 0.075 }, { 1.0, 50.0 } },
-    CorrectionSizeEndDistance = 20.6,
-    CorrectionSizeFactor = 15.0,
-    ScaleFactor = 510.78,
-    BillboardMinMaxSize = { 0.0, 11.15 },
-    EnablePixelSizeControl = true
+    Coloring = {
+      FixedColor = { 1.0, 0.4, 0.2 },
```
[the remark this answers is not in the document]
WeirdRubberDuck: The Color Map overrides the fixed color if it is enabled

data/assets/scene/digitaluniverse/supernovaremnants.asset
```diff
@@ -25,17 +25,18 @@ local Object = {
...
+    Coloring = {
+      FixedColor = { 1.0, 0.5, 0.0 }
+    },
     Opacity = 0.32,
     File = speck .. "snr.speck",
-    Texture = textures .. "point4.png",
     PolygonSides = 7,
     Unit = "pc",
-    ScaleFactor = 424.0,
-    --CorrectionSizeEndDistance = 17.5,
```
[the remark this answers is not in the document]
WeirdRubberDuck: They were used to compute the exponent and factor for the scaling, but were never *actually* set anywhere (commented out or simply not set in all assets). 

image

I have simplified the scaling without these properties, but in a way that (at least to me) is more intuitive.

That said, there is some remaining work here.  I still have to make sure that it makes sense in the dome and that we can set the points size in a way that looks good in there without too much hassle.   As shown in the code snippet above, these factors were used (only) when we used the "CameraPositionNormal" option, and with no pixel size control on, and this is the use case that we have in the dome setup.

alexanderbock:
The "Color Map Enabled" should probably be hidden, right?

image

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    static const openspace::properties::PropertyOwner::PropertyOwnerInfo LabelsInfo = {
+        "Labels",
+        "Labels",
+        "The labels for the points"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo SizeMappingEnabledInfo = {
+        "Enabled",
+        "Size Mapping Enabled",
+        "If this value is set to 'true' and at least on column was loaded as an option "
```
WeirdRubberDuck, line 133:
```suggestion
        "If this value is set to 'true' and at least one column was loaded as an option "
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    //   A relatively small change to this value will lead to a large change in size.
+    //   When no exponent is set, one will be created based on the coordinates in the
+    //   dataset. The points will be visible, but may be appeared as too large or small.
+    //   One option is to not specify the exponent when loading the dataset for the the,
+    //   first time, to make sure the points are visual, and then adapt the value
+    //   interactively when OpenSpace is running until you find a value that you find
+    //   suitable.
+    //
+    // - There is also an option to limit the size of the points based on a given pixel
+    //   size. For now, this only works on flat proejction displays.
```
WeirdRubberDuck, line 241:
```suggestion
    //   size. For now, this only works for flat projection displays.
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+        // The unit used for all distances. Should match the unit of any
+        // distances/positions in the data files
+        std::optional<Unit> unit;
+
+        // [[codegen::verbatim(LabelsInfo.description)]]
+        std::optional<ghoul::Dictionary> labels
+            [[codegen::reference("labelscomponent")]];
+
+        struct SizeSettings {
+            // A list specifying all parameters that may be use for size mapping, i.e.
```
WeirdRubberDuck, line 303:
```suggestion
            // A list specifying all parameters that may be used for size mapping, i.e.
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+            std::optional<bool> enabled;
+
+            // [[codegen::verbatim(FadeInDistancesInfo.description)]]
+            std::optional<glm::dvec2> fadeInDistances;
+
+            // [[codegen::verbatim(InvertFadeInfo.description)]]
+            std::optional<bool> invert;
+        };
+        // Settings related to fading based on camera distance. Can be used to either
+        // fade away or fade int the points when reaching a certain distance from the
```
WeirdRubberDuck, line 346:
```suggestion
        // fade away or fade in the points when reaching a certain distance from the
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    else {
+        // Normal blending, with transparency
+        glDepthMask(true);
+        glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+    }
+
+    _program->activate();
+
+    _program->setUniform(
+        "screenSize",
```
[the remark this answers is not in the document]
WeirdRubberDuck: that is a very good question. Probably only a remnant of how the code was written before (I have added all the things 
 I have added to the uniform cache and removed the things I removed, but this uniform already existed so I probably just didn't think about it) :)

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include <optional>
+
+namespace {
+    constexpr std::string_view _loggerCat = "RenderablePolygonCloud";
+
+    // A RenderablePolygonCloud is a RenderablePointCloud where the shape of the points
+    // is a uniform polygon with a given number of sides instead of a texture. For
+    // instance, PolygonSides = 5 results in the points being rendered as pentagons.
+    // Note that while this renderable inherits the texture property from
+    // RenderablePointCloudany, any added texture value will be ignored in favor of the
```
WeirdRubberDuck, line 42:
```suggestion
    // RenderablePointCloud, any added texture value will be ignored in favor of the
```

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,190 @@
...
+    // polygon shape.
+    //
+    // Se documentation of RenderablePointCloud for details on the other parts of the
```
WeirdRubberDuck, line 45:
```suggestion
    // See documentation of RenderablePointCloud for details on the other parts of the
```

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,190 @@
...
+    _polygonVao = 0;
+
+    glDeleteTextures(1, &_pTexture);
+}
+
+void RenderablePolygonCloud::bindTextureForRendering() const {
+    glBindTexture(GL_TEXTURE_2D, _pTexture);
+}
+
+void RenderablePolygonCloud::createPolygonTexture() {
```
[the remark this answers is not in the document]
WeirdRubberDuck: This depends on what kind of resources we prefer to prioritize, I guess. This textured-based approach is how it was implemented before the rewrite and it seems to make sense to me

The way it works now, the texture is created once (in the initializeGl) and then just used for rendering. If we do it in the geometry shader, that would add extra logic in the shader itself and we would also have to generate more vertices every time the point is rendered (currently, we only need 4, for the texture plane, and this is the same for all the use cases)

I don't know which approach is preferable, but  I like that it does not lead to another path or any special check in the geomoetry shader code that would be special for polygons. But I'm open to suggestions if we think that approach is better

What are your thoughts, @alexanderbock and @Roxeena ?

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,190 @@
...
+    };
+
+    glBufferData(GL_ARRAY_BUFFER, sizeof(VertexData), VertexData.data(), GL_STATIC_DRAW);
+    glVertexAttribPointer(0, 4, GL_FLOAT, GL_FALSE, 4 * sizeof(GLfloat), nullptr);
+    glEnableVertexAttribArray(0);
+    glBindVertexArray(0);
+}
+
+void RenderablePolygonCloud::renderPolygonGeometry(GLuint vao) {
+    std::unique_ptr<ghoul::opengl::ProgramObject> program =
```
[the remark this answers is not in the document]
WeirdRubberDuck: This function is only triggered once, which is from the `initializeGL` function and then the shader is never reused. So it doesn't really matter

modules/base/shaders/billboardpoint_fs.glsl
```diff
@@ -24,45 +24,86 @@
...
+  if (useColorMap) {
+    fullColor *= sampleColorMap(gs_colorParameter);
   }
   else {
     fullColor.rgb *= color;
   }
 
-  float textureOpacity = dot(fullColor.rgb, vec3(1.0));
-  if (textureOpacity == 0.0) {
+  fullColor.a *= alphaValue * fadeInValue;
```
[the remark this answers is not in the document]
WeirdRubberDuck: For the user-facing name, that is true.  In the shaders, it's still a bit mixed though

here I just haven't renamed it, but I can if we want to

src/data/csvloader.cpp
```diff
@@ -0,0 +1,178 @@
...
+                // Check if column should be exluded. Note that list of indices is sorted
+                // so we can do a binary search
+                if (std::binary_search(skipColumns.begin(), skipColumns.end(), i)) {
+                    continue;
+                }
+            }
+
+            const std::string& strValue = row[i];
+
+            // For now, all values are converted to float
```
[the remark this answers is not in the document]
WeirdRubberDuck: Yep.  At least for now. The same goes for SPECK files and have always been the case

The color-mapping would not support categorical data either way, and implementing support for point clouds with categorical data is a bigger project.

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+    if (!file.good()) {
+        return std::nullopt;
+    }
+
+    Dataset result;
+
+    int8_t fileVersion;
+    file.read(reinterpret_cast<char*>(&fileVersion), sizeof(int8_t));
+    if (fileVersion != DataCacheFileVersion) {
+        // Incompatible version and we won't be able to read the file
```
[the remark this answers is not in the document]
WeirdRubberDuck: Nah, I don't think so.  From his/her side it doesnt matter if the cache version of the file works or not, and it will result in a new file being cached. 

In general, the user should not have to know/think/care about the things in the cache folder (unless absolutely necessary, I guess)

src/data/datamapping.cpp
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        return testColumn == column;
+    }
+
+    // This is a data mapping structure that can be used when creating point cloud
+    // datasets, e.g. from a CSV or Speck file.
+    //
+    // It allows specifying things like column names, whether the reading of those
+    // column names hsould be case sensitive, data value that represents missing
```
WeirdRubberDuck, line 75:
```suggestion
    // column names should be case sensitive, data value that represents missing
```

src/rendering/colormappingcomponent.cpp
```diff
@@ -0,0 +1,470 @@
...
+        "Set the data range for the color mapping based on the available data for the "
+        "curently selected data column",
+        openspace::properties::Property::Visibility::User
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo HideOutsideInfo = {
+        "HideValuesOutsideRange",
+        "Hide Values Outside Range",
+        "If true, points with values outside the provided range for the coloring will be "
+        "rendered as transparent, i.e. not shown",
```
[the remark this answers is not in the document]
WeirdRubberDuck: SImplifying it in some way is probably good 😅  Although, personally I prefer when the description includes more information than what is in the name of the property, so would maybe still want some clarification as to what "hidden" means.  

Not a super strong opinion though. Could be ok with just "the points will be hidden" if you think that's clear enough

data/assets/scene/digitaluniverse/tully.asset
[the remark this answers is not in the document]
alexanderbock: ¯\\\_(ツ)\_/¯

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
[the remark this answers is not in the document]
alexanderbock: IIrc, we experimented with Jonathas back when and found that the texture-based approach was running faster than the geometry shader, since we are doing it for a lot of points and if the vertex shader places a point outside the view frustum it might get culled away early, but if a geometry shader is running it can only be culled afterwards since the shader might move it _back_ into the frustum.  So a lot of extra work that impacted performance

modules/base/shaders/billboardpoint_fs.glsl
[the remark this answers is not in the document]
alexanderbock: :+1: I think having the same naming in the shader would be useful. It's a remnant from when we used opacity and transparency sometimes interchangably

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+        res = csv::loadCsvFile(path.string(), specs);
+    }
+    else if (extension == ".speck") {
+        res = speck::loadSpeckFile(path.string(), specs);
+    }
+    else {
+        LERRORC("DataLoader", fmt::format(
+            "Could not read data file {}. File format {} is not supported",
+            path, path.extension()
+        ));
```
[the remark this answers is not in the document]
alexanderbock: Yea, we also use it for if-like switch cases. Like:

```cpp
if (string == "abc") {
  return 1;
}
else if (string == "def") {
  return 2;
}

// shouldn't get here
```

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: Hm, ok.  Although, there is a geometry shader running anyways, to generate the texture plane, so I'm not sure I understand how it would be any different 🤔 

Either way, I don't really think we need to update the implementation

WeirdRubberDuck:
> The "Color Map Enabled" should probably be hidden, right?
> 
> image
> 
> I guess same for "Labels"?
> 
> image

This is a higher discussion to have.  At some point, we made it so that the UI does not hide the Enabled properties. I think for transparency, now that the other checkbox also includes the fading and not the enabling, so the user don't have access to the Enabled status otherwise 🤔 

So if we hide them for these, we should do it for ALL the enabled properties and it should be handled in the UI code

src/data/dataloader.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: 👍

modules/base/shaders/billboardpoint_gs.glsl
```diff
@@ -71,30 +66,15 @@ const vec2 corners[4] = vec2[4](
...
-    // Conversion factor from Parsecs to GigalightYears
-    unit = 306391534.73091 * PARSEC;
-  }
+  gs_colorParameter = colorParameter[0];
 
-  dvec4 dpos = modelMatrix * dvec4(dvec3(pos.xyz) * unit, 1.0);
+  dvec4 dpos = modelMatrix * dvec4(dvec3(pos.xyz), 1.0);
 
-  float scaleMultiply = exp(scaleFactor * 0.10);
+  float scaleMultiply = exp(scaleExponent);
```
WeirdRubberDuck, line 75: Thought when writing the docs for this:   Is it worthwhile to change this computation to be `10^exponent` rather than `e^exponent`?  Would be a bit more intuitive to explain, I think. 

Thoughts @alexanderbock @Roxeena ? :)

modules/base/shaders/billboardpoint_gs.glsl
```diff
@@ -109,18 +89,13 @@ void main() {
...
-      scaleMultiply *= 1.0 / (1.0 + factorVar * exp(expVar));
-    }
-
     scaledRight = scaleMultiply * newRight * 0.5;
     scaledUp = scaleMultiply * newUp * 0.5;
   }
 
-  if (enabledRectSizeControl) {
+  // @TODO: Come up with some better solution for this scaling, that
+  // also work wiht non planar projections and multiple viewport resolutions.
```
WeirdRubberDuck, line 97:
```suggestion
  // also work with non planar projections and multiple viewport resolutions.
```

data/assets/scene/digitaluniverse/globularclusters.asset
```diff
@@ -16,7 +16,7 @@ local speck = asset.resource({
 local Object = {
   Identifier = "GlobularClusters",
   Renderable = {
-    Type = "RenderableBillboardsCloud",
+    Type = "RenderablePolygonCloud",
```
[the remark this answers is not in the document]
Roxeena: I think it could be cool to have at least one simple example to show our users that this is possible to do if they want to.

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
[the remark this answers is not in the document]
Roxeena: I don't have any strong feelings, it just looked like we had two different solutions to the same problem but they are doing different things I suppose.

src/rendering/colormappingcomponent.cpp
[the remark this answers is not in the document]
Roxeena: I think "the points will be hidden" is enough of an explanation, but if you want to do some further explaining that would be fine with me too. I just felt like starting to talk about transparency was a little confusing when the property was simpler than that.

modules/base/shaders/billboardpoint_gs.glsl
Roxeena: Mathematically they are different so which one is being used? But the 10^exponent is more intuitive I think.

data/assets/scene/digitaluniverse/2mass.asset
```diff
@@ -9,28 +9,34 @@ local speck = asset.resource({
...
+        File = speck .. "lss.cmap",
+        ParameterOptions = {
+          { Key = "redshift", Range = { 0.0, 0.075 } },
+          { Key = "prox5Mpc", Range = { 1.0, 50.0 } }
+        }
+      }
+    },
+    SizeSettings = {
+      ScaleExponent = 51.78,
+      BillboardMaxPixelSize = 11.15,
```
WeirdRubberDuck, line 37: How about renaming to just `MaxPixelSize`?  @alexanderbock @Roxeena

data/assets/scene/digitaluniverse/localdwarfs.asset
```diff
@@ -25,17 +25,26 @@ local Object = {
       MinMaxSize = { 7, 20 },
       Unit = "Mpc"
     },
-    Color = { 0.5, 1.0, 0.2 },
-    ColorMap = speck .. "localgroup.cmap",
-    ColorOption = { "association" },
+    Coloring = {
+      FixedColor = { 0.0, 1.0, 0.0 },
+      -- @TODO: This one wasn't actually properly used before the point cloud update.
```
WeirdRubberDuck, line 30: Decide if this should be kept around

data/assets/scene/digitaluniverse/tully.asset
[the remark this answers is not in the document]
WeirdRubberDuck: Alex got the file from Jonathas and there is not that much difference compared to lss.cmap.   I suggest we just don't include it

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+    if (_hasDataFile && _hasColorMapFile) {
+        _colorSettings.colorMapping->initialize(_dataset);
+    }
+
+    if (_hasLabels) {
+        _labels->initialize();
+        _labels->loadLabels();
+    }
+
+    setRenderBin(Renderable::RenderBin::PreDeferredTransparent);
```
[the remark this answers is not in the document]
WeirdRubberDuck: 👍 I think this line can be removed. The explicit renderbin is not really needed.  

Good catch

modules/base/rendering/pointcloud/renderablepointcloud.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: Actually, looks like this uniform is already in the cache and is being set further down in the code as well. oopsies

src/data/dataloader.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: Although, do we really want to add try-catch blocks everywhere we use any of the `loadCachedFile` or `loadFile` functions...?   That would be a lot of code and added complexity for something that is already handled in the code, and would lead to a fatal crash as soon as someone forgets to check it

src/rendering/colormappingcomponent.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: Agreed 👍

data/assets/scene/digitaluniverse/globularclusters.asset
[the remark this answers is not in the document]
WeirdRubberDuck: Sure thing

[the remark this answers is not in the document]
WeirdRubberDuck: Good that you mentioned it. At some point, the polygon version seemed to have stopped working 😄

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1039 @@
...
+
+    GLint viewport[4];
+    glGetIntegerv(GL_VIEWPORT, viewport);
+    _program->setUniform(_uniformCache.screenSize, glm::vec2(viewport[2], viewport[3]));
+
+    bool useTexture = _hasSpriteTexture && _useSpriteTexture;
+    _program->setUniform(_uniformCache.hasSpriteTexture, useTexture);
+    ghoul::opengl::TextureUnit spriteTextureUnit;
+    spriteTextureUnit.activate();
+    bindTextureForRendering();
```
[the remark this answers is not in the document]
WeirdRubberDuck: Answering both your questions here, as they are related. 

Yep, the binding can definitely be moved to an if-statement 

However, I have not found a way to get the rendering to work without creating the TextureUnit and setting that uniform.  Any of my approaches leads to either crashes or GL_INVALID_OPERATION errors... Unfortunately, my understanding of the texture handling is not sufficient to know how to best handle this (I think I just left the code in a state that worked, heh..)

Initiating a discussion IRL with you about this ;)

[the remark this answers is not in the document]
WeirdRubberDuck: Yes, both the binding and activation can definitely be moved

modules/base/shaders/billboardpoint_gs.glsl
WeirdRubberDuck: Currently e^exponent is used (that's what the `exp` function does)

Changng it would involve recomputing all the exponents in the asset files, but I do think it would make a lot more sense

alexanderbock: 10^exponent makes more sense to me, yea

src/data/dataloader.cpp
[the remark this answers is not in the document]
alexanderbock: For the `loadCachedFile`, _that_ funciton should have a try/catch inside and if there is an exception being thrown it should signal back that the cached file wasn't avaiable (i think it's a boolean return value?).  For the `loadFile` I think an exception would be great since it gets caught higher up the chain and logged.

And for the example above, if there is a user-facing way to make `string` not `abc` or `def`, then it should _not_ be a MissingCaseException.  MissingCaseException derives from `std::logic_error` which means that something in the _logic_ of the program went wrong, not something that can be a `std::runtime_error`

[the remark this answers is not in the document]
alexanderbock: Conversely, there should never be a `catch` for a `MissingCaseException` since we want those to fail catastrophically to fix the underlying logic error

data/assets/scene/digitaluniverse/localdwarfs.asset
alexanderbock: What do you mean?  Keeping the ability to have a fixed color?

alexanderbock:
> At some point, we made it so that the UI does not hide the Enabled properties

In that case I misremembered and ignore my comment :)

src/data/dataloader.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: So, the conclusion here is that since the comment here is on the check for the file type, which is the user-facing input, we should not throw a `MissingCaseException`.   Correct?

Also, there are actually cases where we allow calling the `loadFile` function directly (when we don't want to use the caching). So we would still have to catch in manually in some cases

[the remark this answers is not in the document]
WeirdRubberDuck: But rather do a throw when the loading of a cached file fails :) (currently it returns an `std::nullopt` in those scenarios)

data/assets/scene/digitaluniverse/localdwarfs.asset
WeirdRubberDuck: The color map was not correctly read.  

Although, looking at this in more details there are some problems with the speck file for this dataset that I think we need to discuss...

modules/base/shaders/billboardpoint_gs.glsl
WeirdRubberDuck: Fixed

data/assets/scene/digitaluniverse/milkyway_label.asset
```diff
@@ -27,14 +27,21 @@ local HomeLabel = {
       Unit = "Mpc",
       TransformationMatrix = TransformMatrix
     },
-    Color = { 1.0, 0.4, 0.2 },
+    Coloring = {
+      FixedColor = { 1.0, 0.4, 0.2 }
+    },
+    -- Remove? There is no point...
```
WeirdRubberDuck, line 33: TODO: Consider removing it

WeirdRubberDuck: Or at least make sure to remove this comment

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1042 @@
...
+        _labels = std::make_unique<LabelsComponent>(*p.labels);
+        _hasLabels = true;
+        addPropertySubOwner(_labels.get());
+        // Fading of the labels should also depend on the fading of the renderable
+        _labels->setParentFadeable(this);
+    }
+
+    _transformationMatrix = p.transformationMatrix.value_or(_transformationMatrix);
+
+    if (p.sizeSettings.has_value() && (*p.sizeSettings).sizeMapping.has_value()) {
```
alexanderbock, line 537:
```suggestion
    if (p.sizeSettings.has_value() && p.sizeSettings->sizeMapping.has_value()) {
```

?

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -0,0 +1,1042 @@
...
+        funcValue = glm::clamp(funcValue, 0.f, 1.f);
+
+        if (_fading.invert) {
+            funcValue = 1.f - funcValue;
+        }
+
+        fadeValue *= funcValue;
+    }
+
+    return fadeValue;
```
alexanderbock, line 685:
```suggestion
    if (!_fading.enabled) {
        return 1.f;
    }

    float fadeValue = 1.f;
    glm::dmat4 invModelMatrix = glm::inverse(calcModelTransform(data));

    glm::dvec3 cameraPosModelSpace = glm::dvec3(
        invModelMatrix * glm::dvec4(data.camera.positionVec3(), 1.0)
    );

    float distCamera = static_cast<float>(
        glm::length(cameraPosModelSpace) / toMeter(_unit)
    );

    const glm::vec2 fadeRange = _fading.fadeInDistances;
    const float fadeRangeWidth = (fadeRange.y - fadeRange.x);
    float funcValue = (distCamera - fadeRange.x) / fadeRangeWidth;
    funcValue = glm::clamp(funcValue, 0.f, 1.f);

    if (_fading.invert) {
        funcValue = 1.f - funcValue;
    }

    return fadeValue * funcValue;;
```

alexanderbock (commented):
TBC.  Hopping on a plane

data/assets/examples/pointclouds/viridis.cmap
```diff
@@ -0,0 +1,258 @@
+# Viridis Color map
```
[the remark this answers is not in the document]
WeirdRubberDuck: I like that idea.  Maybe both? 

The file we refer to in these examples could be one with fewer colors, to make it easier to understand

data/assets/scene/digitaluniverse/abell.asset
WeirdRubberDuck: Looked at the file itself now and it's one of the complicated ones that I quite don't understand. So I'll remove it for now either way.  It's better to add a new color map if they want it, down the line

(It's prepared by Brent Tully (U Hawaii), Stuart Levy (NCSA/UIUC), who I have know idea who they are. Do you?)

include/openspace/data/datamapping.h
```diff
@@ -0,0 +1,74 @@
...
+namespace ghoul { class Dictionary; }
+
+namespace openspace::dataloader {
+
+struct DataMapping {
+    static DataMapping createFromDictionary(const ghoul::Dictionary& dictionary);
+    static documentation::Documentation Documentation();
+
+    bool hasExcludeColumns() const;
+    bool isExcludeColumn(std::string_view x) const;
```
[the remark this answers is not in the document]
WeirdRubberDuck: 😄 or column

modules/base/rendering/pointcloud/renderablepointcloud.cpp
WeirdRubberDuck: 👍 Forgot that this was possible

modules/base/shaders/billboardpoint_gs.glsl
WeirdRubberDuck: My current thought is to handle that separately from this PR tough

modules/base/rendering/pointcloud/renderablepointcloud.h
```diff
@@ -68,88 +73,100 @@ class RenderableBillboardsCloud : public Renderable {
...
+    } _sizeSettings;
+
+    struct ColorSettings : properties::PropertyOwner {
+        ColorSettings(const ghoul::Dictionary& dictionary);
+        properties::Vec3Property pointColor;
+        std::unique_ptr<ColorMappingComponent> colorMapping;
+    } _colorSettings;
+
+    struct Fading : properties::PropertyOwner {
+        Fading(const ghoul::Dictionary& dictionary);
```
alexanderbock, line 127:
```suggestion
        explicit Fading(const ghoul::Dictionary& dictionary);
```

modules/base/rendering/pointcloud/renderablepointcloud.h
```diff
@@ -68,88 +73,100 @@ class RenderableBillboardsCloud : public Renderable {
...
 
-    properties::FloatProperty _scaleFactor;
-    properties::BoolProperty _useColorMap;
-    properties::Vec3Property _pointColor;
+        properties::BoolProperty pixelSizeControl;
+        properties::FloatProperty maxPixelSize;
+    } _sizeSettings;
+
+    struct ColorSettings : properties::PropertyOwner {
+        ColorSettings(const ghoul::Dictionary& dictionary);
```
alexanderbock, line 121:
```suggestion
        explicit ColorSettings(const ghoul::Dictionary& dictionary);
```

modules/base/rendering/pointcloud/renderablepointcloud.h
```diff
@@ -68,88 +73,100 @@ class RenderableBillboardsCloud : public Renderable {
...
+    bool _hasDataFile = false;
     bool _hasColorMapFile = false;
-    bool _isColorMapExact = false;
     bool _hasDatavarSize = false;
-    bool _hasPolygon = false;
     bool _hasLabels = false;
 
-    int _polygonSides = 0;
+    struct SizeSettings : properties::PropertyOwner {
+        SizeSettings(const ghoul::Dictionary& dictionary);
```
alexanderbock, line 105:
```suggestion
        explicit SizeSettings(const ghoul::Dictionary& dictionary);
```

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void RenderablePolygonCloud::deinitializeGL() {
+    glDeleteBuffers(1, &_polygonVbo);
+    _polygonVbo = 0;
+    glDeleteVertexArrays(1, &_polygonVao);
+    _polygonVao = 0;
+
+    glDeleteTextures(1, &_pTexture);
+}
```
alexanderbock, line 91:
```suggestion

    RenderablePointCloud::deinitializeGL();
}
```

for symmetry with initialize

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
```diff
@@ -0,0 +1,194 @@
...
+    constexpr std::array<GLfloat, 4> VertexData = {
+        // x  y  z  w
+        0.f, 0.f, 0.f, 1.f,
+    };
+
+    glBufferData(GL_ARRAY_BUFFER, sizeof(VertexData), VertexData.data(), GL_STATIC_DRAW);
+    glVertexAttribPointer(0, 4, GL_FLOAT, GL_FALSE, 4 * sizeof(GLfloat), nullptr);
+    glEnableVertexAttribArray(0);
+    glBindVertexArray(0);
+}
```
alexanderbock, line 169: Would it make sense to just move the contents of the function into the previous function instead?  Since we are only calling it once anyway

modules/base/rendering/pointcloud/renderablepolygoncloud.h
```diff
@@ -0,0 +1,71 @@
...
+private:
+    void createPolygonTexture();
+    void renderToTexture(GLuint textureToRenderTo, GLuint textureWidth,
+        GLuint textureHeight);
+    void loadPolygonGeometryForRendering();
+    void renderPolygonGeometry(GLuint vao);
+
+    void bindTextureForRendering() const override;
+
+    int _polygonSides = 3;
```
alexanderbock, line 61:
```suggestion
    int _nPolygonSides = 3;
```

?

src/data/csvloader.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <algorithm>
+#include <cmath>
+#include <cctype>
+#include <fstream>
+#include <functional>
+#include <string_view>
+
+namespace {
+    constexpr std::string_view _loggerCat = "DataLoader: CSV";
+}
```
alexanderbock, line 45:
```suggestion
} // namespace
```

src/data/csvloader.cpp
```diff
@@ -0,0 +1,196 @@
...
+                if (i == yColumn) {
+                    entry.position.y = value;
+                }
+                if (i == zColumn) {
+                    entry.position.z = value;
+                }
+            }
+            else {
+                entry.data.push_back(value);
+            }
```
alexanderbock, line 177:
```suggestion
            if (i == xColumn) {
                entry.position.x = value;
            }
            else if (i == yColumn) {
                entry.position.y = value;
            }
            else if (i == zColumn) {
                entry.position.z = value;
            }
            else {
                entry.data.push_back(value);
            }
```

?

src/data/dataloader.cpp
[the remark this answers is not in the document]
alexanderbock: I think returning an `std::nullopt` is better since the caching-failing is not necessarily an exceptional behavior (for example if the version has changed, it's the expected behavior).

So I'd say neither to both exceptions

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+    ghoul_assert(std::filesystem::exists(path), "File must exist");
+
+    std::ifstream file(path);
+    if (!file.good()) {
+        throw ghoul::RuntimeError(fmt::format("Failed to open dataset file {}", path));
+    }
+
+    std::string extension = path.extension().string();
+    std::transform(extension.begin(), extension.end(), extension.begin(),
+        [](unsigned char c) { return std::tolower(c); });
```
alexanderbock, line 358: Can be replaced with your new function

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+    ghoul_assert(std::filesystem::exists(path), "File must exist");
+
+    std::ifstream file(path);
+    if (!file.good()) {
+        throw ghoul::RuntimeError(fmt::format("Failed to open colormap file {}", path));
+    }
+
+    std::string extension = path.extension().string();
+    std::transform(extension.begin(), extension.end(), extension.begin(),
+        [](unsigned char c) { return std::tolower(c); });
```
alexanderbock, line 479: Can be replaced with your new function

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+
+    ColorMap res;
+    if (extension == ".cmap") {
+        res = speck::loadCmapFile(path.string());
```
alexanderbock, line 483: The `loadCmapFile` should probably take a `std::filesystem::path` as an argument, in which case this could just be `path`

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+
+    if (variableIndex >= entries[0].data.size()) {
+        // The index is not a valid variable index
+        return glm::vec2(0.f);
+    }
+
+    float minValue = std::numeric_limits<float>::max();
+    float maxValue = -std::numeric_limits<float>::max();
+    for (const Dataset::Entry& e : entries) {
+        if (e.data.size() > 0) {
```
alexanderbock, line 665:
```suggestion
        if (!e.data.empty()) {
```

src/data/dataloader.cpp
```diff
@@ -0,0 +1,693 @@
...
+            minValue = std::min(value, minValue);
+            maxValue = std::max(value, maxValue);
+        }
+        else {
+            minValue = 0.f;
+            maxValue = 0.f;
+        }
+    }
+
+    return { minValue, maxValue };
```
alexanderbock, line 679:
```suggestion
    return glm::vec2(minValue, maxValue);
```

src/data/speckloader.cpp
```diff
@@ -357,6 +300,15 @@ Dataset loadFile(std::filesystem::path path, SkipAllZeroLines skipAllZeroLines)
                 valueStream.str(value);
                 valueStream >> entry.data[i];
 
+                // Check if value corresponds to a missing value
+                if (specs.has_value() && (*specs).missingDataValue.has_value()) {
```
alexanderbock, line 304:
```suggestion
                if (specs.has_value() && specs->missingDataValue.has_value()) {
```

src/data/speckloader.cpp
```diff
@@ -357,6 +300,15 @@ Dataset loadFile(std::filesystem::path path, SkipAllZeroLines skipAllZeroLines)
...
+                    float missingDataValue = (*specs).missingDataValue.value();
```
alexanderbock, line 305:
```suggestion
                    float missingDataValue = specs->missingDataValue.value();
```

src/rendering/colormappingcomponent.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+    struct [[codegen::Dictionary(ColorMappingComponent)]] Parameters {
+        // [[codegen::verbatim(EnabledInfo.description)]]
+        std::optional<bool> enabled;
+
+        // [[codegen::verbatim(FileInfo.description)]]
+        std::optional<std::string> file;
+
+        struct ColorMapParameter {
+            // The key for the datavar to use for color
```
alexanderbock, line 154:
```suggestion
            // The key for the data variable to use for color
```

?

src/rendering/colormappingcomponent.cpp
```diff
@@ -0,0 +1,470 @@
...
+            if (_providedParameter.has_value() && *_providedParameter == v.name) {
+                indexOfProvidedOption = i;
+            }
+
+            i++;
+        }
+    }
+    else {
+        // Otherwise, check if the selected columns exist
+        for (int i = 0; i < dataColumn.options().size(); ++i) {
```
alexanderbock, line 425:
```suggestion
        for (size_t i = 0; i < dataColumn.options().size(); ++i) {
```

modules/base/rendering/pointcloud/renderablepolygoncloud.cpp
WeirdRubberDuck: Yep, sure thing

src/data/csvloader.cpp
WeirdRubberDuck: I had it like this and then rewrote it for some reason.. I wonder why 😅 (December brain, maybe...)

But yes, it should be the same

src/data/dataloader.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: Good. I agree.  was a bit confused by this conversation 😅 

I'll consider this resolved by leaving it as is then.

src/data/dataloader.cpp
WeirdRubberDuck: Interestingly, it already does.  But good catch, I'll remove all the extra `.string()`s here

alexanderbock (approved):
Looks good to me!

data/assets/scene/digitaluniverse/localdwarfs.asset
WeirdRubberDuck: https://github.com/OpenSpace/OpenSpace/issues/2983

Issue #1462, opened 2021-01-22
Possiblity to control rendering properties of all exoplanet systems simultaneously


WeirdRubberDuck:
It would be useful to for example change disc texture, opacity or hide all the uncertainty discs or habitable zones of all exoplanet systems in one command. 

A potential solution would be to expose lua functions that does this and add a section for the exoplanet GUI panel with such controls.

alexanderbock:
Another option is to automatically add a tag to all exoplanet-created systems that can be used in such a UI.  Example: https://github.com/OpenSpace/OpenSpace/blob/master/data/assets/scene/solarsystem/planets/mars/mars.asset#L70
and
https://github.com/OpenSpace/OpenSpace/blob/master/data/profiles/default.profile#L43

WeirdRubberDuck:
Additionally, it would be useful to be able to change the default values used for the exoplanet system rendering (like textures, opacities, etc), before the scene graph nodes are created. Perhaps as properties of the ExoplanetsModule, so they are accessible to the user?

WeirdRubberDuck:
Closed in https://github.com/OpenSpace/OpenSpace/pull/2943

Issue #1440, opened 2020-12-18
Add checkbox in exoplanets UI to hide/show all habitable zones


WeirdRubberDuck:
Add a checkbox in the exoplanets UI that controls the visibility of the habitable zone of all added (and future) exoplanets systems. 

Any new added systems should get the same setting as the already existing ones have. So, if I have chosen to hide all habitable zones and add a new system, the habitable zone for that system should not be visible.

PR #2943, opened 2023-11-10
Issue/1440 - Exoplanet Panel and settings update

WeirdRubberDuck:
Add some settings checkboxes to the exoplanet panel, that affect both systems added in the future as well as current systems

image
(These are added in WebGUI PR: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/163, but it's enough to just check out the branch here in OpenSpace to try it out )

also adds some new properties to the exoplanet module, for easier configuration, and changes the default color of the 1 AU size comparison circles

Closes issue #1440 
Closes issue #1462


alexanderbock (approved):

WeirdRubberDuck:
The offset for the checkbox changes with scaling and looks different on the chrome version of the webpage... It's a problem related to updating the checkbox margins (currently set as just a few pixels, added here and there over the entire interface, it seems...) and I'll leave it like this for now.

PR #2947, opened 2023-11-15
More explicit handling of zero duration paths

WeirdRubberDuck:
Make it more obvious when the camera path speed is infinite (duration is zero), and prevent doing redundant computations.  Instead, just jump to the end of the path in traversal

Honestly, it's just lucky that the math worked before... The use of infinite speed was a bit nasty

### DIFF
--- a/src/navigation/path.cpp
+++ b/src/navigation/path.cpp
@@ -120,7 +120,7 @@ Path::Path(Waypoint start, Waypoint end, Type type, std::optional<double> durati
     // computing how much faster/slower it should be
     _speedFactorFromDuration = 1.0;
     if (duration.has_value()) {
-        if (*duration > 0) {
+        if (*duration > 0.0) {
             constexpr double dt = 0.05; // 20 fps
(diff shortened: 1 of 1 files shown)


src/navigation/path.cpp
```diff
@@ -120,14 +120,20 @@ Path::Path(Waypoint start, Waypoint end, Type type, std::optional<double> durati
     // computing how much faster/slower it should be
     _speedFactorFromDuration = 1.0;
     if (duration.has_value()) {
-        constexpr double dt = 0.05; // 20 fps
-        while (!hasReachedEnd()) {
-            traversePath(dt);
-        }
+        if (*duration > 0) {
```
alexanderbock, line 123:
```suggestion
        if (*duration > 0.0) {
```

alexanderbock (approved):

WeirdRubberDuck:
May seem like I very quickly merged this after one approval, but I just really wanted someone to look at the inf handling and see if it was acceptable :) thanks @alexanderbock

PR #2956, opened 2023-11-21
Utilize new functions for converting strings to upper/lower case

WeirdRubberDuck:
Also, changes in ghoul: 
https://github.com/OpenSpace/Ghoul/pull/103

### DIFF
--- a/src/rendering/luaconsole.cpp
+++ b/src/rendering/luaconsole.cpp
@@ -35,8 +35,8 @@
 #include <ghoul/font/fontmanager.h>
 #include <ghoul/font/fontrenderer.h>
 #include <ghoul/misc/clipboard.h>
-#include <ghoul/misc/misc.h>
 #include <ghoul/misc/profiling.h>
+#include <ghoul/misc/stringhelper.h>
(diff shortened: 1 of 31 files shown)


src/rendering/luaconsole.cpp
```diff
@@ -507,20 +508,11 @@ bool LuaConsole::keyboardCallback(Key key, KeyModifier modifier, KeyAction actio
             const size_t fullLength = _autoCompleteInfo.initialValue.length();
             const bool correctLength = command.length() >= fullLength;
 
-            std::string commandLowerCase;
```
WeirdRubberDuck, line 510: @alexanderbock does this change introduce an undesired performance penalty, do you think?

It seems to be the only place where we did not first copy the string explicitly

alexanderbock: My gut feeling is that it is probably faster since we are modifying the string in place rather then back_inserting it.

Either way it's the Lua console, so performance is not really _that_ important here anyway

alexanderbock (approved):
Looks good to me!

Issue #2966, opened 2023-11-29
Read-only properties have inconsistent rendering and bugs


WeirdRubberDuck:
In addition, it's not visually obvious when a property is read-only/disabled in many cases.  

We should go through and clean up the UI code for all the properties so that the rendering of disabled properties is more consistent, problem free and hover behaviors are limited, etc.  

(The main reason that this is not done yet is because the property part of the UI code could need a general rewrite... Currently, every individual property type is responsible for its own rendering and there is a lot of duplicated code, making it difficult and time consuming to implement consistent behaviors between different types of properties)

alexanderbock:
Closed in e3f5b83f7476c8507dccd45ab98e4731504af532

PR #2967, opened 2023-11-29
Add property for actual bounding and interaction sphere value

### LINKED ISSUE #1828: Show the actual value for the bounding sphere in the UI
When the bounding sphere (and interaction sphere) property of a SGN is set to -1, it is computed internally by each renderable. For debugging, it would be useful to see the actual value of the bounding sphere in the UI. 

I recently added an API function to ask OpenSpace for this information (`openspace.boundingSphere(sgnIdentifier)`) which could potentially be used. Alternatively, it would be possible to add a read-only property whose sole purpose is to access the actual bounding sphere of the node, but not change it

WeirdRubberDuck:
Closes #1828 

Have been wanting this for a while, and now when we talked about adding other read-only properties to display information, I thought it could be a good (and simple) approach for this issue

image

### DIFF
--- a/src/scene/scenegraphnode.cpp
+++ b/src/scene/scenegraphnode.cpp
@@ -115,7 +115,7 @@ namespace {
         "This read-only property contains the evaluated value for the bounding sphere. "
         "This is the actual value that is used internally within the software. If the "
         "BoundingSphere property is set to -1, it is the computed value, otherwise it "
-        "should match the one set in the property",
+        "matches the one set in the property",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
@@ -125,7 +125,7 @@ namespace {
         "This read-only property contains the evaluated value for the interaction sphere. "
         "This is the actual value that is used internally within the software. If the "
         "InteractionSphere property is set to -1, it is the computed value, otherwise it "
-        "should match the one set in the property",
+        "matches the one set in the property",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
--- a/src/scene/scenegraphnode.cpp
+++ b/src/scene/scenegraphnode.cpp
@@ -109,6 +109,26 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo EvalBoundingSphereInfo = {
+        "EvaluatedBoundingSphere",
+        "Evaluated Bounding Sphere (Read Only)",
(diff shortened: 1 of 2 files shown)


src/scene/scenegraphnode.cpp
```diff
@@ -109,6 +109,26 @@ namespace {
...
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo EvalInteractionSphereInfo = {
+        "EvaluatedInteractionSphere",
+        "Evaluated Interaction Sphere (Read Only)",
+        "This read-only property contains the evaluated value for the interaction sphere. "
+        "This is the actual value that is used internally within the software. If the "
+        "InteractionSphere property is set to -1, it is the computed value, otherwise it "
+        "should match the one set in the property",
```
alexanderbock, line 128:
```suggestion
        "matches the one set in the property",
```

src/scene/scenegraphnode.cpp
```diff
@@ -109,6 +109,26 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo EvalBoundingSphereInfo = {
+        "EvaluatedBoundingSphere",
+        "Evaluated Bounding Sphere (Read Only)",
```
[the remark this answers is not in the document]
WeirdRubberDuck: Not really.  The better option would be to address this issue https://github.com/OpenSpace/OpenSpace/issues/2966 and make it more obvious which properties are read-only (currently I find it a bit difficult to see in the UI)

WeirdRubberDuck, line 114:
```suggestion
        "Evaluated Bounding Sphere",
```

src/scene/scenegraphnode.cpp
```diff
@@ -109,6 +109,26 @@ namespace {
...
+        "This read-only property contains the evaluated value for the bounding sphere. "
+        "This is the actual value that is used internally within the software. If the "
+        "BoundingSphere property is set to -1, it is the computed value, otherwise it "
+        "matches the one set in the property",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo EvalInteractionSphereInfo = {
+        "EvaluatedInteractionSphere",
+        "Evaluated Interaction Sphere (Read Only)",
```
WeirdRubberDuck, line 124:
```suggestion
        "Evaluated Interaction Sphere",
```

PR #2975, opened 2023-12-08
Reorganization of Spice handling

alexanderbock:
Before this, every place that uses a SPICE name was just providing it as a fixed string which caused some errors to creep in (misspelling, accidentally requiring the wrong kernel asset, etc).  This change has assets that load kernels also define and export the Spice names that are loaded in these kernels as well.  Other assets that want to use a Spice name should access it through the exported name.
Of course the old way still works, but at least for our assets it will reduce the chances of errors.

The branch also includes some various cleanups for asset files


Roxeena (commented):
As I mentioned in the other PR, I think it is confusing to have _spice_ files instead of _kernel_ files, since they have similar names to the _spice/base_ asset file. I prefer having the individual files for a specific mission, planet, etc to be _kernels_ instead of _spice_, I think that is clearer with what it does and contains.

PR #2980, opened 2023-12-14
Issue/2665 - Show timer for how long time left in camera path

### LINKED ISSUE #2665: Camera path - show countdown in dashboard
In the dashboard the camera path could display how many seconds until it reached the target, approaching 0 as it arrives. This would give the presenter an idea of how long they should be talking

WeirdRubberDuck:
Related to WebGui PR: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/compare/issue%2F2665?expand=1

Closes #2665

### DIFF
--- a/modules/server/include/topics/camerapathtopic.h
+++ b/modules/server/include/topics/camerapathtopic.h
@@ -1,0 +1,55 @@
+/*****************************************************************************************
+ *                                                                                       *
+ * OpenSpace                                                                             *
+ *                                                                                       *
+ * Copyright (c) 2014-2023                                                               *
+ *                                                                                       *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this  *
+ * software and associated documentation files (the "Software"), to deal in the Software *
+ * without restriction, including without limitation the rights to use, copy, modify,    *
+ * merge, publish, distribute, sublicense, and/or sell copies of the Software, and to    *
+ * permit persons to whom the Software is furnished to do so, subject to the following   *
+ * conditions:                                                                           *
+ *                                                                                       *
+ * The above copyright notice and this permission notice shall be included in all copies *
+ * or substantial portions of the Software.                                              *
+ *                                                                                       *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,   *
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_MODULE_SERVER___CAMERAPATHTOPIC___H__
+#define __OPENSPACE_MODULE_SERVER___CAMERAPATHTOPIC___H__
+
+#include <modules/server/include/topics/topic.h>
+#include <chrono>
+
+namespace openspace {
+
+class CameraPathTopic : public Topic {
+public:
+    CameraPathTopic();
+    ~CameraPathTopic() override;
+
+    void handleJson(const nlohmann::json& json) override;
+    bool isDone() const override;
+
+private:
+    const int UnsetOnChangeHandle = -1;
+
+    void sendCameraPathData();
+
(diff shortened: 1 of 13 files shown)


modules/server/include/topics/camerapathtopic.h
```diff
@@ -0,0 +1,55 @@
...
+class CameraPathTopic : public Topic {
+public:
+    CameraPathTopic();
+    ~CameraPathTopic() override;
+
+    void handleJson(const nlohmann::json& json) override;
+    bool isDone() const override;
+
+private:
+    const int UnsetOnChangeHandle = -1;
```
alexanderbock, line 42: -> `constexpr` ?

modules/server/src/topics/camerapathtopic.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#include "modules/server/include/topics/CameraPathTopic.h"
+
+#include <modules/server/include/connection.h>
+#include <modules/server/servermodule.h>
+#include <modules/globebrowsing/globebrowsingmodule.h>
+#include <modules/globebrowsing/src/dashboarditemglobelocation.h>
```
alexanderbock, line 30: server <-> globebrowsing  includes can switch places

modules/server/src/topics/camerapathtopic.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    ServerModule* module = global::moduleEngine->module<ServerModule>();
+    _dataCallbackHandle = module->addPreSyncCallback(
+        [this]() {
+            bool isInPath = (global::openSpaceEngine->currentMode()
+                == OpenSpaceEngine::Mode::CameraPath);
+
+            std::chrono::system_clock::time_point now = std::chrono::system_clock::now();
+            if (isInPath && now - _lastUpdateTime > _cameraPathUpdateTime) {
```
alexanderbock, line 82: -> `(now - _lastUpdateTime)` to be extra clear about operator precedence?

modules/server/src/topics/camerapathtopic.cpp
```diff
@@ -0,0 +1,121 @@
...
+                sendCameraPathData();
+                _lastUpdateTime = std::chrono::system_clock::now();
+            }
+        }
+    );
+}
+
+void CameraPathTopic::sendCameraPathData() {
+    using namespace openspace;
```
alexanderbock, line 91: No need for this since we are already in the `openspace` namespace

modules/server/src/topics/camerapathtopic.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    const interaction::PathNavigator& pathNavigator =
+        global::navigationHandler->pathNavigator();
+
+    const interaction::Path* path = pathNavigator.currentPath();
+
+    if (!path) {
+        // This should not happen, as we are only sending this data if the engine mode
+        // is camera path traversal. But guard against it anyways
+        return;
```
alexanderbock, line 101: Should we have a `ghoul_assert` as well if it shouldn't happen?

src/navigation/pathnavigator.cpp
```diff
@@ -208,6 +208,17 @@ bool PathNavigator::isPlayingPath() const {
...
 
+bool PathNavigator::isPaused() const {
+    return hasCurrentPath() && !_isPlaying;
+}
+
+float PathNavigator::estimatedRemainingTimeInPath() const {
+    if (!hasCurrentPath()) {
+        return 0.f;
+    }
+    return _currentPath->estimatedRemainingTime(_speedScale);
```
alexanderbock, line 219:
```cpp
return hasCurrentPath() ? _currentPath->estimatedRemainingTime(_speedScale) : 0.f;
```
?

alexanderbock (approved):
:+1:

modules/server/src/topics/camerapathtopic.cpp
WeirdRubberDuck: Both can actually be removed - they're a remnant from copying from the cameratopic

But I'll change the order in the original file

modules/server/include/topics/camerapathtopic.h
WeirdRubberDuck: Has to be a `static constexpr`, but yes 👍

modules/server/src/topics/camerapathtopic.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+    if (!path) {
+        ghoul_assert(path, "Path must exist");
+        return;
+    }
+
+    // The time is not exact, and we only care about the number of seconds. Also,
+    // any negative values should be interpreted as positive
+    int seconds = static_cast<int>(
+        std::ceil(pathNavigator.estimatedRemainingTimeInPath())
```
WeirdRubberDuck, line 99: Consider changing this rounding to `round` or `floor`. Now it sometimes seems like the path ends about 1 second before it should. Or actually, maybe don't do any rounding at all and just show the integer version of the number.

(The reason I went with ceil was that with sometimes get values that are slightly smaller than zero, as the estimated time is not 100% correct)

PR #2982, opened 2023-12-15
Issue/2949 - Fix for camera transition events

### LINKED ISSUE #2949: "Jump to" navigation mode does not trigger events
When directly jumping to a node, their corresponding enter-exit event are not triggered:

In `default` profile:
1. Being at Earth
2. Shift+Left click on the Plane menu for the Sun
3. Screen is white as the Sun Glare is not faded out with the event

WeirdRubberDuck:
Closes #2949 by not limiting the events being triggered to one per frame (i.e. not assuming a smooth transition between the different spheres)

Also checks if the anchor node has changed since last frame and trigger events accordingly.

### DIFF
--- a/src/navigation/navigationhandler.cpp
+++ b/src/navigation/navigationhandler.cpp
@@ -183,6 +183,7 @@ void NavigationHandler::updateCamera(double deltaTime) {
     // If there is a state to set, do so immediately and then return
     if (_pendingState.has_value()) {
         applyPendingState();
+        updateCameraTransitions();
         return;
     }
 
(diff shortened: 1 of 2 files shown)


src/navigation/navigationhandler.cpp
```diff
@@ -371,10 +372,55 @@ void NavigationHandler::updateCameraTransitions() {
...
+            node,
+            events::EventCameraFocusTransition::Transition::Exiting
         );
+    };
+
+    bool anchorWasChanged = anchorNode() != _lastAnchor;
+    if (anchorWasChanged) {
+        // The anchor was changed between frames, so the transitions we have to check
+        // are a bit different. Just directly trigger the relevant events for the
+        // repsective node
```
alexanderbock, line 384:
```suggestion
        // respective node
```

alexanderbock (approved):
Neat!

PR #2994, opened 2024-01-05
Feature/point cloud scaling

### LINKED ISSUE #2986: Fix point cloud pixel based scaling
It never really worked, and will not as of the [Point Cloud Update](https://github.com/OpenSpace/OpenSpace/pull/2932) either. 

Ideally, we would want to come up with some clever scaling that can be done in world-scale, and equally across all viewports (otherwise, we need to somehow consider different viewport sizes, warping meshes, etc.. And that seems a lot harder)

And it should be easy to use both in domes and for planar projections

WeirdRubberDuck:
Fixes #2986, by introducing a completely world-scale-based computation for the max size of the points, rather than a screen-space-based one. 

As the computation is done in world space, based on the camera distance to the point position, this should work for all types of projection systems. Still got to try it in our dome, however. As well as come up with suitable max size values

The max size is now specified as an angle that the position is allowed to take up out of the camera's horizontal field of view, sort of (but in world-space)


modules/base/shaders/billboardpoint_gs.glsl
```diff
@@ -93,35 +93,24 @@ void main() {
...
+    // to take up. Note that the max size is for the diameter, and we need the radius
+    float desiredAngleRadians = radians(maxAngularSize * 0.5);
+
+    double distanceToCamera = length(dpos.xyz - cameraPosition);
+    double pointSize = length(dvec3(scaledRight));
+    // @TODO (2023-01-05, emmbr) Consider if this atan computation can be optimized using
+    // approximation
+    float angle = atan(float(pointSize / distanceToCamera));
+
+    if ((angle > desiredAngleRadians) && (distanceToCamera > 0)) {
```
alexanderbock, line 107: `0` -> `0.0`

alexanderbock (approved):

alexanderbock:
I think the H-II regions could be a bit bigger when comparing them to the OB Associations 
image

Maybe also making the Open Star Clusters a bit bigger?

alexanderbock:
Putting the 6dF galaxies to 22.5?

alexanderbock:
The Scale Exponent of the Voids doesn't start with a value:

image

WeirdRubberDuck:
> I think the H-II regions could be a bit bigger when comparing them to the OB Associations 
>
>
> Maybe also making the Open Star Clusters a bit bigger?

Not sure the _H-II regions_ should be bigger, as there are some places in the dataset where the points are quite dense and already create quite a bit of overlap, such as here

image

Could increase the size of the _Globular Clusters_ a bit if we want. Maybe to a scale exponent of 18.9, as seen below? 

image

And regarding reducing the size of the _6df galaxies_. I agree. Setting a scale exponent of 22.5 and a max size of 0.2 would make them blend better with the SDSS dataset. 

But in general, I don't feel knowledgeable enough about these datasets to decide what's a good size for them, as it depends on how one intends to view them. I've done my best in trying to preserve the looks of the points compared to the previous scaling, but t would probably be good if someone with an opinion did a pass over them as well after this PR is merged :)

WeirdRubberDuck:
Found the issue with the Voids and it's a bug that is separate form this PR.  I'll give it a look separately from this

WeirdRubberDuck:
Checked the scaling in the dome as well, and did not notice anything weird. It seems to work as expected :)

alexanderbock:
> > I think the H-II regions could be a bit bigger when comparing them to the OB Associations
> > Maybe also making the Open Star Clusters a bit bigger?
> 
> @alexanderbock
> 
> Not sure the _H-II regions_ should be bigger, as there are some places in the dataset where the points are quite dense and already create quite a bit of overlap, such as here
> 
> image
> 
> Could increase the size of the _Globular Clusters_ a bit if we want. Maybe to a scale exponent of 18.9, as seen below?
> 
> image
> 
> And regarding reducing the size of the _6df galaxies_. I agree. Setting a scale exponent of 22.5 and a max size of 0.2 would make them blend better with the SDSS dataset.
> 
> But in general, I don't feel knowledgeable enough about these datasets to decide what's a good size for them, as it depends on how one intends to view them. I've done my best in trying to preserve the looks of the points compared to the previous scaling, but t would probably be good if someone with an opinion did a pass over them as well after this PR is merged :)

H-II regions I agree, I didn't see the small ones

I'd say go ahead with the globular cluster values from the image and the smaller 6df

PR #3004, opened 2024-01-22
Issue/2985 - Add asset with default cmaps

### LINKED ISSUE #2985: Make syncable resource for some default color maps
Can replace the viridis map used in the point clouds example image

Either way, it would be useful to provide a few default color maps that the users can just use out of the box.  

This could be done in an asset that exports named versions of the file path (which could then be imported in other assets)

WeirdRubberDuck:
Adds an asset that gives access to synced filed for the color maps from Matplotlob, using the categories/classes they have in their documentation: https://matplotlib.org/stable/users/explain/colors/colormaps.html

Also adds the functionality to invert color maps, and change the color map file during runtime

Note that the color maps are only given as cmap files, with 256 colors.  We might also want to generate textures for them, eventually

Closes #2985


data/assets/examples/pointclouds/points_datamapping.asset
```diff
@@ -29,7 +30,7 @@ local Example = {
     -- missing data values in a specific color
     Coloring = {
       ColorMapping = {
-        File = asset.resource("viridis.cmap"),
+        File = colormaps.Uniform.magma,
```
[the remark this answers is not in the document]
WeirdRubberDuck: Yes!  To utilize the fact that it is easier to use another color map now

data/assets/util/default_colormaps.asset
```diff
@@ -0,0 +1,199 @@
...
+}
+asset.export("Misc", Misc)
+
+
+
+asset.meta = {
+  Name = "Default Colormaps",
+  Version = "1.0",
+  Description = [[Load a set of default color maps that can be used in other assets. The
+    colormaps are created and categorized based on the ones in Matplotlib.
```
WeirdRubberDuck, line 194:
```suggestion
    color maps are created and categorized based on the ones in Matplotlib.
```

alexanderbock:
Yea, I just feel that it would be more consistent with the rest of the assets, since we always have keys capitalized.

data/assets/util/default_colormaps.asset
```diff
@@ -0,0 +1,199 @@
...
+  turbo = misc .. "turbo.cmap",
+  nipy_spectral = misc .. "nipy_spectral.cmap",
+  gist_ncar = misc .. "gist_ncar.cmap"
+}
+asset.export("Misc", Misc)
+
+
+
+asset.meta = {
+  Name = "Default Colormaps",
```
WeirdRubberDuck, line 191:
```suggestion
  Name = "Default Color Maps",
```

PR #3010, opened 2024-01-24
Add event topic

### LINKED ISSUE #1810: Add topic for events
the topic could have 2 functions.

1) provide a description of actions that are mapped to events.
2) send events that occur over the socket.

engbergandreas:
Add ability to subscribe to events via js-api. Closes #1810 

example: 

```
  const foo = api.startTopic("event", {
    event: ["RenderableEnabled", "RenderableDisabled"],
    status: "start_subscription"
  });
```
where property `event` can be either a string or a list of strings. To subscribe to all events use `"*"` or `"all"`

### DIFF
--- a/include/openspace/events/eventengine.h
+++ b/include/openspace/events/eventengine.h
@@ -36,12 +36,19 @@ namespace events { struct Event; }
 
 class EventEngine {
 public:
+    using ScriptCallBack = std::function<void(ghoul::Dictionary)>;
+
     struct ActionInfo {
         events::Event::Type type;
         uint32_t id = std::numeric_limits<uint32_t>::max();
         bool isEnabled = true;
         std::string action;
         std::optional<ghoul::Dictionary> filter;
+    };
+
+    struct TopicInfo {
+        size_t id;
+        ScriptCallBack callback;
     };
 
     /**
@@ -88,6 +95,16 @@ public:
         std::optional<ghoul::Dictionary> filter = std::nullopt);
 
     /**
+     * Registers a new topic for a specific event type.
+     *
+     * \param topicId The id of the topic that will be triggered
+     * \param type The type for which a new topic is registered
+     * \param callback The callback function that will be called on triggered event
+    */
+    void registerEventTopic(size_t topicId, events::Event::Type type,
+        ScriptCallBack callback);
+
+    /**
      * Removing registration for a type/action combination.
      *
      * \param type The type of the action that should be unregistered
@@ -104,6 +121,14 @@ public:
      * \param identifier The unique identifier of the event that should be removed
      */
     void unregisterEventAction(uint32_t identifier);
+
+    /**
+     * Removing registration for a topic/type combination
+     *
+     * \param topicId The id of the topic that should b unregistered
+     * \param type The type of the topic that should be unregistered
+    */
+    void unregisterEventTopic(size_t topicId, events::Event::Type type);
 
     /**
      * Returns the list of all registered actions, sorted by their identifiers.
@@ -134,6 +159,12 @@ public:
      */
     void triggerActions() const;
 
+    /**
+     * Triggers all topics that are registered for events that are in the current event
+     * queue.
+    */
+    void triggerTopics() const;
+
     static scripting::LuaLibrary luaLibrary();
 
 private:
@@ -148,6 +179,8 @@ private:
     /// to be able to return them to a caller and we want it in this unordered_map to make
     /// the lookup really fast. So having this extra wasted memory is probably worth it
     std::unordered_map<events::Event::Type, std::vector<ActionInfo>> _eventActions;
+
+    std::unordered_map<events::Event::Type, std::vector<TopicInfo>> _eventTopics;
 
     static uint32_t nextRegisteredEventId;
 
--- a/include/openspace/events/eventengine.h
+++ b/include/openspace/events/eventengine.h
@@ -36,7 +36,7 @@ namespace events { struct Event; }
 
 class EventEngine {
 public:
-    using ScriptCallBack = std::function<void(ghoul::Dictionary)>;
+    using ScriptCallback = std::function<void(ghoul::Dictionary)>;
 
     struct ActionInfo {
         events::Event::Type type;
@@ -47,8 +47,8 @@ public:
     };
 
     struct TopicInfo {
-        size_t id;
-        ScriptCallBack callback;
+        uint32_t id = std::numeric_limits<uint32_t>::max();
+        ScriptCallback callback;
     };
 
     /**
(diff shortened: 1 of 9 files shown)


include/openspace/events/eventengine.h
```diff
@@ -36,6 +36,8 @@ namespace events { struct Event; }
 
 class EventEngine {
 public:
+    using ScriptCallBack = std::function<void(ghoul::Dictionary)>;
```
alexanderbock, line 39:
```suggestion
    using ScriptCallback = std::function<void(ghoul::Dictionary)>;
```

include/openspace/events/eventengine.h
```diff
@@ -44,6 +46,11 @@ class EventEngine {
         std::optional<ghoul::Dictionary> filter;
     };
 
+    struct TopicInfo {
+        size_t id;
+        ScriptCallBack callback;
```
alexanderbock, line 51:
```suggestion
        ScriptCallback callback;
```

include/openspace/events/eventengine.h
```diff
@@ -44,6 +46,11 @@ class EventEngine {
         std::optional<ghoul::Dictionary> filter;
     };
 
+    struct TopicInfo {
+        size_t id;
```
alexanderbock, line 50: Should initialize this to some number, similar to the ActionInfo, so maybe:
```suggestion
        uint32_t id = std::numeric_limits<uint32_t>::max();
```
?

include/openspace/events/eventengine.h
```diff
@@ -87,6 +94,16 @@ class EventEngine {
...
 
+    /**
+     * Registers a new topic for a specific event type.
+     *
+     * \param topicId The id of the topic that will be triggered
+     * \param type The type for which a new topic is registered
+     * \param callback The callback function that will be called on triggered event
+    */
+    void registerEventTopic(size_t topicId, events::Event::Type type,
+        ScriptCallBack callback);
```
alexanderbock, line 105:
```suggestion
        ScriptCallback callback);
```

include/openspace/events/eventengine.h
```diff
@@ -105,6 +122,14 @@ class EventEngine {
      */
     void unregisterEventAction(uint32_t identifier);
 
+    /**
+     * Removing registration for a topic/type combination
+     *
+     * \param topicId The id of the topic that should b unregistered
```
alexanderbock, line 128:
```suggestion
     * \param topicId The id of the topic that should be unregistered
```

include/openspace/events/eventengine.h
```diff
@@ -105,6 +122,14 @@ class EventEngine {
      */
     void unregisterEventAction(uint32_t identifier);
 
+    /**
+     * Removing registration for a topic/type combination
```
alexanderbock, line 126: Maybe having a line in here describing what happens if you try to unregister a topicid that doesn't exist, or a topicid that is not mapped to the passed type

modules/server/include/topics/eventtopic.h
```diff
@@ -0,0 +1,52 @@
...
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_MODULE_SERVER___EVENT_TOPIC___H__
+#define __OPENSPACE_MODULE_SERVER___EVENT_TOPIC___H__
+
+#include <modules/server/include/topics/topic.h>
+#include <openspace/events/event.h>
```
alexanderbock, line 29:
```suggestion

#include <openspace/events/event.h>
```

We have an empty line after the include containing the header of the base class(es)

modules/server/include/topics/eventtopic.h
```diff
@@ -0,0 +1,52 @@
...
+class EventTopic : public Topic {
+public:
+    EventTopic() = default;
+    ~EventTopic() override = default;
+
+    void handleJson(const nlohmann::json& json) override;
+    bool isDone() const override;
+
+private:
+    // Returns true if there is atleast one subscription active, false otherwise
```
alexanderbock, line 44:
```suggestion
    // Returns true if there is at least one subscription active, false otherwise
```

modules/server/src/topics/eventtopic.cpp
```diff
@@ -0,0 +1,111 @@
...
+namespace openspace {
+
+bool EventTopic::isDone() const {
+    return !isSubscribed();
+}
+
+void EventTopic::handleJson(const nlohmann::json& json) {
+    std::vector<std::string> events;
+
+    if (json.at("event").is_array()) {
```
alexanderbock, line 53: I think this could crash if someone sends in a JSON object that doesn't contain an "event" key.  So we should guard against that.  Even if the `at` function returns something that returns `false` for the `is_array` function, line 57 will then crash

modules/server/src/topics/eventtopic.cpp
```diff
@@ -0,0 +1,111 @@
...
+                auto type = static_cast<events::Event::Type>(i);
+                events.push_back(std::string(events::toString(type)));
+            }
+        }
+        else {
+            events.push_back(event);
+        }
+    }
+
+    const std::string& status = json.at("status").get<std::string>();
```
alexanderbock, line 72: Same comment here if "status" doesn't exist

modules/server/src/topics/eventtopic.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+bool EventTopic::isSubscribed() const {
+    if (_subscribedEvents.empty()) {
+        return false;
+    }
+
+    bool hasActiveSubscription = std::any_of(
+        _subscribedEvents.begin(),
+        _subscribedEvents.end(),
+        [](const std::pair<events::Event::Type, bool> subscription) {
```
alexanderbock, line 104:
```suggestion
        [](const std::pair<const events::Event::Type, bool>& subscription) {
```

src/events/eventengine.cpp
```diff
@@ -71,6 +71,16 @@ void EventEngine::registerEventAction(events::Event::Type type,
     nextRegisteredEventId++;
 }
 
+void EventEngine::registerEventTopic(size_t topicId, events::Event::Type type,
+                                     ScriptCallBack callback)
```
alexanderbock, line 75:
```suggestion
                                     ScriptCalback callback)
```

alexanderbock (approved):

Issue #1250, opened 2020-07-08
Fix the loading of GIBS globebrowsing layers at startup


alexanderbock:
When starting OpenSpace, the GIBS GetCapabilitiesFile is not loaded correctly and throws this warning:

Could not open dataset: https://gibs.earthdata.nasa.gov/twms/epsg4326/best/twms.cgi?request=GetTileService

This prevents the GlobeBrowsing UI from working

GPayne:
I'm not seeing this error right now with the default profile, at least on linux.

alexanderbock:
It's a very intermittent error that I have been unable to track down.  I have the suspicion that it might be related to how fast the datasets returns results and if they might be split into separate packets and not correctly reassembled maybe

PR #3015, opened 2024-01-31
Issue/1250 increase timeout for gdal http and updated url for loading gibs

micahnyc:
a value of 10 worked for me but I thought to make it 15 for EU users....

### DIFF
--- a/modules/globebrowsing/globebrowsingmodule.cpp
+++ b/modules/globebrowsing/globebrowsingmodule.cpp
@@ -84,6 +84,7 @@
 #endif // _MSC_VER
 
 #include <cpl_string.h>
+#include <cpl_conv.h>
 
 #ifdef _MSC_VER
 #pragma warning (pop)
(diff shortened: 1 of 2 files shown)


micahnyc:
fixes #1250

micahnyc:
want to check if this is affecting the loading of the tiles

modules/globebrowsing/globebrowsingmodule.cpp
```diff
@@ -84,6 +84,7 @@
 #endif // _MSC_VER
 
 #include <cpl_string.h>
+#include <cpl_conv.h>
```
alexanderbock, line 87:
```suggestion
#include <cpl_conv.h>
#include <cpl_string.h>
```

modules/globebrowsing/globebrowsingmodule.cpp
```diff
@@ -258,6 +259,8 @@ void GlobeBrowsingModule::internalInitialize(const ghoul::Dictionary& dict) {
             static_cast<size_t>(CpuCap.installedMainMemory() * 0.25 * 1024 * 1024)
         );
         addPropertySubOwner(GdalWrapper::ref());
+
+        CPLSetConfigOption("GDAL_HTTP_TIMEOUT", "15");
```
alexanderbock, line 263: Would be better to move this to the GDALWrapper class where we already have:

`    CPLSetConfigOption("GDAL_HTTP_TIMEOUT", "3"); // 3 seconds`

alexanderbock (changes requested):

alexanderbock:
Can you squash-merge the commits as the current commit messages are not standard (not starting with a capital letter, for example)

Issue #2117, opened 2022-05-16
ProfileEditor: Load navigation state from a saved file


WeirdRubberDuck:
We can save the current navigation state of the camera using `openspace.navigation.saveNavigationState`

Would be useful to be able to load a saved navigation state directly in the profile editor when setting the Camera field, so the user can just load a file instead of having to type in all the values manually

WeirdRubberDuck:
Note that we might also have to check that the save function works as expected. It is not used very often...

alexanderbock:
I've been using the `openspace.navigation.saveNavigationState` and `openspace.navigation.loadNavigationState` in conjunction quite a lot to be able to redo session recordings from a fixed start position.  Iirc the saveNavigationState returns it in a "legacy" Lua format of the form:
```lua
return {
  ...
}
```
which is not optimal

alexanderbock:
Fixed in 2fbcb0544ef2823dbd13aeadd35d0a466788fff0

PR #3023, opened 2024-02-12
NavigationState handling

alexanderbock:
Change the navigation state format from Lua to JSON (applies to `openspace.navigation.saveNavigationState` and `openspace.navigation.loadNavigationState`. Before, the `yaw` and `pitch` values were ignored if the `up` vector was not specified. Now, these three values are loaded independently.

Add the ability to set the navigation state in the profile editor based on the new JSON format.

### DIFF
--- a/src/navigation/navigationhandler.cpp
+++ b/src/navigation/navigationhandler.cpp
@@ -557,6 +557,10 @@ void NavigationHandler::saveNavigationState(const std::filesystem::path& filepat
     }
 
     std::filesystem::path absolutePath = absPath(filepath);
+    if (!absolutePath.has_extension()) {
+        // Adding the .navstate extension to the filepath if it came without one
+        absolutePath.replace_extension(".navstate");
+    }
     LINFO(fmt::format("Saving camera position: {}", absolutePath));
 
     std::ofstream ofs(absolutePath);
@@ -580,7 +584,7 @@ void NavigationHandler::loadNavigationState(const std::string& filepath) {
 
     std::ifstream f(filepath);
     std::string contents = std::string(
-        (std::istreambuf_iterator<char>(f)),
+        std::istreambuf_iterator<char>(f),
         std::istreambuf_iterator<char>()
     );
     nlohmann::json json = nlohmann::json::parse(contents);
--- a/src/navigation/navigationhandler.cpp
+++ b/src/navigation/navigationhandler.cpp
@@ -567,7 +567,7 @@ void NavigationHandler::saveNavigationState(const std::filesystem::path& filepat
         ));
     }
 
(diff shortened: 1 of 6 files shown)


alexanderbock:
Closes #2117 when merged

src/navigation/navigationhandler.cpp
```diff
@@ -578,22 +578,15 @@ void NavigationHandler::loadNavigationState(const std::string& filepath) {
...
-            "NavigationState"
-        );
-        setNavigationStateNextFrame(NavigationState(navigationStateDictionary));
-    }
-    catch (ghoul::RuntimeError& e) {
-        LERROR(fmt::format("Unable to set camera position: {}", e.message));
-    }
+    std::ifstream f(filepath);
+    std::string contents = std::string(
+        (std::istreambuf_iterator<char>(f)),
```
WeirdRubberDuck, line 583:
```suggestion
        std::istreambuf_iterator<char>(f),
```
?

src/navigation/navigationstate.cpp
```diff
@@ -89,11 +89,39 @@ NavigationState::NavigationState(const ghoul::Dictionary& dictionary) {
...
+
+NavigationState::NavigationState(const nlohmann::json& json) {
+    position.x = json["position"]["x"].get<double>();
+    position.y = json["position"]["y"].get<double>();
+    position.z = json["position"]["z"].get<double>();
+
+    anchor = json["anchor"];
+
+    if (auto it = json.find("referenceframe");  it != json.end()) {
+        referenceFrame = it->get<std::string>();
```
WeirdRubberDuck, line 105: This should be set to the anchor node if the reference frame was not specified

WeirdRubberDuck: Currently, loading a navigation stat ethat looks like this leads to an error:
```
{
  "anchor": "Earth",
  "position": {
    "x": 13514805.101321544,
    "y": 5454659.039307959,
    "z": -18279333.341204002
  },
  "up": {
    "x": 0.8035739201794957,
    "y": 0.0035944893626913277,
    "z": 0.59519411493527
  }
}
```

image

WeirdRubberDuck (changes requested):
Noticed that it is still not saved using a file extension.  Should we?  

Either json, or maybe our custom one 

Also, note to us: this is a highly breaking change and should be included in the list of breaking changes that I have started adding to the docs in this PR: 

https://github.com/OpenSpace/OpenSpace-Docs/pull/11  
(still waiting for someone to read it)

alexanderbock:
If the path provided to the `saveNavigationState` doesn't have an extension, a `.navstate` one is now added. If the user _did_ provide a different extension, that one is used instead

PR #3073, opened 2024-03-07
Add the ability to convert TLE files to SPICE kernels and apply change to ISS

### LINKED ISSUE #1578: ISS mean anomaly / position along orbit is ahead
The position of ISS in OpenSpace is ahead of where it is in the Celestrak visualization.  Micah looking at Kepler Translation as it pertains to mean anomaly.

### LINKED ISSUE #1783: ISS reference frame is doing too much work
Seems like ISS -> EarthInertial -> EarthCenter.

ISS is doing GALACTIC -> J2000
EarthInertial is doing J2000 -> GALACTIC

So if we were to parent the ISS directly to the EarthCenter, we could save two rotations

alexanderbock:
Adds two new SPICE functions `openspace.space.convertTLEtoSPK` and `openspace.space.tleToSpiceTranslation`.  The first converts a TLE file into a type 10 SPK kernel t hat can be used for a `SpiceTranslation`.  The second function uses the former to create a kernel and then returns back a fully formed `SpiceTranslation` table that can be used in an asset.

This also hardcoded loads a geophysical constants kernel needed for SPK10 kernels to work.

This PR also applies those changes to the ISS to both fix its position for a longer time (fixes #1578) and use a new dynamic frame kernel to have a correct orientation for the ISS (fixes #1783)

### DIFF
--- a/data/assets/scene/solarsystem/planets/earth/satellites/misc/iss.asset
+++ b/data/assets/scene/solarsystem/planets/earth/satellites/misc/iss.asset
@@ -20,6 +20,15 @@ local omm = asset.resource({
   SecondsUntilResync = 24 * 60 * 60
 })
 
+local tle = asset.resource({
+  Name = "Satellite TLE Data (ISS)",
+  Type = "UrlSynchronization",
+  Identifier = "satellite_tle_data_iss",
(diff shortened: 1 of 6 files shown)


alexanderbock:
Closes #1578 
Closes #1783

PR #3078, opened 2024-03-12
Include timestamp in navigation state

### LINKED ISSUE #3076: Add simulation time stamp to NavigationState
Now that we've changed the navigation state file format we should also add the simulation timestamp in that file (and in the navigation state data structure?)

WeirdRubberDuck:
And update set/load function to have the ability to include the timestamp

closes #3076 

Also fixes a bug where the absolute path was not correctly used when loading a navstate and makes it so that the `.navstate` file extension is added per default if no extension is included in the filename


include/openspace/navigation/navigationstate.h
```diff
@@ -41,7 +41,8 @@ struct NavigationState {
     explicit NavigationState(const nlohmann::json& json);
     NavigationState(std::string anchor, std::string aim, std::string referenceFrame,
         glm::dvec3 position, std::optional<glm::dvec3> up = std::nullopt,
-        double yaw = 0.0, double pitch = 0.0);
+        double yaw = 0.0, double pitch = 0.0,
+        std::optional<double> timestamp_ = std::nullopt);
```
WeirdRubberDuck, line 45:
```suggestion
        std::optional<double> timestamp = std::nullopt);
```

src/navigation/navigationhandler_lua.inl
```diff
@@ -66,13 +72,21 @@ namespace {
     return state.dictionary();
 }
 
-// Set the navigation state. The argument must be a valid Navigation State.
-[[codegen::luawrap]] void setNavigationState(ghoul::Dictionary navigationState) {
+// Set the navigation state. The first argument must be a valid Navigation State. If
+// usetimeStamp is set to true and the provided navigation state has a timestamp, time
```
alexanderbock, line 76:
```suggestion
// useTimeStamp is set to true and the provided navigation state has a timestamp, time
```

alexanderbock (approved):

PR #3084, opened 2024-03-17
Switch all getline reads to tolerate line endings of any operating system (Issue #2754)

### LINKED ISSUE #2754: Exoplanets have problems on Linux
For some exoplanets in the menu, when selecting and clicking "Add System" errors are logged with the pattern:

```
(E) Lua                  Error executing script '': [string "openspace.addSceneGraphNode({Identifier = 'K2..."]:1: Error loading scene graph node: (RenderableGlobe) Error in specification
(E) Scene                Could not find parent 'K2-105' for 'K2-105_b_Trail'
```

with similar errors as above but ending in `b_Trail`, `b_Disc`, `1AU_Circle`, `HZ_Disc`, and `_Glare`.

GPayne:
Uses a new ghoul::getline that will handle the `\r\n` line ending of Windows, and `\n` of Mac or Linux. Any external files (data, telemetry, recordings, etc) can have any line ending regardless of which operating system OpenSpace runs on.
Fixes #2754


apps/OpenSpace/ext/launcher/src/profile/additionalscriptsdialog.cpp
```diff
@@ -26,6 +26,7 @@
 
 #include "profile/line.h"
 #include "profile/scriptlogdialog.h"
+#include <ghoul/misc/stringhelper.h>
 #include <openspace/scene/profile.h>
```
alexanderbock, line 30:
```suggestion
#include <openspace/scene/profile.h>
#include <ghoul/misc/stringhelper.h>
```

Include file sorting

src/interaction/sessionrecording.cpp
```diff
@@ -409,16 +410,22 @@ bool SessionRecording::startPlayback(std::string& filename,
         LERROR("Unknown data type in header (should be Ascii or Binary)");
         cleanUpPlayback();
     }
-    // throwaway newline character
-    readHeaderElement(_playbackFile, 1);
+    // throwaway newline character(s)
+    std::string lineEnd = readHeaderElement(_playbackFile, 1);
+    bool isDosLineEnding = (lineEnd == "\r");
```
alexanderbock, line 415:
```suggestion
    bool hasDosLineEnding = (lineEnd == "\r");
```

src/interaction/sessionrecording.cpp
```diff
@@ -409,16 +410,22 @@ bool SessionRecording::startPlayback(std::string& filename,
...
+    if (isDosLineEnding) {
```
alexanderbock, line 416:
```suggestion
    if (hasDosLineEnding) {
```

src/interaction/sessionrecording.cpp
```diff
@@ -409,16 +410,22 @@ bool SessionRecording::startPlayback(std::string& filename,
...
 
     if (_recordingDataMode == DataMode::Binary) {
         // Close & re-open the file, starting from the beginning, and do dummy read
         // past the header, version, and data type
         _playbackFile.close();
         _playbackFile.open(_playbackFilename, std::ifstream::in | std::ios::binary);
         size_t headerSize = FileHeaderTitle.length() + FileHeaderVersionLength
-            + sizeof(DataFormatBinaryTag) + sizeof('\n');
+            + sizeof(DataFormatBinaryTag) + sizeof('\n')
+            + ((isDosLineEnding) ? sizeof('\r') : 0);
```
alexanderbock, line 428:
```suggestion
            + (hasDosLineEnding ? sizeof('\r') : 0);
```

src/interaction/sessionrecording.cpp
```diff
@@ -409,16 +410,22 @@ bool SessionRecording::startPlayback(std::string& filename,
...
-    readHeaderElement(_playbackFile, 1);
+    // throwaway newline character(s)
+    std::string lineEnd = readHeaderElement(_playbackFile, 1);
+    bool isDosLineEnding = (lineEnd == "\r");
+    if (isDosLineEnding) {
+        // throwaway the second newline character (\n) also
+        readHeaderElement(_playbackFile, 1);
+    }
 
     if (_recordingDataMode == DataMode::Binary) {
```
alexanderbock, line 421: Are you sure we are actually writing the \r in the case of a binary session recording file?  I thought we'd write strings in this case as a length + the exact number of characters

support/coding/check_style_guide.py
```diff
@@ -342,6 +342,18 @@ def check_for_tab(lines):
   else:
     return ''
 
+
+
+def check_for_std_getline(lines):
+  index = [i for i,s in enumerate(lines)
+          if 'std::getline' in s]
+  if len(index) > 0:
+    return 'File used wrong glm include. Use "#include <ghoul/glm.h>" instead'
```
alexanderbock, line 351:
```suggestion
    return 'File used wrong std::getline function. Use ghoul::getline from the "ghoul/misc/stringhelper.h" instead'
```

src/interaction/sessionrecording.cpp
GPayne: The line `+ (hasDosLineEnding ? sizeof('\r') : 0);` has now been removed (see next comment).

src/interaction/sessionrecording.cpp
GPayne: It's true that when we write the header it has only `\n`. But this could get changed by doing a unix2dos conversion, or modifying an ascii format recording in a text editor.
I did remove the check for the `\r` line ending in the previous comment, though, because I don't see any reasonable way that problem could happen with a binary format recording.

support/coding/check_style_guide.py
GPayne: :man_facepalming:

PR #3085, opened 2024-03-17
Add support for audio playback

alexanderbock:
Adds audio support using the SoLoud library.

Example files available in the data/assets/example/audio folder.
Important new Lua functions added:
 - `openspace.sound.playAudio` / `openspace.sound.stopAudio`
 - `openspace.sound.setGlobalVolume`
 - `openspace.sound.setVolume`
 - `openspace.sound.pauseAudio` / `openspace.sound.resumeAudio`
 - `openspace.sound.playAudio3d`
 - `openspace.sound.set3dSourcePosition`


ylvaselling:
Tested the module and I think it works great! Some thoughts / suggestions:

* I would prefer the module be called `audio` to have the same naming convention as the lua scripting library name. Had a hard time finding it in Cmake because I looked for `audio`. Also rename the classes etc replacing sound with audio
* Imo the lua functions should take an identifier to the audio rather than be handed an identifier (the int that is returned). That seems more in line with the patterns we have for other lua functions, for example `setPropertyValue`. It also means that we can use these functions in the console, without having to read a return value somewhere that is generated by the engine.
* I also think it would be great to have functions for globally pausing and stopping the audio. These functions would not take any identifiers but simply stop all sound.

ylvaselling:
More things after discussion:

* Lua functions to pause and resume all, (and play all from beginning maybe?)
* Change the assets from the user properties to some simple actions

PR #3137, opened 2024-03-24
Add the ability to specifiy local Sun position for the RenderableGlobe and RenderableAtmosphere

### LINKED ISSUE #921: Configurable light source for planetary systems
Planetary systems can have different light sources (position, color, etc). Today the light source for planets is hardcoded as the Sun. It must be an asset property and the color of the light source should be configurable too. The Ptolemaic and binary system is a must.

### LINKED ISSUE #1745: RenderableGlobe objects only receive illumination from the Sun and ignore other light sources
Renderableglobe objects are hardcoded to be illuminated by the Sun and even specifying additional/alternative lightsources has no effect.

For exoplanets in particular it would be good to be able to specify their lightsource as their parent stars for proper display (including transit demos). This could also have long-term visualization use intra-solar system for lunar/planetary reflection as an illumination source for planets & moons.

### LINKED ISSUE #2243: RenderableGlobes not lit by the Sun, but world space (0, 0, 0)
The shading for the globes actually do not use the sun, but the world space origin. 

Add possibility to use specify light source for RenderableGlobes, instead of doing the shading hardcoded based on the worldspace origin, (0, 0, 0).

Comment form previous issue :
Note that just fixing the light source in the asset does not solve all the problems. In addition, it seems like the shading of the RenderableGlobes use (0, 0, 0) for the light source position. [See computation here](https://github.com/OpenSpace/OpenSpace/blob/44f7bd922526b606721ccb93aa1b5ed5d853a25c/modules/globebrowsing/src/renderableglobe.cpp#L1080)

[this was the issue where it was noticed](https://github.com/OpenSpace/OpenSpace/issues/2223)

Should also look through the shadowing code for eclipses

alexanderbock:
Note that for planets with atmospheres, the SunNode has to be specified for _both_ the globe and the atmosphere separately

 - Closes #1745
 - Closes #2243

### DIFF
--- a/modules/atmosphere/rendering/atmospheredeferredcaster.cpp
+++ b/modules/atmosphere/rendering/atmospheredeferredcaster.cpp
@@ -382,8 +382,8 @@ void AtmosphereDeferredcaster::preRaycast(const RenderData& data, const Deferred
             invModelMatrix * glm::dvec4(data.camera.eyePositionVec3(), 1.0);
         program.setUniform(_uniformCache.camPosObj, glm::dvec3(camPosObjCoords));
 
-        SceneGraphNode* node = sceneGraph()->sceneGraphNode("Sun");
-        const glm::dvec3 sunPosWorld = node ? node->worldPosition() : glm::dvec3(0.0);
+        const glm::dvec3 sunPosWorld =
+            _sunNode ? _sunNode->worldPosition() : glm::dvec3(0.0);
 
         glm::dvec3 sunPosObj;
         // Sun following camera position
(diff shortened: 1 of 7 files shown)


WeirdRubberDuck (approved):
Nice :)  

Some thoughts:
1. should we set the sun position for the planets in the solar system? (now they all still use (0, 0, 0), right?)
1. You use three different words for the light source in the code (Light source, sun, illumination). Would be nice to maybe stick to two of those 
1. Speaking of light sources - would we want to ask the user to specify a LightSource object instead of just a SGN name? (or maybe this is overkill for this use case, but I can see how we would want to change it down the line. See comments below)

And some remaining things that we want to do down the line:
- Now there is no way for the user to specify the color of the light source (which I can see being useful for exoplanets). 
- Specify a list of suns/light sources  (in the case of an exoplanet system having more than one star)

With that said, I don't think we want to close #921 with this PR. Otherwise I think it's ok

PR #3147, opened 2024-03-28
Fix loading of cached data values

WeirdRubberDuck:
The `std::memcopy` approach here does not work for std:vectors of floats (it will essentially only work for `char` like things, if I understand the internet correctly. However, a general consensus seems tot be to avoid `std::memcopy` in combination with `std::vector`)

### DIFF
--- a/src/data/dataloader.cpp
+++ b/src/data/dataloader.cpp
@@ -281,6 +281,7 @@ std::optional<Dataset> loadCachedFile(const std::filesystem::path& path) {
 
             // If we have a comment, we need to extract its length's worth of characters
             // from the buffer
+            // @TODO replace this with std::move as well?
             std::memcpy(e.comment->data(), &commentBuffer[commentIdx], e.comment->size());
 
             // and then advance the index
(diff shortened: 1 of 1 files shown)


src/data/dataloader.cpp
```diff
@@ -272,14 +272,16 @@ std::optional<Dataset> loadCachedFile(const std::filesystem::path& path) {
...
+        std::move(startOfCopy, startOfCopy + nValues, e.data.begin());
         valuesIdx += nValues;
 
         if (e.comment.has_value()) {
             ghoul_assert(commentIdx < commentBuffer.size(), "Index too large");
 
             // If we have a comment, we need to extract its length's worth of characters
             // from the buffer
+            // @TODO replace this with std::move as well?
             std::memcpy(e.comment->data(), &commentBuffer[commentIdx], e.comment->size());
```
WeirdRubberDuck, line 285: This just happens to work since the data being copied is a string, if I understand it correctly?  Should we go with the `std::move` approach here too?

PR #3161, opened 2024-04-04
Add asset setting to skip first data point in RenderablePointCloud

WeirdRubberDuck:
To hide our galaxy in the Tully datasets

### DIFF
--- a/modules/base/rendering/pointcloud/renderablepointcloud.cpp
+++ b/modules/base/rendering/pointcloud/renderablepointcloud.cpp
@@ -372,6 +372,9 @@ namespace {
         // [[codegen::verbatim(UseAdditiveBlendingInfo.description)]]
         std::optional<bool> useAdditiveBlending;
 
+        // If true, skip the first data point in the loaded dataset
+        std::optional<bool> skipFirstDataPoint;
+
         enum class [[codegen::map(openspace::DistanceUnit)]] Unit {
             Meter [[codegen::key("m")]],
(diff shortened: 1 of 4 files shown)


modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -701,6 +704,8 @@ RenderablePointCloud::RenderablePointCloud(const ghoul::Dictionary& dictionary)
 
     _useCaching = p.useCaching.value_or(true);
 
+    _skipFirstDataPoint = p.skipFirstDataPoint.value_or(false);
```
alexanderbock, line 707:
```suggestion
    _skipFirstDataPoint = p.skipFirstDataPoint.value_or(_skipFirstDataPoint);
```

That way, if we ever decide to change the default value in the header, we don't have the chance to miss that change here

alexanderbock (approved):

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -701,6 +704,8 @@ RenderablePointCloud::RenderablePointCloud(const ghoul::Dictionary& dictionary)
 
     _useCaching = p.useCaching.value_or(true);
```
WeirdRubberDuck, line 705:
```suggestion
    _useCaching = p.useCaching.value_or(_useCaching);
```

PR #3164, opened 2024-04-04
Updating the star rendering

alexanderbock:
This changes the RenderableStars to render two textures (a halo + a glare) on top of each other. Each texture component is configurable separately.  This change also has a few quality of life features in the code to remove double negations, etc.

### DIFF
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -76,12 +76,12 @@ local SunStar = {
   Renderable = {
     Type = "RenderableStars",
     File = sunspeck .. "sunstar.speck",
-    GlareTexture = textures .. "halo.png",
-    --ShapeTexture = textures .. "disc.png",
+    Halo = {
+      Texture = textures .. "halo.png"
+    },
     ColorMap = colormaps .. "colorbv.cmap",
     MagnitudeExponent = 6.2,
     SizeComposition = "Distance Modulus",
-    RenderMethod = "Texture Based", -- or PSF
     FadeInDistances = { 0.0001, 0.1 },
     RenderableType = "PostDeferredTransparent",
     DataMapping = {
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -32,11 +32,11 @@ local Stars = {
   Renderable = {
     Type = "RenderableStars",
     File = speck .. "stars.speck",
-    Glare = {
+    Halo = {
       Texture = textures .. "halo.png",
       PartialOpacity = 0.875
     },
-    Highlight = {
+    Glare = {
       Texture = textures .. "glare.png",
       PartialOpacity = 5.0,
       Scale = 0.7
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -34,15 +34,15 @@ local Stars = {
     File = speck .. "stars.speck",
     Halo = {
       Texture = textures .. "halo.png",
-      Multiplier = 0.55,
-      Size = 0.95
+      Multiplier = 0.65
     },
     Glare = {
       Texture = textures .. "glare.png",
-      Multiplier = 5.0,
-      Scale = 0.4
+      Multiplier = 15.0,
+      Gamma = 1.66,
+      Scale = 0.18
     },
-    MagnitudeExponent = 6.25,
+    MagnitudeExponent = 6.325,
     ColorMap = colormaps .. "colorbv.cmap",
     OtherDataColorMap = colormaps .. "viridis.cmap",
     SizeComposition = "Distance Modulus",
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -32,13 +32,19 @@ local Stars = {
   Renderable = {
     Type = "RenderableStars",
     File = speck .. "stars.speck",
-    Texture = textures .. "halo.png",
-    --ShapeTexture = textures .. "disc.png",
+    Glare = {
+      Texture = textures .. "halo.png",
+      PartialOpacity = 0.875
+    },
+    Highlight = {
+      Texture = textures .. "glare.png",
+      PartialOpacity = 5.0,
+      Scale = 0.7
+    },
+    MagnitudeExponent = 6.2,
     ColorMap = colormaps .. "colorbv.cmap",
     OtherDataColorMap = colormaps .. "viridis.cmap",
-    MagnitudeExponent = 6.2,
     SizeComposition = "Distance Modulus",
-    RenderMethod = "Texture Based", -- or PSF
     DataMapping = {
       Bv = "colorb_v",
       Luminance = "lum",
@@ -70,7 +76,7 @@ local SunStar = {
   Renderable = {
     Type = "RenderableStars",
     File = sunspeck .. "sunstar.speck",
-    Texture = textures .. "halo.png",
+    GlareTexture = textures .. "halo.png",
     --ShapeTexture = textures .. "disc.png",
     ColorMap = colormaps .. "colorbv.cmap",
     MagnitudeExponent = 6.2,
@@ -101,11 +107,11 @@ local SunStar = {
 
 asset.onInitialize(function()
   openspace.addSceneGraphNode(Stars)
-  openspace.addSceneGraphNode(SunStar)
+  -- openspace.addSceneGraphNode(SunStar)
 end)
 
 asset.onDeinitialize(function()
-  openspace.removeSceneGraphNode(SunStar)
+  -- openspace.removeSceneGraphNode(SunStar)
   openspace.removeSceneGraphNode(Stars)
 end)
 
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -131,7 +131,7 @@ asset.export(SunStar)
 
 asset.meta = {
   Name = "Stars",
-  Version = "2.1",
+  Version = "3.0",
   Description = "Digital Universe asset for the stars",
   Author = "Brian Abbott (AMNH)",
   URL = "https://www.amnh.org/research/hayden-planetarium/digital-universe",
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -78,12 +78,20 @@ local SunStar = {
     Type = "RenderableStars",
     File = sunspeck .. "sunstar.speck",
     Halo = {
-      Texture = textures .. "halo.png"
+      Texture = textures .. "halo.png",
+      Multiplier = 0.55,
+      Gamma = 1.1,
+      Size = 0.95
     },
+    Glare = {
+      Texture = textures .. "glare.png",
+      Multiplier = 1.0,
+      Scale = 0.1
+    },
+    MagnitudeExponent = 6.25,
     ColorMap = colormaps .. "colorbv.cmap",
-    MagnitudeExponent = 6.2,
     SizeComposition = "Distance Modulus",
-    FadeInDistances = { 0.0001, 0.1 },
+    FadeInDistances = { 0.0075, 0.1 },
     RenderableType = "PostDeferredTransparent",
     DataMapping = {
       Bv = "colorb_v",
--- a/data/assets/scene/digitaluniverse/stars.asset
+++ b/data/assets/scene/digitaluniverse/stars.asset
@@ -34,11 +34,11 @@ local Stars = {
     File = speck .. "stars.speck",
     Halo = {
       Texture = textures .. "halo.png",
-      PartialOpacity = 0.875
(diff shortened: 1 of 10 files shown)


data/assets/scene/digitaluniverse/stars.asset
```diff
@@ -32,13 +32,19 @@ local Stars = {
...
     File = speck .. "stars.speck",
-    Texture = textures .. "halo.png",
-    --ShapeTexture = textures .. "disc.png",
+    Halo = {
+      Texture = textures .. "halo.png",
+      PartialOpacity = 0.875
+    },
+    Glare = {
+      Texture = textures .. "glare.png",
+      PartialOpacity = 5.0,
```
WeirdRubberDuck, line 41: Why does the partial opacity go to more than 1?  Confused :)

modules/space/rendering/renderablestars.cpp
```diff
@@ -266,43 +266,21 @@ namespace {
...
 
-    constexpr openspace::properties::Property::PropertyInfo RenderMethodOptionInfo = {
-        "RenderMethod",
-        "Render Method",
-        "Render method for the stars",
+    constexpr openspace::properties::Property::PropertyInfo ScaleInfo = {
+        "Scale",
+        "Scale",
+        "A uniform scale factor that determines how much of the total size of the star "
+        "this component is using. If it is 0, it will be hidden, if it is 1, it will "
```
WeirdRubberDuck, line 273:
```suggestion
        "this component is using. If it is 0, it will be hidden. If it is 1, it will "
```

modules/space/rendering/renderablestars.cpp
```diff
@@ -266,43 +266,21 @@ namespace {
...
-        "User Provided Texture",
-        ""
-    };
-
-    const openspace::properties::PropertyOwner::PropertyOwnerInfo
-        ParametersOwnerOptionInfo =
-    {
+    const openspace::properties::PropertyOwner::PropertyOwnerInfo ParametersOwnerInfo = {
         "ParametersOwner",
         "Parameters Options",
```
WeirdRubberDuck, line 280: While we're making changes, maybe this one can be given a better name?   Something about size composition?

modules/space/rendering/renderablestars.cpp
```diff
@@ -533,100 +458,95 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
-        SizeCompositionOptionInfo,
-        properties::OptionProperty::DisplayType::Dropdown
-    )
-    , _lumCent(LumPercentInfo, 0.5f, 0.f, 3.f)
-    , _radiusCent(RadiusPercentInfo, 0.5f, 0.f, 3.f)
-    , _brightnessCent(BrightnessPercentInfo, 0.5f, 0.f, 3.f)
+    , _halo{
+        properties::PropertyOwner(HaloOwnerInfo),
+        properties::StringProperty(TextureInfo),
+        properties::FloatProperty(PartialOpacityInfo, 1.f, 0.f, 5.f),
```
WeirdRubberDuck, line 464: Again, why is max 5??

modules/space/rendering/renderablestars.cpp
```diff
@@ -533,100 +458,95 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+        properties::FloatProperty(ScaleInfo, 1.f, 0.f, 1.f)
+    }
+    , _glare{
+        properties::PropertyOwner(GlareOwnerInfo),
+        properties::StringProperty(TextureInfo),
+        properties::FloatProperty(PartialOpacityInfo, 1.f, 0.f, 5.f),
```
WeirdRubberDuck, line 470: Se above

modules/space/rendering/renderablestars.cpp
```diff
@@ -533,100 +458,95 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+    _colorTextureFile->setCallback([this]() { _colorTextureIsDirty = true; });
+    _colorTexturePath = p.colorMap.string();
+    _colorTexturePath.onChange([&] {
+        if (std::filesystem::exists(_colorTexturePath.value())) {
+            _colorTextureIsDirty = true;
+        }
+        else {
+            LWARNING(std::format("File not found: {}", _colorTexturePath.value()));
+        }
+        });
```
WeirdRubberDuck, line 511:
```suggestion
    });
```

modules/space/rendering/renderablestars.cpp
```diff
@@ -533,100 +458,95 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
     addProperty(Fadeable::_opacity);
 
-    _dataMapping.bvColor = p.dataMapping.bv.value_or("");
+    _speckFile = p.speckFile.string();
+    _speckFile.onChange([this]() { _speckFileIsDirty = true; });
+    addProperty(_speckFile);
+
+
```
WeirdRubberDuck, line 500:
```suggestion
```

modules/space/rendering/renderablestars.cpp
```diff
@@ -533,100 +458,95 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+        if (std::filesystem::exists(_colorTexturePath.value())) {
+            _colorTextureIsDirty = true;
+        }
+        else {
+            LWARNING(std::format("File not found: {}", _colorTexturePath.value()));
+        }
+        });
+    addProperty(_colorTexturePath);
+
+
```
WeirdRubberDuck, line 514:
```suggestion
```

modules/space/rendering/renderablestars.cpp
```diff
@@ -654,132 +574,99 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
 
-    _moffatPSFParamOwner.addProperty(_FWHMConst);
-    _FWHMConst.onChange(renderPsf);
-    _moffatPSFParamOwner.addProperty(_moffatBetaConst);
-    _moffatBetaConst.onChange(renderPsf);
 
-    _parametersOwner.addPropertySubOwner(_spencerPSFParamOwner);
-    _parametersOwner.addPropertySubOwner(_moffatPSFParamOwner);
+    _parameters.sizeComposition.addOptions({
+        { AppBrightness, "Use Star's Apparent Brightness" },
```
WeirdRubberDuck, line 636: The naming scheme of these should be made more consistent.  

E.g. the first two uses "Use", while the other's don't

modules/space/rendering/renderablestars.cpp
```diff
@@ -654,132 +574,99 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+        { LumSize, "Use Star's Luminosity and Size" },
+        { LumSizeAppBrightness, "Luminosity, Size, App Brightness" },
```
WeirdRubberDuck, line 638:
```suggestion
        { LumSizeAppBrightness, "Luminosity, Size, Apparent Brightness" },
```

If we want to be consistent with using the full word as in the other strings.  However, maybe it becomes too long?

modules/space/rendering/renderablestars.cpp
```diff
@@ -654,132 +574,99 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+    addProperty(_filterOutOfRange);
 
 
-    if (p.sizeComposition.has_value()) {
-        _psfMultiplyOption =
-            static_cast<int>(codegen::map<SizeComposition>(*p.sizeComposition));
-    }
-    else {
-        _psfMultiplyOption = 5;
+    auto markTextureAsDirty = [this]() {_pointSpreadFunctionTextureIsDirty = true; };
```
WeirdRubberDuck, line 604:
```suggestion
    auto markTextureAsDirty = [this]() {_pointSpreadFunctionTextureIsDirty = true; };
    
```
since it does not really belong with the halo texture only

modules/space/rendering/renderablestars.cpp
```diff
@@ -654,132 +574,99 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+        { LumSizeAppBrightness, "Luminosity, Size, App Brightness" },
+        { AbsMagnitude, "Absolute Magnitude" },
+        { AppMagnitude, "Apparent Magnitude" },
+        { DistanceModulus, "Distance Modulus" }
+    });
+    _parameters.sizeComposition =
+        p.sizeComposition.has_value() ?
+        static_cast<int>(codegen::map<SizeComposition>(*p.sizeComposition)) :
+        SizeComposition::DistanceModulus;
+    _parameters.container.addProperty(_parameters.sizeComposition);
```
WeirdRubberDuck, line 647:
```suggestion
    _parameters.container.addProperty(_parameters.sizeComposition);
    
```
Readability

modules/space/rendering/renderablestars.cpp
```diff
@@ -1071,36 +802,38 @@ void RenderableStars::render(const RenderData& data, RendererTasks&) {
...
-            glm::length(data.camera.positionVec3())
-        );
+    if (_enableFadeInDistance) {
+        const float distCam = static_cast<float>(glm::length(data.camera.positionVec3()));
         const glm::vec2 fadeRange = _fadeInDistances;
         const double a = 1.f / ((fadeRange.y - fadeRange.x) * PARSEC);
         const double b = -(fadeRange.x / (fadeRange.y - fadeRange.x));
-        const double funcValue = a * distCamera + b;
-        fadeInVariable *= static_cast<float>(funcValue > 1.f ? 1.f : funcValue);
+        const double funcValue = a * distCam + b;
```
WeirdRubberDuck, line 810: `distCam` here is a float. You likely want it as a double

modules/space/rendering/renderablestars.cpp
```diff
@@ -1071,36 +802,38 @@ void RenderableStars::render(const RenderData& data, RendererTasks&) {
...
+        glareUnit.activate();
+        _glare.texture->bind();
+        _program->setUniform(_uniformCache.glareTexture, glareUnit);
+        _program->setUniform(_uniformCache.glareOpacity, _glare.opacity);
+        _program->setUniform(_uniformCache.glareScale, _glare.scale);
     }
 
-    _program->setUniform(_uniformCache.psfTexture, psfUnit);
+    _program->setUniform(_uniformCache.hasGlare, _glare.texture != nullptr);
+
```
WeirdRubberDuck, line 836:
```suggestion
```

modules/space/rendering/renderablestars.cpp
```diff
@@ -533,100 +458,95 @@ RenderableStars::RenderableStars(const ghoul::Dictionary& dictionary)
...
+        properties::FloatProperty(PartialOpacityInfo, 1.f, 0.f, 5.f),
+        properties::FloatProperty(ScaleInfo, 1.f, 0.f, 1.f)
+    }
+    , _parameters{
+        properties::PropertyOwner(ParametersOwnerInfo),
+        properties::OptionProperty(
+            SizeCompositionOptionInfo,
+            properties::OptionProperty::DisplayType::Dropdown
+        ),
+        properties::FloatProperty(LumPercentInfo, 0.5f, 0.f, 3.f),
```
WeirdRubberDuck, line 479: These need some better documentation for when they are actually applied (i.e. that they come in effect depending on the size composition option). Also about what the "unit" for the value means, sort of

WeirdRubberDuck (changes requested):
Neat!  But found some problems: 

1. Setting the Size Composition to the one shown in the image below leads to my machine first freezing and then OpenSpace crashing due to an unhandled exception. Eek
image
1. Irreversible rendering issues when switching color option (that I wrote to you about in Slack)


The first might have existed in the renderable before this update, but either way it should be fixed and this is a good opportunity to do so

modules/space/rendering/renderablestars.cpp
alexanderbock: Renamed the container to `SizeComposition` and the dropbox to `Method`

modules/space/rendering/renderablestars.cpp
alexanderbock: The double linebreak here (and further down) is on purpose to group the different "groups" of properties visually together. Otherwise it was becoming too much of a mess to quickly parse

data/assets/scene/digitaluniverse/stars.asset
alexanderbock: Agreed.  Renamed it to "Multiplier"

modules/space/rendering/renderablestars.cpp
alexanderbock: This composition method is no more

modules/space/rendering/renderablestars.cpp
alexanderbock: Honestly, I have no idea what the units are there is a lot of mystical math happening in the shader.   If i knew that no user is actually using these, I'd be super happy to get rid of all other size composition options.

WeirdRubberDuck: Ah, ok.  Well that's a future thing then :)

PR #3166, opened 2024-04-07
Fix for reading epoch due to scnlib format changes

GPayne:


modules/space/kepler.cpp
```diff
@@ -266,15 +266,20 @@ namespace {
             epoch.end(),
             [](char c) { return c == '-'; }
         );
-        const std::string format = (nDashes == 2) ? "{:4}-{:2}-{:2}{}" : "{:4}{:2}{:2}{}";
-        auto res = scn::scan<int, int, int, double>(e, scn::runtime_format(format));
+        const std::string format = (nDashes == 2) ? "{:4d}-{:2d}-{}" : "{:4d}{:2d}{}";
+        auto res = scn::scan<int, int, double>(e, scn::runtime_format(format));
```
alexanderbock, line 270:
```suggestion
        auto res = scn::scan<int, int, double>(
            e,
            (nDashes == 2) ? "{:4d}-{:2d}-{}" : "{:4d}{:2d}{}"
        );
```

alexanderbock: This would skip the extra complexity of the `runtime_format`

modules/space/kepler.cpp
```diff
@@ -266,15 +266,20 @@ namespace {
...
         }
-        auto [year, monthNum, dayOfMonthNum, fractionOfDay] = res->values();
+        auto [year, monthNum, dayOfMonth] = res->values();
         const int daysSince2000 = countDays(year);
-        const int wholeDaysInto = daysIntoGivenYear(year, monthNum, dayOfMonthNum);
-        const double daysInYear = static_cast<double>(wholeDaysInto) + fractionOfDay;
+        const int daysInto = daysIntoGivenYear(
+            year,
+            monthNum,
+            static_cast<int>(dayOfMonth)
```
alexanderbock, line 279: Shouldn't `dayOfMonth` already be an `int`?

alexanderbock (approved):

alexanderbock:
To close the issue directly with this PR, you'll need to use the # when referencing the issue number in the title or in the first post

modules/space/kepler.cpp
GPayne: I tried this, but got error:
`the value of ‘nDashes’ is not usable in a constant expression`
so I'll keep it as-is.

modules/space/kepler.cpp
GPayne: I changed it from the previous code which read the integer part first, then the decimal value. `dayOfMonth` now reads the full `double` value from the file as the number of days into the month and the remaining fraction of a day.

Roxeena (commented):
:+1:

PR #3180, opened 2024-04-12
Add examples and documentation for `TimeDependentScale`. Add more strict verifiers

alexanderbock:


engbergandreas (approved):

data/assets/examples/scale/timedependentscale/simple.asset
```diff
@@ -0,0 +1,29 @@
...
+    Scale = {
+      Type = "TimeDependentScale",
+      ReferenceDate = "2000 JAN 01 00:00:00"
+    }
+  },
+  Renderable = {
+    Type = "RenderableCartesianAxes"
+  },
+  GUI = {
+    Name = "TimeDependentScale Example (Simple)",
```
WeirdRubberDuck, line 18: See previous PRs, about removing the word "simple". Still up for discussion though

Just "TimeDependentScale Example" is IMO good enough. And it is fine to have an advanced example as well

modules/base/scale/timedependentscale.cpp
```diff
@@ -60,12 +60,21 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This Scale type provides the ability to scale an object dynamically as time in the
+    // simulation passes. The provided `ReferenceDate`, specifies when the total scale
+    // should be equal to 0 and the scales grows by `Speed` meters for every second in the
+    // simulation. If `ClampToPositive` is specified as `true`, then the scale will only
+    // stay positive. If it is `false` and the simulation time is _before_ the
+    // `ReferenceDate`, the scale value will be negative.
```
WeirdRubberDuck, line 68: What does a negative scale mean here?  Is there a use case for it?  When? 

I'm confused. Maybe provide an example?

WeirdRubberDuck (changes requested):
This gives me errors when starting the default profile: 

image

Also, I get the same errors when trying the created example assets

modules/base/scale/timedependentscale.cpp
alexanderbock: I don't know if there is a use-case for it, but I don't see a reason to prematurely prevent it from happening.  There might be legitimate reasons to, for example, mirror a renderable before the start time.

alexanderbock:
Errors are fixed by extending the allowed strings for the datetime verifier

modules/base/scale/timedependentscale.cpp
WeirdRubberDuck: Ok 👍 I also see that clamp to positive is default, so maybe it's not as confusing as I think it is. 

Looking at the propertyinfo for clamp to positive, that one is confusing without this description and to me feel sort of backwards (says that the scale will never result in negative values rather than that the scale will be clamped to stay positive).

 How about replacing the description there with this text, and shorten this description to something less confusing to a user who just wants to know what time scale does in broad terms without explaining the details of that property? E.g. "Per default, the scale is clamped to positive values"

alexanderbock: Good points.  How about now?

WeirdRubberDuck: Much better!

Roxeena (commented):
:+1:

PR #3181, opened 2024-04-12
Add example and documentation for the StaticScale.  Extract the static scale as a double value instead of a float

alexanderbock:


engbergandreas (approved):

data/assets/examples/scale/staticscale/simple.asset
```diff
@@ -0,0 +1,29 @@
...
+    Scale = {
+      Type = "StaticScale",
+      Scale = 149597870700
+    }
+  },
+  Renderable = {
+    Type = "RenderableCartesianAxes"
+  },
+  GUI = {
+    Name = "StaticScale Example (Simple)",
```
WeirdRubberDuck, line 18: Same comment as for NonUniformStaticScale.  Is "Simple" really necessary here?

WeirdRubberDuck (approved):

modules/base/scale/staticscale.cpp
```diff
@@ -36,9 +36,14 @@ namespace {
         openspace::properties::Property::Visibility::NoviceUser
     };
 
+    // This Scale type scales the scene graph node that it is attached to by a fixed
+    // amount that does not change over time. It is possible to change the fixed scale
+    // after starting the application, but it otherwise remains unchanged. The scaling is
+    // a simple multiplication so that a `Scale` value of 10 means that the object will be
+    // 10 times larger than it should be.
```
ylvaselling, line 43: Rephrase "than it should be" with original size

Roxeena (commented):
:+1:

PR #3185, opened 2024-04-15
Add documentation and example for `RenderableSphereImageOnline`

Roxeena:


modules/base/rendering/renderablesphereimageonline.cpp
```diff
@@ -65,6 +65,9 @@ namespace {
         );
     }
 
+    // A RenderableSphereImageOnline can be used to show an image from an online source
+    // (as a URL) on a sphere in the OpenSpace scene. For more informaiton about a general
```
engbergandreas, line 69:
```suggestion
    // (as a URL) on a sphere in the OpenSpace scene. For more information about a general
```

engbergandreas (changes requested):

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
```diff
@@ -0,0 +1,28 @@
+local Radius = 696340000 * 20 -- The radius of the Sun in meters, times 20
+
+local SimpleSphereOnlineImage = {
+  Identifier = "SimpleSphereOnlineImage",
```
WeirdRubberDuck, line 4:
```suggestion
  Identifier = "RenderableSphereImageOnline_Example",
```
?

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
```diff
@@ -0,0 +1,28 @@
...
+  Renderable = {
+    Type = "RenderableSphereImageOnline",
+    URL = "http://data.openspaceproject.com/examples/renderableplaneimageonline.jpg",
+
+    -- Parameters from the parent class `RenderableSphere`
+    Size = Radius,
+    Segments = 16
+  },
+  GUI = {
+    Name = "Simple Sphere Online Image",
```
WeirdRubberDuck, line 14:
```suggestion
    Name = "Sphere - Online Image",
```
? 
Or do we have a reason to call it "Simple" sphere? 

I also suggest removing the word "simple" from the description

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
```diff
@@ -0,0 +1,28 @@
+local Radius = 696340000 * 20 -- The radius of the Sun in meters, times 20
```
WeirdRubberDuck, line 1: Suggestion to add a summary about the example at the top. Thoughts? :)

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
```diff
@@ -0,0 +1,28 @@
...
+    URL = "http://data.openspaceproject.com/examples/renderableplaneimageonline.jpg",
+
+    -- Parameters from the parent class `RenderableSphere`
+    Size = Radius,
+    Segments = 16
+  },
+  GUI = {
+    Name = "Simple Sphere Online Image",
+    Path = "/Examples",
+    Description = "Simple sphere with an online image of an owl"
```
WeirdRubberDuck, line 16:
```suggestion
    Description = "Sphere using an image from a web URL as texture"
```
? A bit more descriptive, perhaps

WeirdRubberDuck: And maybe bring back the mention of an owl, if you like :)

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
engbergandreas: I think it only makes sense if there are multiple assets where one is using only the required parameters for example. Otherwise, I agree

WeirdRubberDuck: As we discussed during the meeting session - maybe we should discuss this during an upcoming meeting to make sure that we are consistent in naming schemes, etc. 

One suggestion would be to bring it up during our Monday meeting next week. Thoughts @alexanderbock ?

modules/base/rendering/renderablesphereimageonline.cpp
```diff
@@ -65,6 +65,9 @@ namespace {
...
+    // RenderableSphere see [RenderableSphere](#base_renderable_sphere).
```
ylvaselling, line 70: I was wondering if we should mention that the base class RenderableSphere is abstract, so that people don't think it can be instantiated from this comment. I think it is also ok to leave the mentioning about renderable sphere out since we already links to the base classes anyways.

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
alexanderbock: Yea, before merging any of these, let's discuss it on Monday and decide on a format

modules/base/rendering/renderablesphereimageonline.cpp
alexanderbock: We decided to flatten the renderable hierarchy as it is described in the documentation.  So the user of a RenderableSphereImageOnline shouldn't know about the existence of the RenderableSphere abstract class at all.

alexanderbock: The best solution for now would be to do the same as here: https://github.com/OpenSpace/OpenSpace/blob/master/modules/base/rendering/renderableplaneimagelocal.cpp#L67

By adding the `RenderableSphere::Documentation()` as the second argument to the `doc` function, it inserts all of the parameters directly without any need to copy them.

data/assets/examples/renderable/renderablesphereimageonline/simplesphereimageonline.asset
```diff
@@ -0,0 +1,28 @@
+local Radius = 696340000 * 20 -- The radius of the Sun in meters, times 20
+
+local SimpleSphereOnlineImage = {
```
alexanderbock, line 3: One suggestion was to use the `Node` as the name for individual scene graph nodes so that that part looks the same for all?  Thoughts

PR #3191, opened 2024-04-18
Issue/3070 - Outline/border to textured points

### LINKED ISSUE #3070: Add border/outline to textured point clouds
I imagine two versions: 
- Just a border at the bottom
- enforce a circular shape and use the same border as the regular, untextured, points

With two coloring modes 🎨 
- properties for outline color and width
- from color map (and then do not multiply the color with )

WeirdRubberDuck:
Adds border in three different styles and makes it possible to apply color map på border rather than the point body. Closes #3070 

Round, Square, Bottom (with texture and color map to border instead of texture)
image
image
image

And for points without texture (and not using color map, only the previously existing border color)
image
image
image

Notice how the points change from round to square when Square or Bottom border style is chosen

### DIFF
--- a/modules/base/rendering/pointcloud/renderablepointcloud.cpp
+++ b/modules/base/rendering/pointcloud/renderablepointcloud.cpp
@@ -327,8 +327,9 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo OutlineStyleInfo = {
         "OutlineStyle",
         "Outline Style",
-        "This setting decides the shape of points that have an outline (Round, Square, "
-        "or lust a line at the bottom).",
+        "This setting decides the shape of points that have an outline (round, square, "
+        "or lust a line at the bottom). Note that anything but \"Round\" will lead to "
+        "the point being drawed as squares.",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
--- a/modules/base/rendering/pointcloud/renderablepointcloud.cpp
+++ b/modules/base/rendering/pointcloud/renderablepointcloud.cpp
@@ -337,8 +337,8 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo ApplyColorMapToOutlineInfo = {
         "ApplyColorMapToOutline",
         "Apply Color Map to Outline",
-        "If true and outline is enabled, the color map will be applied to the outline "
-        "rather than the point body. Only works if color mapping is enabled.",
+        "If true and the outline is enabled, the color map will be applied to the "
+        "outline rather than the point body. Only works if color mapping is enabled.",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
@@ -488,9 +488,9 @@ namespace {
             std::optional<float> outlineWeight;
 
             enum class [[codegen::map(OutlineStyle)]] OutlineStyle {
-                Round [[codegen::key("Round")]],
-                Square [[codegen::key("Square")]],
-                Bottom [[codegen::key("Bottom")]]
+                Round,
+                Square,
+                Bottom
             };
             // [[codegen::verbatim(OutlineStyleInfo.description)]]
             std::optional<OutlineStyle> outlineStyle;
@@ -598,8 +598,8 @@ RenderablePointCloud::ColorSettings::ColorSettings(const ghoul::Dictionary& dict
         if (settings.colorMapping.has_value()) {
             colorMapping = std::make_unique<ColorMappingComponent>(
                 *settings.colorMapping
-            );
-            addPropertySubOwner(colorMapping.get());
+           );
+           addPropertySubOwner(colorMapping.get());
         }
 
         enableOutline = p.coloring->enableOutline.value_or(enableOutline);
--- a/modules/base/rendering/pointcloud/renderablepointcloud.cpp
+++ b/modules/base/rendering/pointcloud/renderablepointcloud.cpp
@@ -58,7 +58,7 @@
 namespace {
     constexpr std::string_view _loggerCat = "RenderablePointCloud";
 
-    constexpr std::array<const char*, 36> UniformNames = {
+    constexpr std::array<const char*, 37> UniformNames = {
         "cameraViewMatrix", "projectionMatrix", "modelMatrix", "cameraPosition",
         "cameraLookUp", "renderOption", "maxAngularSize", "color", "opacity",
         "scaleExponent", "scaleFactor", "up", "right", "fadeInValue", "hasSpriteTexture",
@@ -66,7 +66,8 @@ namespace {
         "nanColor", "useNanColor", "hideOutsideRange", "enableMaxSizeControl",
         "aboveRangeColor", "useAboveRangeColor", "belowRangeColor", "useBelowRangeColor",
         "hasDvarScaling", "dvarScaleFactor", "enableOutline", "outlineColor",
-        "outlineWeight", "outlineStyle", "aspectRatioScale", "useOrientationData"
+        "outlineWeight", "outlineStyle", "useCmapOutline", "aspectRatioScale",
+        "useOrientationData"
     };
 
     enum RenderOption {
@@ -333,6 +334,14 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo ApplyColorMapToOutlineInfo = {
+        "ApplyColorMapToOutline",
+        "Apply Color Map to Outline",
+        "If true and outline is enabled, the color map will be applied to the outline "
+        "rather than the point body. Only works if color mapping is enabled.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
     // A RenderablePointCloud can be used to render point-based datasets in 3D space,
     // optionally including color mapping, a sprite texture and labels. There are several
     // properties that affect the visuals of the points, such as settings for scaling,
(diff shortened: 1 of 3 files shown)


WeirdRubberDuck:
Not a fan of all the nasty if-statements in the shader (quite hastily implemented), but it works and the path should not change between rendering two points.

alexanderbock:
When selecting the "Bottom" outline style, I would expect an Outline Weight of 1 to cover the entire image, the same as with the Circle and Square.  But only half of the image is covered

image

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -318,6 +325,23 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo OutlineStyleInfo = {
+        "OutlineStyle",
+        "Outline Style",
+        "This setting decides the shape of points that have an outline (round, square, "
+        "or lust a line at the bottom). Note that anything but \"Round\" will lead to "
```
alexanderbock, line 332:
```suggestion
        "or just a line at the bottom). Note that anything but \"Round\" will lead to "
```

I'd also turn around the "Round" sentence.  I would expect that by default the images are rendered as squares, so instead:

"Note that selecting \"Round\" will lead to the points being drawn as ellipses"

or something like that

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -318,6 +325,23 @@ namespace {
...
+        "the point being drawed as squares.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ApplyColorMapToOutlineInfo = {
+        "ApplyColorMapToOutline",
+        "Apply Color Map to Outline",
+        "If true and outline is enabled, the color map will be applied to the outline "
+        "rather than the point body. Only works if color mapping is enabled.",
```
alexanderbock, line 341:
```suggestion
        "If true and the outline is enabled, the color map will be applied to the "
        "outline rather than the point body. Only works if color mapping is enabled.",
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -462,6 +486,17 @@ namespace {
 
             // [[codegen::verbatim(OutlineColorInfo.description)]]
             std::optional<float> outlineWeight;
+
+            enum class [[codegen::map(OutlineStyle)]] OutlineStyle {
+                Round [[codegen::key("Round")]],
+                Square [[codegen::key("Square")]],
+                Bottom [[codegen::key("Bottom")]]
```
alexanderbock, line 493: No need for the `codegen::key` as they already have the correct key

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -543,23 +598,22 @@ RenderablePointCloud::ColorSettings::ColorSettings(const ghoul::Dictionary& dict
         if (settings.colorMapping.has_value()) {
             colorMapping = std::make_unique<ColorMappingComponent>(
                 *settings.colorMapping
-           );
-           addPropertySubOwner(colorMapping.get());
+            );
+            addPropertySubOwner(colorMapping.get());
```
alexanderbock, line 602: I think this was unintentional

alexanderbock (approved):

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -318,6 +325,23 @@ namespace {
...
+        "This setting decides the shape of points that have an outline (round, square, "
+        "or lust a line at the bottom). Note that anything but \"Round\" will lead to "
+        "the point being drawed as squares.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ApplyColorMapToOutlineInfo = {
+        "ApplyColorMapToOutline",
+        "Apply Color Map to Outline",
+        "If true and outline is enabled, the color map will be applied to the outline "
```
Roxeena, line 340:
```suggestion
        "If true, and outline is enabled and the color map will be applied to the outline "
```

Roxeena: Feel free to choose Alex's suggestion though

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -532,9 +567,29 @@ RenderablePointCloud::ColorSettings::ColorSettings(const ghoul::Dictionary& dict
...
+
+    outlineColor.setViewOption(properties::Property::ViewOptions::Color);
+    addProperty(outlineColor);
+
+    addProperty(outlineWeight);
+
+    outlineStyle.addOption(OutlineStyle::Round, "Round");
+    outlineStyle.addOption(OutlineStyle::Square, "Square");
+    outlineStyle.addOption(OutlineStyle::Bottom, "Bottom");
+    outlineStyle = OutlineStyle::Round;
```
Roxeena, line 588: In Alex's previous comment he suggested that the default should be square, should we change that here?

modules/base/shaders/pointcloud/pointcloud_fs.glsl
```diff
@@ -85,21 +92,52 @@ Fragment getFragment() {
...
   // Moving the origin to the center and calculating the length
-  float lengthFromCenter = length((texCoord - vec2(0.5)) * 2.0);
-  if (!hasSpriteTexture && (lengthFromCenter > 1.0)) {
+  vec2 centeredTexCoords = (texCoord - vec2(0.5)) * 2.0;
+  float lengthFromCenter = length(centeredTexCoords);
+
+  bool shouldBeRound = (!hasSpriteTexture && !enableOutline) ||
+    (enableOutline && outlineStyle == OutlineStyleRound);
+
+  if (shouldBeRound && (lengthFromCenter > 1.0)) {
```
Roxeena, line 101: I am not sure but it feels like this should be reversed? If the distance to the center is **smaller** than 1 it should be discarded. Or am I wrong?

modules/base/shaders/pointcloud/pointcloud_fs.glsl
```diff
@@ -85,21 +92,52 @@ Fragment getFragment() {
...
     discard;
   }
 
   vec4 fullColor = vec4(color, 1.0);
+  vec4 cmapColor = vec4(1.0);
   if (useColorMap) {
-    fullColor = sampleColorMap(gs_colorParameter);
+    cmapColor = sampleColorMap(gs_colorParameter);
+    if (!useCmapOutline) {
```
Roxeena, line 109: This also feels reversed, shouldn't the color map be used when the property is enabled?

Roxeena (commented):
:+1: Looks great and tested it out as well. It would be nice with a new example asset that showcases this.

modules/base/shaders/pointcloud/pointcloud_fs.glsl
WeirdRubberDuck: We want to discard all pixels outside the radius == 1, to get a circular point.  So, yes, you're wrong :)

modules/base/shaders/pointcloud/pointcloud_fs.glsl
WeirdRubberDuck: I understand that this can seem confusing. However, the color map color should only be the base color for the point if it isn't applied on the border/outline.  So this is correct

modules/base/rendering/pointcloud/renderablepointcloud.cpp
WeirdRubberDuck: Yep!

WeirdRubberDuck, line 602:
```suggestion
           );
           addPropertySubOwner(colorMapping.get());
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
WeirdRubberDuck, line 493:
```suggestion
                Round,
                Square,
                Bottom
```

WeirdRubberDuck:
```suggestion
                Round,
                Square,
                Bottom
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
WeirdRubberDuck: This gets complicated as the border is not only applied to textured points, but the regular ones as well (which was een supported before this PR).  In that case, I would expect the default value to be "Round" 

And in my opinion, since we are drawing what we call "points", it kind of makes sense to me that the default value always is round.   But if we want to change the default value for when textures are used to be square, that would also be ok

modules/base/rendering/pointcloud/renderablepointcloud.cpp
WeirdRubberDuck: See reply in Alex's previous comment.  My intuition tells me that the default value should always be round rather than square

modules/base/rendering/pointcloud/renderablepointcloud.cpp
WeirdRubberDuck: Need help deciding what to do here

WeirdRubberDuck:
> When selecting the "Bottom" outline style, I would expect an Outline Weight of 1 to cover the entire image, the same as with the Circle and Square. But only half of the image is covered
> 
> image

I know, but this would also result in the border being twice as thick compared to the "Squared" style when using the "Bottom" style, which I found much more annoying than the fact that a value of 1 does not cover the entire image (which to me seems like an unlikely use case that the user would not want to do anyways, so why make that behavior more intuitive?)

Thoughts?

alexanderbock:
Merged it for now.  More fixed tbh

PR #3201, opened 2024-04-19
Add renderableorbitdisc example asset

engbergandreas:


data/assets/examples/renderable/renderableorbitdisc/renderableorbitdisc.asset
```diff
@@ -0,0 +1,22 @@
+local OrbitDisc = {
+    Identifier = "RenderableOrbitDisc_Example",
```
alexanderbock, line 2: We're doing 2 space indentation in asset files

data/assets/examples/renderable/renderableorbitdisc/renderableorbitdisc.asset
```diff
@@ -0,0 +1,22 @@
...
+    Renderable = {
+      Type = "RenderableOrbitDisc",
+      Eccentricity = 0.2,
+      Size = 10000,
+      Offset = { 0.1, 0.7 },
+      Texture = asset.resource("texture.png"),
```
alexanderbock, line 8: I uploaded the texture to data.openspaceproject.com/examples/barber_pole.png.  I think in general we should prevent having binary files in the folders as much as possible.  In this case, it's only 137 bytes, but still?

modules/exoplanets/rendering/renderableorbitdisc.cpp
```diff
@@ -95,7 +95,7 @@ namespace {
         float size;
 
         // [[codegen::verbatim(EccentricityInfo.description)]]
-        float eccentricity;
+        float eccentricity [[codegen::inrange(0.0, 0.99)]];
```
alexanderbock, line 98: `[[codegen::less(1.0)]]`?  Or we should add a bunch of more 9s here to enable as much eccentric orbit as possible

modules/exoplanets/rendering/renderableorbitdisc.cpp
```diff
@@ -116,8 +116,8 @@ RenderableOrbitDisc::RenderableOrbitDisc(const ghoul::Dictionary& dictionary)
     : Renderable(dictionary)
     , _texturePath(TextureInfo)
     , _size(SizeInfo, 1.f, 0.f, 3.0e12f)
-    , _eccentricity(EccentricityInfo, 0.f, 0.f, 1.f)
-    , _offset(OffsetInfo, glm::vec2(0.f), glm::vec2(0.f), glm::vec2(1.f))
+    , _eccentricity(EccentricityInfo, 0.f, 0.f, 0.99f)
```
alexanderbock, line 119: Ditto here

PR #3216, opened 2024-04-23
Update documentation for `LightSource`

micahnyc:
added current types to documentation

### DIFF
--- a/src/scene/lightsource.cpp
+++ b/src/scene/lightsource.cpp
@@ -44,8 +44,9 @@ namespace {
 
     struct [[codegen::Dictionary(LightSource)]] Parameters {
         // The type of the light source that is described in this element. The available
-        // types of light sources depend on the configuration of the application and can
-        // be written to disk on application startup into the FactoryDocumentation
+        // types are 'SceneGraphLightSource' and 'CameraLightSource'. If OpenSpace was
+        // compiled manually, these types may depend on the configuration of the application
+        // and will be written to disk on application startup into the FactoryDocumentation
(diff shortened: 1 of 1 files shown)


src/scene/lightsource.cpp
```diff
@@ -44,8 +44,9 @@ namespace {
 
     struct [[codegen::Dictionary(LightSource)]] Parameters {
         // The type of the light source that is described in this element. The available
-        // types of light sources depend on the configuration of the application and can
-        // be written to disk on application startup into the FactoryDocumentation
+        // types are 'SceneGraphLightSource' and 'CameraLightSource'. If OpenSpace was
```
ylvaselling, line 47: Sorry in advance for long comment 🙏 

I'm a little hesitant to describe the types here... if we add another type later, we have to remember to manually edit this description. Also pretty much all the other categories follow the same paradigm where you specify the type, so if we put in all the types here we would also (presumably) put in all the types in the Renderable base class, the Scale base class, and so on, which will be a lot of manual maintenance. Maybe it is better to teach users about this paradigm somewhere instead of specifying the particular types of each category? Or alternatively we could generate this list in the documentation for each base class, if it is not already clear in the TOC-tree in the sidebar?

On another note, I think the "factory documentation" and the "compiling OpenSpace manually" are two phrasings that I think a new user would get a bit scared by. We don't use the word factory documentation anywhere else, except in the source code, so I think that one we could leave out entirely. As for compiling OpenSpace manually, it might be better to put that kind of information somewhere else, for example in the development section. I think most users will not be people who compile OpenSpace :)

PR #3217, opened 2024-04-24
Make use of the new uniform cache method that uses the variable names as uniform names

alexanderbock:
So far, we needed to separately keep two lists for UniformCaches. One that is the `UniformCache` definition in the header file that has the variable names into which the uniform locations are stored. The second is the array of uniform names that live in the cpp file.  Both lists are sensitive to reordering and changing the order of a variable would lead to subtle and hard to detect errors. Furthermore, we had issues in the past where forgetting a `,` in the uniform name list caused hard-to-detect errors.

This PR uses an updated UniformCache that automatically uses the variables names of the definition as the uniform names, thus saving us from duplicating.

The thing to test in here is that the UniformNames manually specified in the source files must match the variables names specified in the header file.  In the cases where that wasn't true, I renamed them in the UniformCache where possible, with one exception where the naming of the uniform was confusing.

# Before
```cpp
// Header file
    UniformCache(modelViewProjection, modelViewTransform, colorTexture, opacity,
        mirrorBackside, multiplyColor) _uniformCache;

// Source file
    constexpr std::array<const char*, 6> UniformNames = {
        "modelViewProjection", "modelViewTransform", "colorTexture", "opacity",
        "mirrorBackside", "multiplyColor"
    };

    ghoul::opengl::updateUniformLocations(*_shader, _uniformCache, UniformNames);
```

# After
```cpp
// Header file
    UniformCache(modelViewProjection, modelViewTransform, colorTexture, opacity,
        mirrorBackside, multiplyColor) _uniformCache;

// Source file
    ghoul::opengl::updateUniformLocations(*_shader, _uniformCache);
```


modules/spacecraftinstruments/rendering/renderableplanetprojection.h
```diff
@@ -79,12 +79,12 @@ class RenderablePlanetProjection : public Renderable {
...
-    UniformCache(sunPos, modelTransform, modelViewProjectionTransform, hasBaseMap,
+    UniformCache(sun_pos, modelTransform, modelViewProjectionTransform, hasBaseMap,
         hasHeightMap, heightExaggeration, meridianShift, ambientBrightness,
-        projectionFading, baseTexture, projectionTexture, heightTexture)
-        _mainUniformCache;
+        projectionFading, baseTexture, projectionTexture,
+        heightTexture) _mainUniformCache;
 
-    UniformCache(projectionTexture, projectorMatrix, modelTransform, boresight,
+    UniformCache(projectionTexture, ProjectorMatrix, ModelTransform, boresight,
```
ylvaselling, line 87: Why are the first letters in caps?

ext/ghoul
ylvaselling, line 1: I think there is some inconsistencies whether the texture parameter is called tex or texture in different files?

modules/spacecraftinstruments/rendering/renderableplanetprojection.h
alexanderbock: Same for the comment below as well.

I didn't want to introduce any additional bugs if necessary, so for those cases where the variable name in the UniformCache (in this case `projectorMatrix`) disagreed with the uniform name in the shader (in this case `ProjectorMatrix`), I chose to replace the variable name, is a mismatch would cause a compiler error, while changing the shader code to have a better uniform name was a more brittle change.

I'd see going through the individual shaders and cleaning them up as a bigger change that we'd want to more thoroughly test

PR #3220, opened 2024-04-26
Add documentation and example for the `TileProviderByIndex`

alexanderbock:


modules/globebrowsing/src/tileprovider/tileproviderbyindex.cpp
```diff
@@ -67,21 +85,30 @@ documentation::Documentation TileProviderByIndex::Documentation() {
 TileProviderByIndex::TileProviderByIndex(const ghoul::Dictionary& dictionary) {
     ZoneScoped;
 
-    const Parameters p = codegen::bake<Parameters>(dictionary);
+    Parameters p = codegen::bake<Parameters>(dictionary);
+
+    // For now we need to impute the LayerGroupID this way. We don't want it to be part of
```
ylvaselling, line 90: You have a great vocabulary! "impute" 😮 Maybe use a synonym, like represent or assigne? I had to google 😅

modules/globebrowsing/src/tileprovider/tileproviderbyindex.cpp
```diff
@@ -67,21 +85,30 @@ documentation::Documentation TileProviderByIndex::Documentation() {
...
+    // the parameters struct as that would mean it would be visible to the end user, which
+    // we don't want since this value just comes from whoever creates us, not the user
```
ylvaselling, line 92:
```suggestion
    // we don't want since this value just comes from whoever creates it, not the user
```

Is that what you mean?

modules/globebrowsing/src/tileprovider/tileproviderbyindex.cpp
```diff
@@ -67,21 +85,30 @@ documentation::Documentation TileProviderByIndex::Documentation() {
...
+        const std::string type = p.defaultTileProvider.value<std::string>("Type");
         typeID = ghoul::from_string<layers::Layer::ID>(type);
     }
 
-    _defaultTileProvider = createFromDictionary(typeID, p.defaultProvider);
+    p.defaultTileProvider.setValue("LayerGroupID", static_cast<int>(group));
+    _defaultTileProvider = createFromDictionary(typeID, p.defaultTileProvider);
 
-    for (const Parameters::IndexProvider& ip : p.indexTileProviders) {
+    for (Parameters::IndexProvider& ip : p.tileProviders) {
```
ylvaselling, line 107: Why remove the const?

modules/globebrowsing/src/tileprovider/tileproviderbyindex.cpp
```diff
@@ -139,7 +171,16 @@ int TileProviderByIndex::minLevel() {
...
 
 int TileProviderByIndex::maxLevel() {
-    return _defaultTileProvider->maxLevel();
+    int result = _defaultTileProvider->maxLevel();
+
+    using K = TileIndex::TileHashKey;
+    using V = std::unique_ptr<TileProvider>;
+    for (std::pair<const K, V>& it : _providers) {
+        TileIndex index = TileIndex(it.first);
+        result = std::max(result, static_cast<int>(index.level));
```
ylvaselling, line 180: Does this set the max level to the maximum of all the tile providers? Won't that mean that the other tile providers will go out of range?

modules/globebrowsing/src/tileprovider/tileproviderbylevel.cpp
```diff
@@ -50,7 +48,13 @@ TileProviderByLevel::TileProviderByLevel(const ghoul::Dictionary& dictionary) {
 
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
-    const layers::Group::ID group = static_cast<layers::Group::ID>(p.layerGroupID);
+    // For now we need to impute the LayerGroupID this way. We don't want it to be part of
```
ylvaselling, line 51: See above:)

modules/globebrowsing/src/tileprovider/tileproviderbylevel.cpp
```diff
@@ -50,7 +48,13 @@ TileProviderByLevel::TileProviderByLevel(const ghoul::Dictionary& dictionary) {
...
+    // the parameters struct as that would mean it would be visible to the end user, which
+    // we don't want since this value just comes from whoever creates us, not the user
```
ylvaselling, line 53:
```suggestion
    // we don't want since this value just comes from whoever creates it, not the user
```

modules/globebrowsing/src/tileprovider/tileproviderbylevel.cpp
ylvaselling, line 51:
```suggestion
    // For now we need to represent the LayerGroupID this way. We don't want it to be part of
```

data/assets/examples/tileprovider/tileproviderbyindex/example.asset
```diff
@@ -0,0 +1,58 @@
...
+    Path = "/Examples/TileProviderByIndex"
+  }
+}
+
+asset.onInitialize(function()
+  openspace.addSceneGraphNode(Node)
+end)
+
+asset.onDeinitialize(function()
+  openspace.removeSceneGraphNode(Node)
```
ylvaselling, line 57: I tried the example and I think it would be nice if you could go a bit closer to the globe, is that possible? Now the globe is max 20% of my vertical fov, I suppose because of the interaction sphere or something like that. It makes it a bit difficult to see the tiles 
image

modules/globebrowsing/src/tileprovider/tileproviderbyindex.cpp
alexanderbock: But it's such a great word :'(
image

I changed it to "inject"

modules/globebrowsing/src/tileprovider/tileproviderbyindex.cpp
alexanderbock: Yes, but that is fine and already supported by the GlobeBrowsing subsystem.   If a layer is asked to generate a tile that is higher resolution that it is capable of, it will reuse the last possible tile and create smaller and smaller views on it by manipulating the uv coordinates

modules/globebrowsing/src/layer.cpp
```diff
@@ -174,10 +174,6 @@ namespace {
         // Sets the blend mode of this layer to determine how it interacts with other
         // layers on top of this
         std::optional<BlendMode> blendMode;
-
-        // If the primary layer creation fails, this layer is used as a fallback
-        std::optional<ghoul::Dictionary>
```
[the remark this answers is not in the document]
alexanderbock: Oh.. I wanted to do this in a separate commit. The whole "fallback" concept was a thing in the past was a dead parameter for a few years now

PR #3243, opened 2024-05-03
Update Voyager Model, closes #3035

### LINKED ISSUE #3035: Voyager model has box in middle of it
Screenshot 2024-02-20 at 4 16 10 PM

Roxeena:
The "box" in the model was baked into the obj model files, so the solution for this issue is to update the model to be in one piece instead of in many smaller pieces that we used to have.

The new model needs to be uploaded to our servers: 
[Voyager.zip](https://github.com/OpenSpace/OpenSpace/files/15197904/Voyager.zip)

Something to discuss could be the scale and size of the spacecraft. According to a quick search, the spacecraft is rather small (about 2.5 meters with the antenna). The model, right now, looks to be much larger than that but it does match up with the model we previously had with the "box" inside it. Do we want to update it to be in the correct size? Then we would have a similar issue as with another model in that it is too small to get up close to it without turning off or lowering the zoom limit.


ylvaselling:
How can we test this pr?

ylvaselling:
Thank you, I got it working! Looks great! 🎉

alexanderbock (approved):
Data server is updated, too

PR #3246, opened 2024-05-07
Add example for NavigationState

GPayne:


data/assets/examples/navigationstate.asset
```diff
@@ -0,0 +1,34 @@
+-- Basic
+-- Example of using getNavigationState and setNavigationState to manipulate
```
ylvaselling, line 2: Maybe mention that this is an action, so users know where to look for it

data/assets/examples/navigationstate.asset
```diff
@@ -0,0 +1,34 @@
...
+      Pitch = currentNavState["Pitch"],
+      Anchor = currentNavState["Anchor"],
+      Yaw = currentNavState["Yaw"],
+      Position = currentNavState["Position"],
+      Up = {0.0, 0.0, -1.0}
+    }
+    -- Set the navigation state with the new navigation state variable
+    openspace.navigation.setNavigationState(newNavState) 
+  ]],
+  Documentation = "Flips the camera upside-down.",
```
ylvaselling, line 23: I know for scene graph nodes we said to not have any descriptions but maybe for actions we want to?

ylvaselling (approved):

data/assets/examples/navigationstate.asset
GPayne: I added to the comment to say this is an action

alexanderbock (approved):

Issue #1902, opened 2022-03-04
Atmosphere/Globe interaction issue


micahnyc:
Warning bad description, just dont want to miss this issue. TO BE FILLED OUT MORE

-----------
When enabling perform shading, the contrast gets shrunk, then the atmosphere stretches it, and we get an extra loss of detail.

PR #3250, opened 2024-05-10
Fix faulty computation of sun position in atmosphere

WeirdRubberDuck:
Fixes a problem where the light source direction was erroneously computed for the atmosphere. It was not noticeable in our solar system but led to errors when adding an atmosphere to an exoplanet system. 

Reported in support slack
https://openspacesupport.slack.com/archives/C04TY4ZKXV2/p1714145171752719

To test it, you can try loading the assets below, that add an exoplanet system with an atmosphere
[kepler36.zip](https://github.com/OpenSpace/OpenSpace/files/15272781/kepler36.zip)

### DIFF
--- a/modules/atmosphere/rendering/atmospheredeferredcaster.cpp
+++ b/modules/atmosphere/rendering/atmospheredeferredcaster.cpp
@@ -381,13 +381,11 @@ void AtmosphereDeferredcaster::preRaycast(const RenderData& data, const Deferred
         const glm::dvec3 sunPosWorld = node ? node->worldPosition() : glm::dvec3(0.0);
 
         glm::dvec3 sunPosObj;
-        // Sun following camera position
         if (_sunFollowingCameraEnabled) {
-            sunPosObj = invModelMatrix * glm::dvec4(data.camera.eyePositionVec3(), 1.0);
+            sunPosObj = camPosObjCoords;
         }
         else {
-            sunPosObj = invModelMatrix *
-                glm::dvec4((sunPosWorld - data.modelTransform.translation) * 1000.0, 1.0);
+            sunPosObj = invModelMatrix * glm::dvec4(sunPosWorld, 1.0);
         }
 
         // Sun Position in Object Space


ylvaselling:
Looks good! 

Related but not about this issue (I assume) - when testing I noticed some odd behavior from the atmosphere: the sun popping up without fading and the ground being sort of "cut off" from the atmosphere (see video). I didn't find an issue for it except for [this ](https://github.com/OpenSpace/OpenSpace/issues/1902)one . Do you know anything about this / thoughts? I was thinking we should have an issue for it. 

https://github.com/OpenSpace/OpenSpace/assets/24265984/ed309ac1-de93-4f70-a738-8097fb6b01a2

WeirdRubberDuck:
> Looks great! 🎉
> 
> Related but not about this issue (I assume) - when testing I noticed some odd behavior from the atmosphere: the sun popping up without fading and the ground being sort of "cut off" from the atmosphere (see video). I didn't find an issue for it except for [this](https://github.com/OpenSpace/OpenSpace/issues/1902) one . Do you know anything about this / thoughts? I was thinking we should have an issue for it.
> 
> Also not really this issue, but when you enable global illumination, the sun changes its position in the atmosphere (understandably), but it looks a bit funky. Maybe we should have an issue for not rendering the circle of the sun in the atmosphere when global illumination is enabled?

Feel free to go ahead and create issues for those things @ylvaselling. Could be something for @engbergandreas to look into as part of "looking at the atmosphere code" :) 

But I'll consider them separate from this PR.

alexanderbock:
Yea, it's a different problem as we right now need to specify the angular size of the Sun for each atmosphere. For this example, the angular size is too small and should be adjusted in the asset.   Not sure if we want to automatically calculate it as it will require a dependency chain on the planet scene graph node as well as the Sun (not just the sun position)

alexanderbock (approved):

alexanderbock:
Looks great :+1:

PR #3259, opened 2024-05-14
Added extra file extensions for spice kernels

micahnyc:
I got some spice kernels from someone where the extension was .spk 

When I tried to use them, openspace didn't find positions for the object, however when I used brief.exe and spicepy, I did get values. Turns out we only will read positions if the file name ends in .bsp, so I thought since I had to dig into the code to figure this out, maybe we could help any users that this happens to in the future.

### DIFF
--- a/src/util/spicemanager.cpp
+++ b/src/util/spicemanager.cpp
@@ -261,11 +261,17 @@ SpiceManager::KernelHandle SpiceManager::loadKernel(std::filesystem::path filePa
     }
 
     const std::filesystem::path fileExtension = filePath.extension();
-    if (fileExtension == ".bc" || fileExtension == ".BC") {
-        findCkCoverage(filePath); // binary ck kernel
+    if (fileExtension == ".bc" ||
+        fileExtension == ".BC" ||
+        fileExtension == ".ck" ||
+        fileExtension == ".CK") {
+            findCkCoverage(filePath); // binary ck kernel
     }
-    else if (fileExtension == ".bsp" || fileExtension == ".BSP") {
-        findSpkCoverage(filePath); // binary spk kernel
+    else if (fileExtension == ".bsp" ||
+            fileExtension == ".BSP" ||
+            fileExtension == ".spk" ||
+            fileExtension == ".SPK") {
+            findSpkCoverage(filePath); // spk kernel
     }
 
     const KernelHandle kernelId = ++_lastAssignedKernel;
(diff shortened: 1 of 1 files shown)


ylvaselling (approved):

ylvaselling:
Should we test this pr? If yes, could you please provide the spice kernel and instructions on how to use them?

src/util/spicemanager.cpp
```diff
@@ -261,11 +261,17 @@ SpiceManager::KernelHandle SpiceManager::loadKernel(std::filesystem::path filePa
     }
 
     const std::filesystem::path fileExtension = filePath.extension();
-    if (fileExtension == ".bc" || fileExtension == ".BC") {
-        findCkCoverage(filePath); // binary ck kernel
+    if (fileExtension == ".bc" ||
+        fileExtension == ".BC" ||
+        fileExtension == ".ck" ||
+        fileExtension == ".CK") {
+            findCkCoverage(filePath); // binary ck kernel
```
alexanderbock, line 268:
```suggestion
        fileExtension == ".CK")
    {
        findCkCoverage(filePath); // binary ck kernel
```

src/util/spicemanager.cpp
```diff
@@ -261,11 +261,17 @@ SpiceManager::KernelHandle SpiceManager::loadKernel(std::filesystem::path filePa
...
     }
-    else if (fileExtension == ".bsp" || fileExtension == ".BSP") {
-        findSpkCoverage(filePath); // binary spk kernel
+    else if (fileExtension == ".bsp" ||
+            fileExtension == ".BSP" ||
+            fileExtension == ".spk" ||
+            fileExtension == ".SPK") {
+                findSpkCoverage(filePath); // spk kernel
```
alexanderbock, line 274:
```suggestion
            fileExtension == ".SPK")
        {
            findSpkCoverage(filePath); // spk kernel
```

micahnyc:
[MAUNAKEA.zip](https://github.com/OpenSpace/OpenSpace/files/15393943/MAUNAKEA.zip)

Here is a asset and kernels that have the .spk extension

micahnyc:
Yup thats good, its giving the position for MAUNAKEA (the altitude needs to be adjusted based on the heightmap we load vs the one that was used by the person who gave me the kernel, or an offset can be applied to our heightmap).

alexanderbock (approved):

Issue #3257, opened 2024-05-14
Drag-and-drop an .mp4 file would sometimes crash OpenSpace


engbergandreas:
This is the file used, will see if I can find the crashlog. Worked on some of our workshop machines but not on others, very 50/50
https://github.com/OpenSpace/OpenSpace/assets/48772850/8ba13bc6-4f31-4618-969d-cae113f20116

alexanderbock:
Maybe duplicate of #3163

PR #3261, opened 2024-05-16
Update post synchronization from range-based to index-based for loop closes #3163

### LINKED ISSUE #3163: Read access violation when drag-and-dropping a video
**Describe the bug**
In debug mode, when you drag and drop a video .mp4 file, OpenSpace throws an exception due to a bad memory access. In release mode this bug probably happens but it is silent, i.e. nothing happens. Tested on master 2024-04-04. There seems to be some problem with the syncables.

**To Reproduce**
1. Drag and drop a video file into OpenSpace

**Expected behavior**
No exception.

**Screenshots**
image
image

**Desktop**
 - OS: Windows 11 Home
 - Version 10.0.22631

engbergandreas:
closes #3163 and #3257 

The issue is that as we are iterating the syncables in post-synchronization the video player adds a new `syncable` to the `syncEngine` which invalidates the range-based for loop due to vector capacity change. Either we go with this solution of changing it to an index-based for loop (should probably change it in e.g., pre-synchronization as well?) or we add the `syncable` to some queue which we process at the next frame for example.

### DIFF
--- a/src/engine/syncengine.cpp
+++ b/src/engine/syncengine.cpp
@@ -69,8 +69,8 @@ void SyncEngine::preSynchronization(IsMaster isMaster) {
 void SyncEngine::postSynchronization(IsMaster isMaster) {
     ZoneScoped;
 
-    for (Syncable* syncable : _syncables) {
-        syncable->postSync(isMaster);
+    for (size_t i = 0; i < _syncables.size(); i++) {
+        _syncables[i]->postSync(isMaster);
     }
 }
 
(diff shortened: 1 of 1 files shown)


alexanderbock (approved):
For symmetry it would be good to do it in the presync aswell.  I'd also say to add a comment here explaining why we use a raw loop instead or else there is a danger of someone changing it back later and reintroducing the bug

ylvaselling (approved):

PR #3262, opened 2024-05-16
Automatically create bookmarks folder for user 

### LINKED ISSUE #3229: User folder should create bookmarks folder for users
Was with a user who was on v20 last night. We had to move their bookmarks file, but it wasn't clear where to put it. Then once we found out, we had a typo in the folder name we create so it still wanst working

Now that we have bookmarks in version 20 living in the user folder, we should create the /user/bookmarks folder for them, this will stop them from making any typos in the name when creating it themselves. In addition, when we refer to the folder in our documentation, it will exist even if empty for when users go to put something in there. 

It seems like we tried to do that here: https://github.com/OpenSpace/OpenSpace/blob/c09e6da4057798ae4d377c132994921d54fd0023/data/assets/global/localbookmarks.asset#L10 

but maybe its not creating the folder?

engbergandreas:
There is an option to also create all the directories in the specified path but perhaps that should be its own lua function? 
The `createDirectory` requires that the parent directory exists

closes #3229

### DIFF
--- a/data/assets/global/localbookmarks.asset
+++ b/data/assets/global/localbookmarks.asset
@@ -5,6 +5,11 @@ local bookmarkHelper = asset.require("util/generate_bookmarks")
 asset.require("scene/solarsystem/planets/earth/earth")
 
 local localBookmarks = openspace.absPath("${USER}/bookmarks/localbookmarks.csv")
+local bookmarksDirectory = openspace.absPath("${USER}/bookmarks")
+
+if not openspace.directoryExists(bookmarksDirectory) then
+  openspace.createDirectory(bookmarksDirectory)
+end
 
 -- Create bookmarks file if it does not exist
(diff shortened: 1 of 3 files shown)


alexanderbock:
Normally, those would be the same function with an optional parameter to determine if the directories recursively (for example: https://nodejs.org/api/fs.html#fsmkdirsyncpath-options)  I think by default that parameter should be `false` though

alexanderbock (approved):

micahnyc:
this works, but I figured it would happen in this block where we already have a lot of this stuff: https://github.com/OpenSpace/OpenSpace/blob/c3657921d2c2e85a6738057c08f9f2ce300a3c3a/src/engine/openspaceengine.cpp#L887

ylvaselling (approved):

PR #3270, opened 2024-05-20
Add profile information to the visual test files

### LINKED ISSUE #197: Automatic regression testing with rendered images
Having a separate repository with gold-standard images.  With each compile, Windows, Linux, and Mac machines will run OpenSpace test examples and compare them to the gold standard images.

alexanderbock:
Instead of taking the folder name as profile, the folder is now purely for organizational purposes and used to determine the group of the test (see https://github.com/OpenSpace/OpenSpaceVisualTesting/pull/4)

Closes #197


ylvaselling (approved):
Nice!

My only comment is that maybe having the additional interface over the lua scripting might require more maintenance when we update the lua api, than just passing through pure lua? I suppose that would require a wait function though. Just a thought

alexanderbock:
It's a good point and the missing `wait` is definitely something that would prevent us from using it right now.  I'm also thinking that we'd want some user interface to create these tests in the long run, which would be harder if they are just bare Lua scripts.

tests/visual/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+    Script Equivalent: `openspace.time.setDeltaTime`
+
+  - `navigationstate`: Sets the camera to the provided navigation state. The provided value must be an object that must contain at least a `anchor` and `position` and may optionally contain an `aim`, `referenceFrame`, `up`, `yaw`, `pitch`, and `timestamp`. All these values are then used to instantaneously set the position of the camera.
+
+    Example: `{ "type": "navigationstate", "value": { "anchor": "Juno", "pitch": -0.0165756, "position": [ -22.49081, 1.191533, 26.35740 ], "up": [ 0.0288083, 0.999373, -0.0205962 ], "yaw": 0.152454 } }`
+
+    Script Equivalent: `openspace.navigation.setNavigationState`
+
+  - `pause`: Determines whether the in-game clock should be paused or resumed. The provided value must be a boolean that is the paused state after the instruction
```
engbergandreas, line 36:
```suggestion
  - `pause`: Determines whether the in-game clock should be paused or resumed. The provided value must be a boolean that is the clock state after the instruction
```
or some other descriptive word, saying paused here causes some ambiguity

PR #3271, opened 2024-05-20
Add example for the LuaScale class and fix it at the same time

alexanderbock:
Adding documentation for the LuaScale class and fixing some issues with it while already in the code

### DIFF
--- a/data/assets/examples/scale/luascale/example.lua
+++ b/data/assets/examples/scale/luascale/example.lua
@@ -1,0 +1,3 @@
+function scale(simTime, prevSimTime, wallTime)
+  return 10, 1, 1
+end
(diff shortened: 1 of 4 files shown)


data/assets/examples/scale/luascale/example.lua
```diff
@@ -0,0 +1,3 @@
+function scale(simTime, prevSimTime, wallTime)
```
ylvaselling, line 1: Is it possible to have this function as part of the asset? It looks like it only accepts a path but I thought I'd ask anyways. I'm thinking it won't show up in the documentation (unless we make some changes)

ylvaselling (approved):

PR #3284, opened 2024-05-27
Add example for the LuaTranslation class and fix it at the same time

alexanderbock:


ylvaselling (commented):
Tried this with the docs master

It seems the lua comment needs to search for just `--` and not `-- ` for the script to understand that the comment continues

It seems our smart move with the literal include uses the `generated/asset-components` folder as the "base" folder. So it doesn't find it right now:

```
This asset creates a SceneGraphNode that only displays coordinate axes. The coordinate axes are translated by a value determined by executing a Lua file that returns the translation parameters to be used as a table. In order to see the translation, we need to also have a node that does not move so that we can see the relative movement.

System Message: WARNING/2 (C:\Users\Ylva\Documents\OpenSpace\Documentation\OpenSpace-Docs\generated/asset-components/LuaTranslation.md, line 76)

Include file ‘C:\Users\Ylva\Documents\OpenSpace\Documentation\OpenSpace-Docs\generated\asset-components\example.lua’ not found or reading it failed
``` 

An easy solution is to write the relative path 
```
../asset_examples/data/assets/examples/translation/luatranslation/example.lua
```
which yields this result:
image

Otherwise I tested the asset in OS and it worked like a charm

PR #3287, opened 2024-05-27
Issue/3286: renderableOrbitalKepler - optimization of point and trail rendering

Arohdin:
## Trail optimization
Optimization for the trail shader was achieved by using a geometry shader to cull vertices that does not contribute to the final trail. In other words, vertices which part of the trail would be fully faded (discard by fragment shader) is now removed in geometry shader and thus less data is being processed in fragment shader.

## Point optimization
Changed nearly all calculations from double to float. Was using double as a safe-guard during development and never revisited that decision. 

## Performance statistics
Everything is compared to 0.20.0 RC1

### Trail optimization
The new implementation results in a framerate that does not fluctuate depending on how many pixels the trails cover on screen.
Test using the _Main Asteroid Belt_ at different zoom levels within the solar system

**Before:**
- 60 - 170 fps
- 16.67 - 5.88 ms

**After:**
- 156 fps
- 6.41 ms

At most it increased by 0.53 ms and at most it recued by 10.26 ms

### Point optimization
Test using the _Main Asteroid Belt_ at different zoom levels within the solar system

**Before:**
-  86 fps
- 11.63 ms

**After:**
- 167 - 176 fps
- 5.99 - 5.68 ms

At most it reduced by 9.58 ms

### Trail + Point
Using both optimizations and testing with rendering mode "Trails + Points".
Test using the _Main Asteroid Belt_ at different zoom levels within the solar system

**Before:**
-  42 - 75 fps
- 23.81 - 13.33 ms

**After:**
- 94 - 97 fps
- 10.64 - 10.31 ms

At most it reduced by 13.5 ms

##
### Test system
Intel Core 9 12900
64GB RAM
Nvidia RTX 3080 (10GB)
Screen resolution: 2560x1440

### DIFF
--- a/modules/space/rendering/renderableorbitalkepler.cpp
+++ b/modules/space/rendering/renderableorbitalkepler.cpp
@@ -403,8 +403,8 @@ void RenderableOrbitalKepler::initializeGL() {
         _trailProgram->uniformLocation("modelViewTransform");
     _uniformTrailCache.projection =
         _trailProgram->uniformLocation("projectionTransform");
-    _uniformTrailCache.colorFadeCutoffPoint =
-        _trailProgram->uniformLocation("colorFadeCutoffPoint");
+    _uniformTrailCache.colorFadeCutoffValue =
+        _trailProgram->uniformLocation("colorFadeCutoffValue");
     _uniformTrailCache.trailFadeExponent =
         _trailProgram->uniformLocation("trailFadeExponent");
     _uniformTrailCache.inGameTime = _trailProgram->uniformLocation("inGameTime");
@@ -574,7 +574,7 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
         // on the distance from the head of the trail to the part that's being rendered.
         // Value is passed as uniform due to it being used in both geometry and fragment
         // shader.
-        _trailProgram->setUniform(_uniformTrailCache.colorFadeCutoffPoint, 0.05f);
+        _trailProgram->setUniform(_uniformTrailCache.colorFadeCutoffValue, 0.05f);
 
         glLineWidth(_appearance.trailWidth);
 
--- a/modules/space/rendering/renderableorbitalkepler.cpp
+++ b/modules/space/rendering/renderableorbitalkepler.cpp
@@ -123,8 +123,9 @@ namespace {
 
     constexpr openspace::properties::Property::PropertyInfo TrailFadeInfo = {
         "TrailFade",
-        "Trail Fade Factor",
-        "Determines how fast the trail fades out.",
+        "Trail Fade",
+        "Determines how fast the trail fades out. A smaller number shows less of the "
+        "trail and a larger number shows more.",
         openspace::properties::Property::Visibility::User
     };
 
@@ -378,7 +379,8 @@ void RenderableOrbitalKepler::initializeGL() {
            return global::renderEngine->buildRenderProgram(
                "OrbitalKeplerTrails",
                absPath("${MODULE_SPACE}/shaders/debrisVizTrails_vs.glsl"),
-               absPath("${MODULE_SPACE}/shaders/debrisVizTrails_fs.glsl")
+               absPath("${MODULE_SPACE}/shaders/debrisVizTrails_fs.glsl"),
+               absPath("${MODULE_SPACE}/shaders/debrisVizTrails_gs.glsl")
            );
        }
    );
@@ -401,7 +403,8 @@ void RenderableOrbitalKepler::initializeGL() {
         _trailProgram->uniformLocation("modelViewTransform");
     _uniformTrailCache.projection =
         _trailProgram->uniformLocation("projectionTransform");
-    _uniformTrailCache.trailFade = _trailProgram->uniformLocation("trailFade");
+    _uniformTrailCache.trailFadeExponent = _trailProgram->uniformLocation("trailFadeExponent");
+    _uniformTrailCache.colorFadeCutoffPoint = _trailProgram->uniformLocation("colorFadeCutoffPoint");
     _uniformTrailCache.inGameTime = _trailProgram->uniformLocation("inGameTime");
     _uniformTrailCache.color = _trailProgram->uniformLocation("color");
     _uniformTrailCache.opacity = _trailProgram->uniformLocation("opacity");
@@ -562,7 +565,9 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
         const float fade = pow(
             _appearance.trailFade.maxValue() - _appearance.trailFade, 2.f
         );
-        _trailProgram->setUniform(_uniformTrailCache.trailFade, fade);
+        _trailProgram->setUniform(_uniformTrailCache.trailFadeExponent, fade);
+
+        _trailProgram->setUniform(_uniformTrailCache.colorFadeCutoffPoint, 0.05f);
 
         glLineWidth(_appearance.trailWidth);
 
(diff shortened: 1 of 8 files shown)


modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -401,7 +403,8 @@ void RenderableOrbitalKepler::initializeGL() {
         _trailProgram->uniformLocation("modelViewTransform");
     _uniformTrailCache.projection =
         _trailProgram->uniformLocation("projectionTransform");
-    _uniformTrailCache.trailFade = _trailProgram->uniformLocation("trailFade");
+    _uniformTrailCache.trailFadeExponent = _trailProgram->uniformLocation("trailFadeExponent");
+    _uniformTrailCache.colorFadeCutoffPoint = _trailProgram->uniformLocation("colorFadeCutoffPoint");
```
alexanderbock, line 407: Lines too long

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -562,7 +565,9 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
         const float fade = pow(
             _appearance.trailFade.maxValue() - _appearance.trailFade, 2.f
         );
-        _trailProgram->setUniform(_uniformTrailCache.trailFade, fade);
+        _trailProgram->setUniform(_uniformTrailCache.trailFadeExponent, fade);
+
+        _trailProgram->setUniform(_uniformTrailCache.colorFadeCutoffPoint, 0.05f);
```
alexanderbock, line 570: Maybe add a comment here why it is 0.05?  The uniform is never set to another value either, so maybe just declare it as a constant in the shader instead?

modules/space/rendering/renderableorbitalkepler.h
```diff
@@ -112,7 +112,7 @@ class RenderableOrbitalKepler : public Renderable {
     RenderableOrbitalKepler::Appearance _appearance;
 
     // Line cache
-    UniformCache(modelView, projection, trailFade, inGameTime, color, opacity)
+    UniformCache(modelView, projection, trailFadeExponent, colorFadeCutoffPoint, inGameTime, color, opacity)
```
alexanderbock, line 115: Line length too long

modules/space/shaders/debrisVizTrails_gs.glsl
```diff
@@ -0,0 +1,80 @@
...
+
+uniform float trailFadeExponent;
+uniform float colorFadeCutoffPoint;
+
+void main() {
+  float cFrac = currentRevolutionFraction[0];
+  float v0Frac = vertexRevolutionFraction[0];
+  float v1Frac = vertexRevolutionFraction[1];
+
+  // Distance between current recolution fraction and revolution factor for vertex0 and vertex1
```
alexanderbock, line 46: Line too long?

modules/space/shaders/debrisVizTrails_gs.glsl
```diff
@@ -0,0 +1,80 @@
...
+    vd1 += 1.0;
+  }
+  // ==========================
+
+  // Calculates if vertex0 is close enough to NOT be discarded in fragment shader
+  float invert = pow(1.0 - vd1, trailFadeExponent);  
+  float fade = clamp(invert, 0.0, 1.0);
+
+  // Only emit vertices for line segments where both vertices should be rendered
+  if ( (fade > colorFadeCutoffPoint) || (vd0 < vd1) ) {
```
alexanderbock, line 63:
```suggestion
  if ((fade > colorFadeCutoffPoint) || (vd0 < vd1)) {
```

modules/space/rendering/renderableorbitalkepler.cpp
Arohdin: The cutoff value is being used in both geometry and fragment shader for trails.
The reason I made it an uniform was to minimize the risk if it being different in each shader.
I'll add a comment explaining 0.05.

Arohdin: >0.05 is the "alpha value" for which the trail should no longer be rendered.
>The value that's compared to 0.05 is calculated in the shader and depends
>on the distance from the head of the trail to the part that's being rendered.
>Value is passed as uniform due to it being used in both geometry and fragment
>shader.

Too descriptive?

alexanderbock: I think that's great :+1:

alexanderbock (approved):

modules/space/shaders/debrisVizTrails_gs.glsl
```diff
@@ -0,0 +1,81 @@
...
+  float vd0 = cFrac - v0Frac;
+  if (vd0 < 0.0) {
+    vd0 += 1.0;
+  }
+
+  float vd1 = cFrac - v1Frac;
+  if (vd1 < 0.0) {
+    vd1 += 1.0;
+  }
+  // ==========================
```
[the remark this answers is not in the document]
ylvaselling: Aha, thank you for the thorough explanation! 🥳 It makes sense now:) 

Is there any chance we could place a condensed version of this as comments in the code? Just thinking if someone new looks at the code it might be helpful. Maybe just explaining that the `currentRevolutionFraction` is the head of the trail, and that the `v0Frac` and `v1Frac` are the fractions around that head would be helpful at least for me

modules/space/shaders/debrisVizTrails_gs.glsl
```diff
@@ -0,0 +1,81 @@
...
+
+  // Calculates if vertex0 is close enough to NOT be discarded in fragment shader
```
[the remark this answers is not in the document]
ylvaselling: Sounds good 👍

modules/space/shaders/debrisVizTrails_gs.glsl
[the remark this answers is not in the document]
Arohdin: How about this?

> cFrac is how far along the trail orbit the head of the trail is.
> v0Frac and v1Frac are how far the two vertices that creates the current line strip
> are along the trail orbit. The variables span between 0 and 1, where 0 is the
> beginning of the trail and 1 is the end of the trail (a full orbit).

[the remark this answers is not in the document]
ylvaselling: Perfect!

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -401,7 +403,10 @@ void RenderableOrbitalKepler::initializeGL() {
         _trailProgram->uniformLocation("modelViewTransform");
     _uniformTrailCache.projection =
         _trailProgram->uniformLocation("projectionTransform");
-    _uniformTrailCache.trailFade = _trailProgram->uniformLocation("trailFade");
+    _uniformTrailCache.colorFadeCutoffPoint =
+        _trailProgram->uniformLocation("colorFadeCutoffPoint");
```
GPayne, line 407: Since this code deals with rendering trails and points, how about calling this `colorFadeCutoffLevel` or `colorFadeCutoffValue` instead?

Arohdin: Changed to `colorFadeCutoffValue`

GPayne:
Tested on linux. Everything looks great, and performance is definitely better than before.
I have a concern about the point size exponent. I was running with the main asteroid belt (the largest set) and accidentally clicked the exponent slider with too high a value, and was stuck with having to restart my laptop which had ground to a halt. It only has nvidia 1050 graphics, but other users with lesser GPUs might also encounter this problem.
I see the point of having the wide size exponent range to adjust based on the zoom distance. At least for this subject matter it should probably be capped at a lower value.

Arohdin:
> I see the point of having the wide size exponent range to adjust based on the zoom distance. At least for this subject matter it should probably be capped at a lower value.

Yeah, it's an issue for both this renderable as well as the point cloud. Will think about this and try to fix it/work around it asap (at least for this renderable).

PR #3292, opened 2024-05-31
Adds a join server lua function for parallelpeer

engbergandreas:
Includes the necessary server info to join a parallel connection, however does not include the buffer time, and keyframe intervals. 
Also fixes the disconnect function calling connect (?)

### DIFF
--- a/src/network/parallelpeer_lua.inl
+++ b/src/network/parallelpeer_lua.inl
@@ -24,6 +24,20 @@
 
 namespace {
 
+[[codegen::luawrap]] void joinServer(std::string port, std::string address,
+    std::string password, std::string hostpassword = "", std::string name = "Anonymous") {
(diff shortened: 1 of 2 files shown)


src/network/parallelpeer_lua.inl
```diff
@@ -24,6 +24,20 @@
 
 namespace {
 
+[[codegen::luawrap]] void joinServer(std::string port, std::string address,
+    std::string password, std::string hostpassword = "", std::string name = "Anonymous") {
```
alexanderbock, line 28:
```suggestion
[[codegen::luawrap]] void joinServer(std::string port, std::string address,
                                     std::string password, std::string hostpassword = "",
                                     std::string name = "Anonymous")
{
```

alexanderbock (approved):

PR #3298, opened 2024-06-04
added zIndex parsing for .info files

micahnyc:
This is a pull request to be put into this one: https://github.com/OpenSpace/OpenSpace/pull/3281

### DIFF
--- a/modules/globebrowsing/scripts/layer_support.lua
+++ b/modules/globebrowsing/scripts/layer_support.lua
@@ -225,10 +225,7 @@ openspace.globebrowsing.parseInfoFile = function (file)
     local identifier = Identifier
     local zIndex = ZIndex
 
-    if zIndex == "" then
-        zIndex = 999
-    end
-    if identifier == "" then
+    if identifier == nil then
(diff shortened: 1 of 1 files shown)


modules/globebrowsing/scripts/layer_support.lua
```diff
@@ -217,11 +218,16 @@ openspace.globebrowsing.parseInfoFile = function (file)
...
     local Location = rawget(_G, "Location")
+    local ZIndex = rawget(_G, "ZIndex")
 
     -- Now we can start
     local name = Name or Identifier
     local identifier = Identifier
+    local zIndex = ZIndex
 
+    if zIndex == "" then
+        zIndex = 999
```
Roxeena, line 229: I don't think this is needed. When adding a layer without specifying the ZIndex it will be put right after the latest layer that has been loaded, similar to how it worked before. Therefore I don't think we need a default value.

Roxeena (commented):
I was just about to implement this but then noticed this. Thanks!

PR #3306, opened 2024-06-05
Fix renderable sphere example and add one for using local image

WeirdRubberDuck:
The example `RenderableSphereImageOnline` was broken due to missing the previously non-optional `Size` parameter. 

Both `Size` and `Segments` have been marked as an advanced user property and should not be mandatory in the asset component specification, so I changed that and added verifiers. 

 The example also had a comment about properties from `RenderableSphere`, that the user should ont have to know about. removed this. I also changed the default value to the one we wanted to use for this example, to be a bit smoother per default.
 
 Finally, added a similar example for `RenderableSphereImageLocal`.

### DIFF
--- a/data/assets/examples/renderable/renderablesphereimagelocal/sphereimagelocal.asset
+++ b/data/assets/examples/renderable/renderablesphereimagelocal/sphereimagelocal.asset
@@ -1,5 +1,5 @@
 -- Basic
--- This example will show a sphere that is covered with an image which is retrieved from
+-- This example shows a sphere that is covered with an image which is retrieved from
 -- a local file path. The image will be stretched over the entire sphere as an
 -- equirectangular projection.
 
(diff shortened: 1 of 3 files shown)


data/assets/examples/renderable/renderablesphereimagelocal/sphereimagelocal.asset
```diff
@@ -0,0 +1,24 @@
+-- Basic
+-- This example will show a sphere that is covered with an image which is retrieved from
```
alexanderbock, line 2:
```suggestion
-- This example shows a sphere that is covered with an image which is retrieved from
```

data/assets/examples/renderable/renderablesphereimageonline/sphereimageonline.asset
```diff
@@ -1,16 +1,13 @@
 -- Basic
 -- This example will show a sphere that is covered with an image which is retrieved from
```
alexanderbock, line 2:
```suggestion
-- This example shows a sphere that is covered with an image which is retrieved from
```

alexanderbock (approved):

PR #3308, opened 2024-06-07
Add ability for SPICE translation and rotation to account for light travel time

micahnyc:
The use case here was that we wanted to show and asteroid casting a shadow on earth. However, by the time the shadow reaches earth, it will have moved and rotated. So without this change the shadow will land on the wrong part of earth. 

Using this change the body will be moved back in it's orbit by the amount of light travel time.

Following the same thought thru to rotation, the body will be rotated back in time to the rotation that would be observed. Right now if you were to anchor and aim at Jupiter, and change the fov so that you could see it, you would see the rotation of Jupiter _at the current time_, not the rotation at Jupiter _as it would be observed from earth at the current time_  (since you would really be seeing Jupiter 4 minutes ago)

(note when using light travel time for rotation, you must also provide the target/observer body IDs since they are not know based on the rotation frame)

First and second screenshots show Jupiter with its regular rotation, and then with light travel time to the sun accounted for.
Third screenshot shows Neptune (barycenter) moved back by the light travel time to the sun.

(also this PR adds a lightTravelTime function to the spicemanager that is used by the rotation, but could be used in the future by other new functions) 

OpenSpace_000003
OpenSpace_000004
OpenSpace_000005

### DIFF
--- a/include/openspace/util/spicemanager.h
+++ b/include/openspace/util/spicemanager.h
@@ -994,6 +994,55 @@ public:
         int numberOfTerminatorPoints);
 
     /**
+     * The structure retuned by the #lightTime method.
+     */
+    struct LightTimeResult {
+        /// Epoch of the signal at the target.
+        /// This is the time at which the signal would arrive at the target.
+        double epochAtTarget = 0.0;
+
+        /// Time between transmit and receipt of the signal.
+        /// This is the light travel time.
+        double elapsed = 0.0;
+    };
+
+    /**
+     * This method returns time of a signal recieved at a target from an \observer
+     * corrected for light travel time.
+     *
+     * \param observedTime Time at which the signal was received
+     * \param target The target who emited the signal
+     * \param observer The observer who received the signal
+     * \param direction The direction of travel ("->" or "<-")
+     *
+     * \return The LightTimeResult structure that contains information about the time
+     *
+     * \throw SpiceException If \p instrument does not name a valid NAIF object
+     * \post The returned structure has all its values initialized
+     *
+     * \see https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/ltime_c.html
+     */
+    LightTimeResult lightTravelTime(double observedTime, int target, int observer, const std::string& direction) const;
+
+    /**
+    * This method returns time of a signal recieved at a target from an \observer
+    * corrected for light travel time.
+    *
+    * \param observedTime Time at which the signal was received
+    * \param target The target who emited the signal
+    * \param observer The observer who received the signal
+    * \param direction The direction of travel ("->" or "<-")
+    *
+    * \return The LightTimeResult structure that contains information about the time
+    *
+    * \throw SpiceException If \p instrument does not name a valid NAIF object
+    * \post The returned structure has all its values initialized
+    *
+    * \see https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/ltime_c.html
+    */
+    LightTimeResult lightTravelTime(double observedTime, const std::string& target, const std::string& observer, const std::string& direction) const;
+
+    /**
      * Sets the SpiceManager's exception handling. If UseException::No is passed to this
      * function, all subsequent calls will not throw an error, but fail silently instead.
      * If set to UseException::Yes, a SpiceException is thrown whenever an error occurs.
--- a/modules/space/rotation/spicerotation.cpp
+++ b/modules/space/rotation/spicerotation.cpp
@@ -29,9 +29,12 @@
 #include <openspace/util/spicemanager.h>
 #include <openspace/util/time.h>
 #include <openspace/util/updatestructures.h>
+#include <ghoul/logging/logmanager.h>
 #include <optional>
 
 namespace {
+    constexpr std::string_view _loggerCat = "SpiceRotation";
+
     constexpr openspace::properties::Property::PropertyInfo SourceInfo = {
         "SourceFrame",
         "Source",
@@ -63,6 +66,32 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo UseLightTravelTimeInfo = {
+        "UseLightTravelTime",
+        "Use Light Travel Time",
+        "If set to true, the rotation will be updated based on light travel "
+        "time to the observer.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo LightTravelTimeObserverInfo = {
+        "LightTravelTimeObserver",
+        "Light Travel Time Observer",
+        "This is the SPICE NAIF name for the body who observes the rotation."
+        "It can either be a fully qualified name (such as 'EARTH') or a NAIF integer id "
+        "code (such as '399'). Only used when UseLightTravelTime is true",
+        openspace::properties::Property::Visibility::Developer
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo LightTravelTimeTargetInfo = {
+        "LightTravelTimeTarget",
+        "Light Travel Time Target",
+        "This is the SPICE NAIF name for the body who will be rotated."
+        "It can either be a fully qualified name (such as 'EARTH') or a NAIF integer id "
+        "code (such as '399'). Only used when UseLightTravelTime is true",
+        openspace::properties::Property::Visibility::Developer
+    };
+
     struct [[codegen::Dictionary(SpiceRotation)]] Parameters {
         // [[codegen::verbatim(SourceInfo.description)]]
         std::string sourceFrame
@@ -80,6 +109,15 @@ namespace {
         // [[codegen::verbatim(FixedDateInfo.description)]]
         std::optional<std::string> fixedDate
             [[codegen::annotation("A time to lock the rotation to")]];
+
+        // [[codegen::verbatim(UseLightTravelTimeInfo.description)]]
+        std::optional<bool> useLightTravelTime;
+
+        // [[codegen::verbatim(LightTravelTimeObserverInfo.description)]]
+        std::optional<std::string> lightTravelTimeObserver;
+
+        // [[codegen::verbatim(LightTravelTimeTargetInfo.description)]]
+        std::optional<std::string> lightTravelTimeTarget;
     };
 #include "spicerotation_codegen.cpp"
 } // namespace
@@ -94,6 +132,9 @@ SpiceRotation::SpiceRotation(const ghoul::Dictionary& dictionary)
     : _sourceFrame(SourceInfo)
     , _destinationFrame(DestinationInfo)
     , _fixedDate(FixedDateInfo)
+    , _useLightTravelTime(UseLightTravelTimeInfo)
+    , _lightTravelTimeObserver(LightTravelTimeObserverInfo)
+    , _lightTravelTimeTarget(LightTravelTimeTargetInfo)
 {
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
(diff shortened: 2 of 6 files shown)


include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
         AberrationCorrection aberrationCorrection, double ephemerisTime,
         int numberOfTerminatorPoints);
 
+    /**
+     * The structure retuned by the #lightTime method.
```
alexanderbock, line 997:
```suggestion
     * The structure returned by the #lightTime method.
```

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+     */
+    struct LightTimeResult {
+        /// Epoch of the signal at the target.
+        /// This is the time at which the signal would arrive at the target.
+        double epochAtTarget = 0.0;
+
+        /// Time between transmit and receipt of the signal.
```
alexanderbock, line 1004:
```suggestion
        /// Time between transmission and reception of the signal.
```

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
         AberrationCorrection aberrationCorrection, double ephemerisTime,
         int numberOfTerminatorPoints);
 
+    /**
+     * The structure retuned by the #lightTime method.
+     */
+    struct LightTimeResult {
+        /// Epoch of the signal at the target.
```
alexanderbock, line 1000: This would benefit from a better definition what the "signal" is

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+        double epochAtTarget = 0.0;
+
+        /// Time between transmit and receipt of the signal.
+        /// This is the light travel time.
+        double elapsed = 0.0;
+    };
+
+    /**
+     * This method returns time of a signal recieved at a target from an \observer
+     * corrected for light travel time.
```
alexanderbock, line 1011:
```suggestion
     * This method returns time of a signal received at a target from an \p observer
     * corrected for light travel time.
```

Which time is this? The `LightTimeResult` returns back two distinct times

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+     *
+     * \param observedTime Time at which the signal was received
+     * \param target The target who emited the signal
```
alexanderbock, line 1014:
```suggestion
     * \param target The target who emitted the signal
```

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+     * \param direction The direction of travel ("->" or "<-")
+     *
+     * \return The LightTimeResult structure that contains information about the time
+     *
+     * \throw SpiceException If \p instrument does not name a valid NAIF object
+     * \post The returned structure has all its values initialized
+     *
+     * \see https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/ltime_c.html
+     */
+    LightTimeResult lightTravelTime(double observedTime, int target, int observer, const std::string& direction) const;
```
alexanderbock, line 1025: This line looks like it is exceeding our line length, but didn't check

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+    };
+
+    /**
+     * This method returns time of a signal recieved at a target from an \observer
+     * corrected for light travel time.
+     *
+     * \param observedTime Time at which the signal was received
+     * \param target The target who emited the signal
+     * \param observer The observer who received the signal
+     * \param direction The direction of travel ("->" or "<-")
```
alexanderbock, line 1016: The direction parameter should instead be an `enum class` instead, for example the AberrationCorrection::Direction which already does this

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+    * \param direction The direction of travel ("->" or "<-")
+    *
+    * \return The LightTimeResult structure that contains information about the time
+    *
+    * \throw SpiceException If \p instrument does not name a valid NAIF object
+    * \post The returned structure has all its values initialized
+    *
+    * \see https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/ltime_c.html
+    */
+    LightTimeResult lightTravelTime(double observedTime, const std::string& target, const std::string& observer, const std::string& direction) const;
```
alexanderbock, line 1043: These two functions should instead be:
```cpp
LightTimeResult lightTravelTime(double observedTime, const std::variant<std::string, int>& target, const std::variant<std::string, int>& observer, const std::string& direction) const;
```

Right now we can have _only_ strings or _only_ ints, but not mixing them.

include/openspace/util/spicemanager.h
```diff
@@ -993,6 +993,55 @@ class SpiceManager {
...
+     * \throw SpiceException If \p instrument does not name a valid NAIF object
+     * \post The returned structure has all its values initialized
+     *
+     * \see https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/ltime_c.html
+     */
+    LightTimeResult lightTravelTime(double observedTime, int target, int observer, const std::string& direction) const;
+
+    /**
+    * This method returns time of a signal recieved at a target from an \observer
+    * corrected for light travel time.
```
alexanderbock, line 1029: Same comments as above

modules/space/rotation/spicerotation.cpp
```diff
@@ -63,6 +66,32 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo UseLightTravelTimeInfo = {
+        "UseLightTravelTime",
+        "Use Light Travel Time",
+        "If set to true, the rotation will be updated based on light travel "
+        "time to the observer.",
```
alexanderbock, line 73: This needs a lot more description as it is a pretty confusing topic to begin with

modules/space/rotation/spicerotation.cpp
```diff
@@ -63,6 +66,32 @@ namespace {
...
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo LightTravelTimeObserverInfo = {
+        "LightTravelTimeObserver",
+        "Light Travel Time Observer",
+        "This is the SPICE NAIF name for the body who observes the rotation."
```
alexanderbock, line 80:
```suggestion
        "This is the SPICE NAIF name for the body who observes the rotation. "
```

modules/space/rotation/spicerotation.cpp
```diff
@@ -63,6 +66,32 @@ namespace {
...
+        "It can either be a fully qualified name (such as 'EARTH') or a NAIF integer id "
```
alexanderbock, line 81:
```suggestion
        "It can either be a SPICE object name (such as 'EARTH') or a NAIF integer id "
```

modules/space/rotation/spicerotation.cpp
```diff
@@ -63,6 +66,32 @@ namespace {
...
+        "code (such as '399'). Only used when UseLightTravelTime is true",
+        openspace::properties::Property::Visibility::Developer
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo LightTravelTimeTargetInfo = {
+        "LightTravelTimeTarget",
+        "Light Travel Time Target",
+        "This is the SPICE NAIF name for the body who will be rotated."
```
alexanderbock, line 89:
```suggestion
        "This is the SPICE NAIF name for the body that will be rotated. "
```

modules/space/rotation/spicerotation.cpp
```diff
@@ -63,6 +66,32 @@ namespace {
...
+        "It can either be a fully qualified name (such as 'EARTH') or a NAIF integer id "
```
alexanderbock, line 90:
```suggestion
        "It can either be a SPICE object name (such as 'EARTH') or a NAIF integer id "
```

modules/space/rotation/spicerotation.cpp
```diff
@@ -80,6 +109,15 @@ namespace {
...
             [[codegen::annotation("A time to lock the rotation to")]];
+
+        // [[codegen::verbatim(UseLightTravelTimeInfo.description)]]
+        std::optional<bool> useLightTravelTime;
+
+        // [[codegen::verbatim(LightTravelTimeObserverInfo.description)]]
+        std::optional<std::string> lightTravelTimeObserver;
+
+        // [[codegen::verbatim(LightTravelTimeTargetInfo.description)]]
+        std::optional<std::string> lightTravelTimeTarget;
```
alexanderbock, line 120: To prevent new error conditions, I'd suggest instead:
```cpp
struct LightTravelTime {
  std::string observer;
  std::string target;
};

std::optional<LightTravelTime> lightTravelTime;
```
and then `useLightTravelTime = lightTravelTime.has_value()` and we don't need to specify it separately

That way it is impossible to set `useLightTravelTime` to true but then forget to set the observer or target

src/util/spicemanager.cpp
```diff
@@ -1387,6 +1387,17 @@ glm::dmat3 SpiceManager::getEstimatedTransformMatrix(const std::string& fromFram
...
+    return lightTravelTime(observedTime, naifId(observer), naifId(target), direction);
+}
+
+SpiceManager::LightTimeResult SpiceManager::lightTravelTime(double observedTime, int target, int observer, const std::string& direction) const {
+
+    LightTimeResult res;
+    ltime_c(observedTime, observer, direction.c_str(), target, &res.epochAtTarget, &res.elapsed);
+    return res;
+}
+
```
alexanderbock, line 1400: Why do we need the new functions if the `targetPosition` already has the `AberrationCorrection aberrationCorrection` parameter that does the calculations correctly inside SPICE?  We could add the same parameter to the transform matrix functions instead?

modules/space/rotation/spicerotation.cpp
```diff
@@ -136,11 +195,37 @@ glm::dmat3 SpiceRotation::matrix(const UpdateData& data) const {
...
         time = *_fixedEphemerisTime;
     }
-    return SpiceManager::ref().positionTransformMatrix(
-        _sourceFrame,
-        _destinationFrame,
-        time
-    );
+
+    if (_useLightTravelTime) {
+        if ((std::string(_lightTravelTimeTarget).length() < 1) || (std::string(_lightTravelTimeTarget).length() < 1)) {
```
alexanderbock, line 200:
```suggestion
        if (_lightTravelTimeTarget.value().empty() || _lightTravelTimeTarget.value().empty()) {
```

modules/space/translation/spicetranslation.cpp
```diff
@@ -159,11 +178,20 @@ glm::dvec3 SpiceTranslation::position(const UpdateData& data) const {
...
+            SpiceManager::AberrationCorrection::Direction::Reception
+        };
+    }
+
     return SpiceManager::ref().targetPosition(
         _cachedTarget,
         _cachedObserver,
         _cachedFrame,
-        {},
+        corr,
```
alexanderbock, line 194:
```suggestion
        std::move(corr),
```

alexanderbock (changes requested):

modules/space/rotation/spicerotation.cpp
```diff
@@ -111,6 +152,24 @@ SpiceRotation::SpiceRotation(const ghoul::Dictionary& dictionary)
...
+    _lightTravelTimeObserver.onChange([this]() {
+        requireUpdate();
+    });
+    _lightTravelTimeObserver = p.lightTravelTimeObserver.value_or(_lightTravelTimeObserver);
+    addProperty(_lightTravelTimeObserver);
+
+    _lightTravelTimeTarget.onChange([this]() {
+        requireUpdate();
+    });
+    _lightTravelTimeTarget = p.lightTravelTimeObserver.value_or(_lightTravelTimeTarget);
```
Roxeena, line 170:
```suggestion
    _lightTravelTimeTarget = p.lightTravelTimeTarget.value_or(_lightTravelTimeTarget);
```

modules/space/translation/spicetranslation.cpp
```diff
@@ -159,11 +178,20 @@ glm::dvec3 SpiceTranslation::position(const UpdateData& data) const {
     if (_fixedEphemerisTime.has_value()) {
         time = *_fixedEphemerisTime;
     }
+
+    SpiceManager::AberrationCorrection corr = {};
```
Roxeena, line 182: Why does the translation and rotation implement the light travel time correction differently? Here it seems to use SPICE built in functionality but in the rotation it seems to be a more manual process where the time is shifted outside of SPICE.

modules/space/rotation/spicerotation.cpp
alexanderbock: Good catch!

alexanderbock:
Going to close the PR in a month

PR #3325, opened 2024-06-14
Feature/update gui actions

ylvaselling:
Updates the GUI when an action is added or removed. To be used with this PR:

I removed the subscriptions as they are not really needed for actions - not that many actions are added or removed during a regular session. Instead a new topic is created when needed. The UI listens to the event topic which now notifies when an action has been added or removed.

There are some naming conventions we might want to look over for consistency, such as "action" vs "shortcut" and "uri" vs "identifier".

### DIFF
--- a/modules/server/src/topics/shortcuttopic.cpp
+++ b/modules/server/src/topics/shortcuttopic.cpp
@@ -110,7 +110,8 @@ nlohmann::json ShortcutTopic::shortcutJson(const std::string& identifier) const 
         actions.end(),
         [&identifier](const interaction::Action& action) {
             return action.identifier == identifier;
-    });
+        }
+    );
 
     if (found == actions.end()) {
         return json;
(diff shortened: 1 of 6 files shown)


modules/server/src/topics/shortcuttopic.cpp
```diff
@@ -101,23 +101,46 @@ std::vector<nlohmann::json> ShortcutTopic::shortcutsJson() const {
...
+nlohmann::json ShortcutTopic::shortcutJson(const std::string& identifier) const {
+    std::vector<nlohmann::json> json;
+    std::vector<interaction::Action> actions = global::actionManager->actions();
+
+    auto found = std::find_if(
+        actions.begin(),
+        actions.end(),
+        [&identifier](const interaction::Action& action) {
+            return action.identifier == identifier;
+    });
```
alexanderbock, line 113:
```suggestion
        }
    );
```

modules/server/src/topics/shortcuttopic.cpp
```diff
@@ -101,23 +101,46 @@ std::vector<nlohmann::json> ShortcutTopic::shortcutsJson() const {
...
+
+    if (found == actions.end()) {
+        return json;
```
alexanderbock, line 116:
```suggestion
        return {};
```

Making it more clear that it is an empty result

alexanderbock (approved):

PR #3326, opened 2024-06-18
fix parsing TLE files with CRLF line endings on unix

k2d222:
on systems with LF line endings (unix), the parsing would not get rid of the extra CR character causing a parse error for tle files. This PR reads the file in binary mode which is platform independent.
reference: https://en.cppreference.com/w/cpp/io/c/FILE#Binary_and_text_modes

### DIFF
--- a/src/util/spicemanager_lua.inl
+++ b/src/util/spicemanager_lua.inl
@@ -23,6 +23,7 @@
  ****************************************************************************************/
 
 #include <ghoul/misc/stringhelper.h>
+#include <ios>
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


alexanderbock:
@GPayne  Can you give this a testrun as well?

GPayne:
Unfortunately this caused a segfault when initializing the ISS, running on ubuntu. Since the code is removing all '\r' characters from the lines, it might work without the binary read mode.

k2d222:
> Unfortunately this caused a segfault when initializing the ISS, running on ubuntu. Since the code is removing all '\r' characters from the lines, it might work without the binary read mode.

interesting, I'll have a look at this later. But looking at the changes I don't see anything that might be UB, so Im surprised it segfaults.

was it gcc or clang?

k2d222:
should be fixed now @GPayne

It would work without binary mode too, but I find it better this way.

GPayne:
Works well now for me, @k2d222, thanks for the fix.

GPayne (approved):

alexanderbock:
Works on Windows as well :+1:

PR #3327, opened 2024-06-19
Add a point cloud example that changes the point orientation

WeirdRubberDuck:
Note that it is really difficult to see what is going on without a texture, so added a texture to this example as well.

### DIFF
--- a/data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
+++ b/data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
@@ -13,8 +13,8 @@ local Node = {
   Renderable = {
     Type = "RenderablePointCloud",
     File = asset.resource("data/dummydata.csv"),
-    -- Change the oreientation render option to face the camera position instead
-    -- of it's view direction
+    -- Change the orientation render option to face the camera position instead
+    -- of its view direction
     OrientationRenderOption = "Camera Position Normal",
     -- Add a texture so we can more easily see how the orientation is changed
     Texture = {
--- a/data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
+++ b/data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
@@ -9,7 +9,7 @@
 -- See Textured example for more details.
 
 local Node = {
-  Identifier = "RenderablePointCloud_Example",
+  Identifier = "RenderablePointCloud_Example_FaceCameraPosition",
   Renderable = {
     Type = "RenderablePointCloud",
     File = asset.resource("data/dummydata.csv"),
--- a/data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
+++ b/data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
@@ -1,0 +1,37 @@
+-- Orientation - Face Camera Position
+-- This example creates a point cloud where the planes that make up the points are
+-- oriented to face the camera position, rather than the view direction (which is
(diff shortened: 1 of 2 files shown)


data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
```diff
@@ -0,0 +1,37 @@
...
+--
+-- A texture is added so that we can more easily see how the orientation is altered.
+-- See Textured example for more details.
+
+local Node = {
+  Identifier = "RenderablePointCloud_Example",
+  Renderable = {
+    Type = "RenderablePointCloud",
+    File = asset.resource("data/dummydata.csv"),
+    -- Change the oreientation render option to face the camera position instead
```
WeirdRubberDuck, line 16:
```suggestion
    -- Change the orientation render option to face the camera position instead
```

data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
```diff
@@ -0,0 +1,37 @@
...
+    -- of it's view direction
```
alexanderbock, line 17:
```suggestion
    -- of its view direction
```

alexanderbock (approved):

data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
```diff
@@ -0,0 +1,37 @@
...
+-- oriented to face the camera position, rather than the view direction (which is
+-- default). This means that the point will be layed out in a spherical way around the
+-- camera, which works better for spherical displays compared to the default
+-- orientation.
+--
+-- A texture is added so that we can more easily see how the orientation is altered.
+-- See Textured example for more details.
+
+local Node = {
+  Identifier = "RenderablePointCloud_Example",
```
Roxeena, line 12: Won't this identifier collide with other examples if I were to load several of them at the same time?

Roxeena (commented):
I have not been able to run this example yet but a general thought I had was that maybe we should decide on one term to use for this behavior since we have a few versions of it (billboard, face camera normal, face camera position, etc.). They all do/mean the same or similar things and it can be confusing sometimes.

data/assets/examples/renderable/renderablepointcloud/pointcloud_textured_facecameraposition.asset
WeirdRubberDuck: Yes!  I have forgotten to change it form the example I copy-pasted ;)  Thank you

WeirdRubberDuck:
> I have not been able to run this example yet but a general thought I had was that maybe we should decide on one term to use for this behavior since we have a few versions of it (billboard, face camera normal, face camera position, etc.). They all do/mean the same or similar things and it can be confusing sometimes.

Not really sure what you are suggesting that we change. Where should that consistent word be used? 

The word I've used in general for this parameter since adding the "Fixed Rotation" option is "orientation" or "Orientation Mode". The "Fixed Rotation" mode does not lead to billboarding, the points just have a fixed orientation in space. So, I have removed any mentioning of billboarding unless it's in the descriptions of the two orientation modes that have a "billboarding" behavior. In summary, "Camera Position Normal" and "Camera View Direction" are billboarded orientation modes, while "Fixed Rotation" is not. 

If you want to think more about this @Roxeena, I have an open PR in the Docs repo that adds documentation about these things. Would be great if you wanted to give some feedback on that :) 
https://github.com/OpenSpace/OpenSpace-Docs/pull/26

alexanderbock:
Let's go with the different names right now and then revisit this once we have a sufficiently large number of examples.  Then we can identify multiple name changes in one go

PR #3332, opened 2024-06-22
Add Lua function to calculate the number of seconds between dates and use it in assets

alexanderbock:


data/assets/scene/solarsystem/missions/pioneer/pioneer11.asset
```diff
@@ -48,11 +51,12 @@ local Pioneer11Trail = {
       Observer = coreKernels.ID.SolarSystemBarycenter
     },
     Color = { 0.70, 0.50, 0.20 },
-    StartTime = "1973 APR 06 02:25:00.000",
-    EndTime = "1990 JAN 02 00:00:00.000",
+    StartTime = StartTime,
+    EndTime = ,
```
Roxeena, line 55: Missing EndTime

data/assets/scene/solarsystem/interstellar/c-2019_q4_borisov.asset
```diff
@@ -27,7 +27,7 @@ local C2019Q4BorisovTrail = {
     Color = { 0.9, 0.9, 0.0 },
     StartTime = "2015 JAN 01 00:00:00",
     EndTime = "2050 JAN 01 00:00:00",
-    SampleInterval = 60 * 60 * 24 -- = 86 400 seconds in 1 day
+    SampleInterval = 24 * 60 * 60 -- 1 sample per 1 day
```
Roxeena, line 30: Do we have an `openspace.time.secondsPerDay()` function? If not then maybe that could be useful here, and since we have one for a year why should we not have one for a day too? [This](https://github.com/OpenSpace/OpenSpace/blob/55e196e351b443007cc15aad7203686e85b7ff0e/include/openspace/util/timeconversion.h#L39) C++ constant can be useful to implement that.

data/assets/scene/solarsystem/missions/voyagerpioneer/voyager1_2__pioneer10_11.asset
```diff
@@ -13,8 +13,6 @@ local pioneer_horizons = asset.resource({
 })
 
 
-local SampleInterval = 24 * 60 * 60
```
Roxeena, line 16: Why not store it? Now we need to repeat the calculation 4 times and maintain all of them. Or as I suggested earlier, we could just add a function of number of seconds in a day and use that instead.

data/assets/scene/solarsystem/sssb/c2019y4atlas.asset
```diff
@@ -26,10 +29,9 @@ local C2019Y4AtlasTrail = {
     },
     Color = { 0.533333, 0.850980, 0.996078 },
     EnableFade = false,
-    StartTime = "1950 JAN 1 0:00:00",
-    EndTime = "2100 JAN 1 00:00:00",
-    SampleInterval = 52595, -- About 14h 36m 35.5200s in seconds
-    TimeStampSubsampleFactor = 1
+    StartTime = StartTime,
+    EndTime = EndTime,
+    SampleInterval = openspace.time.duration(StartTime, EndTime) / 90000
```
Roxeena, line 34: Why divide by 90000? What does it mean? It could be nice with a comment that describes it. It would be nice also to keep the old comment with how long in times this equals too as well.

src/util/time_lua.inl
```diff
@@ -374,6 +374,28 @@ namespace {
...
+ * time is before the start time, the return value is negative. If the start time is equal
+ * to the end time, the return value is 0.
+ * Both start and end times must be valid ISO 8601 date strings.
+ */
+[[codegen::luawrap]] double duration(std::string start, std::string end) {
+    using namespace openspace;
+
+    const double tStart = Time::convertTime(start);
+    const double tEnd = Time::convertTime(end);
+    return tEnd - tStart;
```
Roxeena, line 388: Do we want to add an `abs` here? In case the user mixes up start and end, or do we put it on the user to make sure it is correct together with the comment above?

data/assets/scene/solarsystem/missions/voyager/voyager1.asset
```diff
@@ -95,8 +95,9 @@ local VoyagerTrailCruiseEarthJupiter = {
     Color = { 0.70, 0.50, 0.20 },
     StartTime = "1977 SEP 05",
     EndTime = "1979 MAR 04",
-    -- 545 is the number of days between the Start and End time
-    SampleInterval = 545 * 2
+    -- 1 sample every 2d
+    SampleInterval = openspace.duration("1977 SEP 05", "1979 MAR 04") /
```
Roxeena, line 99: Maybe we should store the dates in local variables and minimize the risk of errors.

data/assets/scene/solarsystem/missions/voyager/voyager1.asset
```diff
@@ -195,8 +197,10 @@ local VoyagerTrailCruiseSaturnInf = {
     Color = { 0.70, 0.50, 0.20 },
     StartTime = "1980 NOV 16",
     EndTime = "2100 JAN 01",
-    -- 43511 is the number of days between the Start and End time
-    SampleInterval = 43511 * 2
+    -- 1 sample every 2d
+    SampleInterval = openspace.duration("1980 NOV 16", "2100 JAN 01") /
```
Roxeena, line 201: This calculation seems to be very common, in fact, every time the `duration` function is used in this PR it is used for this calculation. Isn't it better then to make a Lua function that does all of these steps instead of just one of them? It decreases the number of lines of code and will be easier to maintain.

src/util/time_lua.inl
```diff
@@ -374,6 +374,28 @@ namespace {
...
+}
+
+/**
+ * Returns the number of seconds in a Julian year, which is equal to 31557600.
+ */
+[[codegen::luawrap]] double secondsPerYear() {
+    // We could use a call to SPICE here, but the value is a constant anyway
+    return 31557600.0;
```
Roxeena, line 396: Shouldn't we just use the already defined constant [here](https://github.com/OpenSpace/OpenSpace/blob/55e196e351b443007cc15aad7203686e85b7ff0e/include/openspace/util/timeconversion.h#L37) instead of repeating the magic number?

Roxeena (commented):
Great feature that will make it easier to create trails with the correct sample interval 👍 I had a few thoughts of how this could be improved even further, see the comments below.

data/assets/scene/solarsystem/interstellar/c-2019_q4_borisov.asset
alexanderbock: We have a function now

data/assets/scene/solarsystem/sssb/c2019y4atlas.asset
alexanderbock: Line 28 just above has the "Step Size: 90000" line.  I wouldn't want to put in the exact number of dates since we use the `StartTime`, `EndTime` to make it more flexible and I think the comment would get out of date

src/util/time_lua.inl
alexanderbock: I'd rather leave it up to the user.  It'd be different if the parameters would already be in a form that would support quickly checking whether the start time is before the end time, but a user would have to call the convertTime function manually to check it in case they are interested in knowing what the ordering of the values is

data/assets/scene/solarsystem/missions/voyager/voyager1.asset
alexanderbock: I opted against that since it would lead to a lot of variables at the top of the file that would hurt readability

data/assets/scene/solarsystem/missions/voyager/voyager1.asset
Roxeena: Maybe we could have the function be named more to what it is used for, such as `getSampleInterval(...)` or `calculateSampleInterval(...)`. Since the function will almost solely be used for this purpose I don't see a problem with naming it for that specifically.

data/assets/scene/solarsystem/interstellar/oumuamua.asset
```diff
@@ -27,8 +27,7 @@ local OumuamuaTrail = {
     Color = { 0.9, 0.9, 0.0 },
     StartTime = "2014 JAN 01 00:00:00",
     EndTime = "2050 JAN 01 00:00:00",
-    SampleInterval = 86400,
-    TimeStampSubsampleFactor = 1
+    SampleInterval = 86400
```
Roxeena, line 30:
```suggestion
    SampleInterval = openspace.time.secondsPerDay()
```

ylvaselling (approved):
Neat!

PR #3340, opened 2024-06-28
Add possiblity to sort SGNs in GUI based on numerical value

WeirdRubberDuck:
Adds property the scene graph nodes for sorting the item in the GUI based on a number instead of alphabetically, with respect to its neighboring scene graph nodes in the menu. 

Let me know what you think of the property name `OrderNumber`. Didn't come up with anything better...

This PR should be combined with the following PR in the WebGUI repo: https://github.com/OpenSpace/OpenSpace-WebGuiFrontend/pull/192

Also applies the new sorting to the grids, to sort them based on their size instead of alphabetically:

image

Note that: 
- Another PR is coming for also controlling the order of the _Groups_ in the UI.
- We do not care about using this ordering in the ImGui UI. Hence why it is not applied there

### DIFF
--- a/include/openspace/scene/scenegraphnode.h
+++ b/include/openspace/scene/scenegraphnode.h
@@ -146,6 +146,8 @@ public:
     Renderable* renderable();
 
     std::string guiPath() const;
+    float guiOrderNumber() const;
+    bool useGuiOrder() const;
     bool hasGuiHintHidden() const;
     void setGuiHintHidden(bool value);
 
@@ -170,15 +172,17 @@ private:
     std::vector<std::string> _onRecedeAction;
     std::vector<std::string> _onExitAction;
 
+    ghoul::mm_unique_ptr<Renderable> _renderable;
+
     // If this value is 'true' GUIs are asked to hide this node from collections, as it
     // might be a node that is not very interesting (for example barycenters)
     properties::BoolProperty _guiHidden;
 
-    ghoul::mm_unique_ptr<Renderable> _renderable;
-
     properties::StringProperty _guiPath;
     properties::StringProperty _guiDisplayName;
     properties::StringProperty _guiDescription;
+    properties::FloatProperty _guiOrderNumber;
+    properties::BoolProperty _useGuiOrder;
 
     // Transformation defined by translation, rotation and scale
     struct {
--- a/include/openspace/scene/scenegraphnode.h
+++ b/include/openspace/scene/scenegraphnode.h
@@ -146,8 +146,6 @@ public:
     Renderable* renderable();
 
     std::string guiPath() const;
-    float guiOrderNumber() const;
-    bool useGuiOrder() const;
     bool hasGuiHintHidden() const;
     void setGuiHintHidden(bool value);
 
@@ -181,8 +179,8 @@ private:
     properties::StringProperty _guiPath;
     properties::StringProperty _guiDisplayName;
     properties::StringProperty _guiDescription;
-    properties::FloatProperty _guiOrderNumber;
-    properties::BoolProperty _useGuiOrder;
+    properties::FloatProperty _guiOrderingNumber;
+    properties::BoolProperty _useGuiOrdering;
 
     // Transformation defined by translation, rotation and scale
     struct {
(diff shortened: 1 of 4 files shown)


alexanderbock:
Could go to "Order" order "Ordering" instead?

include/openspace/scene/scenegraphnode.h
```diff
@@ -170,15 +172,17 @@ class SceneGraphNode : public properties::PropertyOwner {
     std::vector<std::string> _onRecedeAction;
     std::vector<std::string> _onExitAction;
 
+    ghoul::mm_unique_ptr<Renderable> _renderable;
+
```
alexanderbock, line 176: Why moving it further up?

src/scene/scenegraphnode.cpp
```diff
@@ -309,6 +326,15 @@ namespace {
...
             // not display, for example, barycenters
             std::optional<bool> hidden;
+
+            // If this value is specified, the scene graph node will be ordered in
+            // relation to its neighbors in the GUI based on this value, so that nodes
+            // with a higher value appear later in the list. Scene graph noeds with the
+            // same value will be sorted alphabetically.
+            //
+            // The nodes without a given value will be placed at the top at the list and
+            // sorted alphabetically.
```
alexanderbock, line 336: I thought we wanted to move the items without to the bottom of the list?

alexanderbock (approved):

include/openspace/scene/scenegraphnode.h
WeirdRubberDuck: Ah, just to group the gui properties. It bugged me that `_guiHidden` was not placed together with the other ones

But I'm fine with undoing this change

src/scene/scenegraphnode.cpp
WeirdRubberDuck: I did, just forgot to change this comment  :) Thanks

WeirdRubberDuck:
> Could go to "Order" order "Ordering" instead?

Hm, yeah, maybe. Sound a bit less like the kind of number you get when making an online purchase at least 😅

include/openspace/scene/scenegraphnode.h
alexanderbock: Nah.. i was just curious

src/scene/scenegraphnode.cpp
```diff
@@ -175,6 +175,23 @@ namespace {
         openspace::properties::Property::Visibility::Hidden
     };
 
+    constexpr openspace::properties::Property::PropertyInfo GuiOrderInfo = {
+        "GuiOrderNumber",
+        "Gui Order Number",
+        "This is an optional numerical value that will affect the sorting of this scene "
+        "graph node in relation ot its neighbors in the user interface. Nodes with the "
```
ylvaselling, line 182:
```suggestion
        "graph node in relation to its neighbors in the GUI. Nodes with the "
```

src/scene/scenegraphnode.cpp
```diff
@@ -175,6 +175,23 @@ namespace {
...
+        "same value will be sorted alphabetically.",
+        openspace::properties::Property::Visibility::Hidden
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo UseGuiOrderInfo = {
+        "UseGuiOrder",
+        "UseGuiOrder",
```
ylvaselling, line 189:
```suggestion
        "Use Gui Order",
```

src/scene/scenegraphnode.cpp
```diff
@@ -309,6 +326,15 @@ namespace {
             // scene graph node. This is most useful to trim collective lists of nodes and
             // not display, for example, barycenters
             std::optional<bool> hidden;
+
+            // If this value is specified, the scene graph node will be ordered in
+            // relation to its neighbors in the GUI based on this value, so that nodes
+            // with a higher value appear later in the list. Scene graph noeds with the
```
ylvaselling, line 332:
```suggestion
            // with a higher value appear later in the list. Scene graph nodes with the
```

ylvaselling (changes requested):

Issue #3342, opened 2024-07-01
UI issue with OptionProperty that is filled after construction


WeirdRubberDuck:
When adding a PointCloud with color mapping during runtime, the list of parameter options for the color mapping is not correctly filled - _**if it point dataset is not already cached**_. A reload of the UI fixed the problem. 

This likely submerged after update that the UI no longer requires reloads. And note that the problem is for the color mapping component, which initializes the list of options _after the construction_ (in an update call). Could it be that the time it takes to load the full dataset during runtime somehow prevents the UI from updating correctly?

Steps to reproduce: 
1. Empty your cache folder (ur just remove the data file that correpsonds to the asset added in 3., e.g. _dummydataset.csv)_
2. Start OpenSpace. E.g. with the default profile
3. Drag and drop an asset with an color mapping component into the openspace view while the software is running. E.g: "OpenSpace\data\assets\examples\renderable\renderablepointcloud\pointcloud_colormapped.asset"
4. Notice how the Parameter option list for the color mapping is not filled (see image)

![image](https://github.com/OpenSpace/OpenSpace/assets/8808894/9b31f324-5e29-4ba1-b820-6486370f9f10)

⚠️ OBS! 
- If I close down OpenSpace after this and does the same procedure, but without clearing the cache folder, the issue does not occur.
- If the asset is include din the profile, the issue does not occur, even if the data is not cached. 
- Reloading the UI fixes the problem.

ylvaselling:
The UI should probably be notified of the property change with the `notifyChangeListeners` function where the changes are made. This throws an out of range error however (potentially because the initialize function is threaded?)

alexanderbock:
Closed in e3f5b83f7476c8507dccd45ab98e4731504af532

PR #3343, opened 2024-07-01
"Fix" warning from hdf.asset

WeirdRubberDuck:
The "proximity" parameter does not exist in the dataset. Just add all the parameters to be available for color mapping, but not providing a list at all. Default will be `redshift`, which was in the previous list

I wanted this to go through a PR in case anyone has opinions about removing these preset parameters. The color mapping will likely be slightly different, due the previous custom ranges being slightly different than the automatic one (min and max value of the dataset). 

But this asset is not included in any asset per default and the points are hard to see without changing their scale, so personally I don't think it matters.

### DIFF
--- a/data/assets/scene/digitaluniverse/hdf.asset
+++ b/data/assets/scene/digitaluniverse/hdf.asset
@@ -32,11 +32,7 @@ local Object = {
     },
     Coloring = {
       ColorMapping = {
-        File = ColorMap .. "hudf.cmap",
-        ParameterOptions = {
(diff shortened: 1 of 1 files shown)


WeirdRubberDuck:
Also note that the code in this PR is subject to this issue: https://github.com/OpenSpace/OpenSpace/issues/3342.  
You may have to reload the UI if you're adding the asset during runtime and try to change the color mapping parameter from the UI...

alexanderbock (approved):
I think it makes sense :+1:

Roxeena (approved):
👍

Issue #3019, opened 2024-02-08
URL sync failing 


micahnyc:
I was trying to use the url sync to download a file, and it kept failing (my presumption was b/c of timeout).

https://api.globe.gov/search/v1/measurement/protocol/measureddate/?protocols=sky_conditions&startdate=2024-01-22&enddate=2024-01-22&geojson=TRUE&sample=FALSE

when I used openspace.downloadfile instead it worked.

micahnyc:
[globe-2024-01-22.asset.txt](https://github.com/OpenSpace/OpenSpace/files/14213336/globe-2024-01-22.asset.txt)

alexanderbock:
Could you doublecheck that it is really an issue?  For me it downloaded correctly.  I put:

```lua
local dataFolder = asset.syncedResource({
    Identifier = "globe_sky_conditions",
    Name = "globe_sky_conditions",
    Type = "UrlSynchronization",
    UseHash = false,
    FileName = "sky_conditions",
    Override = true,
    Url = "https://api.globe.gov/search/v1/measurement/protocol/measureddate/?protocols=sky_conditions&startdate=2024-01-22&enddate=2024-01-22&geojson=TRUE&sample=FALSE"
})
```
into my base.asset as I didn't have the json.lua that your asset needed.  And it correctly downloaded


Also, you can specify a `FileName` key to override the automatic detection of the filename

alexanderbock:
@micahnyc

micahnyc:
Confirmed as working

PR #3345, opened 2024-07-04
Fix url sync 

Roxeena:
There was an issue with the URL sync when several instances asked Spice for time information at the same time. Since Spice is not multithreaded it could not handle more than 1 request at a time and an exception was thrown occasionally during loading. This did not happen often but can be tested by removing all previous satellite URL sync data from the sync folder and then loading a profile with all satellites. Then all satellites will sync with the URL sync and this crash will most likely happen. This PR fixes this by making the mutex static so all URL sync instances wait their turn before asking Spice for time information. 

I am not sure what we lose by not having the `mutable` keyword for the mutex anymore but from some fast searching, I couldn't find a good way to combine the two keywords.

### DIFF
--- a/modules/sync/syncs/urlsynchronization.cpp
+++ b/modules/sync/syncs/urlsynchronization.cpp
@@ -225,7 +225,7 @@ bool UrlSynchronization::isEachFileValid() {
 
     if (ossyncVersion == "1.0") {
         // We need to mutex-protect the access to the time conversion for now
-        std::lock_guard guard(UrlSynchronization::_mutex);
+        std::lock_guard guard(_mutex);
 
         ghoul::getline(file >> std::ws, line);
         const std::string& fileIsValidToDate = line;
--- a/modules/sync/syncs/urlsynchronization.cpp
+++ b/modules/sync/syncs/urlsynchronization.cpp
@@ -265,7 +265,7 @@ bool UrlSynchronization::isEachFileValid() {
 
 void UrlSynchronization::createSyncFile(bool) const {
     // We need to mutex-protect the access to the time conversion for now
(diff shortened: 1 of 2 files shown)


Roxeena:
I'm not sure if this will also fix issue #3019 or if that is due to some other bug with the URL sync.

modules/sync/syncs/urlsynchronization.cpp
```diff
@@ -225,7 +225,7 @@ bool UrlSynchronization::isEachFileValid() {
 
     if (ossyncVersion == "1.0") {
         // We need to mutex-protect the access to the time conversion for now
-        std::lock_guard guard(_mutex);
+        std::lock_guard guard(UrlSynchronization::_mutex);
```
alexanderbock, line 228:
```suggestion
        std::lock_guard guard(_mutex);
```

No need to specify the class name directly, since this function is already part of the class

modules/sync/syncs/urlsynchronization.cpp
```diff
@@ -265,7 +265,7 @@ bool UrlSynchronization::isEachFileValid() {
 
 void UrlSynchronization::createSyncFile(bool) const {
     // We need to mutex-protect the access to the time conversion for now
-    std::lock_guard guard(_mutex);
+    std::lock_guard guard(UrlSynchronization::_mutex);
```
alexanderbock, line 268:
```suggestion
    std::lock_guard guard(_mutex);
```

Ditto

alexanderbock (approved):
The `mutable` is just there so that we could change it from inside the `const` function `createSyncFile`. But since it is `static` now, `mutable` is not needed anymore isnce it doesn't belong to a `const this` anymore

PR #3346, opened 2024-07-09
Add support for model vertex colors

Roxeena:
This PR adds support to load models with vertex colors. The vertex color is applied as a multiplication factor to any color that already exists for that vertex from any other material/texture. 

To be used together with https://github.com/OpenSpace/Ghoul/pull/111 in the Ghoul repo, more information there. 

I also converted the previous model examples into the new format. 

To test this PR, load the _examples/renderable/renderablemodel/model.asset_ file together with the equivalent PR in the ghoul repo. If the vertex colors are loaded correctly there should be 2 sets of clear checkmarks in RGB color on the model. If the vertex colors fail to load, 3 cross marks in CMY color will appear on top of the first row of checkmarks (see an example of passing test [here](https://github.com/KhronosGroup/glTF-Sample-Models/tree/main/2.0/VertexColorTest)).


data/assets/examples/renderable/renderablemodel/advanced/model_shader.asset
```diff
@@ -0,0 +1,51 @@
+-- Custom Shaders
+-- This example loads a simple model with user custom shaders. The parent is not
+-- set which defaults the position of the model to the center of the Solar System.
```
alexanderbock, line 3:
```suggestion
-- This example loads a simple model with custom shaders.
```

Since we don't set the Parent in any example, we don't need to mention it explicitly

data/assets/examples/renderable/renderablemodel/advanced/model_shader.asset
```diff
@@ -0,0 +1,51 @@
...
+
+-- Load the example model from OpenSpace servers
+-- If you want to use your own model, this block of code can safely be deleted
```
alexanderbock, line 6:
```suggestion
-- If you want to use your own model, this block of code can be safely deleted
```

data/assets/examples/renderable/renderablemodel/model.asset
```diff
@@ -0,0 +1,52 @@
+-- Basic
+-- This example loads a simple model. The parent is not set which defaults the position
+-- of the model to the center of the Solar System.
```
alexanderbock, line 3:
```suggestion
-- This example loads a simple model.
```

data/assets/examples/renderable/renderablemodel/model.asset
```diff
@@ -0,0 +1,52 @@
+-- Basic
```
alexanderbock, line 1: How about moving this into a `model_lighting.asset` and then have another basic one without lighting?  Since that would be a bit simpler still

data/assets/examples/renderable/renderablemodel/model_animation.asset
```diff
@@ -0,0 +1,58 @@
+-- Animation
+-- This example loads a model with a simple keyframe animation. The animation starts at
+-- a set time, in this case "2024 07 09 12:00:00". The parent is not set which defaults
+-- the position of the model to the center of the Solar System.
```
alexanderbock, line 4:
```suggestion
-- a set time, in this case "2024 07 09 12:00:00".
```

data/assets/examples/renderable/renderablemodel/model_animation.asset
```diff
@@ -0,0 +1,58 @@
...
+
+-- Load the asset of the Sun to illuminate the model
+local sun = asset.require("scene/solarsystem/sun/transforms")
```
alexanderbock, line 7: If we don't need the lighting for this, it might be better to not have it so that the example becomes a bit simpler

data/assets/examples/renderable/renderablemodel/model_animation_mode_bounce.asset
```diff
@@ -0,0 +1,61 @@
+-- Animation Bounce From Start
+-- This example loads a model with a simple keyframe animation. The animation starts at
+-- a set time, in this case "2024 07 09 12:00:00" and is set to bounce after that time
+-- (bounce is similar to a boomerang for videos). The parent is not set which defaults
+-- the position of the model to the center of the Solar System.
```
alexanderbock, line 5:
```suggestion
-- (bounce is similar to a boomerang for videos).
```

data/assets/examples/renderable/renderablemodel/model_animation_mode_bounce.asset
```diff
@@ -0,0 +1,61 @@
...
+
+-- Load the asset of the Sun to illuminate the model
+local sun = asset.require("scene/solarsystem/sun/transforms")
```
alexanderbock, line 8: Same here

data/assets/examples/renderable/renderablemodel/model_animation_mode_bounce_infinitely.asset
```diff
@@ -0,0 +1,61 @@
+-- Animation Bounce Infinitely
+-- This example loads a model with a simple keyframe animation. The animation starts at
+-- a set time, in this case "2024 07 09 12:00:00" and is set to bounce both before and
+-- after that time (bounce is similar to a boomerang for videos). The parent is not set
+-- which defaults the position of the model to the center of the Solar System.
```
alexanderbock, line 5:
```suggestion
-- after that time (bounce is similar to a boomerang for videos).
```

data/assets/examples/renderable/renderablemodel/model_animation_mode_bounce_infinitely.asset
```diff
@@ -0,0 +1,61 @@
...
+
+-- Load the asset of the Sun to illuminate the model
+local sun = asset.require("scene/solarsystem/sun/transforms")
```
alexanderbock, line 8: And here

data/assets/examples/renderable/renderablemodel/model_animation_mode_loop.asset
```diff
@@ -0,0 +1,61 @@
+-- Animation Loop From Start
+-- This example loads a model with a simple keyframe animation. The animation starts at
+-- a set time, in this case "2024 07 09 12:00:00" and is set to loop after that time. The
+-- parent is not set which defaults the position of the model to the center of the
+-- Solar System.
```
alexanderbock, line 5:
```suggestion
-- a set time, in this case "2024 07 09 12:00:00" and is set to loop after that time.
```

data/assets/examples/renderable/renderablemodel/model_animation_mode_loop.asset
```diff
@@ -0,0 +1,61 @@
...
+
+-- Load the asset of the Sun to illuminate the model
+local sun = asset.require("scene/solarsystem/sun/transforms")
```
alexanderbock, line 8: And here

data/assets/examples/renderable/renderablemodel/model_animation_mode_loop_infinitely.asset
```diff
@@ -0,0 +1,61 @@
+-- Animation Loop Infinitely
+-- This example loads a model with a simple keyframe animation. The animation starts at
+-- a set time, in this case "2024 07 09 12:00:00" and is set to loop both before and
+-- after that time. The parent is not set which defaults the position of the model to
+-- the center of the Solar System.
```
alexanderbock, line 5:
```suggestion
-- after that time.
```

data/assets/examples/renderable/renderablemodel/model_animation_mode_loop_infinitely.asset
```diff
@@ -0,0 +1,61 @@
...
+
+-- Load the asset of the Sun to illuminate the model
+local sun = asset.require("scene/solarsystem/sun/transforms")
```
alexanderbock, line 8: And here

data/assets/scene/solarsystem/planets/earth/satellites/misc/tiangong.asset
```diff
@@ -23,7 +23,7 @@ local omm = asset.resource({
 local TiangongPosition = {
   Identifier = "TiangongPosition",
   Parent = transforms.EarthInertial.Identifier,
-  BoundingSphere = 54.5, -- half the width
+  BoundingSphere = 27.8, -- half the width
```
alexanderbock, line 26: Why this change?

alexanderbock (approved):

data/assets/scene/solarsystem/planets/earth/satellites/misc/tiangong.asset
Roxeena: Because this was straight copy pasted from the ISS asset and Tiangong is smaller. This is half the size I found on [Wikipedia](https://en.wikipedia.org/wiki/Tiangong_space_station). I felt I was not really able to fly close to it.

WeirdRubberDuck: Maybe this should be a separate commit?

WeirdRubberDuck (approved):

data/assets/scene/solarsystem/planets/earth/satellites/misc/tiangong.asset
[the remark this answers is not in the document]
Roxeena: I think it fits since I think it is a bug and it is related to models.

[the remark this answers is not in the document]
WeirdRubberDuck: Yeah but the fix of the bug becomes "invisible" if its part of this PR - especially since there is no issue for the bug to be closed. You can't see from the history that this was a problem before this PR. Do you get what I mean? 

If you don't want a separate commit for it, at least add a sentence to the description of this PR explaining that this issue is fixed :) For some traceability

data/assets/examples/renderable/renderablemodel/model_animation.asset
```diff
@@ -0,0 +1,49 @@
+-- Animation
+-- This example loads a model with a simple keyframe animation. The animation starts at
```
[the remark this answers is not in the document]
Roxeena: I'm not really sure how to explain the keyframe animation part in a way that does not make it more complex. I wanted to have the keyword "_keyframe_ animation" in there to make it somewhat clearer that we **only** support that type of animation but I could also leave it as only "animation" and leave the documentation for what types are supported.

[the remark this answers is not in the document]
WeirdRubberDuck: 👍 I agree with you, in that it would make it more complex. Removing the "keyframe" would solve it, IMO. 

It's also future-proofing to remove it. Who knows, one day we might support other types of animations and then this example does not have to be updated ;)

PR #3353, opened 2024-07-18
Add option to specify a default texture for added exoplanets

### LINKED ISSUE #3280: Exoplanet module - Default planet texture option
A lot of users like to add their own textures to their exoplanets. Today someone ask they could set a default texture for all the planets that are created. I figured since we have this option for the star texture why not create an option for planet texture.

WeirdRubberDuck:
Add a similar setting like we have for the stars

closes #3280

### DIFF
--- a/modules/exoplanets/exoplanetsmodule.cpp
+++ b/modules/exoplanets/exoplanetsmodule.cpp
@@ -373,7 +373,7 @@ void ExoplanetsModule::internalInitialize(const ghoul::Dictionary& dict) {
     }
 
     if (p.planetDefaultTexture.has_value()) {
-        _noDataTexturePath = p.planetDefaultTexture->string();
+        _planetDefaultTexturePath = p.planetDefaultTexture->string();
     }
 
     if (p.noDataTexture.has_value()) {
--- a/modules/exoplanets/exoplanetsmodule.cpp
+++ b/modules/exoplanets/exoplanetsmodule.cpp
@@ -94,6 +94,15 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo PlanetDefaultTextureInfo = {
+        "PlanetDefaultTexture",
+        "Planet Default Texture",
+        "The path to an image that should be used by default for the planets in all "
+        "added exoplanet systems. If not specified, the planets are rendered without a "
+        "texture when added.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
     constexpr openspace::properties::Property::PropertyInfo OrbitDiscTextureInfo = {
         "OrbitDiscTexture",
         "Orbit Disc Texture",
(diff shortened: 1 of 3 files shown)


modules/exoplanets/exoplanetsmodule.cpp
```diff
@@ -354,6 +372,10 @@ void ExoplanetsModule::internalInitialize(const ghoul::Dictionary& dict) {
         _starGlareTexturePath = p.starGlareTexture->string();
     }
 
+    if (p.planetDefaultTexture.has_value()) {
+        _noDataTexturePath = p.planetDefaultTexture->string();
```
alexanderbock, line 376:
```suggestion
        _planetDefaultTexturePath = p.planetDefaultTexture->string();
```

modules/exoplanets/exoplanetsmodule_lua.inl
```diff
@@ -286,6 +303,19 @@ void createExoplanetSystem(const std::string& starName,
...
+            planetLayers = std::format(
+                "ColorLayers = {{ {} }}",
+                "{"
+                    "Identifier = 'PlanetTexture',"
+                    "FilePath = openspace.absPath('" + formatPathToLua(planetTexture) + "'),"
+                    "BlendMode = 'Color',"
+                    "Enabled = true"
+                "}"
+            );
+        }
```
alexanderbock, line 317:
```suggestion
            planetLayers = std::format(
                "ColorLayers = {{"
                "  {{"
                "    Identifier = 'PlanetTexture',"
                "    FilePath = openspace.absPath('{}'),"
                "    BlendMode = 'Color',"
                "    Enabled = true"
                "  }}"
                "}}",
                formatPathToLua(planetTexture)
            );
        }
```

Might be easier to read?

modules/exoplanets/exoplanetsmodule.cpp
WeirdRubberDuck: D: good 👀

modules/exoplanets/exoplanetsmodule_lua.inl
WeirdRubberDuck: Heh, yeah I agree that this is a bit odd.  
I'd just remove the std::format entirely, I think. To be more consistent with the rest of the function :)

alexanderbock (approved):

modules/exoplanets/exoplanetsmodule.cpp
```diff
@@ -354,6 +372,10 @@ void ExoplanetsModule::internalInitialize(const ghoul::Dictionary& dict) {
         _starGlareTexturePath = p.starGlareTexture->string();
     }
 
+    if (p.planetDefaultTexture.has_value()) {
+        _planetDefaultTexturePath = p.planetDefaultTexture->string();
+    }
+
```
ylvaselling, line 378: Not really about your addition, but seems to me these if statements could use the `value_or` function like the first enabled property?

ylvaselling (approved):

modules/exoplanets/exoplanetsmodule.cpp
WeirdRubberDuck: No, it cannot.  Note that we do not get the value directly, but use the `string()` operator to get a string version of the `std::filesystem:path`

The `StringProperty` is not compatible with values of `std::filesystem:path` directly, but we want to use that to get the correct verification of the value in the asset. (However, as a side note, I think that we should probably have a specialized property for files, since we use them so often and would want special features in the UI that make them easier to use, for example)

PR #3369, opened 2024-08-06
Update to newer Qt version

alexanderbock:
To check here is whether this still compiles with older Qt version.  Whoever tests this, please put your Qt version in a post so that we can determine what the minimum version is


GPayne:
Worked for me on Windows11 with Qt 6.6.2.
Also verified that this doesn't adversely affect the linux build.

PR #3378, opened 2024-08-07
Lua Sandboxing

alexanderbock:
This change disables the `require` function in Lua and the `io`, `os`, and `package` tables in an effort to better secure our Lua interpreter against potentially malicious code. There is a new `openspace.cfg` setting `SandboxedLua` that toggles that behavior, with it being _enabled_ by default. 

This change required changing the constellation asset files as they previously relied on the `io` functions. This dependency has been replaced with OpenSpace-native functions instead.

To test in here is whether we have any asset files that no longer work as they require the use of any of the now-removed tables.  If that is the case, the assets need to be rewritten to use (potentially new) OpenSpace functions instead.

### DIFF
--- a/data/assets/global/localbookmarks.asset
+++ b/data/assets/global/localbookmarks.asset
@@ -13,15 +13,11 @@ end
 
 -- Create bookmarks file if it does not exist
 if not openspace.fileExists(localBookmarks) then
-  local file = io.open(localBookmarks, "w")
-  file:write(
-    "Group (optional),Name (required),Globe (optional),Lat (required if globe)," ..
-    "Lon (required if globe),Altitude (optional if globe),x (required if not globe)," ..
-    "y (required if not globe),z (required if not globe),Scale (optional)," ..
-    "LineWidth (optional)\n" ..
-    "NASA,Kennedy Space Center,Earth,28.6658276,-80.70282839,,,,,,\n"
+  openspace.downloadFile(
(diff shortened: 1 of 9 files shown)


GPayne:
I'm adding this comment here so it's associated with the PR (instead of in the ghoul repo). There aren't too many functions in `io` or `os`, and some of them could still be useful. How about importing both libs, and then only setting these functions to `nil`:
```
io.output
io.write
io.flush
io.popen
io.tmpfile
os.execute
os.exit
os.remove
os.rename
```
This would prevent modifying or deleting files. It would not prevent unauthorized reading of files for example, if we are concerned with that.

alexanderbock:
What are the functions that you are specifically looking for?  I think it will be easier to document if we just say no `io` or `os` at all.  We could add our own `openspace.` functions to do the things that we get rid of this way, for example?

GPayne:
That would leave functions like:
```
io.open
io.read
io.lines
os.clock
os.date
os.time
os.difftime
os.getenv
```
The `io` functions we were already using.

alexanderbock:
I'd say that we'd want to encourage people to use our versions of those functions instead.  For the `io` functions, we have `openspace.readFile` already.  And similar functions for the timing

GPayne:
If we now have our own functions that cover the possible use cases for those `io` and `os` functions, then that would resolve my comment.

PR #3385, opened 2024-08-18
Add example for StaticTranslation

alexanderbock:


data/assets/examples/translation/statictranslation/static.asset
```diff
@@ -0,0 +1,29 @@
+-- Basic
+-- This asset creates a set of coordinate axes that are offset from their origin
```
WeirdRubberDuck, line 2: "that are offset from their origin by a fixed and static amount" is a little confusing to me. I think the reason is that the origin will move with the object, and now it says "their origin", which implies the origin of the coordinate axes

Maybe it's enough to say "that is moved by a fixed amount" or somehow specify that it's "the origin", rather than "their origin"?

WeirdRubberDuck (approved):

data/assets/examples/translation/statictranslation/static.asset
alexanderbock: Changed it to "original position".  Wanted to make sure that it is relative to the parents coordinate system without saying that (since the example doesn't have a parent)

PR #3389, opened 2024-08-22
Improvements to the ScriptLog logging

### LINKED ISSUE #2067: Add "tag" to scripts in scriptLog that are not triggered by user interaction
We have started triggering more and more scripts automatically from the C++, "cluttering" the scriptLog with scripts that might be confusing for the user. They make it harder to find the scripts that are triggered by changing properties in the UI, for example. 

A suggestion is to add a tag in the shape `[[ tag ]]` to the beginning of all scripts that are triggered from the C++ code, for example from the SkyBrowser or an Action

### LINKED ISSUE #3079: The script log could include whether a script was synced or not
For example, using some kind of tag system like `[[local]]` `[[synced]]` at the start of each line

Would be useful when debugging scripts that are triggered from other scripts, for example

alexanderbock:
# No dependent scripts
This PR prevents automatically generated scripts from showing up in the scriptlog, specifically when executing actions. Before this PR, the `triggerAction` command would be shown in the ScriptLog together with the contents of the action script:

```lua
return openspace.action.triggerAction("os.FadeToBlack")
    if openspace.propertyValue("RenderEngine.BlackoutFactor") > 0.5 then
      openspace.setPropertyValueSingle("RenderEngine.BlackoutFactor", 0.0, 3)
    else
      openspace.setPropertyValueSingle("RenderEngine.BlackoutFactor", 1.0, 3)
    end
```

with this PR, only the `triggerAction` remains:
```lua
return openspace.action.triggerAction("os.FadeToBlack")
```

This closes #2067.  The JSON solution in the issue was not attempted since
  1. It is not trivial to know when we can close the scriptlog file to finalize the file and add the closing `]}` to make it a valid JSON file
  2. There is benefit in having the ScriptLog be human-readable and more importantly human-copy-and-pastable, which JSON would prevent.

# Verbose logging
This PR also adds a new setting into the `openspace.cfg` called `VerboseScriptLog` which is disabled by default.  If it is enabled, all of the dependent scripts from above will be printed again with a `--[[generated]]` prefix.  Further more every script will include whether it was `--[[sync]]` or `-[[!sync]]` and if it was `--[[sendRemote]]`/`--[[!sendRemote]]`.  This is primarily used for debugging purposes and I don't forsee a user needing to look into this.  This closes #3079


include/openspace/scripting/scriptengine.h
```diff
@@ -83,18 +85,13 @@ class ScriptEngine : public Syncable {
     void addLibrary(LuaLibrary library);
     bool hasLibrary(const std::string& name);
 
-    bool runScript(const std::string& script,
-        const ScriptCallback& callback = ScriptCallback());
-    bool runScriptFile(const std::filesystem::path& filename);
```
ylvaselling, line 88: Comment at line 44 should be changed if this function is removed

include/openspace/scripting/scriptengine.h
```diff
@@ -49,15 +49,17 @@ namespace openspace::scripting {
...
-        ScriptCallback callback;
+    struct Script {
+        BooleanType(ShouldBeSynchronized);
+        BooleanType(ShouldSendToRemote);
+        BooleanType(ShouldBeLogged);
+        using Callback = std::function<void(ghoul::Dictionary)>;
+
+        std::string code;
+        ShouldBeSynchronized synchronized = ShouldBeSynchronized::Yes;
+        ShouldSendToRemote sendToRemote = ShouldSendToRemote::Yes;
```
ylvaselling, line 60: I think it would be good with a comment to explain the difference between remote and synchronized

src/interaction/keybindingmanager.cpp
```diff
@@ -40,7 +40,7 @@ namespace openspace::interaction {
 void KeybindingManager::keyboardCallback(Key key, KeyModifier modifier, KeyAction action)
 {
     if (action == KeyAction::Press || action == KeyAction::Repeat) {
-        // iterate over key bindings
+        // iterate over kfey bindings
```
ylvaselling, line 43:
```suggestion
        // iterate over key bindings
```

src/rendering/screenspacerenderable.cpp
```diff
@@ -434,11 +434,11 @@ ScreenSpaceRenderable::ScreenSpaceRenderable(const ghoul::Dictionary& dictionary
...
         // when this triggerProperty was pressed in the gui, therefor it has already been
         // synced and sent to the connected nodes and peers
-        global::scriptEngine->queueScript(
-            std::move(script),
-            scripting::ScriptEngine::ShouldBeSynchronized::No,
-            scripting::ScriptEngine::ShouldSendToRemote::No
-        );
+        global::scriptEngine->queueScript({
+            .code = std::move(script),
+            .synchronized = scripting::ScriptEngine::Script::ShouldBeSynchronized::No,
```
ylvaselling, line 439: Isn't this default?

src/scene/scene.cpp
```diff
@@ -554,11 +554,12 @@ void Scene::updateInterpolations() {
...
                 // and peers
-                global::scriptEngine->queueScript(
-                    std::move(i.postScript),
-                    scripting::ScriptEngine::ShouldBeSynchronized::No,
-                    scripting::ScriptEngine::ShouldSendToRemote::No
-                );
+                using Script = scripting::ScriptEngine::Script;
+                global::scriptEngine->queueScript({
+                    .code = std::move(i.postScript),
+                    .synchronized = Script::ShouldBeSynchronized::No,
```
ylvaselling, line 560: also default?

modules/base/rendering/pointcloud/renderableinterpolatedpoints.cpp
```diff
@@ -187,11 +187,11 @@ RenderableInterpolatedPoints::Interpolation::Interpolation()
...
         // No syncing, as this was triggered from a property change (which happened
         // based on an already synced script)
-        global::scriptEngine->queueScript(
-            script,
-            scripting::ScriptEngine::ShouldBeSynchronized::No,
-            scripting::ScriptEngine::ShouldSendToRemote::No
-        );
+        global::scriptEngine->queueScript({
+            .code = script,
+            .synchronized = scripting::ScriptEngine::Script::ShouldBeSynchronized::No,
```
WeirdRubberDuck, line 192: eek, the path to this boolean type became very long now

Oh well

modules/iswa/rendering/iswakameleongroup.cpp
```diff
@@ -180,11 +179,10 @@ void IswaKameleonGroup::clearFieldlines() {
...
             LDEBUG("Removed fieldlines: " + std::get<0>(seedPath.second));
 
-            global::scriptEngine->queueScript(
-                "openspace.removeSceneGraphNode('" + std::get<0>(seedPath.second) + "')",
-                scripting::ScriptEngine::ShouldBeSynchronized::Yes,
-                scripting::ScriptEngine::ShouldSendToRemote::Yes
+            const std::string script = std::format(
+                "openspace.removeSceneGraphNode('{}')", std::get<0>(seedPath.second)
             );
+            global::scriptEngine->queueScript({ script });
```
WeirdRubberDuck, line 185: It's a bit annoying that we have to add these brackets for every `queueScript` now... How about making a version of the `queueScript` that takes just a string and does this for us ?

It is the most common use case, after all.

src/rendering/screenspacerenderable.cpp
WeirdRubberDuck: It seems like as part of this PR, it was set to sync per default.

WeirdRubberDuck (commented):
Not completely done with testing, but leaving for the weekend and don't know when I'll get back to this. 

But shouldn't things like adding exoplanet systems be tagged somehow? 
Does not seem like it is, currently

src/scene/scene.cpp
WeirdRubberDuck: It seems like as part of this PR, is was set to `Yes` per default

src/rendering/screenspacerenderable.cpp
alexanderbock: The default is that the scripts are synchronized and sendtoremote.  In this case we don't want it to happen since the `onChange` will already get triggered correctly through the scripting system

modules/base/rendering/pointcloud/renderableinterpolatedpoints.cpp
alexanderbock: It's partially on purpose to not make it easy to write this, since it could result in hard-to-detect errors and we want to encourage the use of the default values

modules/iswa/rendering/iswakameleongroup.cpp
alexanderbock: Very good point.  Changed that

Roxeena:
I can't run this branch, it crashes in the string library when the script to set up the WebGUI is run, this was with the empty profile.

alexanderbock:
@Roxeena  Weird;  but should be fixed now

src/engine/configuration.cpp
```diff
@@ -134,6 +134,11 @@ namespace {
         // from previous runs) will be silently overwritten
         std::optional<std::string> scriptLog;
 
+        // If this value is set to `true`, the script log will contain scripts that have
```
Roxeena, line 137: I'm confused. So if I set this to true, will the script log work as it used to before this PR again? Logging everything even stuff inside actions etc? From the description of the PR, it sounded like it would just add tags of whether it was synced or sent to remote or not, but it would still behave the same way as if it is false.

Roxeena: To add to this, I think it could be good to have a setting that does turn it back the way it was with logging everything. In case an advanced user (or us) wants to copy a line of code inside an action and make something of their own with it.

Roxeena (commented):
I haven't tested it too much yet but I noticed that if an action has a delayed script, for example, the hide constellation lines action that also disables them when they are faded out fully, the delayed script is still logged. Should it? Since it is part of an action I would intuitively think not.

WeirdRubberDuck:
Feel like this might require a discussion. Currently, only the actions get the `--[[generated]]` tag, but we talked about that maybe being added to any scripts not triggered directly by the user or UI. In my mind, that means that it would for example apply to the scripts that add the scene graph nodes for the exoplanets, which is triggered by another script call (a bit similar to the actions). 

However, making that a "generated" script would make it quite difficult for the user to add their own exoplanet system, without adding the verbose logging. Which, IMO, is a little TOO verbose to want to look at, with the sync tags being added to every line :)  (Is this verboseness something to discuss?)

However, maybe we actually want to make the scriptLog an HTML webpage where the user can filter by or show the different tags, similarly to the other Log page. In that case, the verboseness of the actual data file would not be a problem. 

A bt of a braindump here, but hope you get my point

include/openspace/scripting/scriptengine.h
```diff
@@ -41,23 +41,41 @@ namespace openspace::scripting {
...
+    struct Script {
+        BooleanType(ShouldBeSynchronized);
+        BooleanType(ShouldSendToRemote);
+        BooleanType(ShouldBeLogged);
+        using Callback = std::function<void(ghoul::Dictionary)>;
+
+        /// The Lua script that should be executed
+        std::string code;
+
+        /// Determines whether a script should be sent to computer that are in the same
```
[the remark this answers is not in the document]
alexanderbock: It's actually:
https://docs.openspaceproject.com/en/latest/contribute/development/coding-style.html#member-variables-and-enum-members-are-documented-with-everything-else-with

include/openspace/scripting/scriptengine.h
```diff
@@ -41,23 +41,41 @@ namespace openspace::scripting {
...
+        /// be tightly locked
+        ShouldBeSynchronized synchronized = ShouldBeSynchronized::Yes;
+
+        /// Determines whether a script should be send to a distant OpenSpace instance
+        /// that is not part of the same cluster. This is used in the ParallelConnection
+        /// feature that can connect OpenSpace instances in a less tight connection
+        ShouldSendToRemote sendToRemote = ShouldSendToRemote::Yes;
+
+        /// Determines whether the script should be logged to a local script log file.
+        /// Note that this might be overwritten if the user requested a verbose log file
```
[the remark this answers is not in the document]
alexanderbock: The openspace.cfg has a setting to toggle whether the scriptlog is verbose or not

modules/imgui/src/guispacetimecomponent.cpp
```diff
@@ -210,11 +195,10 @@ void GuiSpaceTimeComponent::render() {
     if (dateChanged) {
         // No sync or send because time settings are always synced and sent to the
         // connected nodes and peers
-        global::scriptEngine->queueScript(
-            std::format("openspace.time.setTime([[{}]])", std::string_view(Buffer)),
-            scripting::ScriptEngine::ShouldBeSynchronized::No,
-            scripting::ScriptEngine::ShouldSendToRemote::No
+        const std::string script = std::format(
+            "openspace.time.setTime([[{}]])", std::string_view(Buffer)
```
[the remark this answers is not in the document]
alexanderbock: Yea, the default value has changed, but there was also a bug where these scripts _should_ have been synchronized, but weren't before

src/engine/configuration.cpp
alexanderbock: I made it a bit more obvious in the comment now.

`false`:  The ScriptLog is missing generated scripts as we define it wherever they occur (with Actions right now)
`true`:  The ScriptLog behaves the same as it did before the PR _and_ the `--[[sync]]`, `--[[generated]]` etc tags are added.

So there is no way to get the generated scripts without them being marked as generated

alexanderbock:
I didn't want to get into the rabbit hole of deciding everywhere whether we want to print the script or not;  i think that is a better discussion to be left for the individual places.

Regarding the HTML page, I was considering making the scriptlog a JSON file instead, having the same filtering possibility there.  However, a big problem that I couldn't fix properly is that we want to scriptlog to be a functional file even though OpenSpace might have crashed before finalizing the scriptlog.  So any file format that has a footer was presenting a problem.

PR #3398, opened 2024-09-06
Sync out the anchor node, issue 2854

### LINKED ISSUE #2854: Night Sky actionissue in dome
The action to show Cardinal directions is local and only shows on master and not on the dome.

Roxeena:
closes #2854.

What caused the issue, was that the anchor node was not properly set on the cluster nodes since the `updateCamera` function is not run on the nodes. To solve this the anchor node is always synced out to the cluster nodes to ensure that all nodes and master have the same anchor node.


src/navigation/orbitalnavigator.cpp
```diff
@@ -2084,6 +2085,17 @@ double OrbitalNavigator::rotationSpeedScaleFromCameraHeight(
         1.0;
 }
 
+// This should only be run on nodes and not on master
```
WeirdRubberDuck, line 2088: Maybe add a `ghoul_assert` for this, instead of this comment?  

Or, put the comment in the header file instead, as part of a doxygen comment for the function. IIRC, we avoid comments over function in the cpp files

alexanderbock: Adding a ghoul_assert makes sense I think

alexanderbock (approved):
:+1:

PR #3417, opened 2024-09-16
Feature/UI launcher accessibility

engbergandreas:
This PR fixes a bunch of our accessibility issues with the launcher UI 

If you have time to test it with a screen reader, please do. Otherwise, the most important thing is to check if the accessibility descriptions make sense and are intuitive. 

Key improvements
* If there are unsaved changes during profile editing and you exit the editor you are prompted to save before exit
* You can exit the launcher via "esc" key
* Tabbing should work in the entire UI except for some places in "Actions and Keybindings" and the "Window config"
* Improved error message outputs


alexanderbock:
In the settings window, the "Property Visibility" and the "Layer Server" dropdown don't get any visual indicator that they are selected when tabbing through the controls.  Is that something that is easy to fix?

alexanderbock:
Got the "unsaved changes" dialog in the profile editor by:
1. Open Launcher
2. With the "default" profile selected, press "Edit"
3. Immediately "Cancel"

engbergandreas:
> Got the "unsaved changes" dialog in the profile editor by:
> 
>     1. Open Launcher
> 
>     2. With the "default" profile selected, press "Edit"
> 
>     3. Immediately "Cancel"

Yep we actually have to update our default profiles, there is a new key in `time: is_paused` that is added when using the save profile function

engbergandreas:
> In the settings window, the "Property Visibility" and the "Layer Server" dropdown don't get any visual indicator that they are selected when tabbing through the controls. Is that something that is easy to fix?

They do get this indicator but perhaps we should update the color since its not very obvious: 
image

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -277,79 +328,52 @@ QWidget* LauncherWindow::createCentralWidget() {
...
+    editProfileButton->setAccessibleName("Edit profile");
 
     QPushButton* newProfileButton = new QPushButton("New", centralWidget);
     connect(
         newProfileButton, &QPushButton::released,
         [this]() {
-            openProfileEditor("", true);
-        }
+        openProfileEditor("", true);
+    }
```
alexanderbock, line 360:
```suggestion
            openProfileEditor("", true);
        }
```

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -368,7 +392,44 @@ QWidget* LauncherWindow::createCentralWidget() {
...
+                this,
+                "Empty Profile",
+                "Cannot launch with an empty profile"
+            );
+        }
+        else {
+            _shouldLaunch = true;
+            close();
+        }
+    }
```
alexanderbock, line 426:
```suggestion
            if (_profileBox->currentText().isEmpty()) {
                QMessageBox::critical(
                    this,
                    "Empty Profile",
                    "Cannot launch with an empty profile"
                );
            }
            else {
                _shouldLaunch = true;
                close();
            }
        }
```

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -781,6 +850,37 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
         savePath,
         this
     );
+
+    // Check whether there are unsaved changes from the profile editor.
+    connect(&editor, &ProfileEdit::raiseExitWindow, this,
+        [this, &editor, &savePath, &p, &profile]() {
```
alexanderbock, line 856:
```suggestion
    connect(
        &editor,
        &ProfileEdit::raiseExitWindow,
        [this, &editor, &savePath, &p, &profile]() {
```

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -781,6 +850,37 @@ void LauncherWindow::openProfileEditor(const std::string& profile, bool isUserPr
...
+        // If this is a new profile we want to prompt the user
+        if (!std::filesystem::exists(origPath)) {
+            editor.promptUserOfUnsavedChanges();
+            return;
+        }
+
+        // Save the profile to a tmp file
+        const std::string tmpPath = std::format("{}{}.tmp",
+            savePath, editor.specifiedFilename()
+            );
```
alexanderbock, line 868:
```suggestion
        const std::string tmpPath = std::format(
            "{}{}.tmp",
            savePath, editor.specifiedFilename()
        );
```

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -416,39 +415,48 @@ bool HorizonsDialog::isValidInput() {
...
+        _stepEdit->setFocus();
         styleLabel(_stepLabel, IsDirty::Yes);
         return false;
     }
     // Numerical
     bool couldConvert = false;
     const int32_t step = _stepEdit->text().toInt(&couldConvert);
     if (!couldConvert) {
-        _errorMsg->setText(QString::fromStdString(std::format(
+        QMessageBox::critical(this, "Error", QString::fromStdString(std::format(
```
alexanderbock, line 443:
```suggestion
        QMessageBox::critical(
            this,
            "Error",
            QString::fromStdString(std::format(
```

apps/OpenSpace/ext/launcher/src/profile/profileedit.cpp
```diff
@@ -188,17 +194,19 @@ void ProfileEdit::createWidgets(const std::string& profileName) {
...
+        _keybindingsEdit->setReadOnly(true);
+        _keybindingsEdit->setAccessibleName("Loaded action and keybindings");
+        container->addWidget(_keybindingsEdit, 1, 0, 1, 3);
+
         QPushButton* keybindingsProperties = new QPushButton("Edit");
         connect(
             keybindingsProperties, &QPushButton::clicked,
             this, &ProfileEdit::openKeybindings
         );
+        keybindingsProperties->setAccessibleName("Edit actions & keybindings");
```
alexanderbock, line 207:
```suggestion
        keybindingsProperties->setAccessibleName("Edit actions and keybindings");
```

alexanderbock:
> > In the settings window, the "Property Visibility" and the "Layer Server" dropdown don't get any visual indicator that they are selected when tabbing through the controls. Is that something that is easy to fix?
> 
> They do get this indicator but perhaps we should update the color since its not very obvious: image

Darkmode strikes again:

image
image

apps/OpenSpace/ext/launcher/resources/qss/launcher.qss
```diff
@@ -22,7 +18,7 @@ LauncherWindow  QLabel {
 }
 
 LauncherWindow  QLabel#label_choose, QLabel#label_options {
-  color: #dddddd;
+  color: rgb(255, 255, 255);
```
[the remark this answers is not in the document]
engbergandreas: 👎 default font color is black

WeirdRubberDuck (approved):
Found some issues related to tabbing and focusing, but I think it makes sense to create new issues for those. So I'll do that 

Especially, since several of them would be solved by slightly redesigning parts of the UI instead of just addressing the specific problem :)

WeirdRubberDuck:
One tab focus-related issue that seems quite small/specific and could be part of this PR: 

Tabbing through the Action pane moves the focus to the save button for an action, even if no action is being edited. That is, this button: 
image

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
```diff
@@ -296,6 +295,7 @@ void HorizonsDialog::createWidgets() {
         _startEdit = new QDateTimeEdit;
         _startEdit->setDisplayFormat("yyyy-MM-dd  T  hh:mm:ss");
         _startEdit->setDate(QDate::currentDate().addYears(-1));
+        _startEdit->setAccessibleName("Set start time");
```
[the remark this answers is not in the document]
engbergandreas: When trying with the NVDA screenreader it will read something like "set start time, 2023-01-09 T 00:00:00" and since the format is fixed maybe it's fine? Tabbing through the input field will go year -> month -> day -> time and at each point, it will do a readout of the value. 

but I guess adding a description like so `setAccessibleDescription("time format: YYYY-MM-DD T HH:MM:SS") isn't a bad idea?

engbergandreas:
> One tab focus-related issue that seems quite small/specific and could be part of this PR:
> 
> Tabbing through the Action pane moves the focus to the save button for an action, even if no action is being edited. That is, this button: 

I'm not sure I can reproduce this? That part of the panel should only be enabled if you have clicked the `+` sign to add a new action, if you cancel it should be disabled again and not be part of the tab order?

apps/OpenSpace/ext/launcher/src/profile/horizonsdialog.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: If the format is read, I guess that's fine. Was mostly a thought. 

And what I meant was rather not having it spell out the entire format, but maybe mention that its's given as a ISO 8601-like date string, in UTC. That is probably information that everyone could benefit from, however, so might be worth adding that information in some other way :)

PR #3427, opened 2024-09-18
Add the ability to query RAM and VRAM usage

alexanderbock:
This PR adds two new Lua functions `openspace.ramInUse()` and `openspace.vramInUse()` that report the amount of system memory and video memory used by OpenSpace (in bytes).

This value is also reported in the "Memory Information" ImGUI panel:
image

And as a plot in Tracy:
image

### DIFF
--- a/modules/imgui/src/guimemorycomponent.cpp
+++ b/modules/imgui/src/guimemorycomponent.cpp
@@ -67,9 +67,9 @@ void GuiMemoryComponent::render() {
     _isCollapsed = ImGui::IsWindowCollapsed();
 
     uint64_t ram = global::openSpaceEngine->ramInUse();
-    ImGui::Text("RAM: %u (%.2f MiB)", ram, ram / (1024.f * 1024.f));
+    ImGui::Text("RAM: %u (%f MiB)", ram, ram / (1024.f * 1024.f));
     uint64_t vram = global::openSpaceEngine->vramInUse();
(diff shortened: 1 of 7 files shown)


modules/imgui/src/guimemorycomponent.cpp
```diff
@@ -65,11 +66,16 @@ void GuiMemoryComponent::render() {
     _isEnabled = v;
     _isCollapsed = ImGui::IsWindowCollapsed();
 
+    uint64_t ram = global::openSpaceEngine->ramInUse();
+    ImGui::Text("RAM: %u (%.2f MiB)", ram, ram / (1024.f * 1024.f));
```
WeirdRubberDuck, line 70: Who's our intended audience for this?  Mebibytes are seemingly quite rarely used as a unit 😅 

And depending who we expect to use this feature, we should probably consider what unit we use. I'm personally fine with whichever

WeirdRubberDuck: Also, I think we can skip the `.2f` here, considering the size and context of the numbers? :)

WeirdRubberDuck (approved):
Nice!  Di not try the tracy version. 

Tested on Windows

modules/imgui/src/guimemorycomponent.cpp
alexanderbock: Removed the .2f.   I'll leave the MiB since that is the correct unit for dividing by 1024 ;)   And also I see this ImGui panel only being used by Devs;  in general a regular user shouldn't need to look at this

PR #3432, opened 2024-09-20
Split up JWST and Lagrange point assets into smaller assets

### LINKED ISSUE #3425: JWST asset needs to be separated into multiple smaller assets
The [James webb asset](https://github.com/OpenSpace/OpenSpace/blob/master/data/assets/scene/solarsystem/telescopes/jwst/jwst.asset) includes a lot more than just the model of the James Webb telescope (which is what I would have expected).

It seemingly loads everything we have related to the James Webb mission, including the Band, the FoV of the telescope, multiple scene graph nodes showing L2, and the actions and mission file. These should all be split into individual assets, so a user can add only the part they want. 

All these individual parts are also enabled per default, which means that if I want to show only the model, I have to first disable all the stuff that I don't want/need in my scene, which is very annoying. 

Short version: 
* There is no way to add just the JWST model. There should be.

WeirdRubberDuck:
And avoid using hardcoded identifier names, or timestamps that are the same across multiple files. Now, it is possible to add just the JWST model, or just the L2 to Sun Line, or just the JWST safe viewing band, etc.

This change leads to no breaking change for users that have included the `jwst.asset` or `lagrange_points.asset`

However, it is a _**breaking change**_ for users that had included the `l2.asset`, `l3.asset`, `l4.asset`, or `l5.asset`assets separately. 

Closes #3425


data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/transforms.asset
```diff
@@ -0,0 +1,57 @@
...
+  GUI = {
+    Name = "L2 Position",
+    Path = "/Solar System/Planets/Earth/Lagrange Points",
+    Hidden = true
+  }
+}
+
+-- This ref. frame co-revolves with L2 and is needed for a more intuitive trail of JWST in
+-- relation to L2
+local L2CoRevFrame = {
```
alexanderbock, line 27: Should we move this over to the JWST portion of the filesystem?

data/assets/scene/solarsystem/telescopes/jwst/fieldofview.asset
```diff
@@ -0,0 +1,59 @@
...
+  Tag = { "mission_jwst_fov" },
+  GUI = {
+    Name = "JWST Field of View",
+    Path = "/Solar System/Telescopes/JWST",
+    Description = [[
+      The field of view for the James Webb Space Telescope at its current position.
+    ]]
+  }
+}
+
```
alexanderbock, line 38:
```suggestion


```

data/assets/scene/solarsystem/telescopes/jwst/jwst.asset
```diff
@@ -1,176 +1,31 @@
...
-asset.export(JWSTLabel)
-
 
 
 asset.meta = {
   Name = "James Webb Space Telescope",
   Description = [[
-    James Webb Space Telescope animated model, safe viewing band, field of view and label.
+    This assets adds the content for the James Webb Space Telescope mission, including
+    the animated model, safe viewing band, field of view, label, mission file and actions.
```
alexanderbock, line 26:
```suggestion
    the animated model, safe viewing band, field of view, label, mission file, and actions.
```

data/assets/scene/solarsystem/telescopes/jwst/model.asset
```diff
@@ -0,0 +1,63 @@
...
+  GUI = {
+    Name = "James Webb Space Telescope Model",
+    Path = "/Solar System/Telescopes/JWST",
+    Description = [[
+      Animated model of the James Webb Space Telescope that shows its deployment in
+      real time.
+    ]]
+  }
+}
+
```
alexanderbock, line 44:
```suggestion


```

data/assets/scene/solarsystem/telescopes/jwst/viewingband.asset
```diff
@@ -0,0 +1,69 @@
...
+
+
+asset.onInitialize(function()
+  openspace.addSceneGraphNode(JWSTBand)
+end)
+
+asset.onDeinitialize(function()
+  openspace.removeSceneGraphNode(JWSTBand)
+end)
+
```
alexanderbock, line 58:
```suggestion

asset.export(JWSTBand)
```
?

alexanderbock (approved):

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/transforms.asset
WeirdRubberDuck: Makes sense to me. @Roxeena

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/l2sunline.asset
```diff
@@ -0,0 +1,44 @@
...
+  Parent = sunTransforms.SolarSystemBarycenter.Identifier,
+  Renderable = {
+    Type = "RenderableNodeLine",
+    Enabled = asset.enabled,
+    StartNode = sun.Sun.Identifier,
+    EndNode = transforms.L2Position.Identifier,
+    Color = { 0.5, 0.5, 0.5 },
+    LineWidth = 2
+  },
+  Tag = { "lagrange_points_earth_l2_small" },
```
Roxeena, line 18: This tag seems to be wrong. This is the node with the sun to L2 line, not the small L2 indicator. Not sure what we want instead, maybe just remove it? Or at least add the `"lagrange_points_earth"` general tag?

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/l2.asset
```diff
@@ -0,0 +1,122 @@
...
+    Enabled = asset.enabled,
+    Text = "L2",
+    FontSize = 50.0,
+    Size = 6.0,
+    MinMaxSize = { 1.0, 100.0 },
+    OrientationOption = "Camera View Direction",
+    BlendMode = "Normal",
+    EnableFading = false
+  },
+  Tag = { "lagrange_points_earth_l2_small" },
```
Roxeena, line 63: Shouldn't the `"lagrange_points_earth"` tag also be included here?

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/l2.asset
```diff
@@ -0,0 +1,122 @@
...
+  Parent = transforms.L2Position.Identifier,
+  Renderable = {
+    Type = "RenderablePlaneImageLocal",
+    Enabled = asset.enabled,
+    Billboard = true,
+    Size = 4000000,
+    Texture = circle .. "circle.png",
+    BlendMode = "Additive"
+  },
+  Tag = { "lagrange_points_earth_l2_small" },
```
Roxeena, line 24: Same here

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/transforms.asset
Roxeena: In one way I agree, right now this is only used for the JWST, but this is a more general thing, and considering that I don't agree. For example, the GAIA trail might want to use this too to have a more intuitive L2 orbit trail than what we currently have. There are other spacecraft already at L2 and future spacecraft that will go there that will have a similar orbit around L2 that might want to use this frame as well. 

Perhaps the comment should be updated to reflect this general use case.

data/assets/scene/solarsystem/telescopes/jwst/trail.asset
```diff
@@ -76,18 +74,19 @@ local JWSTTrailCruise = {
...
     End = EndTime
   },
   Renderable = {
     Type = "RenderableTrailOrbit",
+    Enabled = asset.enabled,
     Translation = {
       Type = "SpiceTranslation",
       Target = kernels.ID.JWST,
-      Observer = transforms.ID.L2,
+      Observer = L2transforms.ID.L2,
```
Roxeena, line 89: I get an error here when trying to run the JWST profile. I think this is supposed to be `L2Kernels.ID.L2` instead.

Roxeena (commented):
I get some errors when trying to run the JWST profile. 
I think most originate from scripts now having incorrect paths to properties, for example, the timelapse script and the trail event probably need to be updated. 

jwst-errors

I'm not sure if it would be part of this PR to fix it or if we should create an issue for it and fix it later.

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/l2.asset
WeirdRubberDuck: 🤷 It wasn't in the original asset. I have literally just copied them over, so I have no idea it should be added or not. 

Should it?

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/transforms.asset
WeirdRubberDuck: I like the idea up updating the comment to make it clearer that it's not JWST specific. 

The JWST specific part could be moved to where this frame is being used instead, in the JWST-related asset :)

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/l2.asset
Roxeena: I would think it should. So I can turn on/off everything related to the Lagrange points in one script.

WeirdRubberDuck:
> I get some errors when trying to run the JWST profile. I think most originate from scripts now having incorrect paths to properties, for example, the timelapse script and the trail event probably need to be updated.
> 
> jwst-errors
> 
> I'm not sure if it would be part of this PR to fix it or if we should create an issue for it and fix it later.

Interesting.  

Looking at the error messages, I don't think those originate from the changes in this PR (all the scene graph nodes should be added just like before). So, then I think they should be fixed outside of this PR. 

But I could investigate a bit. Was quite a while since this PR was created now.. 😅

data/assets/scene/solarsystem/planets/earth/lagrange_points/l2/transforms.asset
WeirdRubberDuck: Rephrased the comment to be a bit more general, but it still mentions JWST for easier understanding. Let me know if you think it looks ok

PR #3446, opened 2024-09-30
Add documentation for TileProviderByLevel

alexanderbock:


data/assets/examples/tileprovider/tileproviderbylevel/bylevel.asset
```diff
@@ -0,0 +1,64 @@
+-- Basic
+-- This example file adds two layers to a globe. The first layer is used for further
+-- distances from the planet and when zooming into the planet, it gets replaced with the
+-- second layer. The place at which this transition occurs is determined by the "level",
+-- a number has to be between 2 and 22. Every increase in level means a transition point
+-- that is closer to the planet by a factor of two. Recommended reading for this example
+-- is the documentation on the DefaultTileProvider.
```
ylvaselling, line 7: Maybe we can skip over the duplicated parts from the description of the tile provider here? To me at least it makes it a bit confusing

Also maybe link to the default tile provider?

modules/globebrowsing/src/tileprovider/tileproviderbylevel.cpp
```diff
@@ -27,11 +27,28 @@
...
 
 namespace {
+    // This tile provider will switch between different tile providers specified within
+    // based on the level of detail that is requested by the Globe. All other things being
+    // equal, this corresponds to the distance of the camera to the planet, with a closer
+    // distance resulting in a higher lever. Due to technical reasons, the available
+    // levels are in the range [2, 22] and each increase in levels corresponds to a
+    // doubling in the effective resolution. For a given requested level, the tile
+    // provider that has the largest `MaxLevel` that is not greater than the requested
+    // level will be used.
```
ylvaselling, line 37: I found this text a bit hard to understand / technical. I tried to simplify this text a bit... a suggestion (is it even correct?)

This tile provider will switch between different tile providers based on the distance of the camera to the globe. Each tile provider will specify the distances in which they are valid. The distance is specified in _levels_. A smaller distance corresponds to a larger level number. The levels are in the range [2, 22]. Each increase in level corresponds to a doubling in the texture resolution. For a given requested level, the tile provider that has the largest `MaxLevel` that is not greater than the requested level will be used.

modules/globebrowsing/src/tileprovider/tileproviderbylevel.cpp
```diff
@@ -27,11 +27,28 @@
...
+            // inclusive, meaning that a value of 4 causes the tile provider to be used
+            // at level 4 but not at level 5.
             int maxLevel [[codegen::greaterequal(0)]];
-            ghoul::Dictionary tileProvider;
+
+            // The tile provider that should be used at this stage.
+            ghoul::Dictionary tileProvider [[codegen::reference("globebrowsing_layer")]];
         };
+
+        // The list of all tile providers that are used by this TileProviderByLevel
```
ylvaselling, line 51:
```suggestion
        // The list of all tile providers that are used by this TileProviderByLevel.
```

modules/globebrowsing/src/tileprovider/tileproviderbylevel.cpp
alexanderbock: The reason its a bit strange is that it doesn't only depend on the distance of the camera as the "TargetLevelOfDetail" slider has a direct impact as well.  And it's on the list to make a tile provider that actually works on camera distance.

How about:

This tile provider will switch between different tile providers specified within based on the required level of detail. The level of details mostly depends on the distance of the camera to the planet, with a closer distance resulting in a higher level. The levels are in the range [2, 22]. Each increase in level corresponds to a doubling in the texture resolution. For a given requested level, the tile provider that has the largest MaxLevel that is not greater than the requested level will be used.
```

PR #3450, opened 2024-10-08
Move Exoplanet creation code from C++ to Lua 

### LINKED ISSUE #3431: Move exoplanet creation code from C++ to a Lua script file
Move exoplanet creation code from C++ to a Lua script file in the module so that the visuals, tags, etc. could more easily be changed by the user. 

The exoplanets module C++ part could instead provide a function that gets all the data for a specific system, which can then be used by the above mentioned Lua function.

WeirdRubberDuck:
So that it can be changed by the user, if desired. Closes #3431 

Please try out the new functions and see if the structure of the new .lua file makes sense, as well as the documentation

Breaking change: `openspace.exoplanet.addExoplanetSystem` no longer have the option of taking a list of star names. INstead, a new function `openspace.exoplanet.addExoplanetSystems` was added to add multiple exoplanet systems in one function call

Note that there are some `TODO` comments still around in the .lua file. These will be addressed by another PR

### DIFF
--- a/modules/exoplanets/datastructure.cpp
+++ b/modules/exoplanets/datastructure.cpp
@@ -35,21 +35,6 @@
 namespace {
     constexpr std::string_view _loggerCat = "ExoplanetsModule";
 
-    // KeplerTranslation requires angles in range [0, 360]
-    float validAngle(float angle, float defaultValue, bool& usedDefault) {
-        if (std::isnan(angle)) {
-            usedDefault = true;
-            return defaultValue;
-        }
-        while (angle < 0.f) {
-            angle += 360.f;
-        }
-        while (angle > 360.f) {
-            angle =- 360.f;
-        }
-        return angle;
-    };
-
     // This is the documentation for the data object that is returned from the
     // `openspace.exoplanets.systemData` function in the Scripting API.
     //
@@ -223,6 +208,8 @@ ghoul::Dictionary ExoplanetSystem::toDataDictionary() const {
     }
 
     ghoul::Dictionary planets;
+    std::vector<float> inclinations;
+    inclinations.reserve(planetNames.size());
 
     for (size_t i = 0; i < planetNames.size(); i++) {
         const ExoplanetDataEntry& data = planetsData[i];
@@ -268,9 +255,26 @@ ghoul::Dictionary ExoplanetSystem::toDataDictionary() const {
                 hasUsedDefaultValues = true;
             }
 
-            float inclination = validAngle(data.i, 90.f, hasUsedDefaultValues);
-            float bigOmega = validAngle(data.bigOmega, 180.f, hasUsedDefaultValues);
-            float omega = validAngle(data.omega, 90.f, hasUsedDefaultValues);
+            // KeplerTranslation requires angles in range [0, 360]
+            auto validAngle = [&hasUsedDefaultValues](float angle, float defaultValue) {
+                if (std::isnan(angle)) {
+                    hasUsedDefaultValues = true;
+                    return defaultValue;
+                }
+                while (angle < 0.f) {
+                    angle += 360.f;
+                }
+                while (angle > 360.f) {
+                    angle = -360.f;
+                }
+                return angle;
+            };
+
+            float inclination = validAngle(data.i, 90.f);
+            float bigOmega = validAngle(data.bigOmega, 180.f);
+            float omega = validAngle(data.omega, 90.f);
+
+            inclinations.push_back(inclination);
 
             planet.setValue("Inclination", static_cast<double>(inclination));
             planet.setValue("AscendingNode", static_cast<double>(bigOmega));
@@ -310,12 +314,9 @@ ghoul::Dictionary ExoplanetSystem::toDataDictionary() const {
     res.setValue("SystemRotation", exoplanetSystemRotation);
 
     float meanInclination = 0.f;
-    for (const ExoplanetDataEntry& p : planetsData) {
-        // Compute a valid inclination value (same as used for the dictionary)
-        bool usedDefault = false; // Dummy value
-        meanInclination += validAngle(p.i, 90.f, usedDefault);
+    for (const float& i : inclinations) {
+        meanInclination += i / static_cast<float>(inclinations.size());
     }
-    meanInclination /= static_cast<float>(planetsData.size());
     const glm::dmat4 rotation = computeOrbitPlaneRotationMatrix(meanInclination);
     const glm::dmat3 meanOrbitPlaneRotationMatrix = static_cast<glm::dmat3>(rotation);
 
--- a/modules/exoplanets/datastructure.cpp
+++ b/modules/exoplanets/datastructure.cpp
@@ -101,7 +101,7 @@ namespace {
 
             // The uncertainty of the semi-major axis, given as an asymmetric uncertainty
             // with two values: the lower and upper uncertainty for the orbit size. The
-            // values are given as values relative to the size of the SemiMajorAxis. For
+            // values are given as values relative to the size of the `SemiMajorAxis`. For
             // example, the lower (first) value is computed as the lower uncertainty
             // value divided by the value for the semi-major axis.
             //
@@ -148,7 +148,7 @@ namespace {
             // periapsis values. Hence, it may be derived from default values.
             glm::dmat3 orbitPlaneRotationMatrix;
 
-            // True if default values have been used for any of the orbit parameters, due
+            // True, if default values have been used for any of the orbit parameters, due
             // to missing data.
             bool hasUsedDefaultValues;
         };
@@ -169,7 +169,7 @@ namespace {
         glm::dmat3 meanOrbitRotation;
 
         // The distance from this system to our solar system, in light years.
-        double distanceToUs;
+        double distance;
     };
 #include "datastructure_codegen.cpp"
 } // namespace
--- a/modules/exoplanets/datastructure.cpp
+++ b/modules/exoplanets/datastructure.cpp
@@ -54,45 +54,45 @@ namespace {
     // specific planets and star, some values may be undefined.
     struct [[codegen::Dictionary(ExoplanetSystemData)]] Parameters {
         // The identifier of the system, which equals the identifier of the host star
-        // created based on the star name
+        // created based on the star name.
         std::string systemId;
 
-        // The position of the host star, expressed in meters
+        // The position of the host star, expressed in meters.
         glm::dvec3 position;
 
         // The number of known planets in the system that have enough data to be
-        // visualized
+        // visualized.
         int numPlanets;
 
-        // The name of the star
+        // The name of the star.
         std::string starName;
 
-        // The radius of the star, in meters, if data exists
+        // The radius of the star, in meters, if data exists.
         std::optional<double> starRadius;
 
         // An RGB color for the host star, computed from the star's B-V value, if data
-        // exists
+        // exists.
         std::optional<glm::dvec3> starColor [[codegen::color()]];
 
-        // The effective temperature of the star, in Kelvin, if data exists
+        // The effective temperature of the star, in Kelvin, if data exists.
         std::optional<double> starTeff;
 
         // The luminosity for the star, in units of solar luminosities [log10(Solar)], if
-        // data exists
+        // data exists.
         std::optional<double> starLuminosity;
 
         struct Planet {
-            // An identifier to use for the planet, created based on the planet name
+            // An identifier to use for the planet, created based on the planet name.
             std::string id;
 
-            // The name of the exoplanet
+            // The name of the exoplanet.
             std::string name;
 
-            // The radius of the planet, in meters, if data exists
+            // The radius of the planet, in meters, if data exists.
             std::optional<double> starRadius;
 
             // The semi-major axis of the planet's orbit, in meters. That is, the orbit
-            // size. This is required to be able to be visualize the planetary orbit
+            // size. This is required to be able to be visualize the planetary orbit.
             double semiMajorAxis;
 
             // The uncertainty of the semi-major axis, given as an asymmetric uncertainty
@@ -106,37 +106,37 @@ namespace {
             std::optional<glm::dvec2> semiMajorAxisUncertainty;
 
             // The eccentricity of the planet's orbit. If it does not exist in the data,
-            // it is given a default value of 0
+            // it is given a default value of 0.
             double eccentricity;
 
             // The inclination of the planet's orbit, given as an angle in degrees in the
             // range 0-360. If it does not exist in the data, it is given a default value
-            // of 90 degrees
-            double inlination;
+            // of 90 degrees.
+            double inclination;
 
             // The longitude of ascending node, that is, the angle for the direction of
-            // the point where the orbit passes through the plane of reference. Also
+            // the point where the orbit passes through a reference plane. Also
             // known as the right ascension of the ascending node. The angle is given as
             // degrees in the range 0-360. If it does not exist in the data, it is given
-            // a default value of 180 degrees
+            // a default value of 180 degrees.
             double ascendingNode;
 
             // The argument of periapsis of the orbit, given as an angle in degrees in
             // the range 0-360. It is the angle between the planet's periapsis (the point
             // where it is closest to its star) and its ascending node, that is, where it
             // passes through its reference plane. If it does not exist in the data, it
-            // is given a default value of 90 degrees
+            // is given a default value of 90 degrees.
             double argumentOfPeriapsis;
 
             // The epoch to use when computing the planet's initial position in its orbit,
             // given as an ISO 8601-like date string of the format YYYY-MM-DDTHH:MN:SS.
             //
             // Computed based on the \"Time of Conjunction (Transit Midpoint)\" value in
-            // the data from the exoplanet archive
+            // the data from the exoplanet archive.
             std::string epoch;
 
             // Time the planet takes to make a complete orbit around the host star or
-            // system, in days
+            // system, in days.
             double period;
 
             // A rotation matrix that represents the rotation of the plane of the orbit.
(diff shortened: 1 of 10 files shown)


modules/exoplanets/exoplanetshelper.h
```diff
@@ -174,6 +63,7 @@ void sanitizeNameString(std::string& s);
  */
 void updateStarDataFromNewPlanet(StarData& starData, const ExoplanetDataEntry& p);
 
+
```
WeirdRubberDuck, line 66:
```suggestion
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+        // An RGB color for the host star, computed from the star's B-V value, if data
+        // exists
+        std::optional<glm::dvec3> starColor [[codegen::color()]];
+
+        // The effective temperature of the star, in Kelvin, if data exists
+        std::optional<double> starTeff;
+
+        // The luminosity for the star, in units of solar luminosities [log10(Solar)], if
+        // data exists
```
alexanderbock, line 81:
```suggestion
        // data exists.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+        std::optional<double> starLuminosity;
+
+        struct Planet {
+            // An identifier to use for the planet, created based on the planet name
```
alexanderbock, line 85:
```suggestion
            // An identifier to use for the planet, created based on the planet name.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            std::string id;
+
+            // The name of the exoplanet
```
alexanderbock, line 88:
```suggestion
            // The name of the exoplanet.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            std::string name;
+
+            // The radius of the planet, in meters, if data exists
```
alexanderbock, line 91:
```suggestion
            // The radius of the planet, in meters, if data exists.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            std::optional<double> starRadius;
+
+            // The semi-major axis of the planet's orbit, in meters. That is, the orbit
+            // size. This is required to be able to be visualize the planetary orbit
```
alexanderbock, line 95:
```suggestion
            // size. This is required to be able to be visualize the planetary orbit.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            double semiMajorAxis;
+
+            // The uncertainty of the semi-major axis, given as an asymmetric uncertainty
+            // with two values: the lower and upper uncertainty for the orbit size. The
+            // values are given as values relative to the size of the SemiMajorAxis. For
```
alexanderbock, line 100:
```suggestion
            // values are given as values relative to the size of the `SemiMajorAxis`. For
```

?

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            // example, the lower (first) value is computed as the lower uncertainty
+            // value divided by the value for the semi-major axis.
+            //
+            // If no value is given, there is no uncertainty and the semi-major axis value
+            // is exact.
+            std::optional<glm::dvec2> semiMajorAxisUncertainty;
+
+            // The eccentricity of the planet's orbit. If it does not exist in the data,
+            // it is given a default value of 0
```
alexanderbock, line 109:
```suggestion
            // it is given a default value of 0.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            double eccentricity;
+
+            // The inclination of the planet's orbit, given as an angle in degrees in the
+            // range 0-360. If it does not exist in the data, it is given a default value
+            // of 90 degrees
```
alexanderbock, line 114:
```suggestion
            // of 90 degrees.
```

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            double inlination;
+
+            // The longitude of ascending node, that is, the angle for the direction of
+            // the point where the orbit passes through the plane of reference. Also
+            // known as the right ascension of the ascending node. The angle is given as
+            // degrees in the range 0-360. If it does not exist in the data, it is given
+            // a default value of 180 degrees
```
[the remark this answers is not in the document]
alexanderbock: Why 180 and not 0?

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    end
+
+    local hasUsedDefaultValuesInfo = ""
+    if planetData.HasUsedDefaultValues then
+      hasUsedDefaultValuesInfo = [[
+        OBS! Default values have been used to visualize the orbit (for example for
+        inclination, eccentricity, or argument of periastron), and hence the
+        data specified for the orbit might not be reliable.
+      ]]
+    end
```
alexanderbock, line 491: Replacing this with a `openspace.ternary` when it is being further below?

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+        table.insert(planetColorLayers, PlanetTextureLayer)
+        ambientIntensity = 0.15
+      end
+
+      local Planet = {
+        Identifier = id,
+        Parent = starIdentifier,
+        Renderable = {
+          Type = "RenderableGlobe",
+          Radii =  planetData.Radius,
```
alexanderbock, line 525:
```suggestion
          Radii = planetData.Radius,
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+          Layers = {
+            ColorLayers = planetColorLayers
+          },
+          LightSourceNode = starIdentifier,
+          AmbientIntensity = ambientIntensity
+        },
+        Tag = { "exoplanet_planet" },
+        Transform = {
+          Translation = PlanetKeplerTranslation
+        },
```
alexanderbock, line 536:
```suggestion
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+          Enabled = true
+        }
+        table.insert(planetColorLayers, PlanetTextureLayer)
+        ambientIntensity = 0.15
+      end
+
+      local Planet = {
+        Identifier = id,
+        Parent = starIdentifier,
+        Renderable = {
```
alexanderbock, line 523:
```suggestion
        Transform = {
          Translation = PlanetKeplerTranslation
        },
        Renderable = {
```

alexanderbock: Just for consistency of ordering

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    --------------------------------------------------------------------
+    local trailResolution = 1000.0
+    -- Increase the resolution of very eccentric orbits
+    local EccentricityThreshold = 0.85
+    if planetData.Eccentricity > EccentricityThreshold then
+      trailResolution = trailResolution * 2.0
+    end
+
+    local orbitDescription = string.format(
+      "The orbit trail of the exoplanet %s. ", planetData.Name
```
alexanderbock, line 570:
```suggestion
      "The orbit trail of the exoplanet %s.", planetData.Name
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    )
+    if planetData.HasUsedDefaultValues then
+      orbitDescription = orbitDescription .. [[
+        OBS! Default values have been used to visualize the orbit (for example for
```
alexanderbock, line 574:
```suggestion
      orbitDescription = orbitDescription .. " " .. [[
        OBS! Default values have been used to visualize the orbit (for example for
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+
+    local Orbit = {
+      Identifier = id .. "_Trail",
+      Parent = starIdentifier,
+      Renderable = {
+        Type = "RenderableTrailOrbit",
+        Period = planetData.Period,
+        Resolution = trailResolution,
+        Translation = PlanetKeplerTranslation,
+        Color = { 1, 1, 1 }
```
alexanderbock, line 588:
```suggestion
        Color = { 1.0, 1.0, 1.0 }
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+        Description = orbitDescription
+      }
+    }
+    openspace.addSceneGraphNode(Orbit)
+
+    --------------------------------------------------------------------
+    -- Planet Orbit Uncertainty
+    --------------------------------------------------------------------
+    local showUncertaintyDisc = openspace.propertyValue("Modules.Exoplanets.ShowHabitableZone")
+    local discTexture = openspace.propertyValue("Modules.Exoplanets.OrbitDiscTexture")
```
alexanderbock, line 603: Checking if the file exists

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+
+    if hasValue(planetData.SemiMajorAxisUncertainty) then
+      local OrbitDisc = {
+        Identifier = id .. "_Disc",
+        Parent = starIdentifier,
+        Renderable = {
+          Type = "RenderableOrbitDisc",
+          Enabled =  showUncertaintyDisc,
```
alexanderbock, line 611:
```suggestion
          Enabled = showUncertaintyDisc,
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+          Eccentricity = planetData.Eccentricity,
+          Offset = planetData.SemiMajorAxisUncertainty,
+          Opacity = 0.25
+        },
+        Transform = {
+          Rotation = {
+            Type = "StaticRotation",
+            Rotation = planetData.OrbitPlaneRotationMatrix
+          }
+        },
```
alexanderbock, line 623:
```suggestion
```

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    -- Planet Orbit Uncertainty
+    --------------------------------------------------------------------
+    local showUncertaintyDisc = openspace.propertyValue("Modules.Exoplanets.ShowHabitableZone")
+    local discTexture = openspace.propertyValue("Modules.Exoplanets.OrbitDiscTexture")
+
+    if hasValue(planetData.SemiMajorAxisUncertainty) then
+      local OrbitDisc = {
+        Identifier = id .. "_Disc",
+        Parent = starIdentifier,
+        Renderable = {
```
alexanderbock, line 609:
```suggestion
        Transform = {
          Rotation = {
            Type = "StaticRotation",
            Rotation = planetData.OrbitPlaneRotationMatrix
          }
        },
        Renderable = {
```

modules/exoplanets/tasks/exoplanetsdatapreparationtask.h
```diff
@@ -27,7 +27,7 @@
 
 #include <openspace/util/task.h>
 
-#include <modules/exoplanets/exoplanetshelper.h>
+#include <modules/exoplanets/datastructure.h>
```
alexanderbock, line 31:
```suggestion
```

scripts/core_scripts.lua
```diff
@@ -88,6 +88,38 @@ openspace.documentation = {
...
+    ]]
+  },
+  {
+    Name = "isEmpty",
+    Arguments = {
+      { "object", "any" }
+    },
+    Return = "Boolean",
+    Documentation = [[
+      A utility function to check whether an object is empty or not. Idetifies `nil`
```
alexanderbock, line 117:
```suggestion
      A utility function to check whether an object is empty or not. Identifies `nil`
```

alexanderbock (approved):

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    return
+  end
+
+  --------------------------------------------------------------------
+  -- Star Globe
+  --------------------------------------------------------------------
+  local starTexture = openspace.propertyValue("Modules.Exoplanets.StarTexture")
+  local starNoDataTexture = openspace.propertyValue("Modules.Exoplanets.NoDataTexture")
+
+  local colorLayers = {}
```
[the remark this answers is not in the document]
WeirdRubberDuck: I thought of that and made sure to test it as well. But actually, we don't need to.  The Renderable (or in this case, the layer, I think) handles that checking and already provides that check. 

Also, I find that when using the `openspace.printError` function, it's pretty difficult to understand where the output is coming from as it has no "category".  And the resulting error is "uglier", as it adds quotes around the message that is passed to the function. See image below

Instead, in the future, it would be nice to add a check for that on input - to prevent setting properties to invalid files or textures. So that, in this case, you would get an error immediately when starting OpenSpace, as the Module's property value is set. That way, you would also connect it to that property in the Module, and not the texture path that is set for this specific layer

Here's how the error looks if I enter an invalid texture file (which I think is good enough) and how a printed error from `openspace.printError` looks like
image

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    }
+  }
+  openspace.addSceneGraphNode(StarLabel)
+
+  --------------------------------------------------------------------
+  -- Star Glare
+  --------------------------------------------------------------------
+  local starGlareTexture = openspace.propertyValue("Modules.Exoplanets.StarGlareTexture")
+
+  if hasValue(data.StarColor) and hasValue(data.StarRadius) then
```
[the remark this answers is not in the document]
WeirdRubberDuck: See comment above - we don't need to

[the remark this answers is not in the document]
WeirdRubberDuck: But it might be good ot handle it better on the module side

modules/exoplanets/scripts/systemcreation.lua
```diff
@@ -0,0 +1,643 @@
...
+    }
+  }
+  openspace.addSceneGraphNode(Circle)
+
+  --------------------------------------------------------------------
+  -- Habitable Zone
+  --------------------------------------------------------------------
+  local showZone = openspace.propertyValue("Modules.Exoplanets.ShowHabitableZone")
+  local zoneOpacity = openspace.propertyValue("Modules.Exoplanets.HabitableZoneOpacity")
+  local zoneTexture = openspace.propertyValue("Modules.Exoplanets.HabitableZoneTexture")
```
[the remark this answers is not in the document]
WeirdRubberDuck: See above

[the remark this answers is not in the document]
WeirdRubberDuck: And, this is analogous as just setting the property values in the Scene Menu.  So, it should work with entering a faulty path (i.e. nothing should break)

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+            double period;
+
+            // A rotation matrix that represents the rotation of the plane of the orbit.
+            // It is computed based on the inclination, ascending node, and argument of
+            // periapsis values. Hence, it may be derived from default values.
+            glm::dmat3 orbitPlaneRotationMatrix;
+
+            // True if default values have been used for any of the orbit parameters, due
+            // to missing data
+            bool hasUsedDefaultValues;
```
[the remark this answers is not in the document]
WeirdRubberDuck: No, rather, the opposite. It says if any of the non-optional values have been set based on default values

Most data points do not have data for e.g. the orbit inclination, argument of periapsis, or ascending node, but we always set that value to a valid one (because otherwise we would not be able to show the orbit of almost any planets).  This parameter is true if any of those variables have been set to a default value.

modules/exoplanets/datastructure.cpp
WeirdRubberDuck: Yes, I agree with this.  We should always use the code style for the member names, since this is how they show up in the documentation :)

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+    sanitizeNameString(sanitizedStarName);
+    res.setValue("StarName", sanitizedStarName);
+
+    if (!std::isnan(starData.radius)) {
+        double radiusInMeter = distanceconstants::SolarRadius * starData.radius;
+        res.setValue("StarRadius", radiusInMeter);
+    }
+
+    if (!std::isnan(starData.bv)) {
+        res.setValue("StarColor", glm::dvec3(computeStarColor(starData.bv)));
```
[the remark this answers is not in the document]
WeirdRubberDuck: Yeah, I actually think it's more consistent to use the constructor version, since thatäs how we do it everywhere for `glm` types

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+    res.setValue("SystemId", makeIdentifier(starName));
+
+    const glm::dvec3 starPosInParsec = static_cast<glm::dvec3>(starData.position);
+
+    if (!isValidPosition(starPosInParsec)) {
+        LERROR(std::format(
+            "Insufficient data available for exoplanet system '{}'. Could not "
+            "determine star position", starName
+        ));
+        return ghoul::Dictionary();
```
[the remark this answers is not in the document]
WeirdRubberDuck: Yeah, that makes more sense :)

modules/exoplanets/datastructure.cpp
```diff
@@ -0,0 +1,332 @@
...
+#include <openspace/util/time.h>
+#include <openspace/util/timeconversion.h>
+#include <ghoul/logging/logmanager.h>
+#include <string_view>
+
+namespace {
+    constexpr std::string_view _loggerCat = "ExoplanetsModule";
+
+    // KeplerTranslation requires angles in range [0, 360]
+    float validAngle(float angle, float defaultValue, bool& usedDefault) {
```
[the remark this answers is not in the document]
WeirdRubberDuck: Hm, I get your point. However, since I also want to keep track of whether the default value was used, that would bloat the code in the function below quite a lot.

Since it's a very specific function for this code only, maybe I should just put it in the `toDataDictionary` function instead? I used to have it as a lambda function that captured the `hasUsedDefaultValues` variable and just altered that variable and maybe that was better

PR #3460, opened 2024-11-21
Alterations made to the GUI name or location of assets.

brianpatrick:
Changes to the name of some assets as they appear in the GUI. 
Changes to a few asset locations as they appear in the GUI/Scene Menu.

All changes cosmetic, no filename or physical location changes to any files.


data/assets/scene/digitaluniverse/backgroundradiation.asset
```diff
@@ -33,7 +33,7 @@ local COBE = {
     FadeInThreshold = 0.4
   },
   GUI = {
-    Name = "1990 COBE CMB",
+    Name = "COBE",
```
alexanderbock, line 36: What's the reason for changing your mind?  You suggested to change it from COBE to 1990 COBE CMB in https://github.com/OpenSpace/OpenSpace/commit/babbb4f97c37c68aa26e77b7917180cde2fc2153#diff-5297116b943caeb4f4f994d9be25c49bab9ec51c183c75dff7b8ff94f0c62c48R36

(same for WMAP and Planck)

data/assets/scene/digitaluniverse/milkyway_arm_labels.asset
```diff
@@ -39,7 +39,7 @@ local Object = {
     },
   },
   GUI = {
-    Name = "Milky Way Arms Labels",
+    Name = "Milky Way Arm Labels",
```
alexanderbock, line 42:
```suggestion
    Name = "Milky Way Arms Labels",
```

The scene graph node does show multiple arms

data/assets/scene/digitaluniverse/sdss.asset
```diff
@@ -43,7 +43,7 @@ local Object = {
     }
   },
   GUI = {
-    Name = "Sloan Digital Sky Survey",
+    Name = "Sloan Galaxies",
```
alexanderbock, line 46:
```suggestion
    Name = "Sloan Digital Sky Survey",
```

It is the name of the survey

data/assets/scene/solarsystem/sun/sun.asset
```diff
@@ -18,7 +18,7 @@ local Sun = {
   },
   ApproachFactor = 15.0,
   GUI = {
-    Name = "Sun",
+    Name = "Solar Surface",
```
alexanderbock, line 21:
```suggestion
    Name = "Sun",
```

It is the sun, the texture that we put on by default happens to be the photosphere, but other users can put other textures on it

data/assets/scene/digitaluniverse/backgroundradiation.asset
brianpatrick: Because I'm going to order them, as mentioned above.

data/assets/scene/digitaluniverse/milkyway_arm_labels.asset
brianpatrick: Yes, but we don't say pulsars labels, we say pulsar labels. A small grammatical detail, but I'm kind of a stickler for those. I didn't catch it in the last update.

data/assets/scene/digitaluniverse/sdss.asset
brianpatrick: The problem here is Sloan DSS is much larger than just galaxies. I used to have a data set of Sloan stars long ago. So, I want to keep our options open that way. Also, these asset names take on the quality of a name... they need not be absolutely precise, but must have a name that is simple and meaningful. I think Sloan Galaxies is unmistakably from the SDSS.

data/assets/scene/digitaluniverse/alternatestarlabels.asset
```diff
@@ -23,7 +23,7 @@ local Object = {
     Unit = "pc"
   },
   GUI = {
-    Name = "Stars Labels - Alternate",
+    Name = "Alternate Stars Labels",
```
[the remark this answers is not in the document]
brianpatrick: I'm also planning to possibly overhaul these alternate labels in the future. So, the whole star names situation may change at some point.

data/assets/scene/solarsystem/sun/sun.asset
brianpatrick: I've tossed this around in my brain... I see your point. We don't say "Earth Surface" or "Mars Surface."

data/assets/scene/digitaluniverse/alternatestarlabels.asset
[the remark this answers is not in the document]
brianpatrick: Right now I put it in the `base.asset`. Have not committed those changes yet, tho. If it's in the DU asset will it load anywhere? `digitaluniverse.asset` seems to be a subset that is not used. Plus, I also set some ordering in the Solar System.

data/assets/scene/digitaluniverse/sdss.asset
brianpatrick: That is *very* verbose. Also, until a short time ago, the quasar catalog was from Sloan too. So, we always had "Sloan Galaxies" and Sloan Quasars". How about "Sloan DSS Galaxies"? Though I feel the "DSS" is a bit redundant and perhaps cryptic to those unfamiliar with the acronym SDSS.

In the end, the actual name doesn't much matter. We just want to identify that these are the galaxies that the Sloan Foundation funded, and we want something that rolls off the tongue, and is concise. Each of these names has to fit on one of those "cards" in the documentation, and a lengthy name just doesn't look good.

data/assets/scene/digitaluniverse/alternatestarlabels.asset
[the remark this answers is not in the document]
alexanderbock: How about making the digitaluniverse.asset load the same assets as the base and then do the reordering in there?  It doesn't seem very "base-y" to me to also muck with the order of the scene graph nodes and that asset it starting to do a lot

data/assets/scene/digitaluniverse/sdss.asset
alexanderbock: I think Sloan DSS Galaxies can work for me :+1:

data/assets/scene/digitaluniverse/alternatestarlabels.asset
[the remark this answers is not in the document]
brianpatrick: I do not much care *where* this happens, but that it happens and is basically the default on any generic launch of OS. A few reasons for this:

One, seeing the data sets all jumbled in the scene list is off-putting. It is akin to reading a book on C and randomly, or alphabetically, ordering the chapters, then the sections within each chapter. It denies the natural progression that the material exists within. It is far more elegant to adhere to that natural order and endow this list with a sensible order. Yes, it could demand more curation as new sets are added, but those curation steps will bring profound meaning to the user, rather than a haphazard, meaningless order.

Two, I've structured the entire content chapter on this ordering in the guide. The whole progression of these sections reflects the natural progression of these data sets.

I'm open to implementation options... you just tell me how such that it proliferates throughout the OS ecosystem, generally.

data/assets/scene/digitaluniverse/sdss.asset
brianpatrick: OK. I'll make those changes.

data/assets/scene/digitaluniverse/alternatestarlabels.asset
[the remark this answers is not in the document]
alexanderbock: :+1:  Yea, I think in that case we can have your ordering code in the `digitaluniverse.asset` instead and then `asset.require` the digitaluniverse asset from the `base.asset` instead.  Right now the `base.asset` has its own list of DU assets to include and I think it would be better to just replace that list with a single `require` and fix the digitaluniverse asset instead

alexanderbock:
@brianpatrick ?

ylvaselling (approved):

Issue #251, opened 2017-02-25
OSC (Open Sound Control) scripting support


alexanderbock:
Useful for control by Eluminati kiosk-mode GUI.  Either by providing a bridge between OSC and Lua or direct integration into the SGCT external control

alexanderbock:
Not sure how close we'd be to this by now?  I feel that it's basically done @Roxeena ?

PR #3461, opened 2024-12-02
Sonification/Telemetry

Roxeena:
# Telemetry and OSC modules
This PR adds two modules, one to send OSC messages and one to monitor the state of the engine/scene and send that out using the OSC module. This has been used to send scene information to another software, SuperCollider, that has then used the information to create a sonification of the planets in different aspects. However, the thought is that any software capable of receiving OSC messages can use this information for anything.

The goal with this module is that it is general enough that other sonifications can be created without us needing to implement anything extra on the OpenSpace side. It will also be a way to distribute the sonifications we have already created so that anyone can use them. 

Note that the Sonification module has been renamed to Telemetry to make it clearer that we do not produce the sonifications in OpenSpace, but rather, we send out data via OSC to a sonification software that, in turn, creates the sonification. This means that to run this and test it, you need to download another software [SuperCollider](https://supercollider.github.io/). 

Since I am starting to write the documentation for this module, I would appreciate any comments or suggestions to make anything clearer. Especially the _TelemetryModule.h_ file and the util files for the module contain quite a bit of documentation text that could be good to get other eyes on. 

## Steps to test it:
1. You need to build OpenSpace in CMake with the OSC and Telemetry module enabled 
2. Download and install [SuperCollider](https://supercollider.github.io/)
3. Run the SuperCollider test file _OpenSpace-OSC-example.scd_ located in _data\assets\examples\sonification_.
    - To run the file in SuperCollider, click any line inside any OSC definitions (for example, line 31) and press CTRL + ENTER.
    - The SuperCollider console should respond with: `-> OSCdef(Neptune, /Neptune, nil, nil, nil).`
4. Run OpenSpace with the sonification profile.
5. To test the general nodes sonification, you need to drag and drop the asset file _data\assets\examples\sonification\nodes.asset_
6. Turn on the Telemetry module under _Settings/Modules_ in the GUI
7. Turn on all the telemetries/sonifications that are separate checkboxes in the telemetry module settings
8. Fly around in OpenSpace, and you should see messages being printed in the SuperCollider console window
9. Now, you can freely try the different settings for the different telemetries/sonifications, read the tooltips, and stress-test it in general.
10. **Before you shut down OpenSpace, you will need to turn off the telemetry module in the settings menu.** If OpenSpace is shut down while the telemetry module is still enabled, it will crash. This is an issue with the Scene not being safe with multiple threads, especially around startup and shutdown. 

## Things that are still in progress:
* Adjusting the OpenSpace planet SuperCollider sonification code to the new OSC structure and adding it into the repository, the plan is to put it in _data\assets\modules\telemetry\sonification_ 
* Write extensive documentation about the new module, especially about the math and the different angle modes
* Add a scaled-down sonification example for the general nodes sonification


data/assets/examples/sonification/OpenSpace-OSC-example.scd
```diff
@@ -0,0 +1,391 @@
...
+ *                                                                                       *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,   *
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A         *
+ * PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT    *
+ * HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF  *
+ * CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE  *
+ * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
+ ****************************************************************************************/
+
+NetAddr.langPort;
```
WeirdRubberDuck, line 25: This file needs some description about what it is and what it is used for :) Especially since it is among "examples"

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+    constexpr openspace::properties::Property::PropertyInfo IpAddressInfo = {
+        "IpAddress",
+        "Ip address",
+        "The network ip address that the telemetry osc messages is sent to"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo PortInfo = {
+        "Port",
+        "Port",
+        "The network port that the telemetry osc messages is sent to"
```
alexanderbock, line 68:
```suggestion
        "The network port that the telemetry osc messages are sent to."
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo AngleCalculationModeInfo = {
+        "AngleCalculationMode",
+        "Angle Calculation Mode",
+        "This setting changes the method to calculate any angles in the telemetries. "
+        "The Horizontal mode, generally works well for flat displays or forward facing "
+        "immersive envierments. The Circular mode, generally works well for centered "
```
alexanderbock, line 76:
```suggestion
        "immersive environments. The Circular mode, generally works well for centered "
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+        "fisheye displays or omnidirectional immersive environments"
```
alexanderbock, line 77:
```suggestion
        "fisheye displays or omnidirectional immersive environments."
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+    };
+
+    struct [[codegen::Dictionary(TelemetryModule)]] Parameters {
+        // [[codegen::verbatim(IpAddressInfo.description)]]
+        std::optional<std::string> ipAddress;
+
+        // [[codegen::verbatim(PortInfo.description)]]
+        std::optional<int> port;
+
+        enum class [[codegen::map(openspace::TelemetryModule::AngleCalculationMode)]] AngleCalculationMode {
```
alexanderbock, line 95:
```suggestion
        enum class
        [[codegen::map(openspace::TelemetryModule::AngleCalculationMode)]]
        AngleCalculationMode
    {
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    while (isRunning) {
+        // Wait for the main thread
+        //LDEBUG("The telemetry thread is waiting for a signal from the main thread");
+        std::unique_lock<std::mutex> lk(mutexLock);
+        syncToMain.wait(lk);
+        //LDEBUG(
+        //    "The telemetry thread is working after having received a signal from "
+        //    "the main thread"
+        //);
```
alexanderbock, line 263:
```suggestion
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+
+        // Process the telemetries
+        if (isInitialized) {
+            for (TelemetryBase* telemetry : _telemetries) {
+                if (!telemetry) {
+                    continue;
+                }
+                telemetry->update(camera);
```
alexanderbock, line 301:
```suggestion
                if (telemetry) {
                    telemetry->update(camera);
                }
                
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * \return The Lua libraries for all telemetries available in the telemetry module
+     */
+    std::vector<scripting::LuaLibrary> luaLibraries() const override;
+
+    virtual void internalInitialize(const ghoul::Dictionary& dictionary) override;
+    virtual void internalDeinitialize() override;
+
+    /**
+     * Get the list of telemetries that are currently registered in the module
```
alexanderbock, line 94:
```suggestion
     * Get the list of telemetries that are currently registered in the module.
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * \param id The identifier of the telemetry to fetch
+     *
+     * \return The requested telemetry
+     */
+    const TelemetryBase* telemetry(std::string id) const;
+    TelemetryBase* telemetry(std::string id);
+
+    /**
+     * Get the current angle calculation mode used in the telemetry module
```
alexanderbock, line 111:
```suggestion
     * Get the current angle calculation mode used in the telemetry module.
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     * \return The angle calculation mode
+     */
+    AngleCalculationMode angleCalculationMode() const;
+
+    /**
+     * Return whether any elevation angles are being caclulated and sent over the osc
+     * connection or not
+     *
+     * \return True if elevation angles are being caclulated and sent over the osc
+     *         connection, false otherwise
```
alexanderbock, line 122:
```suggestion
     * \return `true` if elevation angles are being calculated and sent over the osc
     *         connection, `false` otherwise
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     */
+    bool includeElevationAngle() const;
+
+private:
+    /**
+     * Main update function that keeps track of all telemetries and keeps the update
+     * thread running and synced to the OpenSpace main thread
```
alexanderbock, line 129:
```suggestion
     * thread running and synced to the OpenSpace main thread.
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * \param isRunning Whether the thread should be kept running or shut down and joined
+     */
+    void update(std::atomic<bool>& isRunning);
+
+    /**
+     * Add a given telemetry to the list of registered telemetries in the module
```
alexanderbock, line 136:
```suggestion
     * Add a given telemetry to the list of registered telemetries in the module.
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * \param telemetry The telemetry to register in the module
+     */
+    void addTelemetry(TelemetryBase* telemetry);
+
+    /**
+     * Function that gets called when the angle calculation mode is changed in the GUI
```
alexanderbock, line 143:
```suggestion
     * Function that gets called when the angle calculation mode is changed in the GUI.
```

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     */
+    void guiOnChangeAngleCalculationMode();
+
+    // To sync the sonificaiton thread with the main thread
+    std::mutex mutexLock;
+    std::condition_variable syncToMain;
+
+    // Properties
```
alexanderbock, line 151:
```suggestion
```

modules/telemetry/src/specific/planetscomparesonification.cpp
```diff
@@ -0,0 +1,299 @@
...
+    _connection->send(label, data);
+}
+
+void PlanetsCompareSonification::planetSelectionChanged(
+                                                properties::OptionProperty& changedPlanet,
+                                             properties::OptionProperty& notChangedPlanet,
+                                                        std::string& prevChangedPlanet)
+{
+    if (changedPlanet != 0 && changedPlanet == notChangedPlanet) {
+        LINFO("Cannot compare a planet to itself");
```
WeirdRubberDuck, line 216: Should this be a warning?

modules/telemetry/src/specific/planetsoverviewsonification.cpp
```diff
@@ -0,0 +1,190 @@
...
+namespace {
+    constexpr std::string_view _loggerCat = "PlanetsOverviewSonification";
+
+    static const openspace::properties::PropertyOwner::PropertyOwnerInfo
+        PlanetsOverviewSonificationInfo =
+    {
+        "PlanetsOverviewSonification",
+        "Planets Overview Sonification",
+        "Sonification that gives an overview of the planets in the solar system with "
+        "simpler sounds"
```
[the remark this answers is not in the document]
WeirdRubberDuck: Simpler than what?  Change that phrasing slightly, please :)

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+        "Sonification of the planets in the solar system"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ToggleAllInfo = {
+        "ToggleAll",
+        "All",
+        "Toggle all sonifications for all planets"
+    };
+
+    // Planets
```
WeirdRubberDuck, line 54:
```suggestion
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+        "Toggle the moons sonification for the planet"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo RingsInfo = {
+        "Rings",
+        "Rings",
+        "Toggle the rings sonification for the planet"
+    };
+
+    // Precision settings
```
WeirdRubberDuck, line 146:
```suggestion
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+        "Precision",
+        "Precision",
+        "Settings for the distance and angle precision of the planets sonification"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo LowDistancePrecisionInfo = {
+        "LowDistancePrecision",
+        "Distance Precision (Low)",
+        "The precision in meters used to determin when to send updated distance data "
+        "over the OSC connection. This is the lower precision used, for planets that are "
```
WeirdRubberDuck, line 157: See my previous comments about the high VS low precisions

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    constexpr openspace::properties::Property::PropertyInfo HighAnglePrecisionInfo = {
+        "HighAnglePrecision",
+        "Angle Precision (High)",
+        "The precision in radians used to determin when to send updated angle data "
+        "over the OSC connection. This is the higher precision used, for the planet that "
+        "is the current focus",
+        openspace::properties::Property::Visibility::User
+    };
+
+    struct [[codegen::Dictionary(PlanetsSonification)]] Parameters {
```
WeirdRubberDuck, line 189: This is the first type whose settings are added in a `Parameters` object.  The others won't be createable or possible to set from an asset file. Should they be?

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+        // The name of the planet
+        std::string name;
+
+        // Names of the moons for the planet
+        std::optional<std::vector<std::string>> moons;
+    };
+#include "planetssonification_codegen.cpp"
+} // namespace
+#include "planetssonification_lua.inl"
```
WeirdRubberDuck, line 198: I think this should be among the other includes? At the very end, after an empty line

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    , _venusProperty(PlanetsSonification::PlanetProperty(VenusInfo))
+    , _earthProperty(PlanetsSonification::PlanetProperty(EarthInfo))
+    , _marsProperty(PlanetsSonification::PlanetProperty(MarsInfo))
+    , _jupiterProperty(PlanetsSonification::PlanetProperty(JupiterInfo))
+    , _saturnProperty(PlanetsSonification::PlanetProperty(SaturnInfo))
+    , _uranusProperty(PlanetsSonification::PlanetProperty(UranusInfo))
+    , _neptuneProperty(PlanetsSonification::PlanetProperty(NeptuneInfo))
+    , _precisionProperty(PlanetsSonification::PrecisionProperty(PrecisionInfo))
+{
+    // Toggle all
```
WeirdRubberDuck, line 215:
```suggestion
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    }
+}
+
+PlanetsSonification::PrecisionProperty::PrecisionProperty(
+                               properties::PropertyOwner::PropertyOwnerInfo precisionInfo)
+    : properties::PropertyOwner(precisionInfo)
+    , lowDistancePrecision(LowDistancePrecisionInfo, 10000.0, 0, 1e+25)
+    , highDistancePrecision(HighDistancePrecisionInfo, 1000.0, 0, 1e+25)
+    , lowAnglePrecision(LowAnglePrecisionInfo, 0.1, 0, 10)
+    , highAnglePrecision(HighAnglePrecisionInfo, 0.05, 0, 10)
```
WeirdRubberDuck, line 347:
```suggestion
    , lowDistancePrecision(LowDistancePrecisionInfo, 10000.0, 0.0, 1e+25)
    , highDistancePrecision(HighDistancePrecisionInfo, 1000.0, 0.0, 1e+25)
    , lowAnglePrecision(LowAnglePrecisionInfo, 0.1, 0.0, 10)
    , highAnglePrecision(HighAnglePrecisionInfo, 0.05, 0.0, 10)
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    if (planetInfo.guiName != "Mercury" && planetInfo.guiName != "Venus") {
+        addProperty(moonsEnabled);
+    }
+    // Saturn is the only planet with rings (visible in OpenSpace)
+    if (planetInfo.guiName == "Saturn") {
+        addProperty(ringsEnabled);
+    }
+}
+
+PlanetsSonification::PrecisionProperty::PrecisionProperty(
```
WeirdRubberDuck, line 341:
```suggestion
PlanetsSonification::PrecisionProperties::PrecisionProperties(
```
See previous comment

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    _neptuneProperty.gravityEnabled.onChange([this]() { onNeptuneSettingChanged(); });
+    _neptuneProperty.temperatureEnabled.onChange([this]() { onNeptuneSettingChanged(); });
+    _neptuneProperty.atmosphereEnabled.onChange([this]() { onNeptuneSettingChanged(); });
+    _neptuneProperty.moonsEnabled.onChange([this]() { onNeptuneSettingChanged(); });
+    // No rings visible in OpenSpace
+    addPropertySubOwner(_neptuneProperty);
+
+    // Precision
+    _anglePrecision = _precisionProperty.lowAnglePrecision;
+    _distancePrecision = _precisionProperty.lowDistancePrecision;
```
WeirdRubberDuck, line 301:
```suggestion
    _anglePrecision = _precisionProperties.lowAnglePrecision;
    _distancePrecision = _precisionProperties.lowDistancePrecision;
```
See previous comment

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+{
+    addProperty(lowDistancePrecision);
+    lowDistancePrecision.setExponent(20.f);
+
+    addProperty(highDistancePrecision);
+    highDistancePrecision.setExponent(20.f);
+
+    addProperty(lowAnglePrecision);
+
+    addProperty(highAnglePrecision);
```
WeirdRubberDuck, line 357:
```suggestion
    addProperty(lowDistancePrecision);
    lowDistancePrecision.setExponent(20.f);
    addProperty(highDistancePrecision);
    highDistancePrecision.setExponent(20.f);
    addProperty(lowAnglePrecision);
    addProperty(highAnglePrecision);
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    std::string label = "/" + _planets[planetIndex].name;
+    std::vector<OscDataType> data;
+
+    // The total size of the data vector is NumDataItemsPlanet for the
+    // planet, and then NumDataItemsMoon per moon
+    data.reserve(
+        NumDataItemsPlanet + NumDataItemsMoon * _planets[planetIndex].moons.size()
+    );
+
+    // Distance
```
WeirdRubberDuck, line 454: Are these comments really needed?

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    }
+    else {
+        throw ghoul::MissingCaseException();
+    }
+
+    return osc::Blob(settings, NumSettings);
+}
+
+void PlanetsSonification::sendPlanetData(int planetIndex) {
+    if (planetIndex < 0 || _planets.size() <= planetIndex) {
```
WeirdRubberDuck, line 440:
```suggestion
    if (planetIndex < 0 ||  planetIndex > _planets.size() - 1) {
```
easier to read, IMO

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    _neptuneProperty.gravityEnabled.setValue(_neptuneProperty.toggleAll);
+    _neptuneProperty.temperatureEnabled.setValue(_neptuneProperty.toggleAll);
+    _neptuneProperty.atmosphereEnabled.setValue(_neptuneProperty.toggleAll);
+    _neptuneProperty.moonsEnabled.setValue(_neptuneProperty.toggleAll);
+}
+void PlanetsSonification::onNeptuneSettingChanged() {
+    sendPlanetData(NeptuneIndex);
+}
+
+bool PlanetsSonification::getData(const Camera* camera, int planetIndex,
```
WeirdRubberDuck, line 588: see previous comments 
```suggestion
bool PlanetsSonification::updateData(const Camera* camera, int planetIndex,
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+        );
+    }
+
+    if (std::abs(distance) < std::numeric_limits<double>::epsilon()) {
+        // The scene is likely not yet initialized
+        return false;
+    }
+
+    // Also calculate angle to moons
+    bool shouldSendData = false;
```
WeirdRubberDuck, line 615:
```suggestion
    bool shouldSendData = false;

    // Also calculate angle to moons
```

WeirdRubberDuck: And why do we also compute the angle too moons?  Could this comment clarify that?

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+}
+
+void PlanetsSonification::update(const Camera* camera) {
+    TelemetryModule* module = global::moduleEngine->module<TelemetryModule>();
+    if (!module) {
+        LERROR("Could not find the TelemetryModule");
+        return;
+    }
+
+    // The compare planets and planets overview sonifications depends on this
```
WeirdRubberDuck, line 697:
```suggestion
    // The compare planets and planets overview sonifications depend on this
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    // sonification. If any of them are enabled, then this sonificaiton sends planetary
+    // data over the osc connection, even if this sonificaiton is disabled.
```
WeirdRubberDuck, line 699:
```suggestion
    // data over the OSC connection, even if this sonificaiton is disabled.
```

modules/telemetry/src/specific/planetssonification_lua.inl
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Adds the given list of planets to the PlanetsSonification internal list of Planets
+ * and Moons
+ */
+[[codegen::luawrap]] void addPlanets(ghoul::Dictionary planets) {
+    openspace::TelemetryModule* module =
+        openspace::global::moduleEngine->module<openspace::TelemetryModule>();
+    if (!module) {
+        LERRORC("PlanetsSonification_lua", "Could not find the TelemetryModule");
```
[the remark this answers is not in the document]
WeirdRubberDuck: Maybe use a `throw ghoul::lua::LuaError` here instead? To make it clear that the error comes from the Lua function call.

modules/telemetry/src/specific/planetssonification_lua.inl
```diff
@@ -0,0 +1,54 @@
...
+        return;
+    }
+    openspace::TelemetryBase* ptr = module->telemetry("PlanetsSonification");
+    if (!ptr) {
+        LERRORC("PlanetsSonification_lua", "Could not find the PlanetsSonification");
```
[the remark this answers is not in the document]
WeirdRubberDuck: Same here, `throw ghoul::lua::LuaError` ?

modules/telemetry/src/telemetrybase.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <openspace/engine/moduleengine.h>
+#include <openspace/scene/scenegraphnode.h>
+#include <openspace/query/query.h>
+
+namespace {
+    constexpr std::string_view _loggerCat = "TelemetryBase";
+
+    constexpr openspace::properties::Property::PropertyInfo EnabledInfo = {
+        "Enabled",
+        "Is Enabled",
```
WeirdRubberDuck, line 37:
```suggestion
        "Enabled",
```
?

modules/telemetry/src/telemetrybase.cpp
```diff
@@ -0,0 +1,73 @@
...
+    : properties::PropertyOwner(info)
+    , _enabled(EnabledInfo, false)
+{
+    _identifier = info.identifier;
+    _connection = new OscConnection(ip, port);
+
+    addProperty(_enabled);
+    _enabled.onChange([this]() {
+        if (!_enabled) {
+            // Disable sending of data
```
WeirdRubberDuck, line 55: Do we need to re-enable it if enabled is true?

modules/telemetry/src/util.cpp
```diff
@@ -0,0 +1,478 @@
...
+    double calculateDistanceTo(const Camera* camera, glm::dvec3 nodePosition,
+                               DistanceUnit unit)
+    {
+        if (glm::length(nodePosition)<std::numeric_limits<glm::f64>::epsilon()) {
+            return 0.0;
+        }
+
+        // Calculate distance to the node from the camera
+        glm::dvec3 cameraToNode = nodePosition - camera->positionVec3();
+        double distance = glm::length(cameraToNode);
```
WeirdRubberDuck, line 79: There is a `glm::distance` function that takes two positions

WeirdRubberDuck, line 79:
```suggestion
        double distance = glm::distance(nodePosition, camera->positionVec3());
```
?

modules/telemetry/src/util.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+        // Convert from meters to desired unit
+        return convertMeters(distance, unit);
+    }
+
+    // Horizontal angles
```
WeirdRubberDuck, line 85:
```suggestion
```
?   Seems to not only do the horizontal calculations.

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include <openspace/engine/windowdelegate.h>
+#include <openspace/rendering/renderengine.h>
+#include <openspace/scene/scene.h>
+
+namespace {
+    constexpr std::string_view _loggerCat = "TelemetryModule";
+
+    // The default OSC reciver is SuperCollider with these default values.
+    // However, the user can define any reciver in the openspace.cfg file as the
+    // ModuleConfiguration for the Telemetry module.
```
WeirdRubberDuck, line 49:
```suggestion
    // The default OSC receiver is SuperCollider with these default values.
    // However, the user can define any receiver in the openspace.cfg file as the
    // ModuleConfiguration for the Telemetry module.
```

Then we should not mention "over the OSC connection" in the comments in all the other files.

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+    constexpr openspace::properties::Property::PropertyInfo EnabledInfo = {
+        "Enabled",
+        "Enabled",
+        "Enable or disable all gathering of telemetry information"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo IpAddressInfo = {
+        "IpAddress",
+        "Ip address",
+        "The network ip address that the telemetry osc messages is sent to"
```
WeirdRubberDuck, line 62:
```suggestion
        "The network IP address that the telemetry OSC messages is sent to."
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    constexpr openspace::properties::Property::PropertyInfo EnabledInfo = {
+        "Enabled",
+        "Enabled",
+        "Enable or disable all gathering of telemetry information"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo IpAddressInfo = {
+        "IpAddress",
+        "Ip address",
```
WeirdRubberDuck, line 61:
```suggestion
        "IP Address",
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+    constexpr openspace::properties::Property::PropertyInfo PortInfo = {
+        "Port",
+        "Port",
+        "The network port that the telemetry osc messages is sent to"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo AngleCalculationModeInfo = {
+        "AngleCalculationMode",
+        "Angle Calculation Mode",
+        "This setting changes the method to calculate any angles in the telemetries. "
```
WeirdRubberDuck, line 74:
```suggestion
        "This setting changes the method for how the angles in the telemetries are computed. "
```

Not completely sure on this phrasing. Is "in the telemetries" correct here? 🙃

modules/telemetry/telemetrymodule.cpp
WeirdRubberDuck, line 76:
```suggestion
        "The Horizontal mode generally works well for flat displays or forward facing "
        "immersive environments. The Circular mode generally works well for centered "
```

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+        "The Horizontal mode, generally works well for flat displays or forward facing "
+        "immersive envierments. The Circular mode, generally works well for centered "
+        "fisheye displays or omnidirectional immersive environments"
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo IncludeElevationAngleInfo = {
+        "IncludeElevationAngle",
+        "Include Elevation Angle",
+        "This setting determines if an additional elevation angle should be calculated "
```
WeirdRubberDuck, line 83: What is the elevation angle, though?  I guess this is what your docs page will explain

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+    _telemetries.push_back(telemetry);
+    addPropertySubOwner(telemetry);
+}
+
+void TelemetryModule::internalDeinitialize() {
+    // Stop the loop before joining and tell the thread it is ok to run last itteration
+    _isRunning = false;
+    syncToMain.notify_one();
+
+    // Join the thread
```
WeirdRubberDuck, line 216: This comment seems superfluous, unless it is elaborated on, e.g. about what "the thread" is

modules/telemetry/telemetrymodule.cpp
```diff
@@ -0,0 +1,314 @@
...
+        // Wait for the main thread
+        //LDEBUG("The telemetry thread is waiting for a signal from the main thread");
+        std::unique_lock<std::mutex> lk(mutexLock);
+        syncToMain.wait(lk);
+        //LDEBUG(
+        //    "The telemetry thread is working after having received a signal from "
+        //    "the main thread"
+        //);
+
+        // Check if the module is even enabled
```
WeirdRubberDuck, line 265:
```suggestion
        // No need to update if the module isn't currently enabled
```
?

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+    constexpr static const char* Name = "Telemetry";
+
+    /**
+     * This mode only affects telemetries that send angle information. For example, the
+     * NodesTelemetry and the PlanetsSonification. More documentation for the elevation
+     * part of the angles is located further below.
+     *
+     * Horizontal: Calculate the angle to the object in the horizontal direction within
+     *             the camera plane. This mode is suitable for flat displays or forward
+     *             facing immersive envierments (such as the Norrköping dome theater).
```
WeirdRubberDuck, line 54:
```suggestion
     *             facing immersive environments (such as the Norrköping dome theater).
```

WeirdRubberDuck: Should we really refer to the Norrköping dome here?  Could instead say ", like a uni-driectional dome theater", to make it meaningful for a larger number of people that are not familiar with our specific dome

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Get the current angle calculation mode used in the telemetry module
+     *
+     * \return The angle calculation mode
+     */
+    AngleCalculationMode angleCalculationMode() const;
+
+    /**
+     * Return whether any elevation angles are being caclulated and sent over the osc
```
WeirdRubberDuck, line 118:
```suggestion
     * Return whether any elevation angles are being caclulated and sent over the OSC
```

however, apparently it could be something else than OSC?

modules/telemetry/telemetrymodule.h
```diff
@@ -0,0 +1,182 @@
...
+     *
+     * True: An additional angle in the vertical direction within the camera view + up
+     *       plane is sent over the osc connection. The camera view + up plane is the
+     *       plane of the camera view direction + camera up direction, with the camera
+     *       right direction as the normal. The elevation angle goes from -pi/2 to pi/2 in
+     *       radians and zero is straight forward in the camera view direction. Positive
+     *       elevation angles is in the up direction of the camera, and negative angles
+     *       are in the down direction of the camera.
+     *
+     * False: The elevation angle sent over the osc connection is always set to 0.0
```
WeirdRubberDuck, line 169:
```suggestion
     * False: The elevation angle sent over the OSC connection is always set to 0.0
```

WeirdRubberDuck (changes requested):
Did not have the time to try it out before the holiday, but here are some code comments at least! 


See you in the new near - in which I could also potentially try it out

modules/telemetry/include/general/anglemodetelemetry.h
```diff
@@ -0,0 +1,81 @@
...
+    AngleModeTelemetry(const std::string& ip, int port);
+    virtual ~AngleModeTelemetry() override = default;
+
+    /**
+     * Main update function to gather angle mode telemetry information (Angle calculation
+     * mode, and whether elevation angle is included) and send it via the osc connection.
+     *
+     * \param camera The camera in the scene (not used in this case)
+     */
+    virtual void update(const Camera*) override;
```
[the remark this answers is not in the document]
Roxeena: Sometimes, the camera is not used, such as in this case, wouldn't this create a warning of an unused parameter?

modules/telemetry/src/general/nodestelemetry.cpp
```diff
@@ -0,0 +1,280 @@
...
+        "HighAnglePrecision",
+        "Angle Precision (High)",
+        "The precision in radians used to determin when to send updated angle data "
+        "over the OSC connection. This is the higher precision used, for the current "
+        "focus node",
+        openspace::properties::Property::Visibility::User
+    };
+
+} // namespace
+#include "nodestelemetry_lua.inl"
```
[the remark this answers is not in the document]
Roxeena: Isnt this how we always include `_lua.inl` files? See https://github.com/OpenSpace/OpenSpace/blob/d4c4c92a5eead01ea1f8403144b614ad70c99330/modules/base/rendering/screenspacedashboard.cpp#L63 and https://github.com/OpenSpace/OpenSpace/blob/d4c4c92a5eead01ea1f8403144b614ad70c99330/src/scene/scene.cpp#L57

modules/telemetry/src/specific/planetscomparesonification.cpp
```diff
@@ -0,0 +1,299 @@
...
+        "When a planet is selected to be compared in any of the drop down menus below, "
+        "it is also upscaled as a visual indicator of which planets are currently being "
+        "compared. This property determines over how many seconds the scaling animation "
+        "should play."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo FirstOptionInfo = {
+        "FirstOption",
+        "Choose a planet to compare",
+        "Choose a planet in the given list to compare"
```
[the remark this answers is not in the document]
Roxeena: I don't like how this end up looking in the GUI. I dont want the user to have to read `'SecondOption'`. And also it is not clear what the `'SecondOption'` is anyway so I am thinking of keeping this as is.

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    _venusProperty.toggleAll.setValue(_toggleAll);
+    _earthProperty.toggleAll.setValue(_toggleAll);
+    _marsProperty.toggleAll.setValue(_toggleAll);
+    _jupiterProperty.toggleAll.setValue(_toggleAll);
+    _saturnProperty.toggleAll.setValue(_toggleAll);
+    _uranusProperty.toggleAll.setValue(_toggleAll);
+    _neptuneProperty.toggleAll.setValue(_toggleAll);
+}
+
+// Mercury
```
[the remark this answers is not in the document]
Roxeena: I would like to keep these in since I think it makes it easier to know where you are when you scroll through this swamp of "repetitive but not quite the same" code. And there arent any equally clear markers already in the code that replaces them. So I think it adds redability in this case to have them.

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+    _neptuneProperty.atmosphereEnabled.setValue(_neptuneProperty.toggleAll);
+    _neptuneProperty.moonsEnabled.setValue(_neptuneProperty.toggleAll);
+}
+void PlanetsSonification::onNeptuneSettingChanged() {
+    sendPlanetData(NeptuneIndex);
+}
+
+bool PlanetsSonification::getData(const Camera* camera, int planetIndex,
+                               TelemetryModule::AngleCalculationMode angleCalculationMode,
+                                  bool includeElevation) {
```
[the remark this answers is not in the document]
Roxeena: This needs to be clearified in the coding style. Because yes argument 2 does not fit and is therefor right-aligned. But argument 3 does fit, shouldn't that then be aligned with the paranthesis as normal? Or is the rule the once one argument i right aligned then all of them are? Or are all the following arguments after the right align? This is what the code style says right now (that I could find) and it is not clear what happens after the last argument here: 

```
// 2. When parameters are split up, the
// following parameters are all aligned
// with the opening ( of the parameter list
// or right-aligned, if they don't fit
ReturnType function(int param1, int p2,
                    int p3, int p4,
                    int parameter2,
               int reallyLongParameterName) 
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -0,0 +1,779 @@
...
+
+void PlanetsSonification::addPlanet(ghoul::Dictionary dict) {
+    const Parameters p = codegen::bake<Parameters>(dict);
+    DataBody planet(p.name);
+
+    if (p.moons.has_value()) {
+        for (const std::string& moon : *p.moons) {
+            planet.moons.push_back(moon);
+        }
+    }
```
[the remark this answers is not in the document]
Roxeena: I tried this and it complaines there there is no matching operator with these types. And besides I took away the default empty string argument so this would not work anyway then, right?

modules/telemetry/src/util.cpp
```diff
@@ -0,0 +1,478 @@
...
+    constexpr std::string_view _loggerCat = "TelemetryUtil";
+
+    /**
+     * Get the position of the node with the given identifier
+     *
+     * \param nodeIdentifier The identifier of the node to get the position for
+     *
+     * \return The position of the node with the given identifier
+     */
+    glm::dvec3 getNodePosition(std::string nodeIdentifier) {
```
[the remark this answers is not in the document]
Roxeena: I made it into:
`glm::dvec3 nodePosition(const std::string& nodeIdentifier) {`
Since I will use it as a string here anyway:
`openspace::SceneGraphNode* node = openspace::sceneGraphNode(nodeIdentifier);`
Let me know what you think tho.

modules/telemetry/include/specific/planetscomparesonification.h
```diff
@@ -0,0 +1,140 @@
...
+#define __OPENSPACE_MODULE_TELEMETRY___PLANETSCOMPARESONIFICATION___H__
+
+#include <modules/telemetry/include/telemetrybase.h>
+
+#include <openspace/properties/scalar/doubleproperty.h>
+#include <openspace/properties/optionproperty.h>
+
+namespace openspace {
+
+class PlanetsCompareSonification : public TelemetryBase {
```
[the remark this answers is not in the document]
Roxeena: The specific telemetries that are tailored made for a sonification keeps the sonification name since they are heavily tailored to the perticular purpose.

I have this information already in the documentaiton page I am writing. I can add it here too but it might be "kaka på kaka" and more copy pasted code/text.

modules/telemetry/src/general/anglemodetelemetry.cpp
```diff
@@ -0,0 +1,110 @@
...
+        _includeElevation = includeElevation;
+        shouldSendData = true;
+    }
+
+    // Make sure that the first message is sent, even if the values are defualt and no
+    // change has been detected
+    if (!_isInitialized) {
+        shouldSendData = true;
+        _isInitialized = true;
+    }
```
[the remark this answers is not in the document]
Roxeena: The data needs to be update before it can be sent. The data is updated in the two if statements above this one. So this one needs to be after that, otherwise it will send "empty" data.

modules/telemetry/src/specific/planetscomparesonification.cpp
Roxeena: Since there is no risk of anything breaking I dont think it should be a warning. Why do you think it should be a warning?

modules/telemetry/src/specific/planetssonification.cpp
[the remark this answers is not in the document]
alexanderbock: Yea, the rule is that once a parameter is too long, it switches over to "all remaining parameters are right aligned".  I'll clarify it in the document

modules/telemetry/src/specific/planetssonification.cpp
Roxeena: The other telemetries/sonifications does not have "data objets" in the same way, only the NodesTelemetry and the PlanetsSonification has that. All other telemetries/sonifications are based on the general state of OpenSpace or some GUI settings.

modules/telemetry/src/specific/planetssonification.cpp
Roxeena: See my reply to Alex's comment above:

> Isnt this how we always include `_lua.inl` files? See
> 
> https://github.com/OpenSpace/OpenSpace/blob/d4c4c92a5eead01ea1f8403144b614ad70c99330/modules/base/rendering/screenspacedashboard.cpp#L63
> and
> 
> https://github.com/OpenSpace/OpenSpace/blob/d4c4c92a5eead01ea1f8403144b614ad70c99330/src/scene/scene.cpp#L57

modules/telemetry/src/util.cpp
Roxeena: So this is another issue I have (especially in the documentation) there are "horizontal angles" and "vertical angles" and both of them can be calculated in a `Horizontal` or `Circular` mode. And the names "horizontal" and "vertical" does not really make sense in the circular case... I am having a really hard time finding the best words to use for all of these cases. Maybe something more abstract like `Linear` and `Radial` is better words for the modes? And then just "main angle" and "elevation angle"? Please let me know if you have any ideas.

modules/telemetry/telemetrymodule.h
Roxeena: I tried to combine both here. Since I want to mention that we have tailored it to certain environments so it might not work flawlessly or wihtout additionall setup for any other venue. However, at the same time I wanted to keep it general for that type of venue. I could also move this information to the docs page solely or have it in both places, what doe you think?

WeirdRubberDuck: Either way I think we should avoid referring to "the Norrköping dome" since this implies that the reader is familiar with our specific dome and we can't assume that all people reading this are.  This goes for both comments in code and on the documentation page. 

You could instead say something like "such as a unidirectional dome with forward-facing seats". 

I think having the comments in the code as well as on the docs makes sense here, since a person who finds this code might not know that the documentation page exists.

modules/telemetry/src/general/anglemodetelemetry.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: But if any of those if-statements are triggered, `shouldSendData` should already be set to true? So then why do we need to set it here? 🤔  A bit confused about this part fo the code. 

But maybe it's not really a big deal

data/assets/examples/sonification/OpenSpace-OSC-example.scd
```diff
@@ -0,0 +1,391 @@
...
+OSCdef.new(
+	\Mode,
+	{
+		arg msg;
+
+		// msg[0] = OSC Label
+		// msg[1] = Angle Calculation Mode. Determines what type of angle the other
+		//          Sonifications are sending. More documentaiton on how these angles work
+		//          can be found on the general OpenSpace documentation page
+		//          https://docs.openspaceproject.com/en/latest/
```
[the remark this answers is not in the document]
WeirdRubberDuck: Shouldn't it link to the page you created that describe the angles?

modules/telemetry/src/specific/planetscomparesonification.cpp
WeirdRubberDuck: Nothing specific really, just wanted to check :)

modules/telemetry/src/specific/planetssonification.cpp
WeirdRubberDuck: Hm, ok, so the others are not added from an asset at all?  Thinking if it's worth to add an empty struct or not, since this creates a pa

For this one, you can add a comment describing it above this line.  That text will then show up on the Asset component page in the documentation :) Here's an example: https://github.com/OpenSpace/OpenSpace/blob/d4c4c92a5eead01ea1f8403144b614ad70c99330/modules/exoplanets/tasks/exoplanetsdatapreparationtask.cpp#L65

And note that you need to add the documentation function as well and add it to the documentation for this module as well.

modules/telemetry/src/specific/planetssonification.cpp
WeirdRubberDuck: So, scene.cpp does it correctly here (note that the incude is BEFORE the empty `namespace`), while screenspacedashboard does not. At least to how I understand that we do it

modules/telemetry/src/specific/planetscomparesonification.cpp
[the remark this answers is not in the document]
WeirdRubberDuck: "... to compare to the other selected planet" ? 

It would be good to make it clearer what it is comparing to. It's also a bit confusing that the "name" part of the two properties are the same

modules/telemetry/include/specific/planetsoverviewsonification.h
```diff
@@ -0,0 +1,96 @@
...
+     * Create an osc::Blob object with the current planets overview sonification settings.
+     * Order of settings: Mercury, Venus, Earth, Mars, Jupiter, Saturn, Uranus, Neptune
+     *
+     * \return An osc::Blob object with the current planets overview sonification settings
+     */
+    osc::Blob createSettingsBlob() const;
+
+    /**
+     * Send current planets overview sonification settings over the osc connection
+     * Order of data: Planets overview settings
```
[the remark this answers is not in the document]
Roxeena: I rephrased it as: 
`The order of sent data is as follows:`
Is that better?

modules/telemetry/src/specific/planetsoverviewsonification.cpp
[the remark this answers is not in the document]
Roxeena: Is this clearer?
```
"Sonification that gives an overview of the planets in the solar system with "
"fewer audible features for each planet compared to the regular planets "
"sonification."
```

modules/telemetry/telemetrymodule.cpp
Roxeena: Is this small change any better? (Copilot thought so...)
```
"This setting changes the method used to calculate any angles in the "
"telemetries.
```

modules/telemetry/telemetrymodule.cpp
Roxeena: Should this still be explained briefly here? It can be very specific to surround sound setups but could also be interesting for other purposes. Or should I just include a link to the docs?

modules/telemetry/src/general/nodestelemetry.cpp
```diff
@@ -147,35 +143,61 @@ NodesTelemetry::PrecisionProperties::PrecisionProperties(
...
+    }
+    TelemetryModule::AngleCalculationMode angleMode = module->angleCalculationMode();
+    bool includeElevation = module->includeElevationAngle();
 
-    // Vertical Angle
-    data[VerticalAngleIndex] = _nodes[nodeIndex].data[VerticalAngleIndex];
+    // Update data for all nodes
+    for (int i = 0; i < _nodes.size(); ++i) {
+        // Increase precision if the node is in focus
+        if (focusNode->identifier() == _nodes[i].identifier) {
```
alexanderbock, line 172:
```suggestion
    for (const TelemetryNode& node : _nodes) {
        // Increase precision if the node is in focus
        if (focusNode->identifier() == node.identifier) {
```

modules/telemetry/src/general/nodestelemetry.cpp
```diff
@@ -147,35 +143,61 @@ NodesTelemetry::PrecisionProperties::PrecisionProperties(
...
 }
 
-bool NodesTelemetry::getData(const Camera* camera, int nodeIndex,
-                             TelemetryModule::AngleCalculationMode angleCalculationMode,
-                             bool includeElevation)
+// Empty overidded functions
+bool NodesTelemetry::updateData(const Camera*) {
+    return false;
+}
+void NodesTelemetry::sendData() {}
```
alexanderbock, line 197:
```suggestion
void NodesTelemetry::sendData() {}

```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -379,21 +378,98 @@ PlanetsSonification::PrecisionProperties::PrecisionProperties(
...
+    TelemetryModule* module = global::moduleEngine->module<TelemetryModule>();
+    if (!module) {
+        LERROR("Could not find the TelemetryModule");
+        return;
+    }
+
+    // The compare planets and planets overview sonifications depend on this
+    // sonification. If any of them are enabled, then this sonificaiton sends planetary
+    // data to the Open Sound Control receiver, even if this sonificaiton is disabled.
+    // @NOTE (malej 2024-11-28): Enabeling any combination of these three sonifications
```
alexanderbock, line 398:
```suggestion
    // @NOTE (malej 2024-11-28): Enabling any combination of these three sonifications
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -379,21 +378,98 @@ PlanetsSonification::PrecisionProperties::PrecisionProperties(
...
+
+    if (!focusNode) {
+        // The scene is likely not yet initialized
+        return;
+    }
+
+    // Update data for all planets
+    for (int i = 0; i < _planets.size(); ++i) {
+        // Increase presision if the planet is in focus
+        if (focusNode->identifier() == _planets[i].name) {
```
alexanderbock, line 433:
```suggestion
    for (const DataBody& planet : _planets) {
        // Increase precision if the planet is in focus
        if (focusNode->identifier() == planet.name) {
```

modules/telemetry/src/specific/planetssonification.cpp
```diff
@@ -379,21 +378,98 @@ PlanetsSonification::PrecisionProperties::PrecisionProperties(
...
+            sendData(i);
+        }
+    }
+}
+
+void PlanetsSonification::stop() {
+    _toggleAll = false;
+}
+
+void PlanetsSonification::addPlanet(ghoul::Dictionary dict) {
```
alexanderbock, line 455:
```suggestion
void PlanetsSonification::addPlanet(const ghoul::Dictionary& dict) {
```

modules/telemetry/src/telemetrybase.cpp
```diff
@@ -58,9 +60,24 @@ TelemetryBase::TelemetryBase(properties::PropertyOwner::PropertyOwnerInfo info,
...
+    stop();
     delete _connection;
 }
 
+void TelemetryBase::update(const Camera* camera) {
+    if (!_enabled) {
+        return;
+    }
+
+    bool dataWasUpdated = updateData(camera);
```
alexanderbock, line 72:
```suggestion
    const bool dataWasUpdated = updateData(camera);
```

alexanderbock (approved):

modules/telemetry/src/general/nodestelemetry.cpp
Roxeena: I need the node index when calling the `updateData` function further down:
`bool dataWasUpdated = updateData(camera, i, angleMode, includeElevation);`
So I want to keep it as is instead of having to introduce a counter to keep track of the index anyway.

modules/telemetry/src/specific/planetssonification.cpp
Roxeena: Same here, I need the `i` for the `updateData` function.

data/assets/examples/sonification/OpenSpace-OSC-example.scd
```diff
@@ -0,0 +1,391 @@
...
+		// msg[3] = The Z position of the camera in the world
+		// msg[4] = The angle in radians for the current quaternion rotation of the
+		//          camera (w)
+		// msg[5] = The X component of the unit rotation axis in the current quaternion
+		//          rotation of the camera
+		// msg[6] = The Y component of the rotation axis in the quaternion rotation
+		// msg[7] = The Z component of the rotation axis in the quaternion rotation
+		// msg[8] = The current speed of the camera,
+		//          specified in the selected distance unit per second
+		// msg[9] = The selected distance unit for the speed of the camera, as a string
```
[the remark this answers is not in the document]
Roxeena: Should I add a new documentation page about it too then? I was thinking that it is the equivalent of "längdenhet" in swedish and that you would understand what type of unit that would be. I could also just add an example of it here as I have done now for the ducumentation page, would that make it clearer?

alexanderbock:
Closes #251 when merged

PR #3465, opened 2024-12-09
Use a pointing hand GLFW cursor for the web grab cursor

WeirdRubberDuck:
This changes the cursor to a pointing hand whenever a "grab" or "grabbing" is requested from the browsers. 

Note that there seemingly is no grab cursor in GLFW ([source](https://www.glfw.org/docs/latest/group__shapes.html)), and we currently use the regular "pointer" as default. Changing from a regular pointer to a pointing hand makes it much clearer that things that represent "drag handles" can be interacted with, IMO.   

Website: 

https://github.com/user-attachments/assets/5fe19313-a146-43ae-b092-eddd0a1ec51b

OpenSpace

https://github.com/user-attachments/assets/152e51cc-3c79-4ee9-aa10-2a743b4de126


ylvaselling:
It is also possible to add our own custom cursors. We could add the dragging cursor later on?

alexanderbock (approved):

PR #3467, opened 2024-12-16
Add jump to time lua function

engbergandreas:
Adds a lua function to interpolate or fade to a specified time example use: 
Automatically decide if it should interpolate or fade using default values
`openspace.time.jumpToTime('1968-12-24T22:20:00.000')`

Sets the threshold to 12h the fade duration to 2 seconds and the interpolation duration to 10 seconds
`openspace.time.jumpToTime('1968-12-24T10:20:00.000', 43200, 2, 10)`

### DIFF
--- a/src/navigation/navigationhandler.cpp
+++ b/src/navigation/navigationhandler.cpp
@@ -100,6 +100,16 @@ namespace {
         "again.",
         openspace::properties::Property::Visibility::User
     };
+
+    constexpr openspace::properties::Property::PropertyInfo FadeJumpThresholdInfo = {
+        "FadeJumpThreshold",
+        "Fade Jump Threshold",
+        "The threshold that determines when a transition should fade instead of "
+        "interpolate. If the time difference exceeds this threshold, the rendering will "
+        "fade to black, set time, and then fade in again. Otherwise the transition will "
+        "interpolate without fading.",
+        openspace::properties::Property::Visibility::User
+    };
 } // namespace
 
 namespace openspace::interaction {
@@ -111,6 +121,7 @@ NavigationHandler::NavigationHandler()
     , _disableJoystickInputs(DisableJoystickInputInfo, false)
     , _useKeyFrameInteraction(FrameInfo, false)
     , _jumpToFadeDuration(JumpToFadeDurationInfo, 1.f, 0.f, 10.f)
+    , _fadeJumpThreshold(FadeJumpThresholdInfo, 86400.f, 0.f, 86400.f, 1.f)
 {
     addPropertySubOwner(_orbitalNavigator);
     addPropertySubOwner(_pathNavigator);
@@ -120,6 +131,7 @@ NavigationHandler::NavigationHandler()
     addProperty(_disableJoystickInputs);
     addProperty(_useKeyFrameInteraction);
     addProperty(_jumpToFadeDuration);
+    addProperty(_fadeJumpThreshold);
 }
 
 NavigationHandler::~NavigationHandler() {}
@@ -188,6 +200,10 @@ float NavigationHandler::jumpToFadeDuration() const {
 
 float NavigationHandler::interpolationTime() const {
     return _orbitalNavigator.retargetInterpolationTime();
+}
+
+float NavigationHandler::fadeJumpThreshold() const {
+    return _fadeJumpThreshold;
 }
 
 void NavigationHandler::setInterpolationTime(float durationInSeconds) {
--- a/src/util/time_lua.inl
+++ b/src/util/time_lua.inl
@@ -23,6 +23,7 @@
  ****************************************************************************************/
 
 #include <openspace/util/timeconversion.h>
+#include <openspace/navigation/navigationhandler.h>
 
 namespace {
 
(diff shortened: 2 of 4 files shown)


src/navigation/navigationhandler.cpp
```diff
@@ -100,6 +100,16 @@ namespace {
         "again.",
         openspace::properties::Property::Visibility::User
     };
+
+    constexpr openspace::properties::Property::PropertyInfo FadeJumpThresholdInfo = {
```
WeirdRubberDuck, line 104: Oh, do we really want a property for this?  It's a bit unclear that it is related to time, and it's not really related to the NavigationHandler at all.  **If we want to keep it, it should potentially be moved to the `TimeMananger` instead. ** To me, that seems like a more natural placement, since the function is located in `openspace.time`. Thoughts?  

However, IIRC we has a discussion of whether we even want this feature of automatically fading or not depending on time. It can seem a bit "magic" or confusing to our users.

src/navigation/navigationhandler.cpp
```diff
@@ -100,6 +100,16 @@ namespace {
...
+        "FadeJumpThreshold",
+        "Fade Jump Threshold",
+        "The threshold that determines when a transition should fade instead of "
```
WeirdRubberDuck, line 107: Clarify that this goes for a transition when setting the _time_

However, I still feel like we should remove this property and feature. If we keep it, I have a feeling that it should be applied in more cases where we set the time, but that would be too confusing.

src/util/time_lua.inl
```diff
@@ -327,6 +328,66 @@ namespace {
...
+ * \param fadeDuration The duration (in seconds) of the fade transition, if triggered.
+ *                     If excluded, the default fade duration specified in the
+ *                     NavigationHandler is used.
+ * \param interpolationDuration The duration (in seconds) over which interpolation should
+ *                               occur if a fade transition is not triggered. If excluded,
+ *                               the default value for time interpolation specified in the
+ *                               TimeManager is used.
+ */
+[[codegen::luawrap]] void jumpToTime(std::variant<double, std::string> time,
+    std::optional<double> fadeJumpThreshold,
```
WeirdRubberDuck, line 356: See previous comment. Remove this ?

src/util/time_lua.inl
```diff
@@ -327,6 +328,66 @@ namespace {
...
+ * If the time difference between the current time and the target time exceeds the
+ * specified threshold, a fade transition will be triggered. Otherwise, interpolation is
+ * used.
+ *
+ * \param time The target time to jump to. If the parameter is a number, it represents the
+ *             number of seconds past the J2000 epoch. If it is a string, it must be a
+ *             valid ISO 8601-like date string of the format YYYY-MM-DDTHH:MN:SS.
+ * \param fadeJumpThreshold The time difference threshold (in seconds) to determine
+ *                          whether to perform a fade transition. If excluded, the default
+ *                          value specified in the NavigationHandler is used.
```
WeirdRubberDuck, line 346: Maybe TimeManager, if we want to keep it

WeirdRubberDuck (changes requested):
To decide:  Do we want to keep the feature of either interpolating or fading, or do we always want to fade?

Either way, there are some changes that I consider needed.

alexanderbock:
> To decide: Do we want to keep the feature of either interpolating or fading, or do we always want to fade?
> 
> Either way, there are some changes that I consider needed.

I think we definitely need the ability to either interpolate or fade depending on the use-case.  Thinking of different kinds of exhibitions for example that want to do one type or the other

WeirdRubberDuck:
> > To decide: Do we want to keep the feature of either interpolating or fading, or do we always want to fade?
> > Either way, there are some changes that I consider needed.
> 
> I think we definitely need the ability to either interpolate or fade depending on the use-case. Thinking of different kinds of exhibitions for example that want to do one type or the other

To clarify, I meant the feature of fading or interpolating _based on whether the time difference exceeds a given value_. Is this what you mean as well, @alexanderbock ?

But yeah, I definitely agree that we need both options, in general. Note that we already have the `openspace.time.interpolateTime` function for interpolating

src/util/time_lua.inl
```diff
@@ -327,6 +328,66 @@ namespace {
...
+ *                               the default value for time interpolation specified in the
+ *                               TimeManager is used.
+ */
+[[codegen::luawrap]] void jumpToTime(std::variant<double, std::string> time,
+    std::optional<double> fadeJumpThreshold,
+    std::optional<double> fadeDuration,
+    std::optional<double> interpolationDuration)
+{
+    using namespace openspace;
+    double t;
```
ylvaselling, line 361: Rename t to something more descriptive, like newTime or something?

src/util/time_lua.inl
```diff
@@ -327,6 +328,66 @@ namespace {
...
+    if (timeDiffInSeconds > threshold) {
+        global::navigationHandler->triggerFadeToTransition(
+            std::format("openspace.time.setTime({})", t),
+            static_cast<float>(fadeTime)
+        );
+    }
+    else {
+        interpolateTime(t, interpolationDuration);
+    }
+}
```
ylvaselling, line 389:
```suggestion
    const double currentTime = global::timeManager->time().j2000Seconds();

    const double timeDiffInSeconds = std::abs(currentTime - t);
    
    // Fade transition
    if (timeDiffInSeconds > threshold) {
        const double fadeTime = fadeDuration.value_or(
            global::navigationHandler->jumpToFadeDuration()
        );
        global::navigationHandler->triggerFadeToTransition(
            std::format("openspace.time.setTime({})", t),
            static_cast<float>(fadeTime)
        );
    }
    else {
        interpolateTime(t, interpolationDuration);
    }
}
```

ylvaselling:
> > > To decide: Do we want to keep the feature of either interpolating or fading, or do we always want to fade?
> > > Either way, there are some changes that I consider needed.
> > 
> > 
> > I think we definitely need the ability to either interpolate or fade depending on the use-case. Thinking of different kinds of exhibitions for example that want to do one type or the other
> 
> To clarify, I meant the feature of fading or interpolating _based on whether the time difference exceeds a given value_. Is this what you mean as well, @alexanderbock ?
> 
> But yeah, I definitely agree that we need both options, in general. Note that we already have the `openspace.time.interpolateTime` function for interpolating

We could also make the function take a bool that sets if it should fade or not, instead of the threshold, if we don't want to have a threshold

ylvaselling (approved):
👍

alexanderbock:
> > > > To decide: Do we want to keep the feature of either interpolating or fading, or do we always want to fade?
> > > > Either way, there are some changes that I consider needed.
> > > 
> > > 
> > > I think we definitely need the ability to either interpolate or fade depending on the use-case. Thinking of different kinds of exhibitions for example that want to do one type or the other
> > 
> > 
> > To clarify, I meant the feature of fading or interpolating _based on whether the time difference exceeds a given value_. Is this what you mean as well, @alexanderbock ?
> > But yeah, I definitely agree that we need both options, in general. Note that we already have the `openspace.time.interpolateTime` function for interpolating
> 
> We could also make the function take a bool that sets if it should fade or not, instead of the threshold, if we don't want to have a threshold

I had in mind a global setting that determines the threshold above which we fade and below we interpolate. By making that value a property somewhere in the "Settings", we allow people that always want to fade to set this value to `0` and people that always want to interpolate to set it to `FLT_MAX`

PR #3478, opened 2025-01-07
Polish `RenderableCartesianAxes` - documentation, examples, and make fading work correctly

WeirdRubberDuck:
* Add opacity property and make axes fade like every other renderable
* Small update to existing documentation to utilize Markdown features and links to other classes
* Add examples for specifying a Parent node and using custom colors

### DIFF
--- a/data/assets/examples/renderable/renderablecartesianaxes/cartesianaxes_customcolor.asset
+++ b/data/assets/examples/renderable/renderablecartesianaxes/cartesianaxes_customcolor.asset
@@ -1,7 +1,7 @@
 -- Custom Colors
 -- This example creates a set of cartesian coordinate axes with specified colors, instead
 -- of using the default which is red (X), green (Y), and blue (Z). The parent is not set
--- which defaults to placing the axes at the center of the Sun.
+-- which defaults to placing the axes at the center of the Solar System.
 
 local Node = {
   Identifier = "RenderableCartesianAxes_Example_CustomColors",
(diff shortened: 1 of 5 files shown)


ylvaselling (approved):

data/assets/examples/renderable/renderablecartesianaxes/cartesianaxes_parent.asset
```diff
@@ -0,0 +1,29 @@
+-- With Parent
+-- This example creates a SceneGraphNode that displays coordinate axes of the given parent
+-- node, in this case Earth.
+
+local Node = {
+  Identifier = "RenderableCartesianAxes_Example_Parent",
+  Parent = "Earth",
```
ylvaselling, line 7: Do we want to use the identifier imported from the earth asset here?

WeirdRubberDuck: Yeah, that would be better.

WeirdRubberDuck: However, I do find myself wondering what we said about this when we decided to keep the examples to do a minimal amount of things 🤔 

But in general, to me it seems better to write the examples like we would write them if it was any other asset, to avoid discrepancies and risk of teaching "bad" asset writing to users

data/assets/examples/renderable/renderablecartesianaxes/cartesianaxes_customcolor.asset
```diff
@@ -0,0 +1,32 @@
+-- Custom Colors
+-- This example creates a set of cartesian coordinate axes with specified colors, instead
+-- of using the default which is red (X), green (Y), and blue (Z). The parent is not set
+-- which defaults to placing the axes at the center of the Sun.
```
alexanderbock, line 4:
```suggestion
-- which defaults to placing the axes at the center of the Solar System.
```

data/assets/examples/renderable/renderablecartesianaxes/cartesianaxes_parent.asset
alexanderbock: In other cases I did add the `asset.require`.  I think it is more important to have the examples be loadable in all cases

alexanderbock (approved):

PR #3479, opened 2025-01-07
Docs and examples for `RenderableBoxGrid`

WeirdRubberDuck:
Adds examples for `RenderableBoxGrid`, as well as a short description for the docs page. 

Edit:  Concluded to remove the Labels component. Below are the notes I added in the PR originally. 
❗ Breaking change:  `RenderableBoxGrid` no longer supports labels 
> _OBS!_ I noticed that the renderable has the option to add a labels component.  However, we don't use this feature anywhere, and when doing a (very small and not at all thorough) attempt to add labels I couldn't get them to show. 
> 
> Do we want to keep the Labels Component feature for `RenderableBoxGrid`? My suggestion would be to remove it for now, since it's not used and adding labels is a bit unintuitive* with the current label code. And then add it again once the labels code has been rewritten (and we add labels to all the grid types 🙂). Thoughts?
> 
> (* The reason I find adding the label unintuitive is that it requires a separate file, and it's not clear how that file connects to the rendered box)

### DIFF
--- a/data/assets/examples/renderable/renderableboxgrid/boxgrid.asset
+++ b/data/assets/examples/renderable/renderableboxgrid/boxgrid.asset
@@ -1,0 +1,31 @@
+-- Basic
+-- This example adds a box grid, which is a 3D box rendered using grid lines, to the
+-- scene.
+--
+-- Per default the box will be given a size of 1x1x1 meter, and here it has been scaled
+-- up by a factor of 100. It will hence have a size of 100x100x100 meters.
+
+local Node = {
+  Identifier = "RenderableBoxGrid_Example",
+  Transform = {
+    Scale = {
+      Type = "StaticScale",
+      Scale = 100
+    }
+  },
+  Renderable = {
+    Type = "RenderableBoxGrid"
+  },
+  GUI = {
+    Name = "RenderableBoxGrid - Basic",
+    Path = "/Examples"
+  }
+}
+
+asset.onInitialize(function()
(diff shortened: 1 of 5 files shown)


data/assets/examples/renderable/renderableboxgrid/boxgrid.asset
```diff
@@ -0,0 +1,31 @@
+-- Basic
+-- This example adds a box grid, which is a 3D box rendered using grid lines, to the
+-- scene.
+--
+-- Per default the box will be given a size of 1x1x1 meter, and here it is scaled up by a
```
WeirdRubberDuck, line 5:
```suggestion
-- Per default, the box will be given a size of 1x1x1 meters, and here it is scaled up by a
```

data/assets/examples/renderable/renderableboxgrid/boxgrid_size.asset
```diff
@@ -0,0 +1,32 @@
+-- With Non-uniform Size
+-- This example created a box grid with a non-uniform size. The size has been set so the
```
ylvaselling, line 2:
```suggestion
-- This example creates a box grid with a non-uniform size. The size has been set so the
```

ylvaselling (approved):

alexanderbock:
Yea, I think it's for the better to remove the labels part from this

WeirdRubberDuck:
Removed the code for labels

PR #3482, opened 2025-01-07
Docs and examples for `RenderableDisc`

WeirdRubberDuck:
Note that this also includes changes to the default and/or min values for the `Size` and `Width`. 

Replaces the old combined example asset with three individual ones


data/assets/examples/renderable/renderabledisc/disc_hollow.asset
```diff
@@ -0,0 +1,30 @@
+-- Hollow
```
ylvaselling, line 1: How about "ring" instead of "hollow"? "hollow" to me sounds like a 3d object which is not filled ("ihålig"). Or "Disc with hole"

modules/base/rendering/renderabledisc.cpp
```diff
@@ -68,6 +68,11 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This renderable can be used to create a circular disc that is colored based on a
+    // one-dimensional texture.
+    //
+    // The disc will be filled i.e. a full circle, per default, but may also be made
+    // hollow using the `Width` parameter.
```
ylvaselling, line 75: see comment above

data/assets/examples/renderable/renderabledisc/disc.asset
```diff
@@ -0,0 +1,27 @@
+-- Basic
+-- This example creates a disc with a single color and a radius of 100 meters.
+
```
ylvaselling, line 3:
```suggestion
--
```

ylvaselling (changes requested):
Missing -- in the descriptions

data/assets/examples/renderable/renderabledisc/disc.asset
WeirdRubberDuck: Oops! Thanks :)

data/assets/examples/renderable/renderabledisc/disc_hollow.asset
WeirdRubberDuck: I would prefer "Disc With Hole" in that case 🤔

WeirdRubberDuck: Although ,technically, the hollow disc _is actually_ a 3D object which is not filled ;) So personally I find that name pretty suitable

Google seems to do so too. But I'm fine with changing it 
image

alexanderbock: I'd go with "hollow disk" "disk with hole" as well.  "Ring" sound more 3Dish to me than "Disk"

data/assets/examples/renderable/renderabledisc/disc_hollow.asset
```diff
@@ -0,0 +1,30 @@
...
+-- This example creates a disc that is made hollow by specifying a `Width` of 0.5, which
+-- will result in the disc stopping halfway from the edge to the center point.
+-- The disc is rendered with a single color and an outer radius of 100 meters.
+--
+-- This renderable requires a texture to be loaded, even for just a single color. Use the
+-- utility function that exists for creating single color textures for this purpose.
+local cyanTexture = openspace.createSingleColorImage("example_disc_color1", { 0.0, 1.0, 1.0 })
```
alexanderbock, line 8: Does it matter that this file name clashes with the base file example file?

alexanderbock (approved):

data/assets/examples/renderable/renderabledisc/disc_hollow.asset
WeirdRubberDuck: CHanged to "With Hole"

PR #3483, opened 2025-01-08
Make debug plane a `ScreenSpaceRenderable` and add documentation

WeirdRubberDuck:
Change the debug plane from a `Renderable` to a `ScreenSpaceRenderable` and add documentation, including an example.   

The change to screen space makes it easier to use, since there is no longer any need to find its position in 3D space.

**Importantly, this PR also adds our first `ScreenSpaceRenderable` example. There was no given format for this, so let me know what you think about the format.** Some things to note that are different from the examples for renderables: 
* I used `Object` instead of `Node` for the table variable name, since the thing added is not a scene graph node. 
* Since a `ScreenSpaceRenderable` has no GUI path, I added "Example" after the type in the name. Otherwise, the format is the same as for the other example names.

❗ Note that this is technically a breaking change since an existing renderable type has been removed


ylvaselling (approved):
Tried the example and it looks good to me!

modules/debugging/rendering/screenspacedebugplane.cpp
```diff
@@ -34,37 +34,34 @@ namespace {
...
-    );
+documentation::Documentation ScreenSpaceDebugPlane::Documentation() {
+    return codegen::doc<Parameters>("debugging_screenspace_debugplane");
 }
 
-RenderableDebugPlane::RenderableDebugPlane(const ghoul::Dictionary& dictionary)
-    : RenderablePlane(dictionary)
+ScreenSpaceDebugPlane::ScreenSpaceDebugPlane(const ghoul::Dictionary& dictionary)
+    : ScreenSpaceRenderable(dictionary)
     , _texture(TextureInfo, -1, -1, 512)
```
alexanderbock, line 54:
```suggestion
    , _texture(TextureInfo, -1, -1, 4096)
```

Should probably increase that since we already are using more that 512 textures

alexanderbock (approved):

PR #3486, opened 2025-01-09
Add a description for the `Renderable` base class

WeirdRubberDuck:


src/rendering/renderable.cpp
```diff
@@ -73,16 +73,27 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This is the base class for all `Renderable` types. These objects are responsible
+    // for rendering the visuals for scene graph nodes in the 3D rendering. Different
+    // `Renderable` types create various kinds of visuals, sometimes based on external
+    // data files or resources (such as images, 3D models, color maps, or datasets).
+    //
+    // A `Renderable` is created by adding it as a component to a scene graph node in an
```
ylvaselling, line 81:
```suggestion
    // A `Renderable` is created by adding it to a scene graph node in an
```
Don't think we use the word components in this case?

ylvaselling (approved):

alexanderbock (approved):

PR #3500, opened 2025-01-20
Automatically reload texture for `RenderableSphereImageLocal` when file changes

WeirdRubberDuck:
Utilize the existing `TextureComponent` to clean up the code and make the texture reload automatically if the file is changed on disk. 

Note that the `LazyLoading` input parameter has been removed. However, it was never actually used.

### DIFF
--- a/modules/base/rendering/renderablesphereimagelocal.cpp
+++ b/modules/base/rendering/renderablesphereimagelocal.cpp
@@ -44,11 +44,6 @@ namespace {
     struct [[codegen::Dictionary(RenderableSphereImageLocal)]] Parameters {
         // [[codegen::verbatim(TextureInfo.description)]]
         std::filesystem::path texture;
-
-        // If true, the image for this sphere will not be loaded at startup but rather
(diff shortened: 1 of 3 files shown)


modules/base/rendering/renderablesphereimagelocal.cpp
```diff
@@ -70,24 +65,24 @@ RenderableSphereImageLocal::RenderableSphereImageLocal(
...
     addProperty(_texturePath);
-    _textureFile->setCallback([this]() { _textureIsDirty = true; });
 }
 
 bool RenderableSphereImageLocal::isReady() const {
     return RenderableSphere::isReady() && _texture;
 }
 
+void RenderableSphereImageLocal::initialize() {
+    _texture = std::make_unique<TextureComponent>(2);
```
alexanderbock, line 77: Can we do all of this in the `initialize` or does it have to go into the `initializeGL`?

alexanderbock (approved):

PR #3503, opened 2025-01-29
cmake: honor OPENSPACE_HAVE_TESTS at more places

dawagner:
- don't build the codegen unit tests if the OpenSpace unit tests are disabled
- propagate the value of this option to Ghoul's corresponding option


dawagner:
On second thought, I'm not sure the current diff does what I'm trying to achieve. I marked it as draft for now.

dawagner:
Fixed it :)

PR #3505, opened 2025-01-29
Utilize new docs functionality in API scripts that are defined in Lua files

WeirdRubberDuck:
Just found an old branch from October last year where I had started making some updates to the Lua API functions that are defined in `.lua` files. 

The changes include using the new documentation features like adding parameter descriptions and using Markdown notation. It also formats the descriptions to be more consistent with the other script descriptions (with a dot at the end). 

This PR also:
* Fixes some issues with the gaia module's filtering functions. They were previously broken. 
* Makes it possible to create a `RenderableBoxGrid` that is smaller than 1.


alexanderbock (approved):
:+1:

PR #3507, opened 2025-01-31
Update time conversion to mirror distance conversion

Roxeena:
It made me sad that the new nice update to the distance conversion util file was not reflected in its sibling time conversion, so I made it so. Now, I am happy.

### DIFF
--- a/modules/base/rendering/renderablemodel.cpp
+++ b/modules/base/rendering/renderablemodel.cpp
@@ -351,26 +351,15 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
         if (std::holds_alternative<float>(*p.animationTimeScale)) {
             _animationTimeScale = std::get<float>(*p.animationTimeScale);
         }
-        else if (std::holds_alternative<std::string>(*p.animationTimeScale))
-        {
+        else if (std::holds_alternative<std::string>(*p.animationTimeScale)) {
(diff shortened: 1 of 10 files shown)


modules/base/rendering/renderablemodel.cpp
```diff
@@ -351,26 +351,15 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
-                _animationTimeScale = static_cast<double>(
-                    convertTime(1.0, timeUnit, TimeUnit::Second)
-                );
+            try {
+                TimeUnit timeUnit = timeUnitFromString(stringUnit);
+                _animationTimeScale =
+                    static_cast<double>(convertTime(1.0, timeUnit, TimeUnit::Second));
             }
-            else {
+            catch (const ghoul::MissingCaseException& e) {
```
alexanderbock, line 362: I think it might be better to explicitly check whether the element exist and not rely on caching the missing case exception.  That exception type is meant to propagate up to the base and actually crash, since it is more of a logic error than a runtime error

alexanderbock (approved):

WeirdRubberDuck (changes requested):
Looks good. However, I noticed that we have tests for the time conversion and many report a FAILED result. 

Looking at the changes, it seems like maybe these tests broke earlier. But please, take the opportunity to fix them as part of this PR :) 

image

PR #3508, opened 2025-02-04
Add new property hint to determine if a scene graph node should focusable

alexanderbock:
For use in the new UI implementation?   I don't think we need to make use of this in the current implementation.  The idea being that the "Focus" and "Fly to" and "Frame" buttons wouldn't show up for the scene graph nodes that have this property false.  They'd also not show up in the Focus list

### DIFF
--- a/include/openspace/scene/scenegraphnode.h
+++ b/include/openspace/scene/scenegraphnode.h
@@ -181,6 +181,7 @@ private:
     properties::StringProperty _guiDisplayName;
     properties::StringProperty _guiDescription;
     properties::BoolProperty _useGuiOrdering;
+    properties::BoolProperty _guiFocussable;
     properties::FloatProperty _guiOrderingNumber;
 
     // Transformation defined by translation, rotation and scale
--- a/include/openspace/scene/scenegraphnode.h
+++ b/include/openspace/scene/scenegraphnode.h
@@ -181,7 +181,7 @@ private:
     properties::StringProperty _guiDisplayName;
     properties::StringProperty _guiDescription;
     properties::BoolProperty _useGuiOrdering;
-    properties::BoolProperty _guiFocussable;
+    properties::BoolProperty _guiFocusable;
     properties::FloatProperty _guiOrderingNumber;
 
     // Transformation defined by translation, rotation and scale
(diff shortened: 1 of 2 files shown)


include/openspace/scene/scenegraphnode.h
```diff
@@ -181,6 +181,7 @@ class SceneGraphNode : public properties::PropertyOwner {
     properties::StringProperty _guiDisplayName;
     properties::StringProperty _guiDescription;
     properties::BoolProperty _useGuiOrdering;
+    properties::BoolProperty _guiFocussable;
```
WeirdRubberDuck, line 184: sanity check - but surely, it should be "focusable" with one s, right?  That's how English works? 🤔 :)

src/scene/scenegraphnode.cpp
```diff
@@ -193,6 +193,15 @@ namespace {
         openspace::properties::Property::Visibility::Hidden
     };
 
+    constexpr openspace::properties::Property::PropertyInfo GuiFocussableInfo = {
+        "IsFocussable",
+        "Focussable Hint",
+        "This value serves as a hint to determine if it makes sense to focus the camera "
```
WeirdRubberDuck, line 199: This description should mention that it affects how the object appears in the GUI, if that's what we want to use it for

It would be an easy addition if we want to include it in the new UI

include/openspace/scene/scenegraphnode.h
alexanderbock: Yea.. after being awake for about 30h, my English proficiency goes down the drain ;)

alexanderbock:
Second thing to get feedback on:

I moved all of the GUI related properties into a separate owner to mimick the structure in the asset.  So it is now:
`Earth.Gui.Path` rather than `Earth.GuiPath`   ,  `Earth.Gui.Name` instead of `Earth.GuiName`

Thoughts?

src/scene/scenegraphnode.cpp
```diff
@@ -328,6 +337,9 @@ namespace {
             // not display, for example, barycenters
             std::optional<bool> hidden;
 
+            // [[codegen::verbatim(GuiFocusableInfo.description)]]
+            std::optional<bool> focussable;
```
WeirdRubberDuck, line 341:
```suggestion
            std::optional<bool> focusable;
```

alexanderbock:
Moved the property owner restructuring to a new branch /feature/gui-subowner

Issue #3518, opened 2025-02-12
elapsed time dashboard: `LowestTimeUnit` property can't be set as a string via the scripting API


dawagner:
Using version 0.20.1:

`openspace.setPropertyValue("Dashboard.ElapsedTime.LowestTimeUnit", "hour")` (or `"Hour"`) fails with: `Property 'Dashboard.ElapsedTime.LowestTimeUnit' does not accept input of type 'String'. Requested type: Number` so one has to set it as a number, (e.g. `5` for hours), which is not very convenient and brittle. I'm aware that I can programatically retrieve the mapping from the property's definition but that's not very convenient either.

I suppose other similar properties have the same issue.

PR #3523, opened 2025-02-14
Launcher cleanup

alexanderbock:
This PR adds two new features to the launcher, but then also contains a cleaning up of the code as a whole.

The two new features:
 - Support folders for profiles and window configurations both in files provided by us as well as in user folders
 - Add more finegrained control over whether a window should render 2d or 3d components in favor over the previous checkbox+combobox option

PR #3524, opened 2025-02-14
Use composable Lua types in Property definitions.  Allow OptionProperty to receive either indices or descriptions

alexanderbock:
This change enables OptionProperties to be set using the description of a value, in addition to than just via indices.

For example:
```lua
openspace.setPropertyValueSingle("Dashboard.ElapsedTime.LowestTimeUnit", "Hour")
```
instead of (this is still valid)
```lua
openspace.setPropertyValueSingle("Dashboard.ElapsedTime.LowestTimeUnit", 4.000000)
```

Closes issue #3518

### DIFF
--- a/include/openspace/properties/list/doublelistproperty.h
+++ b/include/openspace/properties/list/doublelistproperty.h
@@ -35,7 +35,7 @@ public:
         std::vector<double> values = std::vector<double>());
 
     std::string_view className() const override;
-    int typeLua() const override;
+    ghoul::lua::LuaTypes typeLua() const override;
 
     using TemplateProperty<std::vector<double>>::operator std::vector<double>;
     using TemplateProperty<std::vector<double>>::operator=;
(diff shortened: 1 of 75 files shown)


WeirdRubberDuck:
> Looks good, but I can't run it for some reason.
> 
> Seems to be due to the change to the `walkDirectory` function in ghoul, but I don't get why it doesn't find the correct constructor of the `BooleanType´s. Any ideas? Have tried to do a clean build and still get the same problem

Ah, looks like it might need the same changes as in this PR... https://github.com/OpenSpace/OpenSpace/pull/3523

alexanderbock:
Yea, I turned them from `bool` parameters into `BooleanType` and `BooleanType` only has an explicit constructor on purpose to make this type of code illegal

WeirdRubberDuck (approved):
It's great!  We should consider having the UI use this version of setting the value instead, to make the scripts in the log more understandable :)

PR #3525, opened 2025-02-16
General code spring cleaning

alexanderbock:
- Introducing `[[likely]]` and `[[unlikely]]` for cache variables
  - Removing explicitly duplicating uniform names where it was not necessary
  - Various smaller code fixes
  - Adds the base class documentation for DashboardItems


WeirdRubberDuck:
Add to the description that this PR adds the base class documentation for DashboardItems?  (did we have an issue for this?  I can't find it now though)

alexanderbock:
I couldn't find that issue either

modules/base/lightsource/scenegraphlightsource.cpp
```diff
@@ -64,23 +64,21 @@ documentation::Documentation SceneGraphLightSource::Documentation() {
...
+    const Parameters p = codegen::bake<Parameters>(dictionary);
+
+    _intensity = p.intensity.value_or(_intensity);
     addProperty(_intensity);
+
     _sceneGraphNodeReference.onChange([this]() {
         _sceneGraphNode =
             global::renderEngine->scene()->sceneGraphNode(_sceneGraphNodeReference);
-    });
+        });
```
WeirdRubberDuck, line 80:
```suggestion
    });
```

modules/base/rendering/grids/renderablegrid.cpp
```diff
@@ -188,7 +187,7 @@ RenderableGrid::RenderableGrid(const ghoul::Dictionary& dictionary)
 }
 
 bool RenderableGrid::isReady() const {
-    return _hasLabels ? _gridProgram && _labels->isReady() : _gridProgram != nullptr;
+    return _gridProgram && _hasLabels ? _labels->isReady() : true;
```
WeirdRubberDuck, line 190: confused about the order here. May need a parenthesis?

```suggestion
    return _gridProgram && (_hasLabels ? _labels->isReady() : true);
```

modules/base/rendering/grids/renderablesphericalgrid.cpp
```diff
@@ -165,7 +165,7 @@ RenderableSphericalGrid::RenderableSphericalGrid(const ghoul::Dictionary& dictio
 }
 
 bool RenderableSphericalGrid::isReady() const {
-    return _hasLabels ? _gridProgram && _labels->isReady() : _gridProgram != nullptr;
+    return _gridProgram && _hasLabels ? _labels->isReady() : true;
```
WeirdRubberDuck, line 168: Same here 
```suggestion
    return _gridProgram && (_hasLabels ? _labels->isReady() : true);
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -1511,9 +1511,7 @@ void RenderablePointCloud::updateBufferData() {
 }
 
 void RenderablePointCloud::updateSpriteTexture() {
-    bool shouldUpdate = _hasSpriteTexture && _spriteTextureIsDirty;
-
-    if (!shouldUpdate) {
+    if (!(_hasSpriteTexture && _spriteTextureIsDirty)) [[likely]] {
```
WeirdRubberDuck, line 1514: I personally find this a lot harder to read. Maybe get rid of the parenthesis if you don't want the variable?

```suggestion
    if (!_hasSpriteTexture || !_spriteTextureIsDirty) [[likely]] {
```
but all the negating makes it harder to read than with the variable IMO

modules/base/rendering/renderablemodel.cpp
```diff
@@ -297,17 +297,14 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
-        }
+    _forceRenderInvisible = p.forceRenderInvisible.value_or(_forceRenderInvisible);
+    if (p.forceRenderInvisible.has_value() && !_forceRenderInvisible) {
+        // Asset file have specifically said to not render invisible parts, do not notify
+        // in the log if invisible parts are detected and dropped
+        _notifyInvisibleDropped = false;
     }
 
-    _file = absPath(p.geometryFile);
+    _file = p.geometryFile;
```
WeirdRubberDuck, line 307: I guess this could be considered a breaking change.  Now the user would have to do the `openspace.absPath` call. Doesn't look like it's an issue with any of our models though

Just want to ask - is this how we want all our file inputs in assets to work?  Thinking that there may be a mix as of now

modules/base/rendering/renderablemodel.cpp
```diff
@@ -366,14 +363,14 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
             }
             else {
-                std::string message = std::format("The given unit name '{}' does not "
-                    "match any currently supported unit names", stringUnit);
-                LERROR(message);
+                LERROR(std::format(
+                    "The given unit name '{}' does not match any currently supported "
+                    "unit names",
+                    stringUnit
+                ));
```
WeirdRubberDuck, line 373:
```suggestion
                LERROR(std::format(
                    "The given unit name '{}' does not match any currently supported "
                    "unit names", stringUnit
                ));
```
more relaxed rules for logs

modules/base/rendering/screenspaceimagelocal.cpp
```diff
@@ -92,7 +92,7 @@ ScreenSpaceImageLocal::ScreenSpaceImageLocal(const ghoul::Dictionary& dictionary
 
     if (p.texturePath.has_value()) {
         if (std::filesystem::is_regular_file(absPath(*p.texturePath))) {
-            _texturePath = absPath(*p.texturePath).string();
```
WeirdRubberDuck, line 95: Same here. Breaking change since now the user would have to do the `openspace.absPath`call

modules/base/rendering/screenspaceimageonline.cpp
```diff
@@ -90,59 +90,62 @@ bool ScreenSpaceImageOnline::deinitializeGL() {
...
                 _textureIsDirty = false;
-                LERRORC(e.component, e.message);
             }
         }
+        catch (const ghoul::io::TextureReader::InvalidLoadException& e) {
+            _textureIsDirty = false;
+            LERRORC(e.component, e.message);
+        }
     }
+
```
WeirdRubberDuck, line 148:
```suggestion
```

modules/base/rotation/timelinerotation.cpp
```diff
@@ -68,9 +68,7 @@ TimelineRotation::TimelineRotation(const ghoul::Dictionary& dictionary)
         ghoul::mm_unique_ptr<Rotation> rotation = Rotation::createFromDictionary(
             kf.second
         );
-        if (rotation) {
-            _timeline.addKeyframe(t, std::move(rotation));
-        }
+        _timeline.addKeyframe(t, std::move(rotation));
```
WeirdRubberDuck, line 71: Do we want to add it even if it couldn't be created?  Maybe this is not an issue and is caught somewhere higher up

modules/base/translation/timelinetranslation.cpp
```diff
@@ -67,9 +67,7 @@ TimelineTranslation::TimelineTranslation(const ghoul::Dictionary& dictionary)
 
         ghoul::mm_unique_ptr<Translation> translation =
             Translation::createFromDictionary(kf.second);
-        if (translation) {
-            _timeline.addKeyframe(t, std::move(translation));
-        }
+        _timeline.addKeyframe(t, std::move(translation));
```
WeirdRubberDuck, line 70: Same here - are we sure we're not moving a nullptr?

modules/globebrowsing/src/globelabelscomponent.cpp
```diff
@@ -322,7 +322,7 @@ void GlobeLabelsComponent::initialize(const ghoul::Dictionary& dictionary,
         return;
     }
 
-    const bool loadSuccess = loadLabelsData(absPath(*p.fileName));
+    const bool loadSuccess = loadLabelsData(*p.fileName);
```
WeirdRubberDuck, line 325: Same here, breaking since now the user has to specify the `openspace.absPath`?

modules/globebrowsing/src/layergroupid.cpp
```diff
@@ -31,66 +31,66 @@ namespace openspace::globebrowsing::layers {
 
 static_assert(
     std::is_sorted(
-        Groups.begin(),
-        Groups.end(),
```
WeirdRubberDuck, line 35: Just curious - is there some type of optimization/performance gain from using cbegin over begin? Due to the guaranteed constness?

modules/globebrowsing/src/rawtiledatareader.cpp
```diff
@@ -235,14 +234,14 @@ RawTile::ReadError postProcessErrorCheck(const RawTile& rawTile,
     if (hasMissingData && onHighLevel) {
         return RawTile::ReadError::Fatal;
     }
-    return RawTile::ReadError::None;
+    else {
```
WeirdRubberDuck, line 237: Why add the else here?

modules/space/rendering/renderableconstellationbounds.cpp
```diff
@@ -113,7 +113,7 @@ RenderableConstellationBounds::RenderableConstellationBounds(
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
     // Avoid reading files here, instead do it in multithreaded initialize()
-    _vertexFilename = absPath(p.file).string();
+    _vertexFilename = p.file.string();
```
WeirdRubberDuck, line 116: And here. abspath

modules/space/rendering/renderableconstellationsbase.cpp
```diff
@@ -105,7 +105,7 @@ RenderableConstellationsBase::RenderableConstellationsBase(
 
     // Avoid reading files here, instead do it in multithreaded initialize()
     if (p.namesFile.has_value()) {
-        _namesFilename = absPath(*p.namesFile).string();
+        _namesFilename = p.namesFile->string();
```
WeirdRubberDuck, line 108: :) abspath. breaking

modules/space/rendering/renderableeclipsecone.cpp
```diff
@@ -127,6 +123,41 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+
```
WeirdRubberDuck, line 126: Nitpicky, but do we want the extra newlines around this function?

modules/space/rendering/renderablerings.cpp
```diff
@@ -234,29 +229,25 @@ void RenderableRings::update(const UpdateData& data) {
...
+    using namespace ghoul::opengl;
+    std::unique_ptr<Texture> texture = TextureReader::ref().loadTexture(
+        _texturePath.value(),
+        1
+    );
+
+    if (texture) {
+        LDEBUGC(
+            "RenderableRings",
+            std::format("Loaded texture from '{}'", absPath(_texturePath))
```
WeirdRubberDuck, line 242: Why keep the abspath here but not above? I.e. in `loadTexture`

modules/space/translation/gptranslation.cpp
```diff
@@ -55,7 +55,7 @@ namespace {
 namespace openspace {
 
 documentation::Documentation GPTranslation::Documentation() {
-    return codegen::doc<Parameters>("space_transform_gp");
+    return codegen::doc<Parameters>("space_translation_gp");
```
WeirdRubberDuck, line 58: Note: changed key => should always be checked in documentation and codegen links. Doubt that we've referrred to this anywhere though

src/rendering/dashboarditem.cpp
```diff
@@ -42,12 +42,17 @@ namespace {
...
 
     struct [[codegen::Dictionary(DashboardItem)]] Parameters {
+        // Determines the type of the DashbordItem that should be created
         std::string type;
 
+        // The unique identifier for this DashboardItem
         std::string identifier [[codegen::identifier()]];
 
+        // The name for the DashboardItem that is used when showing it in the user
+        // interface
```
WeirdRubberDuck, line 52:
```suggestion
        // Determines the type of the DashbordItem that should be created.
        std::string type;

        // The unique identifier for this DashboardItem.
        std::string identifier [[codegen::identifier()]];

        // The name for the DashboardItem that is used when showing it in the user
        // interface.
```

WeirdRubberDuck (commented):
Getting a couple of warnings from this PR ,which should be fixed:
image

But otherwise it looks good to me. Had some questions about potentially breaking changes related to absPaths

modules/base/rendering/renderablemodel.cpp
alexanderbock: It's not a breaking change since the parameters struct already accepted a `std::filesystem::path` before;  the FileVerifier (verifier.cpp:256) will reject an input that is not a valid, existing file. So a path like `${DATA}/test.jpg` would have been rejected even before we got to this line, already making the `openspace.absPath` call necessary in the asset file.   So the `absPath` call in here will always return the same functional path (it will remove some indirections, like    `C:/abc/def/../ghi` ->`C:/abc/ghi`  but I wouldn't consider this a breaking change if we nolonger remove those

And yes, we want our asset files to provide a valid path to the dictionary.  Almost all of the time we are not using path tokens in the file path we are paying the cost of an abspath every time

modules/base/rendering/screenspaceimagelocal.cpp
alexanderbock: Made it even shorter since I didn't look at the `is_regular_file` check, which is also superfluous as the `codegen::bake` will already throw an exception.  so the `else` case here will never fire anyway

modules/base/rotation/timelinerotation.cpp
alexanderbock: This was a remnant from how the `createFromDictionary` function worked before.  In the past, it returned a `nullptr` if the creation of the object failed.  By now all of those functions are instead throwing an exception if the creation failed, meaning that if we get to the if-statement we didn't get an exception, meaning that the pointer is going to be `!= nullptr`

modules/globebrowsing/src/layergroupid.cpp
alexanderbock: I did it for "its the right thing to do" reasons, exactly as you are saying. I was now curious as well and did some microbenchmarking, and here and there it does make a bit of a difference (depending on the run in either direction), the differences are purely academical.

Moving forward we should just use `.begin()` and slowly migrate over to this.  I'll change these examples back to how they were before

modules/globebrowsing/src/rawtiledatareader.cpp
alexanderbock: Good question.  This function looked more like a "Choose this or that" if statement.   The way it was written before it read more of a `return RawTile::ReadError::Fatal` is the early out of the function and the `RawTile::ReadError::None` is the "normal control flow" of it.

modules/space/rendering/renderableeclipsecone.cpp
alexanderbock: Nope, we don't :D

modules/space/rendering/renderablerings.cpp
alexanderbock: Oversight.  `_texturePath` is a StringProperty and we don't expect people to enter path tokens into the UI, so I removed the bottom `absPath` as well.

Once the UI rewrite is done, I'd like to start working on a `FilePath` view for the StringProperty to properly represent that this property expects a file and not just any string

modules/space/translation/gptranslation.cpp
alexanderbock: Agreed and I did check

modules/base/rendering/renderablemodel.cpp
WeirdRubberDuck: Gotcha

modules/globebrowsing/src/layergroupid.cpp
WeirdRubberDuck: Confused. Do you mean that we should use `.begin()` or `.cbegin()`?

alexanderbock: Using `.begin()` is easier to parse and read then `.cbegin()` and doesn't seem to make any difference in speed.  All microbenchmarks I did were coming up as them being the same

PR #3527, opened 2025-02-17
Add the ability to limit-rate the dashboard updating

### LINKED ISSUE #3517: dashboard items whose values change too often are hard to make sense of
**Is your feature request related to a problem? Please describe.**
I'm currently playing around with the dashboard items (e.g. elapsed time, distance) but when the value displayed by these items vary rapidly, they may be hard to make sense of. The `LowestTimeUnit` property of the "elapsed time" dashboard item helps but the `RequestedUnit` property of the "distance" item, not much so: my use-case is to show the distance between the Apollo 8 craft and the Earth, and I'm requesting the distance to be displayed in km; the next biggest unit is AU, which is not a very good order of magnitude in this case.

**Describe the solution you'd like**
An independent and configurable refresh rate for the dashboard (a single one for the whole dashboard seems fine to me) might allow for a better experience.

**Describe alternatives you've considered**
- a way to periodically call a lua function could be used to manually construct and update dashboard text items but that moves some complexity to the scripting side and since I can't think of any other use-case for that, it feels like over-engineering.
- a property to round the distance (e.g. to a hundred kms) and/or to display the thousands separator might help as well.

alexanderbock:
Adds the ability to limit the update rate of the entire dashboard. Two of the dashboard items are not affected by it as they construct their content too dynamically to cache it between updates, but those (DashboardItemInstruments, DashboardItemMission) are also used seldomly enough and also in need of a bigger rewrite

Closes #3517

### DIFF
--- a/src/rendering/dashboard.cpp
+++ b/src/rendering/dashboard.cpp
@@ -49,6 +49,15 @@ namespace {
         "in x and y-direction on screen.",
         openspace::properties::Property::Visibility::User
     };
+
+    constexpr openspace::properties::Property::PropertyInfo RefreshRateInfo = {
+        "RefreshRate",
+        "Refresh Rate (in ms)",
+        "If this value is bigger than 0, the value represents the number of milliseconds "
+        "between refreshs of the dashboard items. If the value is 0 the dashbaord is "
+        "refreshed at the same rate as the main rendering.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
 } // namespace
 
 namespace openspace {
@@ -57,9 +66,12 @@ Dashboard::Dashboard()
     : properties::PropertyOwner({ "Dashboard" })
     , _isEnabled(EnabledInfo, true)
     , _startPositionOffset(StartPositionOffsetInfo, glm::ivec2(10, -10))
+    , _refreshRate(RefreshRateInfo, 0, 0, 1000)
+    , _lastRefresh(std::chrono::high_resolution_clock::now())
 {
     addProperty(_isEnabled);
     addProperty(_startPositionOffset);
+    addProperty(_refreshRate);
 }
 
 void Dashboard::addDashboardItem(std::unique_ptr<DashboardItem> item) {
@@ -137,8 +149,18 @@ void Dashboard::render(glm::vec2& penPosition) {
         return;
     }
 
+    auto now = std::chrono::high_resolution_clock::now();
+    // count returns values in nanoseconds, _refreshRate is in milliseconds
+    const bool needsUpdate = (now - _lastRefresh).count() > _refreshRate * 1e6;
+    if (needsUpdate) {
+        _lastRefresh = now;
+    }
+
     for (const std::unique_ptr<DashboardItem>& item : _items) {
         if (item->isEnabled()) {
+            if (needsUpdate) {
+                item->update();
+            }
             item->render(penPosition);
         }
     }
--- a/src/rendering/dashboard.cpp
+++ b/src/rendering/dashboard.cpp
@@ -53,8 +53,8 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo RefreshRateInfo = {
         "RefreshRate",
         "Refresh Rate (in ms)",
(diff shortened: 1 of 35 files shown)


src/rendering/dashboard.cpp
```diff
@@ -49,6 +49,15 @@ namespace {
         "in x and y-direction on screen.",
         openspace::properties::Property::Visibility::User
     };
+
+    constexpr openspace::properties::Property::PropertyInfo RefreshRateInfo = {
+        "RefreshRate",
+        "Refresh Rate (in ms)",
+        "If this value is bigger than 0, the value represents the number of milliseconds "
+        "between refreshs of the dashboard items. If the value is 0 the dashbaord is "
+        "refreshed at the same rate as the main rendering.",
```
WeirdRubberDuck, line 58:
```suggestion
        "The number of milliseconds between refreshes of the dashboard items. If the value is 0 
        "the dashboard is refreshed at the same rate as the main rendering.",
```
?

WeirdRubberDuck: Note that if you want to keep the old phrasing, there are also some typos
```suggestion
        "If this value is larger than 0, the value represents the number of milliseconds "
        "between refreshes of the dashboard items. If the value is 0 the dashboard is "
        "refreshed at the same rate as the main rendering.",
```

src/rendering/dashboardtextitem.cpp
```diff
@@ -85,4 +87,10 @@ DashboardTextItem::DashboardTextItem(const ghoul::Dictionary& dictionary, float
     _font = global::fontManager->font(_fontName, _fontSize);
 }
 
+void DashboardTextItem::render(glm::vec2& penPosition) {
```
WeirdRubberDuck, line 90: Curious, since I don't really understand it.  What is the "penPosition" ? While we're refactoring, could we maybe give this variable a more understandable name or add a Doxygen comment over the base render function to explain it?

WeirdRubberDuck (approved):
👍 Do we also want to provide a non-zero value for the setting?

src/rendering/dashboard.cpp
```diff
@@ -137,8 +149,18 @@ void Dashboard::render(glm::vec2& penPosition) {
         return;
     }
 
+    auto now = std::chrono::high_resolution_clock::now();
```
dawagner, line 152: IMHO, it should be a `steady_clock` to ensure it is not going backwards in time.

src/rendering/dashboard.cpp
```diff
@@ -137,8 +149,18 @@ void Dashboard::render(glm::vec2& penPosition) {
...
+    // count returns values in nanoseconds, _refreshRate is in milliseconds
+    const bool needsUpdate = (now - _lastRefresh).count() > _refreshRate * 1e6;
```
dawagner, line 154: I don't believe there's any guarantee about the period unit is for `high_resolution_clock` (or `steady_clock`). If you want to be sure, you'd rather use `duration_cast` or cast the `_refreshRate` variable:

```suggestion
    const bool needsUpdate = (now - _lastRefresh) > std::chrono::milliseconds(_refreshRate);
```

(untested)

dawagner (commented):
Awesome, thanks!

src/rendering/dashboardtextitem.cpp
alexanderbock: I kept any refactoring out of this PR on purpose or it would have grown into something much bigger.  Added a documentation entry to the root DashboardItem class

```
    /**
     * Renders this DashboardItem at the provided \p penPosition. The position indicates
     * where this DashboardItem should render itself and is provided in pixel-coordinates
     * with the top-left corner of the screen being at (0,0). Each derived subclass shall
     * update the \p penPosition according to the items that it should render. If, for
     * example, a single line of text is rendered, the \p penPosition shall be updated by
     * one line's worth of vertical separation.
     *
     * \p penPosition The position at which this DashboardItem should be rendered
     */
```

dawagner:
Thanks, I'll give it a try very soon!

PR #3534, opened 2025-02-21
Add Lua function `openspace.navigation.setFocus(identifier, shouldRetarget, shouldResetVelocities)`

WeirdRubberDuck:
Add a Lua API function for setting the focus node, which includes options to control whether to retarget and/or reset camera velocities

Will be used by the new UI to greatly simplify the code in the navigation panel, and can hopefully be useful in general for our users as well. Maybe in the ShowBuilder, for example?

The code for dealing with the anchor/aim updates has always been confusing, so I also did some minor tidying among the function names to try to reduce that confusion.

### DIFF
--- a/src/navigation/navigationhandler_lua.inl
+++ b/src/navigation/navigationhandler_lua.inl
@@ -205,10 +205,22 @@ namespace {
 }
 
 /**
- * TODO
+ * Set the current focus node for the navigation, or re-focus on it if it was already the
+ * focus node.
+ *
+ * Per default, the camera will retarget to center the focus node in the view. The
+ * velocities will also be reset so that the camera stops moving after any retargetting
+ * is done. However, both of these behaviors may be skipped using the optional arguments.
+ *
+ * \param identifier The identifier of the scene graph node to focus
+ * \param shouldRetarget If true, retarget the camera to look at the focus node
+ * \param shouldResetVelocities If true, reset the camera velocities so that the camera
+ *                              stops after its done retagetting (or immediately if
+ *                              retargetting is not done)
+ *
  */
-[[codegen::luawrap]] void setFocusNode(std::string identifier, bool shouldRetarget = true,
-                                       bool shouldResetVelocities = true)
+[[codegen::luawrap]] void setFocus(std::string identifier, bool shouldRetarget = true,
+                                   bool shouldResetVelocities = true)
 {
     using namespace openspace;
     SceneGraphNode* node = sceneGraphNode(identifier);
--- a/src/navigation/navigationhandler_lua.inl
+++ b/src/navigation/navigationhandler_lua.inl
@@ -205,6 +205,28 @@ namespace {
 }
 
 /**
+ * TODO
+ */
+[[codegen::luawrap]] void setFocusNode(std::string identifier, bool shouldRetarget = true,
+                                       bool shouldResetVelocities = true)
+{
+    using namespace openspace;
+    SceneGraphNode* node = sceneGraphNode(identifier);
+    if (!node) {
+        throw ghoul::lua::LuaError("Unknown node: " + identifier);
+    }
+
+    global::navigationHandler->orbitalNavigator().setFocusNode(
+        node,
+        shouldResetVelocities
+    );
+
+    if (shouldRetarget) {
+        global::navigationHandler->orbitalNavigator().startRetargetAnchor();
+    }
+}
+
+/**
  * Bind an axis of a joystick to be used as a certain type, and optionally define
  * detailed settings for the axis.
  *
--- a/src/navigation/navigationhandler_lua.inl
+++ b/src/navigation/navigationhandler_lua.inl
@@ -209,14 +209,14 @@ namespace {
  * focus node.
  *
  * Per default, the camera will retarget to center the focus node in the view. The
- * velocities will also be reset so that the camera stops moving after any retargetting
+ * velocities will also be reset so that the camera stops moving after any retargeting
  * is done. However, both of these behaviors may be skipped using the optional arguments.
  *
  * \param identifier The identifier of the scene graph node to focus
  * \param shouldRetarget If true, retarget the camera to look at the focus node
  * \param shouldResetVelocities If true, reset the camera velocities so that the camera
- *                              stops after its done retargetting (or immediately if
- *                              retargetting is not done)
+ *                              stops after its done retargeting (or immediately if
+ *                              retargeting is not done)
  *
  */
 [[codegen::luawrap]] void setFocus(std::string identifier, bool shouldRetarget = true,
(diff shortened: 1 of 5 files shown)


src/navigation/navigationhandler_lua.inl
```diff
@@ -204,6 +204,40 @@ namespace {
...
+ * focus node.
+ *
+ * Per default, the camera will retarget to center the focus node in the view. The
+ * velocities will also be reset so that the camera stops moving after any retargetting
+ * is done. However, both of these behaviors may be skipped using the optional arguments.
+ *
+ * \param identifier The identifier of the scene graph node to focus
+ * \param shouldRetarget If true, retarget the camera to look at the focus node
+ * \param shouldResetVelocities If true, reset the camera velocities so that the camera
+ *                              stops after its done retagetting (or immediately if
```
alexanderbock, line 218:
```suggestion
 *                              stops after its done retargetting (or immediately if
```

src/navigation/orbitalnavigator.cpp
```diff
@@ -1083,45 +1083,39 @@ glm::dvec3 OrbitalNavigator::cameraToSurfaceVector(const glm::dvec3& cameraPos,
...
+void OrbitalNavigator::setFocusNode(const SceneGraphNode* node, bool resetVelocities) {
+    setFocusNode(node->identifier(), resetVelocities);
 }
 
-void OrbitalNavigator::setFocusNode(const std::string& focusNode, bool) {
-    _anchor = focusNode;
+void OrbitalNavigator::setFocusNode(const std::string& identifier, bool resetVelocities) {
+    _resetVelocitiesOnAnchorChange = resetVelocities;
+    _anchor = identifier;
     _aim = std::string("");
```
alexanderbock, line 1093:
```suggestion
    _aim = std::string();
```

While you are already in here?

src/navigation/orbitalnavigator.cpp
```diff
@@ -1083,45 +1083,39 @@ glm::dvec3 OrbitalNavigator::cameraToSurfaceVector(const glm::dvec3& cameraPos,
...
-void OrbitalNavigator::setAnchorNode(const SceneGraphNode* anchorNode,
-                                     bool resetVelocitiesOnChange)
+void OrbitalNavigator::updateAnchorNode(const SceneGraphNode* anchorNode)
 {
     if (!_anchorNode) {
         _directlySetStereoDistance = true;
     }
 
-    const bool changedAnchor = _anchorNode != anchorNode;
+    bool changedAnchor = _anchorNode != anchorNode;
```
alexanderbock, line 1102: Any reason to not make it `const` anymore?

alexanderbock (approved):

src/navigation/orbitalnavigator.cpp
WeirdRubberDuck, line 1102:
```suggestion
    const bool changedAnchor = _anchorNode != anchorNode;
```

src/navigation/navigationhandler_lua.inl
```diff
@@ -204,6 +204,40 @@ namespace {
     global::navigationHandler->orbitalNavigator().startRetargetAnchor();
 }
 
+/**
+ * Set the current focus node for the navigation, or re-focus on it if it was already the
+ * focus node.
+ *
+ * Per default, the camera will retarget to center the focus node in the view. The
+ * velocities will also be reset so that the camera stops moving after any retargetting
```
ylvaselling, line 212:
```suggestion
 * velocities will also be reset so that the camera stops moving after any retargeting
```

src/navigation/orbitalnavigator.cpp
```diff
@@ -1113,15 +1104,18 @@ void OrbitalNavigator::setAnchorNode(const SceneGraphNode* anchorNode,
     _syncedAnchorNode = anchorNode ? anchorNode->identifier() : "";
 
     // Need to reset velocities after the actual switch in anchor node,
-    // since the reset behavior depends on the anchor node.
-    if (changedAnchor && resetVelocitiesOnChange) {
+    // since the reset behavior depends on the anchor node
+    if (_resetVelocitiesOnAnchorChange) {
```
ylvaselling, line 1108: Since there is a bool here that is stored between when the lua function call is made and the update happens, is there a risk that if you call two lua calls after each other, with different settings for the `_resetVelocitiesOnAnchorChange`, the bool will only take the last value? Or will this not be a problem? I assume that all the calls happen somewhere in a batch in the render loop

WeirdRubberDuck: Good question. Will do some testing

src/navigation/navigationhandler_lua.inl
```diff
@@ -204,6 +204,40 @@ namespace {
...
+ * is done. However, both of these behaviors may be skipped using the optional arguments.
+ *
+ * \param identifier The identifier of the scene graph node to focus
+ * \param shouldRetarget If true, retarget the camera to look at the focus node
+ * \param shouldResetVelocities If true, reset the camera velocities so that the camera
+ *                              stops after its done retargetting (or immediately if
```
WeirdRubberDuck, line 218:
```suggestion
 *                              stops after its done retargeting (or immediately if
```

src/navigation/navigationhandler_lua.inl
```diff
@@ -204,6 +204,40 @@ namespace {
...
+ *                              retargetting is not done)
```
WeirdRubberDuck, line 219:
```suggestion
 *                              retargeting is not done)
```

src/navigation/orbitalnavigator.cpp
alexanderbock: I think it'll only make a difference if the Lua call is called in the same frame with different parameters and then the last one would could.

So:
_Frame i_
 1. Lua function is queued to reset velocties
 1. Lua function is queued to not reset velocities
 1. Navigation or rendering doesn't know about the Lua functions yet

_Frame i+1_
  1. Camera is updated
  1. Queued Lua functions from _i_ are applied and navigation is changed with not resetting velocities
  1. Rendering with probably(?) old state

_Frame i+2_
  1. ...
  1. Rendering with new state

WeirdRubberDuck: So the way it works is that once the retargeting and/or velocity reset has been triggered, it's irreversible. So when calling for example: 
```
openspace.navigation.setFocus("ISS", true, true);openspace.navigation.setFocus("Earth", false, false);
```
the "true, true" triggers the reset and retargeting, but the anchor node will be set to "Earth".  

I guess this COULD be considered unexpected behavior, but it's also a bit of a weird use case. We already before had some weird behaviors if setting the node while retargeting is being done, and setting the focus twice in the same frame is even more unexpected. 

So, maybe don't an issue, I think

For symmetry, not that

alexanderbock: Yea, I think even though it is unexpected behavior, I'd chuck it up to "people shouldn't do that" and it's fine to leave it

PR #3536, opened 2025-02-25
Add new class TimelineScale

alexanderbock:
There already was a TimelineTranslation and TimelineRotation, so adding this for symmetry

### DIFF
--- a/data/assets/examples/scale/timelinescale/no-interpolate.asset
+++ b/data/assets/examples/scale/timelinescale/no-interpolate.asset
@@ -23,7 +23,7 @@ local Node = {
         -- The second timeline entry
         ["2000 JAN 01 12:00:00"] = {
           Type = "StaticScale",
-          Scale = 0.0
+          Scale = 10.0
         },
         -- The third timeline entry
         ["2000 JAN 01 23:59:59"] = {
--- a/data/assets/examples/scale/timelinescale/no-interpolate.asset
+++ b/data/assets/examples/scale/timelinescale/no-interpolate.asset
@@ -1,0 +1,52 @@
+-- No Interpolation
+-- This asset creates a SceneGraphNode that only displays coordinate axes. The scale of
+-- the coordinate axes are determined by a timeline of individual scales that are used
+-- without interpolating between the timeline entries. These scales are keyframes that are
+-- used to change between different sizes. This example transitions between three sizes.
+-- In this example, the interpolation between entries is disabled, which will cause the
+-- coordinate axes to change their size abruptly when the scale changes. If the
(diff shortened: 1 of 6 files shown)


data/assets/examples/scale/timelinescale/no-interpolate.asset
```diff
@@ -0,0 +1,52 @@
...
+local Node = {
+  Identifier = "TimelineScale_Example_NoInterpolation",
+  Transform = {
+    Scale = {
+      Type = "TimelineScale",
+      Keyframes = {
+        -- The first timeline entry
+        ["2000 JAN 01 00:00:00"] = {
+          Type = "StaticScale",
+          Scale = 10.0
```
Roxeena, line 21:
```suggestion
          Scale = 20.0
```

data/assets/examples/scale/timelinescale/no-interpolate.asset
```diff
@@ -0,0 +1,52 @@
...
+        },
+        -- The second timeline entry
+        ["2000 JAN 01 12:00:00"] = {
+          Type = "StaticScale",
+          Scale = 0.0
```
Roxeena, line 26:
```suggestion
          Scale = 10.0
```
I think it will be very strange for the user if the example disappears due to the 0 scale, so I suggest an alternative timeline. Since this example does not interpolate, it is going to stay "disappeared" for quite a long time in contrast to the example that interpolates.

data/assets/examples/scale/timelinescale/no-interpolate.asset
```diff
@@ -0,0 +1,52 @@
...
+        },
+        -- The third timeline entry
+        ["2000 JAN 01 23:59:59"] = {
+          Type = "StaticScale",
+          Scale = -10.0
```
Roxeena, line 31:
```suggestion
          Scale = 20.0
```

modules/base/scale/timelinescale.cpp
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return glm::mix(prev->data->scaleValue(data), next->data->scaleValue(data), t);
+    }
+    else {
+        if (prevTime <= now && now < nextTime) {
+            return prev->data->scaleValue(data);
+        }
+        else if (nextTime <= now) {
+            return next->data->scaleValue(data);
+        }
```
Roxeena, line 115: I think we will need a case for if `now` is before `prevTime` somewhere here. When I try the non-interpolating example, it disappears when the current time is set to before the first keyframe. This is not the case when I use the interpolating example.

PR #3543, opened 2025-02-27
assetmanager: add a 'reload' function, available as a scripting API

dawagner:
Reloads an asset by removing it and adding its new definition back.

### DIFF
--- a/src/scene/assetmanager.cpp
+++ b/src/scene/assetmanager.cpp
@@ -144,6 +144,90 @@ void AssetManager::deinitialize() {
     _toBeDeleted.clear();
 }
 
+void AssetManager::runRemoveQueue() {
+    ZoneScoped;
+    for (const std::string& asset : _assetRemoveQueue) {
+        std::filesystem::path path = generateAssetPath(_assetRootDirectory, asset);
+
+        const auto it = std::find_if(
+            _assets.cbegin(),
+            _assets.cend(),
+            [&path](const std::unique_ptr<Asset>& a) { return a->path() == path; }
+        );
+        if (it == _assets.cend()) {
+            LWARNING(std::format("Tried to remove unknown asset '{}'. Skipping", asset));
+            continue;
+        }
+
+        Asset* a = it->get();
+        auto jt = std::find(_rootAssets.cbegin(), _rootAssets.cend(), a);
+        if (jt == _rootAssets.cend()) {
+            // Trying to remove an asset from the middle of the tree might have some
+            // unexpected behavior since if we were to remove an asset with children, we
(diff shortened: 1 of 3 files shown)


dawagner:
The diff is a bit of a pain to read but it is essentially:
- extracting the loop on the "add" and "remove" queues in new, private methods for readability
- calling them in the `update` method *but swapped* (i.e. "remove" first)
- adding the `reload` function.

dawagner:
@alexanderbock : what do you think?

src/scene/assetmanager.cpp
```diff
@@ -144,52 +144,55 @@ void AssetManager::deinitialize() {
...
-        _toBeDeleted.clear();
-    }
-
-    // Initialize all assets that have been loaded and synchronized but that not yet
-    // initialized
-    for (auto it = _toBeInitialized.cbegin(); it != _toBeInitialized.cend(); it++) {
-        ZoneScopedN("Initializing queued assets");
-        Asset* a = *it;
+void AssetManager::runRemoveQueue() {
+    ZoneScopedN("Removing queued assets");
```
alexanderbock, line 148:
```suggestion
    ZoneScoped;
```

The `N` version was previously needed since it was embedded in a different function.  Now that it is its own function we can use the `ZoneScoped` which takes the name from the function directly

src/scene/assetmanager.cpp
```diff
@@ -144,52 +144,55 @@ void AssetManager::deinitialize() {
...
+            a->unload();
+        }
+        global::profile->removeAsset(asset);
     }
+    _assetRemoveQueue.clear();
+}
 
-    // Add all assets that have been queued for loading since the last `update` call
+void AssetManager::runAddQueue() {
+    ZoneScopedN("Adding queued assets");
```
alexanderbock, line 194:
```suggestion
    ZoneScoped;
```

Same here

alexanderbock (commented):
Looks great and works great!

src/scene/assetmanager_lua.inl
```diff
@@ -41,6 +41,13 @@ namespace {
     openspace::global::openSpaceEngine->assetManager().remove(assetName);
 }
 
+/**
+ * Reloads (remove+add) the asset with the specfied name.
```
alexanderbock, line 45:
```suggestion
 * Reloads the asset with the specified name. If the asset was previously loaded explicity
 * it will be removed and then re-added. If the asset was not previously loaded, it will only
 * be loaded instead.
```

dawagner:
@alexanderbock : Thanks for the review - I addressed your comments in the new push.

PR #3544, opened 2025-02-27
Spice transform: add a "time offset" option

### LINKED ISSUE #3531: feature request: relative time translation/rotation transforms
**Is your feature request related to a problem? Please describe.**
I would like to show "shadow" bodies or trails that are ahead (or, less frequently, late) of time compared to the "normal" body or trail. SpiceTranslation/SpiceRotation only allow setting a fixed time, which is not enough for this need.

**Describe the solution you'd like**
SpiceTranslation/SpiceRotation could gain a new parameter to compute the rotation/translation with a time offset.

dawagner:
Lets SpiceTranslation and SpiceRotation compute the transformation with a given time offset compared to the simulation time (or fixed date if one is configured).

Closes #3531.

### DIFF
--- a/data/assets/examples/rotation/spicerotation/timeoffset.asset
+++ b/data/assets/examples/rotation/spicerotation/timeoffset.asset
@@ -1,0 +1,37 @@
+-- Time offset
+-- This asset creates a rotation provided by a SPICE kernel and applies it to a
+-- SceneGraphNode that only displays coordinate axes. The rotation of the coordinate axes
+-- are determined by SPICE, in this case pretending that the coordinate axes are rotating
+-- at the same rate as Earth. In this specific example, the orientation is offset 8h back
+-- compared to the actual in-game time in OpenSpace.
+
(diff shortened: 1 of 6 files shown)


dawagner:
@alexanderbock : I've been busy with the flu... but here it is, finally. If the principle is ok with you I'll add the same for SpiceRotation and provide an example asset.

alexanderbock:
Yea, that looks awesome!  🚀

dawagner:
Updated with the rotation and examples - do I need to do anything in order for the example to end up in the generated documentation?

alexanderbock:
Nope, the examples in the `data/examples` folder will automatically get picked up based on the next two folder names;  first the type "Translation" and then the actual subclass "SpaceTranslation".  All files in the resulting folder will be automatically added.

If you are curious, the code for it is here: https://github.com/OpenSpace/OpenSpace-Docs/blob/master/_ext/generate_docs.py

alexanderbock (approved):

dawagner:
Thanks :)

PR #3548, opened 2025-03-01
add a cast from json to std::string for anchor

zingale:
this fixes compilation with GCC 14.2.1


alexanderbock (approved):
Thanks for the fix!  Was this the only compilation issue with 14?  Our continuous integration machines are running on 13, so we missed this one

zingale:
yes, this was the only issue I encountered.

PR #3550, opened 2025-03-03
Window Configuration Generation

### LINKED ISSUE #3547: openspace-window-configuration.json is invalid JSON because of locale-dependent generation
**Describe the bug**

Starting OpenSpace (latest) on a system configured in French fails with:

```
(F) main                 [ReadConfig] (6082): [json.exception.parse_error.101] parse error at line 17, column 31: syntax error while parsing object key - unexpected number literal; expected string literal
```

unless the launcher is bypassed.

This is due to `openspace-window-configuration.json` containing this:

```
      "windows": [
        {
          "name": "OpenSpace",
          "pos": { "x": 50, "y": 50 },
          "size": { "x": 1280,0, "y": 720,0 },
                             ^--- this is wrong
```

and the root cause seems to be how this json file, and this line in particular, is generated:

`scripts/configuration_helper.lua` : 
```
  table.insert(result,   [[          "size": { "x": ]] .. sx .. [[, "y": ]] .. sy .. [[ },]])
```

where `sx` and `sy` are probably floats and Lua correctly formats them with a comma, since that's the decimal separator in fr_FR.

The fix that was implemented for https://github.com/OpenSpace/OpenSpace/issues/3474 probably only kicks in later, I guess.

Here, the proper fix, IMO, would be to use a json library rather than attempting to write json by hand - no locale should need to be overriden as any bug-free json lib should render numbers following the json spec.

alexanderbock:
Use built-in to JSON converter instead of manually creating the string as that had issues with locale-specific number formats (closes #3547). The configuration_helper now uses the nlohmann library exposed through a new Lua function.

While in there, I also removed the ability to create fisheye configuration files from the Lua script directly as we have moved past wanting users to edit the `openspace.cfg` file for these things.

Note:  For this change to work, a change in the serialization of `ghoul::Dictionary`s was necessary.  When serializing a Dictionary, we now check whether we have an array-style dictionary (= only numerical keys that are starting at 1 and are increasing in values without gaps) and if so, serialize it as an array instead.   Prior to this, a Dictionary like:
```cpp
ghoul::Dictionary d;
d.setValue("1", "abc");
d.setValue("2", "def");
d.setValue("3", "ghi");
```
would have been serialized to JSON as:  `{ "1": "abc", "2": "def", "3": "ghi" }`.  After this change it will be `[ "abc", "def", "ghi" ]`.

### DIFF
--- a/openspace.cfg
+++ b/openspace.cfg
@@ -8,7 +8,7 @@
 -- or a dome cluster system
 
 -- A regular 1280x720 window
-SGCTConfig = sgct.config.single{vsync=true, fov={left=12, right=43, up=22, down= 33}}
+SGCTConfig = sgct.config.single{vsync=false}
 -- A regular 1920x1080 window
 -- SGCTConfig = sgct.config.single{1920, 1080}
 -- A windowed 1920x1080 fullscreen
--- a/openspace.cfg
+++ b/openspace.cfg
@@ -8,7 +8,7 @@
 -- or a dome cluster system
 
 -- A regular 1280x720 window
-SGCTConfig = sgct.config.single{vsync=false}
+SGCTConfig = sgct.config.single{vsync=true, fov={left=12, right=43, up=22, down= 33}}
 -- A regular 1920x1080 window
 -- SGCTConfig = sgct.config.single{1920, 1080}
 -- A windowed 1920x1080 fullscreen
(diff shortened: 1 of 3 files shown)


dawagner:
Do you need help testing this MR? I could try to pull it tomorrow or next week.

alexanderbock:
If you'd want to try it out, that would be great!

dawagner:
I confirm it fixes the issue for me :) Thanks.

Issue #3538, opened 2025-02-25
Actions Panel - User defined colors for actions


Arohdin:
Make it possible to define custom colors (background and text) for actions.
Makes it easier to organize and visually distinguish types of actions when there are many.
Should probably only be defined in asset and not user changeable within OpenSpace.

However, maybe bad in terms of accessibility (?)

**Maybe not the most pretty examples, but an example nonetheless**

![Image](https://github.com/user-attachments/assets/712b45f6-1283-4220-9703-4987165107c1)

![Image](https://github.com/user-attachments/assets/92bebf2c-2f13-478e-90ac-22e7e592029b)

![Image](https://github.com/user-attachments/assets/3313eb22-2fe7-47f7-ad54-8c7a65c79eb6)

PR #3551, opened 2025-03-03
Specify background and text colors for actions

alexanderbock:
Add the ability to the backend to specify a background color and a text color for actions.  Addresses the backend work for #3538

### DIFF
--- a/src/interaction/actionmanager_lua.inl
+++ b/src/interaction/actionmanager_lua.inl
@@ -88,6 +88,16 @@ struct [[codegen::Dictionary(Action)]] Action {
     // not provided, the default value is /
     std::optional<std::string> guiPath;
 
+    /// This parameter, if specified, will be used as a hint to any potential user
+    /// interface as a desired background color. This can be used, for example, to
+    /// visually group similar Actions together
+    std::optional<glm::vec4> color [[codegen::color()]];
+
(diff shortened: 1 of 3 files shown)


src/interaction/actionmanager_lua.inl
```diff
@@ -88,6 +88,16 @@ struct [[codegen::Dictionary(Action)]] Action {
     // not provided, the default value is /
     std::optional<std::string> guiPath;
 
+    /// This parameter, if specified, will be used as a hint to any potential user
+    /// interface as a desired background color. This can be used, for example, to
+    /// visually group similar Actions together
+    std::optional<glm::vec4> color [[codegen::color()]];
```
ylvaselling, line 94: The documentation for actions are not on the docs page afaik, so maybe we need a docs page for the actions? I don't think these comments end up in the docs page?

Or, we could make the comments to the registerAction lua function more explicit. Either way it would be nice to have the struct for Action on the docs page somehow

ylvaselling (approved):

src/interaction/actionmanager_lua.inl
alexanderbock: Yea, it's a bigger redesign that is necessary that would go together with changing the `ghoul::Dictionary` in Lua functions to more explicit types as well.  I made an issue to track it #3552

PR #3553, opened 2025-03-03
TimeFrame specification and status

alexanderbock:
This PR adds two separate but related features that would have been more errorprone to have in two separate but dependent PRs.

  1. Makes Translation/Scale/Rotation all able to have a TimeFrame, not just the SpiceRotation. Also adds examples for these
  2. Add read-only BoolProperty to the TimeFrame that indicates whether the current time is a in the timeframe

### DIFF
--- a/data/assets/examples/scale/staticscale/static_timeframe.asset
+++ b/data/assets/examples/scale/staticscale/static_timeframe.asset
@@ -3,8 +3,7 @@
 -- 2002 JAN 01. The coordinate axes of this scene graph node normally have a length of 1
 -- meter and are scaled in this example by a factor of 149597870700, which means they
 -- will be 149597870700 m (1 AU) long, thus reaching the same distance as Earth's orbit
--- around the Sun. In this example, the scale is only applied between 2000 JAN 01 and just
--- prior to 2002 JAN 01.
+-- around the Sun.
 
 local Node = {
   Identifier = "StaticScale_TimeFrame",


data/assets/examples/rotation/staticrotation/timeframe.asset
```diff
@@ -0,0 +1,35 @@
...
+    }
+  },
+  Renderable = {
+    Type = "RenderableCartesianAxes"
+  },
+  GUI = {
+    Name = "StaticRotation - Euler Angles",
+    Path = "/Examples"
+  }
+}
```
WeirdRubberDuck, line 27: I think the making of this example happened a bit quickly...?   Se suggestion below.

Also, rename the file to:  `static_timeframe.asset`

```suggestion
-- Time Frame
-- This asset creates a rotation and applies a time frame to it, meaning that rotation will 
-- only be valid at specific times. In this example, the rotation is only applied in the time  
-- interval between 2000 JAN 01 and just prior to 2002 JAN 01.

local Node = {
  Identifier = "StaticRotation_Example_TimeFrame",
  Transform = {
    Rotation = {
      Type = "StaticRotation",
      Rotation = { math.pi / 2.0, 0.0, math.pi },
      TimeFrame = {
        Type = "TimeFrameInterval",
        Start = "2000 JAN 01",
        End = "2002 JAN 01"
      }
    }
  },
  Renderable = {
    Type = "RenderableCartesianAxes"
  },
  GUI = {
    Name = "StaticRotation - Time Frame",
    Path = "/Examples"
  }
}
```

WeirdRubberDuck: Please do this for all the other examples as well. I'll stop reading here and review after that has been changed :)

alexanderbock: Updated the text in the file.

data/assets/examples/translation/statictranslation/static_timeframe.asset
```diff
@@ -0,0 +1,35 @@
+-- Time Frame
+-- This asset creates a set of coordinate axes that are offset from their original
```
WeirdRubberDuck, line 2: This example is a good example of why I think the first sentence should explain the entire example, briefly :) 

All sentences start with either "this asset" or "this example", which to me suggested that the flow of the explanation is wrong

WeirdRubberDuck (approved):
Nice :)  Looks good. Please just update the example descriptions

src/scene/translation.cpp
```diff
@@ -77,10 +91,20 @@ void Translation::update(const UpdateData& data) {
...
+
+    if (_timeFrame) {
+        _timeFrame->update(data.time);
+    }
+
     const glm::dvec3 oldPosition = _cachedPosition;
-    _cachedPosition = position(data);
-    _cachedTime = data.time.j2000Seconds();
-    _needsUpdate = false;
+    if (_timeFrame && !_timeFrame->isActive()) {
```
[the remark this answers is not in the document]
alexanderbock: Which previous comment?  There is no for loop here; it's just an "update then check the results" paradigm

data/assets/examples/scale/staticscale/static_timeframe.asset
```diff
@@ -0,0 +1,37 @@
+-- Time Frame
+-- This asset creates a scale that is only applied between 2000 JAN 01 and just prior to
+-- 2002 JAN 01. The coordinate axes of this scene graph node normally have a length of 1
+-- meter and are scaled in this example by a factor of 149597870700, which means they
+-- will be 149597870700 m (1 AU) long, thus reaching the same distance as Earth's orbit
+-- around the Sun. In this example, the scale is only applied between 2000 JAN 01 and just
+-- prior to 2002 JAN 01.
```
WeirdRubberDuck, line 7:
```suggestion
-- around the Sun.
```

WeirdRubberDuck (approved):

PR #3562, opened 2025-03-10
Turn the time selector in the Profile Editor into a tabbed window

### LINKED ISSUE #3444: Profile editor time menu should only show the input type currently selected
Now it always shows both input types, but the user can only interact with the one that is currently selected.  Since both look visually enabled, this is confusing to the user. 

Instead, we should either add tabs similarly to how the camera input works. Or, just hide the input method that is not currently selected. 

image

alexanderbock:
image

image

Closes #3444


WeirdRubberDuck:
How about also adding some descriptions of valid values to the "Relative" tab? 😇

PR #3567, opened 2025-03-17
Move the OpenSpaceData folder by default into the user folder

### LINKED ISSUE #3512: Move OpenSpaceData folder into main folder or into user folder
Currently the OpenSpaceData folder gets created outside the main OpenSpace-*.*.*/ folder, and it feels a bit inconsistent with how our sync/user folders/variables work, and it also feels like we are violating the system a bit by creating this folder outside our folder, but not in the system user place. 

I think we should have OpenSpace-0.20.1/user/maps or OpenSpace-0.20.1/globebrowsing/ or something like that

(I am requesting this behavior on when OPENSPACE_GLOBEBROWSING is NOT set)

alexanderbock:
It is not by default placed in `user/globebrowsing`

Closes #3512

### DIFF
--- a/openspace.cfg
+++ b/openspace.cfg
@@ -120,7 +120,7 @@ Paths = {
     SCRIPTS = "${BASE}/scripts",
     SHADERS = "${BASE}/shaders",
     TEMPORARY = "${BASE}/temp",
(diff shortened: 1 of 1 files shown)


WeirdRubberDuck (approved):
👍 

We need to make to mark this as a __breaking change__ for the release, and update the  "Additional Datasets" part when publishing the next release to the website

image

PR #3569, opened 2025-03-18
Add a new panel to the Launcher to control which WebUI button appears on the bottom bar

### LINKED ISSUE #3566: Add a new panel to the Launcher to control which WebUI button appears on the bottom bar

alexanderbock:
Closes #3566


apps/OpenSpace/ext/launcher/src/profile/uipanelsdialog.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <ghoul/filesystem/filesystem.h>
+#include <QCheckBox>
+#include <QDialogButtonBox>
+#include <QLabel>
+#include <QVBoxLayout>
+#include <fstream>
+#include <string_view>
+
+namespace {
+    constexpr std::string_view DefaultPanelPath = "${DATA}/web/default_panels.json";
```
WeirdRubberDuck, line 38: Do we want to rename this file to 
```suggestion
    constexpr std::string_view DefaultPanelPath = "${DATA}/web/default_ui_panels.json";
```
?   Now there's not really anything that connects it to the UI

apps/OpenSpace/ext/launcher/src/profile/uipanelsdialog.cpp
```diff
@@ -0,0 +1,128 @@
...
+        }
+        else {
+            box->setChecked(panel.isVisible);
+        }
+
+        layout->addWidget(box);
+        _checkboxToId[box] = panel.id;
+    }
+    layout->addWidget(new Line);
+    {
```
WeirdRubberDuck, line 110:
```suggestion
    }
    
    layout->addWidget(new Line);
    
    {
```
?

data/web/default_panels.json
```diff
@@ -0,0 +1,17 @@
...
+    "6": { "id": "screenSpaceRenderables", "name": "ScreenSpace Renderables", "visible": false, "enabled": true },
+    "7": { "id": "exoplanets", "name": "Exoplanets", "visible": false, "enabled": true },
+    "8": { "id": "userPanels", "name": "User Panels", "visible": false, "enabled": true },
+    "9": { "id": "actions", "name": "Actions", "visible": false, "enabled": true },
+    "10": { "id": "skyBrowser", "name": "SkyBrowser", "visible": false, "enabled": true },
+    "11": { "id": "mission", "name": "Mission", "visible": false, "enabled": false },
+    "12": { "id": "flightControl", "name": "Flight Control", "visible": false, "enabled": true },
+    "13": { "id": "keybindingsLayout", "name": "Keybindings", "visible": false, "enabled": true },
+    "14": { "id": "gettingStartedTour", "name": "Getting Started Tour", "visible": false, "enabled": true }
+}
```
WeirdRubberDuck, line 17: Having to write indexes makes me a bit sad and makes the file look very busy... Seems hard to maintain as well

Did you discuss using a format that uses an array? Maybe it's annoying, code-wise or otherwise, to have that extra level to the data?
```
{
    "data": [
        { "id": "scene", "name": "Scene", "visible": true, "enabled": true },
        { "id": "settings", "name": "Settings", "visible": false, "enabled": true },
        { "id": "navigation", "name": "Navigation", "visible": true, "enabled": true },
        { "id": "timePanel", "name": "Date & Time", "visible": true, "enabled": true },
        { "id": "sessionRecording", "name": "Session Recording", "visible": false, "enabled": true },
        { "id": "geoLocation", "name": "Geo Location", "visible": false, "enabled": true },
        { "id": "screenSpaceRenderables", "name": "ScreenSpace Renderables", "visible": false, "enabled": true },
        { "id": "exoplanets", "name": "Exoplanets", "visible": false, "enabled": true },
        { "id": "userPanels", "name": "User Panels", "visible": false, "enabled": true },
        { "id": "actions", "name": "Actions", "visible": false, "enabled": true },
        { "id": "skyBrowser", "name": "SkyBrowser", "visible": false, "enabled": true },
        { "id": "mission", "name": "Mission", "visible": false, "enabled": false },
        { "id": "flightControl", "name": "Flight Control", "visible": false, "enabled": true },
        { "id": "keybindingsLayout", "name": "Keybindings", "visible": false, "enabled": true },
        { "id": "gettingStartedTour", "name": "Getting Started Tour", "visible": false, "enabled": true }
    ]
}
```

or using the id as the key?

WeirdRubberDuck: Oh, or maybe I take that back. The numbers are supposed to represent the order...?   Feel free to ignore this comment in that case

data/web/default_panels.json
```diff
@@ -0,0 +1,17 @@
...
+    "1": { "id": "settings", "name": "Settings", "visible": false, "enabled": true },
+    "2": { "id": "navigation", "name": "Navigation", "visible": true, "enabled": true },
+    "3": { "id": "timePanel", "name": "Date & Time", "visible": true, "enabled": true },
+    "4": { "id": "sessionRecording", "name": "Session Recording", "visible": false, "enabled": true },
+    "5": { "id": "geoLocation", "name": "Geo Location", "visible": false, "enabled": true },
+    "6": { "id": "screenSpaceRenderables", "name": "ScreenSpace Renderables", "visible": false, "enabled": true },
+    "7": { "id": "exoplanets", "name": "Exoplanets", "visible": false, "enabled": true },
+    "8": { "id": "userPanels", "name": "User Panels", "visible": false, "enabled": true },
+    "9": { "id": "actions", "name": "Actions", "visible": false, "enabled": true },
+    "10": { "id": "skyBrowser", "name": "SkyBrowser", "visible": false, "enabled": true },
```
WeirdRubberDuck, line 12: All of these were ones that we had visible per default before. Should they still be? 

At least actions, session recording, screen space renderables?

If not, should we update some of the profiles? E.g. to show the skybrowser and exoplanets in default?  Maybe we don't need it?

Having some conflicting thoughts when writing this, so sorry if it's hard to understand :)

data/web/default_panels.json
alexanderbock: @ylvaselling  Should answer this.  I got this file from her as a result of the "Save Layout..." button the new UI.

It was a conscious decision to make this file the same as the output of the layout for future easy of upgrading

data/web/default_panels.json
alexanderbock: I changed their values to mimic the behavior of the previous UI.   In case we have the time to make a full pass through the profiles to adjust, we can change it back again

data/web/default_panels.json
ylvaselling: The number is indeed representing the order, but it is probably possible to use the array too. It was originally written to this format as we loaded the function with our lua function, but since we don't use that anymore, we don't really need the indices like this.

data/web/default_panels.json
ylvaselling: Maybe there is some confusion regarding enabled and visible. Enabled doesn't really do much anymore. We've had some discussions about removing certain panels when we don't want to open them (exoplanets and skybrowser) but I think we've landed on that we want to enable all panels always, but that some are visible. 

Visible on the other hand, is if it is shown in the bottom bar.

alexanderbock: I updated the file between the comments.. A bunch of the things were accidentally actually disabled, but I updated the file since to represent what our default value are

data/web/default_panels.json
alexanderbock: Do you want to update it? Since the JSON is so fundamentally different that writing an updater later will not be much of a problem if we decide to change it after release

ylvaselling: 🤷 I don't really see the problem with the current format but I'm open to changing it. One thing to keep in mind is that our lua function read / write arrays like the one above I'm pretty sure

ylvaselling (approved):

data/web/default_panels.json
alexanderbock: Then let's leave the format as it is :+1:

PR #3579, opened 2025-03-26
Improve performance of `setPropertyValue` function

### LINKED ISSUE #2255: Calling setPropertyValue Causes a Noticeable Skip in Framerate
Setting a property using `setPropertyValue` causes a skip in rendering (a fraction of a second depending on the machine it's running on), which is probably caused by the function traversing the scenegraphnodes to find matches.

This purpose of this issue is to reduce the time that this takes. Possible solutions include preprocessing of the nodes at startup (and when nodes are added, etc.).

alexanderbock:
This PR improves the performance of the `setPropertyValue` by about 20x through three steps:
1. The `Property::uri` function no longer creates the full URI with every call but uses a cached version that gets invalidated when the parentage of the Property changes
2. The list of all properties is now cached and gets invalidated whenever a Property or PropertyOwner gets added or removed
3. The execution of the loop over all Properties in the `findMatchesInAllProperties` is multithreaded

Needs some proper double-checking to make sure that the removal of SceneGraphNodes, ScreenSpaceRenderable, and any other thing is handled correctly. The two potential error modes would be that a property set via `openspace.setPropertyValue` would return an error that a property does not exist (even though it was added) or that we get a different type of errors for a property that actually doesn't exist anymore but is still left in the cache.

Before:
image (2)

After:
image (1)

Closes #2255


src/scene/scene_lua.inl
```diff
@@ -140,63 +141,64 @@ std::vector<openspace::properties::Property*> findMatchesInAllProperties(
...
+        properties.cbegin(),
+        properties.cend(),
+        [&](properties::Property* prop) {
+            // Check the regular expression for all properties
+            const std::string_view id = prop->uri();
+
+            if (isLiteral && id != propertyName) {
+                return;
+            }
+            else if (!propertyName.empty()) {
```
WeirdRubberDuck, line 156: I can see that this hasn't changed from the previous implementation, but there is a lot of confusion in here to me.... Can we maybe rewrite it or rename some variables to make it clearer what is going on? E.g: 

1. `prop->uri()` is actually the full URI of the property, but here its' stored in a variable `id` which makes one think it's the identifier. How about renaming that variable to `uri`?
2. What is `propertyName`? It's not the name of the property, since it seems like here we compare it again `id`, which is really the URI.  Should `noedName` and `propertyName` above be renamed to something else, that more clearly describe what they are (e.g. name, identifier, or URI)?

src/scene/scene_lua.inl
```diff
@@ -140,63 +141,64 @@ std::vector<openspace::properties::Property*> findMatchesInAllProperties(
...
+                size_t propertyPos = id.find(propertyName);
+                if (
+                    (propertyPos == std::string::npos) ||
+                    // Check that the propertyName fully matches the property in id
+                    ((propertyPos + propertyName.length() + 1) < id.length()) ||
+                    // Match node name
+                    (!nodeName.empty() && id.find(nodeName) == std::string::npos))
```
WeirdRubberDuck, line 163: Do we want to consider splitting up this if statement using variables? It's a beast (but surprisingly easy to understand thanks to the comments so maybe not an issue. Maybe add a comment to the first line as well?  )

src/scene/scene_lua.inl
```diff
@@ -547,7 +551,7 @@ namespace {
     std::vector<std::string> res;
     for (properties::Property* prop : props) {
         // Check the regular expression for all properties
-        const std::string& id = prop->uri();
+        const std::string_view id = prop->uri();
```
WeirdRubberDuck, line 554: Same thing here, could we rename the variable to make it clearer that we're working with the full uri and not the identifier?
```suggestion
        const std::string_view uri = prop->uri();
```

src/scene/scene_lua.inl
alexanderbock: For 1:  Done
For 2:  I'd prefer to do that change in a second PR instead, including a bunch of new unit tests to validate that everything works.  There is a bunch of changes that doesn't make too much sense as is, but adding that in here would dilute this PR a lot

src/scene/scene_lua.inl
alexanderbock: The point of turning it into a single statement is to make use of shortcircuiting of the statements.  If the first check already hits, we don't need to do the remaining checks and can save some performance.

src/scene/scene_lua.inl
alexanderbock: https://github.com/OpenSpace/OpenSpace/issues/3588

src/scene/scene_lua.inl
WeirdRubberDuck: Yeah, good point 👍

WeirdRubberDuck:
Tried to do some more testing - the speedup is awesome! 

Trying to look at removing and adding different objects as well. What do we expect is correct handling when this happens? And is there something specific we want tested .

For example, if I remove the Earth SGN and then trigger the global illumination action, we get this error:
image
indicating that Earth is still among the list of matches for the `planet_solarsystem` tag, even though it no longer exists. 

Note, however, that this was an issue before this PR as well. Tested in master and got the same issue

WeirdRubberDuck (approved):
I've been trying to do some testing using the existing actions. The speedup is great!

Tried removing e.g. Earth and then running the "All globes global illumination" action, which calls `openspace.setPropertyValue("{planet_solarSystem}.Renderable.PerformShading", false)`. That seems to work just fine.

The exoplanet settings button in the old UI also work the way they should. Tried it also after removing an exoplanet system

PR #3581, opened 2025-03-27
Add documentation for `ColorMappingComponent`

WeirdRubberDuck:
Add documentation for the `ColorMappingComponent` type in "Other", and update some of the descriptions a bit to better match the format on the docs page. 

Note that specifying a file is now required. It didn't really make sense that it was optional, and this simplified the documentation writing.

### DIFF
--- a/src/rendering/colormappingcomponent.cpp
+++ b/src/rendering/colormappingcomponent.cpp
@@ -35,34 +35,30 @@ namespace {
     constexpr openspace::properties::Property::PropertyInfo EnabledInfo = {
         "Enabled",
         "Color Map Enabled",
-        "If this value is set to 'true', the provided color map is used (if one was "
-        "provided in the configuration). If no color map was provided, changing this "
-        "setting does not do anything.",
+        "Decides if the color mapping should be used or not.",
         openspace::properties::Property::Visibility::NoviceUser
     };
 
     constexpr openspace::properties::Property::PropertyInfo FileInfo = {
         "File",
         "Color Map File",
-        "The path to the color map file to use for coloring the points.",
+        "The path to the color map file to use. Should be a .cmap file",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
     constexpr openspace::properties::Property::PropertyInfo ParameterInfo = {
         "Parameter",
         "Parameter",
-        "This value determines which paramenter is used for coloring the points based "
-        "on the color map. The property is set based on predefined options specified in "
-        "the asset file. When changing the parameter, the value range to used for the"
-        "mapping will also be changed. Per default, it is set to the last parameter in "
-        "the list of options.",
+        "The paramenter in the dataset to use for the color mapping. On change, the "
+        "value range to used for the mapping will also be changed.",
         openspace::properties::Property::Visibility::User
     };
 
     constexpr openspace::properties::Property::PropertyInfo RangeInfo = {
         "ValueRange",
         "Value Range",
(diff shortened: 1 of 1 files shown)


src/rendering/colormappingcomponent.cpp
```diff
@@ -150,30 +146,55 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This component handles settings for dynamic color mapping of data points based on
+    // the columns in a dataset. It is for example utilized by the
+    // [RenderablePointCloud](#base_renderablepointcloud) renderable type.
+    //
+    // The option for the parameters will be initialized based on the dataset used for
+    // the specific renderable. The value range will be set based on the min and max
+    // value for the respective column in the dataset, but can be manually adjusted.
```
alexanderbock, line 155:
```suggestion
    // values for the respective column in the dataset, but can be manually adjusted.
```

src/rendering/colormappingcomponent.cpp
```diff
@@ -150,30 +146,55 @@ namespace {
...
+    // Per default, all columns will be loaded, but this can also be adjusted.
```
alexanderbock, line 156: How? I can only see the default column setting in the struct?

src/rendering/colormappingcomponent.cpp
```diff
@@ -150,30 +146,55 @@ namespace {
...
+    //
+    // In addition to specifying a color map (.cmap) file and changing the parameter, this
+    // component includes features such as:
```
alexanderbock, line 159:
```suggestion
    // In addition to specifying a color map (.cmap) file and changing the parameter used
    // for determining the color, this component includes features such as:
```

src/rendering/colormappingcomponent.cpp
```diff
@@ -150,30 +146,55 @@ namespace {
...
-            // dataset
+            // dataset.
             std::optional<glm::vec2> range;
         };
-        // A list of options for color parameters to use for color mapping, that will
-        // appear as options in the drop-down menu in the user interface. Per default,
-        // the first option in the list is selected. Each option is a table in the form
-        // { Key = \"theKey\", Range = {min, max} }, where the value range is optional.
-        // If added, this range will automatically be set when the option is selected
+        // A list of options for parameters to use for color mapping, that will appear
```
alexanderbock, line 185:
```suggestion
        // A list of options for parameters to use for color mapping that will appear
```

src/rendering/colormappingcomponent.cpp
```diff
@@ -318,9 +339,7 @@ ColorMappingComponent::ColorMappingComponent(const ghoul::Dictionary& dictionary
         _hasBelowRangeColorInAsset = true;
     }
 
-    if (p.file.has_value()) {
-        colorMapFile = p.file->string();
-    }
+    colorMapFile = p.file.string();
```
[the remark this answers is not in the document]
WeirdRubberDuck: Yep, but we would still need a string property that can be changed during runtime to update the file used. 

Worth it?

src/rendering/colormappingcomponent.cpp
alexanderbock: Yea :+1:

src/rendering/colormappingcomponent.cpp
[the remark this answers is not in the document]
alexanderbock: Nah.. Agree that it isn't worth it.  Worth considering adding a `PathProperty` or a different view.  But that's a story for a different PR

PR #3582, opened 2025-03-27
Add type documentation for the different light sources

WeirdRubberDuck:


modules/base/lightsource/cameralightsource.cpp
```diff
@@ -37,9 +37,12 @@ namespace {
         openspace::properties::Property::Visibility::NoviceUser
     };
 
+    // This `LightSource` type represents a light source placed at the position of the
+    // camera. An object with this light source will always be illuminated from the
+    // current view direction.
     struct [[codegen::Dictionary(CameraLightSource)]] Parameters {
         // [[codegen::verbatim(IntensityInfo.description)]]
-        std::optional<float> intensity;
+        std::optional<float> intensity [[codegen::inrange(0.0, 1.0)]];
```
WeirdRubberDuck, line 45: Is this correct?  Looks like we can theoretically specify an intensity larger than 1

alexanderbock: I don't think we need to limit the intensitiy to `[0,1]`.  In some examples I've used light intensities >1 to make models brighter.   And I could imagine values <0 for dramatic effect (to achieve an inverse diffuse lighting, for example)

modules/base/lightsource/scenegraphlightsource.cpp
```diff
@@ -41,18 +41,22 @@ namespace {
...
-    constexpr openspace::properties::Property::PropertyInfo NodeCameraStateInfo = {
+    constexpr openspace::properties::Property::PropertyInfo NodeInfo = {
         "Node",
         "Node",
         "The identifier of the scene graph node to follow.",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This `LightSource` type represents a light source placed at the position of a
+    // scene graph node. That is, it will follow an existing object in the scene.
```
alexanderbock, line 52:
```suggestion
    // scene graph node. That is, it will follow an existing object in the scene. Note
    // that the brightness of the light source does not depend on the distance between
    // the two scene graph nodes. Only the `Intensity` value has an impact on the
    // brightness.
```

?

src/scene/lightsource.cpp
```diff
@@ -42,13 +42,20 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This is the base class of all `LightSource` types, which are components that can be
+    // added to certain `Renderable` types to add lighting effects.
+    //
+    // A `LightSource` is in this case just a table that describes properties such as the
```
alexanderbock, line 48:
```suggestion
    // A `LightSource`, in this case, is just a table that describes properties such as the
```

src/scene/lightsource.cpp
```diff
@@ -42,13 +42,20 @@ namespace {
...
+    // location of the light. It _does not physically exist in the scene_, and the table
+    // of parameters have to be added to each `Renderable` the that should be affected by
```
alexanderbock, line 50:
```suggestion
    // of parameters have to be added to each `Renderable` that should be affected by
```

alexanderbock (approved):

modules/base/lightsource/cameralightsource.cpp
WeirdRubberDuck: Removing this change then :)

modules/base/lightsource/scenegraphlightsource.cpp
WeirdRubberDuck: Good addition 👍

WeirdRubberDuck: Added your suggestion, and updated the description a bit to make it a bit clearer from that start that the _direction_  is what is determined from the other scene graph node. Let me know if it looks ok. 

Also, do you think we need to make the same clarification for the camera light source?

alexanderbock: I don't think we need to add it to the camera light source as people (a) don't generally change the FOV values and (b) wouldn't expect that the light changes based on that.   Using another scene graph node as the light source, since it lives in 3D space conjures more of the feeling of it being a "real" light source with falloff

PR #3587, opened 2025-04-01
Enhance ScreenSpaceRenderable with Ordering & Alignment Controls,  Resolves #2250

### LINKED ISSUE #2250: Update screenspace renderables to increase ease of use
Want to make it easier to put screenspace images for example at a corner of the screen. Some concrete things to update: 

- Remove Z/radius property (that currently affect the visual size of the renderable) and instead order based on order in the list 
- Add option to reorder screen space renderables list similarly to layers (consequence of above) - OBS includes WebUI update
- Add dropdowns to set anchor/align points. 9 points for the image and window, rerspectively: BottomRight, BottomLeft, BottomCenter, LeftCenter, Center, RightCenter, etc.. (see image below) 
- Separate cartesian and radiusAzimuthElevation (RAE) logic. Use RAE automatically if using weird (non-planar) proejctions, and cartiesuan for planar projections
- Add trigger property to copy cartiesian position to RAE (using the current conversion)

To place an image X pixels from edge of window
- Add pixel padding property to the ScreenSpaceImages 
- helper function to compute cartesian position from window coordinates (e.g.  `openspace.screenspace.createRelativePosition("BottomRight", 50, 50)`)

To figure out: 
- Do we still want the "face Camera" checkbox?

Note: No changes to the current coordinate systems 

image

anmol-bhalla:
### **Changes Made**  
✅ **Added Render Ordering**  
- New `RenderOrder` property (lower values render first)  
- Allows better control over renderable layering  

✅ **Implemented 9-Point Alignment System**  
- Supports positions: `TopLeft`, `Center`, `BottomRight`, etc.  
- Simplified positioning logic  

✅ **Removed Deprecated RAE Coordinate System**  
- Cartesian coordinates only now  
- Cleaner, more maintainable code  

### **Testing Performed**  
- Manually verified render ordering works  
- Checked backward compatibility with existing scripts  

### **Impacted Files**  
- [`screenspacerenderable.h`](https://github.com/buddy326/OpenSpace/blob/screenspace-enhancements/include/openspace/rendering/screenspacerenderable.h)  
- [`screenspacerenderable.cpp`](https://github.com/buddy326/OpenSpace/blob/screenspace-enhancements/src/rendering/screenspacerenderable.cpp)  

### **Notes for Reviewers**  
- All new methods are documented in the header  
- Alignment offsets assume screen coordinates:  
  - X: `-1` (left) to `1` (right)  
  - Y: `-1` (bottom) to `1` (top)

### DIFF
--- a/src/rendering/screenspacerenderable.cpp
+++ b/src/rendering/screenspacerenderable.cpp
@@ -60,6 +60,20 @@ namespace {
         openspace::properties::Property::Visibility::User
     };
 
+    constexpr openspace::properties::Property::PropertyInfo RenderOrderInfo = {
+        "RenderOrder",
+        "Render Order",
+        "Lower values render first (behind higher values)",
+        openspace::properties::Property::Visibility::User
+    };
+    
+    constexpr openspace::properties::Property::PropertyInfo AlignmentInfo = {
+        "Alignment",
+        "Alignment",
+        "Anchor point for positioning",
+        openspace::properties::Property::Visibility::User
+    };
+
 //    constexpr openspace::properties::Property::PropertyInfo
 //        UseRadiusAzimuthElevationInfo =
 //    {
@@ -258,6 +272,10 @@ namespace {
         // `ScreenSpaceRenderable`, thus making it possible to address multiple, separate
         // Renderables with a single property change.
         std::optional<std::variant<std::string, std::vector<std::string>>> tag;
+
+        std::optional<float> renderOrder;
+        std::optional<std::string> alignment;
+
     };
 #include "screenspacerenderable_codegen.cpp"
 } // namespace
@@ -341,7 +359,18 @@ ScreenSpaceRenderable::ScreenSpaceRenderable(const ghoul::Dictionary& dictionary
         glm::vec4(0.f),
         glm::vec4(1.f)
     )
-    , _delete(DeleteInfo)
+    , _delete(DeleteInfo),
+    , _renderOrder(RenderOrderInfo, 0.f, -1000.f, 1000.f)
+    , _alignment(AlignmentInfo, {
+        { static_cast<int>(Alignment::TopLeft), "TopLeft" },
+        { static_cast<int>(Alignment::TopCenter), "TopCenter" },
+        { static_cast<int>(Alignment::TopRight), "TopRight" },
+        { static_cast<int>(Alignment::CenterLeft), "CenterLeft" },
+        { static_cast<int>(Alignment::Center), "Center" },
+        { static_cast<int>(Alignment::CenterRight), "CenterRight" },
+        { static_cast<int>(Alignment::BottomLeft), "BottomLeft" },
+        { static_cast<int>(Alignment::BottomCenter), "BottomCenter" },
+        { static_cast<int>(Alignment::BottomRight), "BottomRight" }
 {
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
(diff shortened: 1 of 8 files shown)


alexanderbock:
Awesome!  Will give full comments on the code a bit later, but a first thing is that the spherical coordinates are not deprecated.  They are necessary if you want to place ScreenSpaceRenderables in a dome environment, for example. In those situation you don't have a flat plane in front on which to move things in x and y, but rather want to rotate them in a spherical coordinate system around you.

I'd go with the following:
  - Readd the properties for RAE positioning
  - Follow @WeirdRubberDuck's suggestion in the issue "Add trigger property to copy Cartesian position to RAE (using the current conversion)" instead of converting the value automatically.

src/rendering/screenspacerenderable.cpp
```diff
@@ -341,7 +359,18 @@ ScreenSpaceRenderable::ScreenSpaceRenderable(const ghoul::Dictionary& dictionary
         glm::vec4(0.f),
         glm::vec4(1.f)
     )
-    , _delete(DeleteInfo)
+    , _delete(DeleteInfo),
+    , _renderOrder(RenderOrderInfo, 0.f, -1000.f, 1000.f)
+    , _alignment(AlignmentInfo, {
```
Roxeena, line 364: There is no constructor for the `OptionProperty` that takes in all the options. These will all have to be added later with the `addOption` function.

src/rendering/screenspacerenderable.cpp
```diff
@@ -448,6 +477,15 @@ ScreenSpaceRenderable::ScreenSpaceRenderable(const ghoul::Dictionary& dictionary
         });
     });
     addProperty(_delete);
+
+    _renderOrder = p.renderOrder.value_or(_renderOrder);
+    addProperty(_renderOrder);
+
+    if (p.alignment.has_value()) {
+        _alignment.set(alignmentFromString(*p.alignment));
```
Roxeena, line 485: There is no `set` function.

src/rendering/screenspacerenderable.cpp
```diff
@@ -547,6 +586,37 @@ void ScreenSpaceRenderable::createShaders(ghoul::Dictionary dict) {
...
+
+void ScreenSpaceRenderable::setAlignment(Alignment alignment) {
+    _alignment = static_cast<int>(alignment);
+}
+
+Alignment ScreenSpaceRenderable::alignment() const {
+    return static_cast<Alignment>(_alignment.value());
+}
+
+Alignment ScreenSpaceRenderable::alignmentFromString(const std::string& alignmentStr) {
```
Roxeena, line 605: This function needs to be registered in the .h file.

src/rendering/screenspacerenderable.cpp
```diff
@@ -561,7 +631,19 @@ glm::mat4 ScreenSpaceRenderable::scaleMatrix() {
...
 
 glm::vec2 ScreenSpaceRenderable::screenSpacePosition() {
-    return glm::vec2(_cartesianPosition.value());
+   // Convert from [-1,1] to [0,1] range
+   return glm::vec2(
+    _cartesianPosition.value().x * 0.5f + 0.5f,
+    _cartesianPosition.value().y * 0.5f + 0.5f
+);
+}
+void ScreenSpaceRenderable::setScreenPosition(glm::vec2 pos) {
```
Roxeena, line 640: This function also needs to be registered in the .h file

Roxeena (changes requested):
There are several compilation errors that need to be fixed before I can test this PR, I pointed some out in the comments. Also, the `renderOrder` functionality does not seem to be fully implemented yet. And as Alex mentioned before, the RAE functionality is needed in plentarium environments and cannot be removed.

alexanderbock:
Closing due to inactivity and being out of scope

PR #3589, opened 2025-04-03
Fix out of bounds crash and remove duplicate notifyChangeListeners

ylvaselling:
* Fixes hard crash when adding options dynamically with for example color mapping asset (out of bounds crash)
* Removes duplicate notifyChangeListeners (the NumericalProperty::SetValue function already calls notifyChangeListeners)


WeirdRubberDuck (approved):

WeirdRubberDuck:
Big oops from my side - never imagined it would lead to a crash. Should have tested more thoroughly. 

Thanks for fixing it 🙏

PR #3590, opened 2025-04-03
Rewrite shortcuts topic to ActionKeybind and pass data in a better way

ylvaselling:
This will break the UI so needs to be used with new UI on PR in that repo

* Renames the shortcut topic to ActionKeybind topic
* Passes data in a better way so less handling on the UI side
* Passes color of action to the UI

(Note that I noticed we are not able to set the color of the action in the profile. Is this something we want?)

To test the color: set something like this on an action:

```
Color = {1.0, 0.0, 1.0, 1.0},  
```

### DIFF
--- a/modules/server/include/topics/actionkeybindtopic.h
+++ b/modules/server/include/topics/actionkeybindtopic.h
@@ -22,26 +22,26 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#ifndef __OPENSPACE_MODULE_SERVER___ACTIONKEYBIND_TOPIC___H__
-#define __OPENSPACE_MODULE_SERVER___ACTIONKEYBIND_TOPIC___H__
+#ifndef __OPENSPACE_MODULE_SERVER___SHORTCUT_TOPIC___H__
+#define __OPENSPACE_MODULE_SERVER___SHORTCUT_TOPIC___H__
 
 #include <modules/server/include/topics/topic.h>
 
 namespace openspace {
 
-class ActionKeybindTopic : public Topic {
+class ShortcutTopic : public Topic {
 public:
-    ~ActionKeybindTopic() override = default;
+    ~ShortcutTopic() override = default;
 
     void handleJson(const nlohmann::json& input) override;
     bool isDone() const override;
 
 private:
     void sendData(nlohmann::json data) const;
-    nlohmann::json allActionsKeybinds() const;
-    nlohmann::json action(const std::string& identifier) const;
+    nlohmann::json shortcutsJson() const;
+    nlohmann::json shortcutJson(const std::string& identifier) const;
 };
 
 } // namespace openspace
 
-#endif // __OPENSPACE_MODULE_SERVER___ACTIONKEYBIND_TOPIC___H__
+#endif // __OPENSPACE_MODULE_SERVER___SHORTCUT_TOPIC___H__
--- a/modules/server/src/topics/actionkeybindtopic.cpp
+++ b/modules/server/src/topics/actionkeybindtopic.cpp
@@ -22,7 +22,7 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#include <modules/server/include/topics/actionkeybindtopic.h>
+#include <modules/server/include/topics/shortcuttopic.h>
 
 #include <modules/server/include/connection.h>
 #include <openspace/engine/globals.h>
@@ -113,7 +113,6 @@ nlohmann::json ActionKeybindTopic::allActionsKeybinds() const {
 }
 
 nlohmann::json ActionKeybindTopic::action(const std::string& identifier) const {
-    std::vector<nlohmann::json> json;
     std::vector<interaction::Action> actions = global::actionManager->actions();
 
     auto found = std::find_if(
@@ -129,9 +128,7 @@ nlohmann::json ActionKeybindTopic::action(const std::string& identifier) const {
     }
     interaction::Action action = *found;
 
-    json.push_back(jsonAction(action));
-
-    return json;
+    return jsonAction(action);
 }
 
 void ActionKeybindTopic::sendData(nlohmann::json data) const {
--- a/modules/server/CMakeLists.txt
+++ b/modules/server/CMakeLists.txt
@@ -45,7 +45,7 @@ set(HEADER_FILES
   include/topics/missiontopic.h
   include/topics/sessionrecordingtopic.h
   include/topics/setpropertytopic.h
-  include/topics/shortcuttopic.h
+  include/topics/actionkeybindtopic.h
   include/topics/subscriptiontopic.h
   include/topics/timetopic.h
   include/topics/skybrowsertopic.h
@@ -75,7 +75,7 @@ set(SOURCE_FILES
   src/topics/missiontopic.cpp
   src/topics/sessionrecordingtopic.cpp
   src/topics/setpropertytopic.cpp
-  src/topics/shortcuttopic.cpp
+  src/topics/actionkeybindtopic.cpp
   src/topics/subscriptiontopic.cpp
   src/topics/timetopic.cpp
   src/topics/skybrowsertopic.cpp
--- a/modules/server/src/connection.cpp
+++ b/modules/server/src/connection.cpp
@@ -99,7 +99,7 @@ Connection::Connection(std::unique_ptr<ghoul::io::Socket> s, std::string address
     _topicFactory.registerClass<MissionTopic>("missions");
     _topicFactory.registerClass<SessionRecordingTopic>("sessionRecording");
     _topicFactory.registerClass<SetPropertyTopic>("set");
-    _topicFactory.registerClass<ShortcutTopic>("shortcuts");
+    _topicFactory.registerClass<ActionKeybindTopic>("actionsKeybinds");
     _topicFactory.registerClass<SkyBrowserTopic>("skybrowser");
     _topicFactory.registerClass<SubscriptionTopic>("subscribe");
     _topicFactory.registerClass<TimeTopic>("time");


alexanderbock:
Something is odd..  The shortcuttopic files are not used in the CMakeLists anymore but are still in the repository. The actionkeybindtopic.h defines a class `ShortcutTopic`, the shortcuttopic.h defines `ActionKeybindTopic`, actionkeybindtopic.cpp includes shortcuttopic.h and shortcuttopic.cpp includes actionkeybindtopic.h

modules/server/include/topics/actionkeybindtopic.h
```diff
@@ -0,0 +1,47 @@
...
+ ****************************************************************************************/
+
+#ifndef __OPENSPACE_MODULE_SERVER___SHORTCUT_TOPIC___H__
+#define __OPENSPACE_MODULE_SERVER___SHORTCUT_TOPIC___H__
+
+#include <modules/server/include/topics/topic.h>
+
+namespace openspace {
+
+class ShortcutTopic : public Topic {
```
WeirdRubberDuck, line 32: Rename this class to match the new topic name?

WeirdRubberDuck: Or, as Alex said - something is wrong in the file structure here. Seems like the code that is in shortcuttopic.h should be put in this file and that file should be removed?

WeirdRubberDuck (changes requested):
Not testing right now due to broken/weird file structure. Please fix it :)

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -0,0 +1,149 @@
...
+        colorVec = { color.r, color.g, color.b, color.a };
+    }
+
+    const nlohmann::json shortcutJson = {
+        { "identifier", action.identifier },
+        { "name", action.name },
+        { "synchronization", static_cast<bool>(!action.isLocal) },
+        { "documentation", action.documentation },
+        { "guiPath", action.guiPath },
+        { "color", colorVec }
```
WeirdRubberDuck, line 55: now we would send an empty array if there is no color. Might be better to not include it at all in that case, maybe?  Easier to deal with undefined/null than an empty array and it makes it more clear when an action has a color or not

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+nlohmann::json jsonAction(const interaction::Action& action) {
+    // Convert to std as nlohmann doesn't understand glm
+    std::vector<float> colorVec;
+    if (action.color.has_value()) {
+        glm::vec4 color = action.color.value();
+        colorVec = { color.r, color.g, color.b, color.a };
+    }
+
+    const nlohmann::json shortcutJson = {
```
WeirdRubberDuck, line 49: rename to just `json`? At least remove "shortcut" and replace with action

or maybe even return directly, without specifying a named variable ?

WeirdRubberDuck:
Also, maybe this PR should be merged into `2025-ui-rewrite` instead of `master? (You can set that directly in the PR) :)

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -34,12 +34,56 @@ using nlohmann::json;
...
     return true;
 }
 
-std::vector<nlohmann::json> ShortcutTopic::shortcutsJson() const {
-    std::vector<nlohmann::json> json;
+nlohmann::json jsonAction(const interaction::Action& action) {
+    // Convert to std as nlohmann doesn't understand glm
+    nlohmann::json colorJson;
+    if (action.color.has_value()) {
+        glm::ivec4 color = action.color.value();
```
engbergandreas, line 45: What is the value range for action.color.value() 0-1 or 0-255? Since we are converting from float to int here won't this cause issues?

engbergandreas: eg. (0.1f, 0.5, 1.0f, 1.0f) => (0, 0 ,1, 1) => (0, 0, 255,255) ?

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -34,12 +34,56 @@ using nlohmann::json;
...
+        // Convert to [0, 255] color range as that is the format the UI wants
+        color *= 255.0;
```
WeirdRubberDuck, line 47:
```suggestion
        color *= 255.f;
```
using floats here :)

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -34,12 +34,56 @@ using nlohmann::json;
...
+        colorJson = { "color", { color.r, color.g, color.b, color.a } };
```
WeirdRubberDuck, line 48: But here, you should convert to ints before returning the value. '

However, My suggestion would be to maybe leave this conversion to 0-255 for the UI side, in case we want to use the serialization of the action colors for something else. Then, it would be better to not modify the data format for the color as part of the serialization.

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -34,12 +34,56 @@ using nlohmann::json;
...
+            { "key", ghoul::to_string(k.key) },
+            { "modifiers",
+                {
+                    { "shift" , hasKeyModifier(k.modifier, KeyModifier::Shift) },
+                    { "control" , hasKeyModifier(k.modifier, KeyModifier::Control) },
+                    { "alt" , hasKeyModifier(k.modifier, KeyModifier::Alt) },
+                    { "super" , hasKeyModifier(k.modifier, KeyModifier::Super) }
+                }
+            },
+            { "action", action.identifier },
```
WeirdRubberDuck, line 81:
```suggestion
        { "key", ghoul::to_string(k.key) },
        { "modifiers",
            {
                { "shift" , hasKeyModifier(k.modifier, KeyModifier::Shift) },
                { "control" , hasKeyModifier(k.modifier, KeyModifier::Control) },
                { "alt" , hasKeyModifier(k.modifier, KeyModifier::Alt) },
                { "super" , hasKeyModifier(k.modifier, KeyModifier::Super) }
            }
        },
        { "action", action.identifier }
```

WeirdRubberDuck (changes requested):
Not quite done yet. Please address:

1. There are some issues with setting the colors, due to precision being lost in conversion from [0,1] to [0,255]
2. Code structure => move json serialization to the dedicated file
3. Conversion on color format should happen at the final step in the pipeline, e.g. in ActoinsButton.tsx

modules/server/src/topics/actionkeybindtopic.cpp
engbergandreas: Can also swap the order and do multiplication first followed by conversion to ivec4 for nicer values on the frontend

WeirdRubberDuck: As long as we do the conversion explicitly, either works. :) 

However, see my other comment. I personally think we should do the conversion to [0,255] values _on the frontend_, when we set the value that CSS needs

WeirdRubberDuck: In OpenSpace we always represent color in range [0,1], so to be consistent, that's the data we should send

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -117,28 +133,21 @@ nlohmann::json ShortcutTopic::shortcutJson(const std::string& identifier) const
...
 
-    json.push_back({
-        { "identifier", action.identifier },
-        { "name", action.name },
-        { "synchronization", static_cast<bool>(!action.isLocal) },
-        { "documentation", action.documentation },
-        { "guiPath", action.guiPath },
-    });
-    return json;
+    return jsonAction(action);
```
alexanderbock, line 136:
```suggestion
    nlohmann::json a = jsonAction(action);
    return a;
```

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -117,28 +133,21 @@ nlohmann::json ShortcutTopic::shortcutJson(const std::string& identifier) const
...
 }
 
-void ShortcutTopic::sendData(nlohmann::json data) const {
-    _connection->sendJson(wrappedPayload({ { "shortcuts", data } }));
+void ActionKeybindTopic::sendData(nlohmann::json data) const {
+    _connection->sendJson(wrappedPayload({ data }));
```
alexanderbock, line 140:
```suggestion
    nlohmann::json payload = wrappedPayload({ data });
    _connection->sendJson(std::move(payload));
```

@engbergandreas  Might agree with how annoying it can be to try to debug these issues ;)

alexanderbock (approved):

modules/server/src/topics/actionkeybindtopic.cpp
```diff
@@ -75,33 +88,13 @@ std::vector<nlohmann::json> ShortcutTopic::shortcutsJson() const {
...
-                    { "shift" , hasKeyModifier(k.modifier, KeyModifier::Shift) },
-                    { "control" , hasKeyModifier(k.modifier, KeyModifier::Control) },
-                    { "alt" , hasKeyModifier(k.modifier, KeyModifier::Alt) },
-                    { "super" , hasKeyModifier(k.modifier, KeyModifier::Super) }
-                }
-            },
-            { "action", action.identifier },
-        };
-        json.push_back(shortcutJson);
+        nlohmann::json keybindJson = jsonKeybind(keyBinding.first, keyBinding.second);
```
WeirdRubberDuck, line 91: Do we want to keep Alex's TODO-comment here? 

(I'm personally not sure. It's not completely obvious what he means with it, but I guess it's the pair representation that could be rewritten?)

modules/server/src/jsonconverters.cpp
```diff
@@ -69,6 +70,31 @@ void to_json(json& j, const PropertyOwner* p) {
...
+            glm::vec4 color = action.color.value();
+            j["color"] = { color.r, color.g, color.b, color.a };
+        }
+    }
+
+    void to_json(json& j, const Action* pA) {
+        // Use reference converter instead of pointer
+        j = *pA;
+    }
+} // namespace openspace::interaction
```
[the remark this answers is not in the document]
WeirdRubberDuck: Also, sorry for requesting more things, but how about changing `synchronization` to `isLocal` while we're making changes?  

That aligns better with the data for the action and with what we want to represent in the UI. 

Found a bug in the UI code that needs to be fixed related to this. This check should be inversed:
image

WeirdRubberDuck (changes requested):
Looks great adnworks well!  One final request for changes though:  We should change `synchronization` to `isLocal` in the json-serialized data for the action

WeirdRubberDuck (approved):
Remember to update the UI PR to include the change to `isLocal` from  `synchronization`

PR #3594, opened 2025-04-08
Add handling for apostrophe code

### LINKED ISSUE #2761: Can't enter " characters in StringProperty inside WebUI
Trying to enter a " is prohibited even in cases where it would be legal to (for example StringPropertys that are not used as files)

ylvaselling:
Fixes the issue of not being able to use English keyboard and writing " in CEF. 

To test:
* Write in Swedish and English keyboard to see nothing else broke
* Switch to english keyboard and write ' and " and see that it works now

Closes #2761


alexanderbock (approved):
Works for me :+1:

PR #3598, opened 2025-04-11
Add distance-based scaling to RenderablePlane

lundkvistarn:
This PR adds optional distance-based scaling to RenderablePlane, enabling it to scale dynamically based on its distance from the camera. Also added example asset for it.

scalebydistance


data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_scalebydistance.asset
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+asset.onInitialize(function()
+    openspace.addSceneGraphNode(Node)
+end)
+
+asset.onDeinitialize(function()
+    openspace.removeSceneGraphNode(Node)
+end)
```
alexanderbock, line 39:
```suggestion
  Identifier = "RenderablePlaneImageLocal_Example_ScaleByDistance",
  Parent = earth.Earth.Identifier,
  Transform = {
    Translation = {
      Type = "GlobeTranslation",
      Globe = earth.Earth.Identifier,
      Altitude = 100000,
    }
  },
  Renderable = {
    Type = "RenderablePlaneImageLocal",
    Size = 1,
    Texture = openspace.absPath("${DATA}/test2.jpg"),
    ScaleByDistance = true,
    ScaleRatio = 0.01,
    ScaleByDistanceMaxHeight = 2000000,
    ScaleByDistanceMinHeight = 30000
  },
  GUI = {
    Name = "RenderablePlaneImageLocal - ScaleByDistance",
    Path = "/Examples"
  }
}

asset.onInitialize(function()
  openspace.addSceneGraphNode(Node)
end)

asset.onDeinitialize(function()
  openspace.removeSceneGraphNode(Node)
end)
```

We use 2 character indentation for Lua files

modules/base/rendering/renderableplane.cpp
```diff
@@ -29,6 +29,7 @@
 #include <openspace/documentation/verifier.h>
 #include <openspace/engine/globals.h>
 #include <openspace/rendering/renderengine.h>
+#include <openspace/engine/windowdelegate.h>
```
alexanderbock, line 32:
```suggestion
#include <openspace/engine/windowdelegate.h>
#include <openspace/engine/globals.h>
#include <openspace/rendering/renderengine.h>
```

Sorting

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
         "aspect ratio of the content. Otherwise it will remain in the given size."
     };
 
+    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceInfo = {
+        "ScaleByDistance",
+        "Scale By Distance",
+        "Decides whether the plane should automatically adjust in size to match the "
+        "distance to the camera. Otherwise it will remain in the given size."
```
alexanderbock, line 86:
```suggestion
        "Decides whether the plane should automatically adjust in size to based on "
        "the distance to the camera. Otherwise it will remain in the given size."
```

Since it doesn't match the distance of the camera.

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
...
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleRatioInfo = {
+        "ScaleRatio",
+        "Scale Ratio",
+        "The scale ratio for scaling a plane by distance to camera."
```
alexanderbock, line 92: Could you add another sentence here to add more information about permissible value ranges and what the ratio does?

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
...
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMaxHeightInfo = {
+        "ScaleByDistanceMaxHeight",
+        "Scale By Distance Max Height",
+        "The maximum height a plane can get while using the scale by distance."
```
alexanderbock, line 98: Assuming this is in meters?  Would be great to mention that here

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
...
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMinHeightInfo = {
+        "ScaleByDistanceMinHeight",
+        "Scale By Distance Min Height",
+        "The minimum height a plane can get while using the scale by distance."
```
alexanderbock, line 104: Same here

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
...
+        "distance to the camera. Otherwise it will remain in the given size."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleRatioInfo = {
+        "ScaleRatio",
+        "Scale Ratio",
+        "The scale ratio for scaling a plane by distance to camera."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMaxHeightInfo = {
```
alexanderbock, line 95:
```suggestion
    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMaxHeightInfo =
    {
```
Exceeding the line width

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
...
+        "ScaleByDistanceMaxHeight",
+        "Scale By Distance Max Height",
+        "The maximum height a plane can get while using the scale by distance."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMinHeightInfo = {
```
alexanderbock, line 101:
```suggestion
    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMinHeightInfo =
    {
```

Exceeding the line width

modules/base/rendering/renderableplane.cpp
```diff
@@ -187,6 +229,18 @@ RenderablePlane::RenderablePlane(const ghoul::Dictionary& dictionary)
     _autoScale = p.autoScale.value_or(_autoScale);
     addProperty(_autoScale);
 
+    _scaleByDistance = p.scaleByDistance.value_or(_scaleByDistance);
+    addProperty(_scaleByDistance);
+
+    _scaleRatio = p.scaleRatio.value_or(_scaleRatio);
+    addProperty(_scaleRatio);
+
+    _scaleByDistanceMaxHeight = p.scaleByDistanceMaxHeight.value_or(_scaleByDistanceMaxHeight);
```
alexanderbock, line 238:
```suggestion
    _scaleByDistanceMaxHeight =
        p.scaleByDistanceMaxHeight.value_or(_scaleByDistanceMaxHeight);
```

modules/base/rendering/renderableplane.cpp
```diff
@@ -187,6 +229,18 @@ RenderablePlane::RenderablePlane(const ghoul::Dictionary& dictionary)
...
+    addProperty(_scaleByDistanceMaxHeight);
+
+    _scaleByDistanceMinHeight = p.scaleByDistanceMinHeight.value_or(_scaleByDistanceMinHeight);
```
alexanderbock, line 241:
```suggestion
    _scaleByDistanceMinHeight =
        p.scaleByDistanceMinHeight.value_or(_scaleByDistanceMinHeight);
```

modules/base/rendering/renderableplane.cpp
```diff
@@ -262,6 +316,33 @@ void RenderablePlane::render(const RenderData& data, RendererTasks&) {
     cameraOrientedRotation[1] = glm::dvec4(newUp, 0.0);
     cameraOrientedRotation[2] = glm::dvec4(normal, 0.0);
 
+    if (_scaleByDistance) {
+        const glm::dvec3 cameraPosition = data.camera.positionVec3();
+        const glm::dvec3 modelPosition = data.modelTransform.translation;
+
+        const float fovDegrees = global::windowDelegate->getHorizFieldOfView();
```
alexanderbock, line 323: This fails in Fisheye mode, where the horizontal field of view returned is not valid.  I'd look at the RenderablePointCloud class to see how it does the scaling and if you can do it without needing the FOV here.  @WeirdRubberDuck  probably has more insight on this

modules/base/rendering/renderableplane.cpp
```diff
@@ -262,6 +316,33 @@ void RenderablePlane::render(const RenderData& data, RendererTasks&) {
...
+            projectedHeight,
+            _scaleByDistanceMinHeight.value(),
+            _scaleByDistanceMaxHeight.value()
+        );
+
+        glm::vec2 currentSize = _size.value();
+
+        if (currentSize.y != 0.f) {
+            float scaleFactor = projectedHeight / currentSize.y;
+            glm::vec2 scaledSize = currentSize * scaleFactor;
```
alexanderbock, line 341: I'd make all of these calculations `double` precision until the very last one to prevent the error of floating point inaccuracies

modules/base/rendering/renderableplane.h
```diff
@@ -73,6 +73,10 @@ class RenderablePlane : public Renderable {
     properties::BoolProperty _mirrorBackside;
     properties::Vec2Property _size;
     properties::BoolProperty _autoScale;
+    properties::BoolProperty _scaleByDistance;
+    properties::FloatProperty _scaleRatio;
+    properties::FloatProperty _scaleByDistanceMaxHeight;
+    properties::FloatProperty _scaleByDistanceMinHeight;
```
WeirdRubberDuck, line 79: Since all these properties are related, how about collecting them in a subpropertyowner?

modules/base/rendering/renderableplane.cpp
```diff
@@ -262,6 +316,33 @@ void RenderablePlane::render(const RenderData& data, RendererTasks&) {
...
+        float projectedHeight = 2.0f * distance * halfFovTan * _scaleRatio.value();
+        projectedHeight = std::clamp(
+            projectedHeight,
+            _scaleByDistanceMinHeight.value(),
+            _scaleByDistanceMaxHeight.value()
+        );
+
+        glm::vec2 currentSize = _size.value();
+
+        if (currentSize.y != 0.f) {
```
WeirdRubberDuck, line 339:
```suggestion
        if (currentSize.y > 0.f) {
```
? don't think it may be negative?

modules/base/rendering/renderableplane.cpp
WeirdRubberDuck: Here's the code that does it for the points, using the position of the camera.  The angle that you have here is computed in world space from the position of the camera, the object, and a vector describin the size of the object

https://github.com/OpenSpace/OpenSpace/blob/17279d2b4ad8b1672087c04e0742fbea2e460107/modules/base/shaders/pointcloud/pointcloud_gs.glsl#L135

Let me know if you want to discuss it @lundkvistarn !

modules/base/rendering/renderableplane.cpp
```diff
@@ -136,6 +174,10 @@ RenderablePlane::RenderablePlane(const ghoul::Dictionary& dictionary)
     , _mirrorBackside(MirrorBacksideInfo, false)
     , _size(SizeInfo, glm::vec2(10.f), glm::vec2(0.f), glm::vec2(1e25f))
     , _autoScale(AutoScaleInfo, false)
+    , _scaleByDistance(ScaleByDistanceInfo, false)
+    , _scaleRatio(ScaleRatioInfo, 0.01f)
+    , _scaleByDistanceMaxHeight(ScaleByDistanceMaxHeightInfo, 200000.f)
+    , _scaleByDistanceMinHeight(ScaleByDistanceMinHeightInfo, 30000.f)
```
WeirdRubberDuck, line 180: These properties need min and max values for the inputs in the UI to be useful

modules/base/rendering/renderableplane.cpp
```diff
@@ -110,6 +136,18 @@ namespace {
...
+        std::optional<bool> scaleByDistance;
+
+        // [[codegen::verbatim(ScaleRatioInfo.description)]]
+        std::optional<float> scaleRatio;
+
+        // [[codegen::verbatim(ScaleByDistanceMaxHeightInfo.description)]]
+        std::optional<float> scaleByDistanceMaxHeight;
+
+        // [[codegen::verbatim(ScaleByDistanceMinHeightInfo.description)]]
+        std::optional<float> scaleByDistanceMinHeight;
```
WeirdRubberDuck, line 149: Add `[[codegen::greater(0.f)]]` for some of these?

modules/base/rendering/renderableplane.cpp
```diff
@@ -78,6 +79,31 @@ namespace {
...
+        "ScaleByDistanceMaxHeight",
+        "Scale By Distance Max Height",
+        "The maximum height a plane can get while using the scale by distance."
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo ScaleByDistanceMinHeightInfo = {
+        "ScaleByDistanceMinHeight",
+        "Scale By Distance Min Height",
+        "The minimum height a plane can get while using the scale by distance."
+    };
```
WeirdRubberDuck, line 105: add property visibility level to all the new properties

modules/base/rendering/renderableplane.cpp
WeirdRubberDuck: Yes please, I personally do not understand it. 

Is this the thing that decides how big the plane is allowed to become?  Maybe it can be changed to be a bit more intuitive?  Now it's unclear, at least to me, how this relates to the min and max size properties

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_scalebydistance.asset
```diff
@@ -0,0 +1,39 @@
+-- Scaling Image
+-- This example shows how to create a textured plane at Earth, where the texture is
+-- loaded from a local file on disk and the plane changes its scale based on its distance
+-- to the camera.
```
WeirdRubberDuck, line 4: Rewrite this so that the "special" thing about this example is as early as possible in the sentence. I also suggest that you remove the ``Parent and `Transform` that places it close to Earth, to make the example as minimal as possible. 

The description could then look something like:
```suggestion
-- This example creates a textured plane that is scaled based on the distance to the
-- camera, so that it stays a constant size in screen space. The scale is limited so that
-- the plane does not become larger or smaller than a given max height and min height, in
-- meters.
```

And the code below just: 
```
local Node = {
  Identifier = "RenderablePlaneImageLocal_Example_ScaleByDistance",
  Renderable = {
    Type = "RenderablePlaneImageLocal",
    Size = 1,
    Texture = openspace.absPath("${DATA}/test2.jpg"),
    ScaleByDistance = true,
    ScaleRatio = 0.01,
    ScaleByDistanceMaxHeight = 2000000,
    ScaleByDistanceMinHeight = 30000
  },
  GUI = {
    Name = "RenderablePlaneImageLocal - ScaleByDistance",
    Path = "/Examples"
  }
}
```

(But according to some of my comments below, the scale properties should be grouped into one object)

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_scalebydistance.asset
```diff
@@ -0,0 +1,39 @@
+-- Scaling Image
```
WeirdRubberDuck, line 1:
```suggestion
-- Scale by Distance to Camera
```

modules/base/rendering/renderableplane.cpp
WeirdRubberDuck, line 149: These should be grouped into a struct (and also a property ower in the UI). Se for example how it's done for the size settings in the point cloud renderable: 
https://github.com/OpenSpace/OpenSpace/blob/17279d2b4ad8b1672087c04e0742fbea2e460107/modules/base/rendering/pointcloud/renderablepointcloud.cpp#L434

I'll gladly discuss how to do it, if you want to!  Have some ideas

modules/base/rendering/renderableplane.cpp
WeirdRubberDuck: Also, I think that it might maybe be worth to move this computation to a geometry shader. It's currently not that smooth when interacting. See this video

https://github.com/user-attachments/assets/89d6363c-3345-45be-980e-eb2688a3224b

WeirdRubberDuck (changes requested):
Awesome, but needs some improvements (such as adding min max values and grouping properties into a subpropertyowner)

I'll gladly help out and discuss in case the comments are unclear @lundkvistarn !

modules/base/rendering/renderableplane.cpp
WeirdRubberDuck: Also, the min and max for the max and min height should be something more general. These values are a lot higher than the default size, for example. 

Maybe a min of 1 and a max of 10?

PR #3601, opened 2025-04-14
Add submenu to the "New" buttons

### LINKED ISSUE #3578: Better handling of Edit/New for Profiles
Instead of editing existing profiles and saving them under a new name, it would be better to prevent editing of data/profiles and instead provide the ability to use them as templates for the "New" option.

alexanderbock:
Add a new menu to the New button in the Launcher for the New profile to provide the option to either create an empty profile or duplicate an existing one.  Disable the "Edit" button when a built-in profile is selected

image

Closes #3578

### DIFF
--- a/apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
+++ b/apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
@@ -35,6 +35,7 @@
 #include <QFile>
 #include <QKeyEvent>
 #include <QLabel>
+#include <QMenu>
 #include <QMessageBox>
 #include <QPushButton>
 #include <QStandardItemModel>
@@ -195,6 +196,20 @@ LauncherWindow::LauncherWindow(bool profileEnabled, const Configuration& globalC
             newProfileButton, &QPushButton::released,
             this, &LauncherWindow::newProfile
         );
+
+        QMenu* menu = new QMenu;
+        QAction* newClean = new QAction("New profile");
+        connect(
+            newClean, &QAction::triggered,
+            this, &LauncherWindow::newProfile
+        );
+        QAction* newFromCurrent = new QAction("Duplicate profile");
+        connect(
+            newFromCurrent, &QAction::triggered,
+            this, &LauncherWindow::editProfile
+        );
+        menu->addActions({ newClean, newFromCurrent });
+        newProfileButton->setMenu(menu);
     }
 
     // Creating the profile box _after_ the Edit and New buttons as the comboboxes
@@ -391,7 +406,19 @@ void LauncherWindow::selectProfile(std::optional<std::string> selection) {
     ghoul_assert(selection.has_value(), "No special item in the profiles");
     if (selection.has_value()) {
         // Having the `if` statement here to satisfy the MSVC code analysis
-        _editProfileButton->setEnabled(std::filesystem::exists(*selection));
+
+        // Enable the Edit button only for the user profiles
+        const bool isUser = selection->starts_with(_userProfilePath.string());
+        _editProfileButton->setEnabled(isUser);
+
+        if (isUser) {
+            _editProfileButton->setToolTip("");
+        }
+        else {
+            _editProfileButton->setToolTip(
+                "Cannot edit the selected profile as it is one of the built-in profiles"
+            );
+        }
     }
 }
 
@@ -407,8 +434,7 @@ void LauncherWindow::selectConfiguration(std::optional<std::string> selection) {
         // If the configuration is a default configuration, we don't allow editing
         _editWindowButton->setEnabled(false);
         _editWindowButton->setToolTip(
-            "Cannot edit since the selected configuration is one of the files "
-            "provided by OpenSpace"
+            "Cannot edit the selected configuration as it is one of the built-in profiles"
         );
     }
     else {


ylvaselling (commented):
Looks good!

Some thoughts (maybe out of scope of this pr?):
* The popup about discarding changes shows up if you opened a new profile but didn't change anything. If you open a duplicate profile but don't change anything it doesn't show up. Make it not show up for the new profile as well?
* Should it be called "Empty profile" rather than "New profile"? Now it's New > New Profile which doesn't really make sense to me
* Also I just noticed that the cursor is a pointer on the buttons below the profile name but not the profile name button
* The text alignment seems a little off to me?
image
* Is it possible to style it in a similar way to the other dropdowns or buttons? Now it kind of introduces another style

WeirdRubberDuck:
> Looks good!
> 
> Some thoughts (maybe out of scope of this pr?):
> 
> * The popup about discarding changes shows up if you opened a new profile but didn't change anything. If you open a duplicate profile but don't change anything it doesn't show up. Make it not show up for the new profile as well?
> * Should it be called "Empty profile" rather than "New profile"? Now it's New > New Profile which doesn't really make sense to me
> * Also I just noticed that the cursor is a pointer on the buttons below the profile name but not the profile name button
> * The text alignment seems a little off to me?
>   image
> * Is it possible to style it in a similar way to the other dropdowns or buttons? Now it kind of introduces another style

➕on all these comments. And to me, fixing at least the last 4 seems to be within the scope of the PR?

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -391,7 +406,19 @@ void LauncherWindow::selectProfile(std::optional<std::string> selection) {
...
+        _editProfileButton->setEnabled(isUser);
+
+        if (isUser) {
+            _editProfileButton->setToolTip("");
+        }
+        else {
+            _editProfileButton->setToolTip(
+                "Cannot edit the selected profile as it is one of the built-in profiles"
+            );
+        }
```
WeirdRubberDuck, line 421: This seems to only be applied after selecting a profile. I can still edit the default profile when starting the launcher, before selecting a profile in the list. 

Please fix :)

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -195,6 +196,20 @@ LauncherWindow::LauncherWindow(bool profileEnabled, const Configuration& globalC
             newProfileButton, &QPushButton::released,
             this, &LauncherWindow::newProfile
         );
+
+        QMenu* menu = new QMenu;
+        QAction* newClean = new QAction("New profile");
```
WeirdRubberDuck, line 201:
```suggestion
        QAction* newClean = new QAction("Empty profile");
```

WeirdRubberDuck (changes requested):
Agreeign with Ylva that it's a bit weird that the style of this dropdown differs compares to the other in the menu

Also, this spacing looks quite weird. Why is it so wide?
image

WeirdRubberDuck:
Also, pressing enter on the "New" button before it is expanded currently opens a new empty profile. I don't think this is expected behavior? 

Would have expected the dropdown to open, but it seems like all our dropdowns open only on SPACE rather than ENTER

WeirdRubberDuck:
> Can someone check this on an older version of Qt (running 6.9.0) as it smells like a Qt bug to me: When moving into the profile name box _from the top_ the cursor doesn't change. For the window config box it works and from any other side it works as well
> 
>  2025-04-22.15-16-28.mp4

Yep, not an issue in Qt 6.7.2

apps/OpenSpace/ext/launcher/resources/qss/launcher.qss
```diff
@@ -61,6 +61,17 @@ LauncherWindow  QComboBox#config:disabled {
...
+  background: rgb(50, 50, 50);
+  min-width: 8em;
+  max-width: 8em;
+  color: rgb(255, 255, 255);
+}
+
+LauncherWindow  QMenu#newprofile::item:selected {
+  background: rgb(100, 100, 100);
+}
+
```
WeirdRubberDuck, line 74: We're getting dangerously close to 50-shades-of-gray territory here 😅 >.> 

image

apps/OpenSpace/ext/launcher/resources/qss/launcher.qss
```diff
@@ -61,6 +61,17 @@ LauncherWindow  QComboBox#config:disabled {
   color: rgb(225, 225, 225);
 }
 
+LauncherWindow  QMenu#newprofile {
+  background: rgb(50, 50, 50);
+  min-width: 8em;
+  max-width: 8em;
+  color: rgb(255, 255, 255);
```
WeirdRubberDuck, line 68:
```suggestion
  color: white;
```
to be clearer when it's pure white and when it's a bright gray?   maybe we should avoid RGB values completely, really

alexanderbock: used white and black where we can.

Unfortunately, QSS doesn't have a way to do that except for grafting on some text-replacement, which feels iffy to do.

https://stackoverflow.com/questions/10898399/using-variables-in-qt-stylesheets

PR #3604, opened 2025-04-15
Remove "Up" keybind and retire `property_helper` and `renderable_helper` 

WeirdRubberDuck:
This PR includes two changes, but I guess they might need come discussion: 

1. Remove the Up keybind that interpolates the time to now. We all agreed on this, since it's dangerous and annoying to press by accident
    - I also removed the action that triggered the interpolation. Let me know if this should be brought back (I personally don't think so)
    
2. I also removed the `property_helper` and `renderable_helper` assets. `property_helper` was only used for inverting boolean property values, which we now have a Lua API function for in core_scripts. It was also imported in a bunch of assets where it was no longer used. The `renderable_helper` was not used at all. 
    - In conclusion, to me, it seemed like it was time for retirement for these trusty old guys :) I hope you agree?
    - Might be a **breaking change** for users that have built assets with the property helper

## Discussion / TODO
I was also on the verge of moving the actions form the `default_keybindings.asset` file to another asset, to split up the actions themselves and the keybinding. It would be nice to allow users to load all these essential actions without having them be bound to a key. Hoewver, left this as of now, since I felt it needed discussion on where to put the actions instead (we already have a "default_actions" asset, which is different than these default keybindings...)


alexanderbock (approved):

WeirdRubberDuck:
> 👍 For removing the helper assets 👍 For removing the Up keybind
> 
> How about renaming the `default_actions.asset` -> `solarsystem_actions.asset` and then separating the `default_keybindings.asset` into `default_keybindings` and `default_actions`?

Big thumbs up on that suggestion. I'll go head and do it as part of this PR?

WeirdRubberDuck:
> Can you add the removal of the assets to the https://internal.openspaceproject.com/en/breaking-changes-for-next-version page?

And of course :)

ylvaselling (approved):

WeirdRubberDuck:
> > 👍 For removing the helper assets 👍 For removing the Up keybind
> > How about renaming the `default_actions.asset` -> `solarsystem_actions.asset` and then separating the `default_keybindings.asset` into `default_keybindings` and `default_actions`?
> 
> Big thumbs up on that suggestion. I'll go head and do it as part of this PR?

Ended up doing a new PR for this as the current one was already approved

PR #3605, opened 2025-04-15
Feature/notify meta data change

ylvaselling:
To be used with this PR: https://github.com/OpenSpace/OpenSpace-WebGui/pull/120

Changes:
* Notifies when meta data changes
* Changes structure of JSON passed to UI to the format we actually want in the UI
    * Note that this doesn't really match the structure we have in the Property class (_metaData struct). Input is welcome
* Only passes value when value changes and only passes meta data when meta data changes
* Removes `generateMetaDataJson` method 
* Changes JSONs to camel case


include/openspace/properties/numericalproperty.inl
```diff
@@ -61,6 +61,7 @@ T NumericalProperty<T>::minValue() const {
 
 template <typename T>
 void NumericalProperty<T>::setMinValue(T value) {
+    Property::notifyMetaDataChangeListeners();
     _minimumValue = std::move(value);
```
alexanderbock, line 65: I feel we should call the notify function _after_ setting the value if we can?  Otherwise if someone is listening to the notify change and requests the meta data, they will get the old value back

include/openspace/properties/numericalproperty.inl
```diff
@@ -61,6 +61,7 @@ T NumericalProperty<T>::minValue() const {
 
 template <typename T>
 void NumericalProperty<T>::setMinValue(T value) {
+    Property::notifyMetaDataChangeListeners();
```
alexanderbock, line 64:
```suggestion
    notifyMetaDataChangeListeners();
```

include/openspace/properties/numericalproperty.inl
```diff
@@ -71,6 +72,7 @@ T NumericalProperty<T>::maxValue() const {
 
 template <typename T>
 void NumericalProperty<T>::setMaxValue(T value) {
+    Property::notifyMetaDataChangeListeners();
```
alexanderbock, line 75:
```suggestion
    notifyMetaDataChangeListeners();
```

include/openspace/properties/numericalproperty.inl
```diff
@@ -81,6 +83,7 @@ T NumericalProperty<T>::steppingValue() const {
 
 template <typename T>
 void NumericalProperty<T>::setSteppingValue(T value) {
+    Property::notifyMetaDataChangeListeners();
```
alexanderbock, line 86:
```suggestion
    notifyMetaDataChangeListeners();
```

include/openspace/properties/numericalproperty.inl
```diff
@@ -118,17 +121,18 @@ void NumericalProperty<T>::setExponent(float exponent) {
     }
 
     _exponent = exponent;
+    Property::notifyMetaDataChangeListeners();
```
alexanderbock, line 124:
```suggestion
    notifyMetaDataChangeListeners();
```

include/openspace/properties/property.h
```diff
@@ -236,6 +246,21 @@ class Property {
...
+    /**
+     * This method registers a \p callback function that will be called every time the
+     * notifyMetaDataChangeListener function was called. The callback can be removed by
+     * calling the removeOnMetaDataChange method with the OnMetaDataChangeHandle that was
+     * returned here.
+     *
+     * \param callback The callback function that is called when the meta data of the
+     * property has been changed.
+     * \return An OnMetaDataChangeHandle that can be used in subsequent calls to remove a
+     * callback
```
alexanderbock, line 258:
```suggestion
     * \param callback The callback function that is called when the meta data of the
     *        property has been changed.
     * \return An OnMetaDataChangeHandle that can be used in subsequent calls to remove a
     *         callback
```

include/openspace/properties/property.h
```diff
@@ -261,6 +286,19 @@ class Property {
      */
     void removeOnDelete(OnDeleteHandle handle);
 
+    /**
+     * This method deregisters a callback that was previously registered with the onMetaDataChange
+     * method. If OnMetaDataChangeHandleAll is passed to this function, all registered callbacks
+     * are removed.
+     *
+     * \param handle An OnMetaDataChangeHandle that was returned from a previous call to onMetaDataChange
```
alexanderbock, line 294: This line looks like it is too long?

include/openspace/properties/property.h
```diff
@@ -509,6 +541,10 @@ class Property {
     /// The callback functions that will be invoked whenever the value changes
     std::vector<std::pair<OnDeleteHandle, std::function<void()>>> _onDeleteCallbacks;
 
+    /// The callback functions that will be invoked whenever the value changes
```
alexanderbock, line 544:
```suggestion
    /// The callback functions that will be invoked whenever the meta data changes
```

src/properties/property.cpp
```diff
@@ -193,6 +205,29 @@ void Property::removeOnDelete(OnDeleteHandle handle) {
...
+        ghoul_assert(
+            it != _onMetaDataChangeCallbacks.end(),
+            "handle must be a valid callback handle"
+        );
+
+        _onMetaDataChangeCallbacks.erase(it);
+    }
+}
+
+
```
alexanderbock, line 230:
```suggestion
```

src/properties/property.cpp
```diff
@@ -207,6 +242,12 @@ void Property::notifyChangeListeners() {
     }
 }
 
+void Property::notifyMetaDataChangeListeners() {
+    for (const std::pair<OnMetaDataChangeHandle, std::function<void()>>& p : _onMetaDataChangeCallbacks) {
```
alexanderbock, line 246: This line looks too long?

src/properties/property.cpp
```diff
@@ -215,61 +256,48 @@ void Property::resetToUnchanged() {
...
-        { Visibility::User, "User" },
-        { Visibility::AdvancedUser, "AdvancedUser" },
-        { Visibility::Developer, "Developer" },
-        { Visibility::Hidden, "Hidden" }
+       { Visibility::Always, "Always" },
+       { Visibility::NoviceUser, "NoviceUser" },
+       { Visibility::User, "User" },
+       { Visibility::AdvancedUser, "AdvancedUser" },
+       { Visibility::Developer, "Developer" },
+       { Visibility::Hidden, "Hidden" }
```
alexanderbock, line 266:
```suggestion
        { Visibility::Always, "Always" },
        { Visibility::NoviceUser, "NoviceUser" },
        { Visibility::User, "User" },
        { Visibility::AdvancedUser, "AdvancedUser" },
        { Visibility::Developer, "Developer" },
        { Visibility::Hidden, "Hidden" }
```

I think the indentation was correct before

include/openspace/properties/numericalproperty.inl
WeirdRubberDuck: Yes, the notification has to happen after the change!  Now, we send invalid information 

To make sure it works correctly @ylvaselling you may use the renderablenodearrow example and set the Use relative Length or Use relative Offsets property. This renderable has a function that sets the min and max when changing two properties to be "relative" or not. The min and max settings are currently not correct

https://github.com/OpenSpace/OpenSpace/blob/93ef632663009f0b3b31eff4f311560764c5fcce/modules/base/rendering/renderablenodearrow.cpp#L174

WeirdRubberDuck (changes requested):
Not completely finished with my review, but found some issues

src/properties/property.cpp
ylvaselling: It's two tabs in now 🤔

include/openspace/properties/numericalproperty.inl
ylvaselling: I'm afraid I can't do that 
image

PR #3611, opened 2025-04-17
Launcher should indicate whether a profile is a standard profile or a user profile 

### LINKED ISSUE #3576: Launcher should indicate whether a profile is a standard profile or a user profile
Right now when starting the launcher, it is not possible to see whether a profile "default" would load the `default.profile` from the `data/profiles` folder or the `user/data/profiles` folder. The user can see this by opening the dropdown menu, but otherwise it is not possible.

Related issue is that we are only storing the final location of the profile in the settings file, so if someone started the `data/profiles/default.profile` the previous time, it would start the `user/data/profiles/default.profile` the next time.

alexanderbock:
Add icon to all user-generated profile and configurations.  The icon is a U for "user generated content" within a circle of one of the logo's primary colors.

Open for comments on the look of the logo.  We can control the size, aspect ratio, and basically draw whatever we'd like, but the limitation is that it has to be on the left side of the text.

Another option would have been to use `setLabelDrawingMode` of the QComboBox, but that was only introduced in the latest Qt version 6.9 and still has some bugs with it.  It would be be possible to customize the rendering of the text with it, that that were to work.

image

Closes #3576

### DIFF
--- a/apps/OpenSpace/ext/launcher/src/splitcombobox.cpp
+++ b/apps/OpenSpace/ext/launcher/src/splitcombobox.cpp
@@ -24,6 +24,7 @@
 
 #include "splitcombobox.h"
 
+#include "usericon.h"
 #include <ghoul/filesystem/filesystem.h>
 #include <ghoul/misc/assert.h>
 #include <QPainter>
@@ -68,28 +69,14 @@ void SplitComboBox::populateList(const std::string& preset) {
     clear();
 
     // Create "icons" that we use to indicate whether an item is built-in or user content
-    QIcon userIcon = []() {
-        QPixmap px = QPixmap(40, 50);
-        px.fill(Qt::transparent);
-
-        QPainter painter = QPainter(&px);
-        painter.setBrush(QColor(183, 211, 149, 255));
-        painter.drawEllipse(0, 10, 38, 38);
-
-        QFont f = QFont("Arial");
-        f.setPixelSize(28);
-        f.setBold(true);
-        painter.setFont(f);
-        painter.drawText(0, 10, 40, 40, Qt::AlignCenter, "U");
-        return QIcon(px);
-    }();
+    QIcon icon = userIcon();
 
     //
     // Special item (if it was specified)
     if (!_specialFirst.empty()) {
         if (_specialFirst.starts_with(_userPath.string())) {
             addItem(
-                userIcon,
+                icon,
                 QString::fromStdString(_specialFirst),
                 QString::fromStdString(_specialFirst)
             );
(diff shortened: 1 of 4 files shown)


ylvaselling (approved):
👍

ylvaselling:
I ran Cmake and built it and got this error:  

image

image

ylvaselling (changes requested):
Changing my review since I got some thrown errors and couldn't run OpenSpace

alexanderbock:
@ylvaselling  Try again now.  Hadn't updated this branch to the newest `master` in which this issue was fixed yet

ylvaselling:
Looks good!

Some potential polishing:

* Aligning the U symbol in the dropdown list to the first character in the other names
* Maybe the symbol less pixel-y?
 
image

apps/OpenSpace/ext/launcher/src/splitcombobox.cpp
```diff
@@ -66,14 +67,39 @@ void SplitComboBox::populateList(const std::string& preset) {
...
+        painter.setBrush(QColor(183, 211, 149, 255));
+        painter.drawEllipse(0, 5, 19, 19);
+
+        QFont f = QFont("Arial");
+        f.setPixelSize(14);
+        f.setBold(true);
+        painter.setFont(f);
+        painter.drawText(0, 5, 20, 20, Qt::AlignCenter, "U");
+        return QIcon(px);
+    }();
```
WeirdRubberDuck, line 85: Maybe this icon should live somewhere else?  Thinking if we want to use it inside of the editor as well, for things that live in the user folder

alexanderbock:
I won't even try to align those as that will be extremely difficult.  In the two pictures above, there are already different paddings, and here is yet a third where it already is aligned

image

Qt is really not made for these microadjustments. It's more like all of nothing with them

apps/OpenSpace/ext/launcher/src/splitcombobox.cpp
alexanderbock: Moved the icon creation to a separate file.

The component came to be as a deduplication of the code that existed there before.  It's basically a combobox that can display to distinct lists of objects (hence SplitComboBox)

apps/OpenSpace/ext/launcher/src/splitcombobox.cpp
```diff
@@ -66,14 +67,39 @@ void SplitComboBox::populateList(const std::string& preset) {
...
 
     //
     // Special item (if it was specified)
```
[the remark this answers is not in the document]
alexanderbock: Its the item that is set if OpenSpace is started with a commandline option that overwrites the selection.  That file doesn't have to follow the normal conventions (not from `data/profile` or `user/data/profile` for example, which is why we have to treat it separately

ylvaselling (approved):

PR #3621, opened 2025-04-23
Add time varying dashboard item text class 

lundkvistarn:
New dashboard item class, DashboardItemTimeVaryingText, designed to display numeric values from a JSON time series, dynamically updating based on time. Also added a example for this.

Useful for displaying dynamic values like KP index or solar wind parameters in sync with time.

timevarying


modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+namespace {
+    constexpr std::string_view _loggerCat = "DashboardItemTimeVaryingText";
+
+    constexpr openspace::properties::Property::PropertyInfo TextBeforeInfo = {
+        "TextBefore",
+        "Text Before",
+        "Optional text before the time varying text to be displayed.",
+        openspace::properties::Property::Visibility::User
+    };
```
alexanderbock, line 46: To be more consistent with other Dashboard items (DashboardItemDate, DashboardItemDistance, etc) it would be better to have this as a `FormatString` instead.  So in the example above it would be `"Observed KP index: {}"` instead. That would allow a user to also add a suffix through `"{} is the observed KP index"` for example.

You can lift the details straight from the DashboardItemDate.
Property: https://github.com/OpenSpace/OpenSpace/blob/master/modules/base/dashboard/dashboarditemdate.cpp#L39-L46
Parameters: https://github.com/OpenSpace/OpenSpace/blob/master/modules/base/dashboard/dashboarditemdate.cpp#L63-L64
Resolving the format string: https://github.com/OpenSpace/OpenSpace/blob/master/modules/base/dashboard/dashboarditemdate.cpp#L103-L109

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    constexpr openspace::properties::Property::PropertyInfo DataFileInfo = {
```
alexanderbock, line 47:
```suggestion

    constexpr openspace::properties::Property::PropertyInfo DataFileInfo = {
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        "base_dashboarditem_timevaryingtextBefore",
+        DashboardTextItem::Documentation()
+    );
+}
+
+DashboardItemTimeVaryingText::DashboardItemTimeVaryingText(
+                                                      const ghoul::Dictionary& dictionary)
+    : DashboardTextItem(dictionary)
+    , _textBefore(TextBeforeInfo, "")
+    , _dataFile(DataFileInfo, "")
```
alexanderbock, line 76: Either the `_dataFile` property needs an `onChange` handler that reloads the file whenever the property changes, it shouldn't be a property, or it should be read-only (in that order of preference).  Right now a user would be able to enter a new path, but nothing would happen since the `loadDataFromJson` is only done in the constructor

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        std::string dataFile;
+    };
+#include "dashboarditemtimevaryingtext_codegen.cpp"
+} // namespace
+
+namespace openspace {
+
+documentation::Documentation DashboardItemTimeVaryingText::Documentation() {
+    return codegen::doc<Parameters>(
+        "base_dashboarditem_timevaryingtextBefore",
```
alexanderbock, line 67:
```suggestion
        "base_dashboarditem_timevaryingtext",
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        "DataFile",
+        "Data File Path",
+        "The file path to the JSON data.",
+        openspace::properties::Property::Visibility::User
+    };
+
+    struct [[codegen::Dictionary(DashboardItemTimeVaryingText)]] Parameters {
+        // [[codegen::verbatim(TextBeforeInfo.description)]]
+        std::optional<std::string> textBefore;
+        // [[codegen::verbatim(DataFileInfo.description)]]
```
alexanderbock, line 57:
```suggestion

        // [[codegen::verbatim(DataFileInfo.description)]]
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        openspace::properties::Property::Visibility::User
+    };
+    constexpr openspace::properties::Property::PropertyInfo DataFileInfo = {
+        "DataFile",
+        "Data File Path",
+        "The file path to the JSON data.",
+        openspace::properties::Property::Visibility::User
+    };
+
+    struct [[codegen::Dictionary(DashboardItemTimeVaryingText)]] Parameters {
```
alexanderbock, line 54:
```suggestion
    // This `DashboardItem` displays text based on the content of a provided data file. The
    // value that is displayed depends on the current in-game simulation time.
    struct [[codegen::Dictionary(DashboardItemTimeVaryingText)]] Parameters {
```

Should add some more description about the file format once that is settled (see below)

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    : DashboardTextItem(dictionary)
+    , _textBefore(TextBeforeInfo, "")
+    , _dataFile(DataFileInfo, "")
+{
+    const Parameters p = codegen::bake<Parameters>(dictionary);
+    _textBefore = p.textBefore.value_or(_textBefore);
+    addProperty(_textBefore);
+    _dataFile = absPath(p.dataFile).string();
+    addProperty(_dataFile);
+    loadDataFromJson(_dataFile);
```
alexanderbock, line 83: This loading should happen in the `initialize` method so that it is multithreaded

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+                                                      const ghoul::Dictionary& dictionary)
+    : DashboardTextItem(dictionary)
+    , _textBefore(TextBeforeInfo, "")
+    , _dataFile(DataFileInfo, "")
+{
+    const Parameters p = codegen::bake<Parameters>(dictionary);
+    _textBefore = p.textBefore.value_or(_textBefore);
+    addProperty(_textBefore);
+    _dataFile = absPath(p.dataFile).string();
+    addProperty(_dataFile);
```
alexanderbock, line 82:
```suggestion
    _textBefore = p.textBefore.value_or(_textBefore);
    addProperty(_textBefore);
    
    _dataFile = absPath(p.dataFile).string();
    addProperty(_dataFile);
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    loadDataFromJson(_dataFile);
+}
+
+void DashboardItemTimeVaryingText::update() {
+    if (_startTimes.empty()) {
+        _buffer = "";
```
alexanderbock, line 88:
```suggestion
        _buffer.clear();
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    if (current >= first && current < last) {
+        int newIdx = updateActiveTriggerTimeIndex(current);
+        if (newIdx != _activeTriggerTimeIndex) {
+            _activeTriggerTimeIndex = newIdx;
+            double timeKey = _startTimes[_activeTriggerTimeIndex];
+            double value = _data[timeKey];
+
+            std::ostringstream oss;
+            oss << value;
+            _buffer = _textBefore.value() + oss.str();
```
alexanderbock, line 105: This would be the place for the string replacement mentioned above

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+    else {
+        _activeTriggerTimeIndex = -1;
+        _buffer = "";
```
alexanderbock, line 110:
```suggestion
        _buffer.clear();
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+void DashboardItemTimeVaryingText::loadDataFromJson(const std::string& filePath) {
+    std::ifstream file(filePath);
```
alexanderbock, line 115:
```suggestion
    std::ifstream file = std::ifstream(filePath);
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    if (!file.is_open()) {
+        LERROR(std::format(
+            "Time varying text, '{}' is not a valid JSON file",
+            filePath
+        ));
```
alexanderbock, line 120:
```suggestion
        throw ghoul::Runtime(std::format(
            "Time varying text, '{}' is not a valid JSON file",
            filePath
        ));
```

Since it is not a recoverable error if the JSON file doesn't load

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        return;
+    }
+
+    nlohmann::json jsonData;
+    file >> jsonData;
+
+    _data.clear();
+    _startTimes.clear();
+
```
alexanderbock, line 129:
```suggestion
```

As this is only done once, both containers have to be empty

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+    for (const auto& item : jsonData["data"]) {
```
alexanderbock, line 130:
```suggestion
    for (const nlohmann::json& item : jsonData["data"]) {
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+        LERROR(std::format(
+            "Time varying text, '{}' is not a valid JSON file",
+            filePath
+        ));
+        return;
+    }
+
+    nlohmann::json jsonData;
+    file >> jsonData;
+
```
alexanderbock, line 126:
```suggestion

    if (jsonData.find("data") != jsonData.end()) {
        throw ghoul::RuntimeError(std::format(
            "Error loading JSON file. No 'data' was found in '{}', filePath
        ));
    }

```

modules/base/dashboard/dashboarditemtimevaryingtext.h
```diff
@@ -0,0 +1,62 @@
...
+
+     properties::StringProperty _textBefore;
+     properties::StringProperty _dataFile;
+
+     std::unordered_map<double, double> _data;
+     std::vector<double> _startTimes;
+
+     int _activeTriggerTimeIndex = -1;
+     double _sequenceEndTime = 0.0;
+     mutable double _lastValue = 0.0;
```
alexanderbock, line 58:
```suggestion
```

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+void DashboardItemTimeVaryingText::computeSequenceEndTime() {
+    if (_startTimes.size() > 1) {
+        double first = _startTimes.front();
+        double last = _startTimes.back();
+        double avgDuration = (last - first) / static_cast<double>(_startTimes.size() - 1);
+        _sequenceEndTime = last + avgDuration;
+    }
+}
```
alexanderbock, line 149:
```suggestion
    if (_startTimes.size() > 1) {
        double first = _startTimes.front();
        double last = _startTimes.back();
        double avgDuration = (last - first) / static_cast<double>(_startTimes.size() - 1);
        _sequenceEndTime = last + avgDuration;
    }
}

```

alexanderbock: Why is the average duration added to the end end time?  That should be documented somewhere

alexanderbock (changes requested):

alexanderbock (approved):

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
lundkvistarn: I added a comment for it. It's added so that the last value is shown for the average duration a value gets shown and not disappearing instantly.

modules/base/dashboard/dashboarditemtimevaryingtext.cpp
```diff
@@ -0,0 +1,186 @@
...
+        "with the values read from the file provided in `DataFile`",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo DataFileInfo = {
+        "DataFile",
+        "Data File Path",
+        "The file path to the JSON data. The JSON must contain a 'data' array with "
+        "timestamp-value pairs. Example format: "
+        "{\"data\": [[\"2024-05-10T00:00:00Z\", 2.33], [\"2024-05-10T03:00:00Z\", 3.0]]}",
```
alexanderbock, line 51: I'd suggest to instead put this into the description of the class as a while (before the `[[codegen::dictionary]]` part as part of a bigger description what the _class_ does

PR #3628, opened 2025-04-30
Move profile name from top bar center to File menu 

WeirdRubberDuck:
Move the profile name away from the top bar, to be less visually intrusive, and add some more information about the profile, such as the description, file path, author, and license. 

I skipped showing the version in the UI as of now, thinking that this might be more confusing than useful. I also updated the profile description so they all end with punctuation, as it looked weird when they didn't. 

Uses this UI PR: OpenSpace/OpenSpace-WebGui#133

This is what it looks like: 
image

### DIFF
--- a/data/profiles/missions/juno.profile
+++ b/data/profiles/missions/juno.profile
@@ -48,7 +48,7 @@
   ],
   "meta": {
     "author": "OpenSpace Team",
-    "description": "A work in progress profile for juno.",
+    "description": "A work in progress profile for Juno.",
     "license": "MIT License",
     "name": "Juno",
     "url": "https://www.openspaceproject.com",
--- a/data/profiles/missions/juno.profile
+++ b/data/profiles/missions/juno.profile
@@ -48,7 +48,7 @@
   ],
   "meta": {
     "author": "OpenSpace Team",
-    "description": "work in progress scene for juno",
+    "description": "A work in progress profile for juno.",
     "license": "MIT License",
     "name": "Juno",
     "url": "https://www.openspaceproject.com",
--- a/data/profiles/missions/messenger.profile
+++ b/data/profiles/missions/messenger.profile
@@ -42,7 +42,7 @@
   ],
   "meta": {
     "author": "OpenSpace Team",
-    "description": "This scene contains model and trajectory of the NASA MESSENGER spacecraft with craft pointing data from 2011-03 to 2011-06. In addition, a rendering of Mercury's magnetosphere based on data recorded by MESSENGER can be enabled and viewed around the planet. Along with the mission data, additional maps were added to Mercury showing mineral abundances on the surface and a multi-color mosaic from the MDIS instrument",
+    "description": "This scene contains model and trajectory of the NASA MESSENGER spacecraft with craft pointing data from 2011-03 to 2011-06. In addition, a rendering of Mercury's magnetosphere based on data recorded by MESSENGER can be enabled and viewed around the planet. Along with the mission data, additional maps were added to Mercury showing mineral abundances on the surface and a multi-color mosaic from the MDIS instrument.",
     "license": "MIT License",
     "name": "Messenger",
     "url": "https://www.openspaceproject.com",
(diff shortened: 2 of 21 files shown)


WeirdRubberDuck:
FYI I plan to update the UI hash in this PR for easier testing, but had to wait until the UI is finished building

data/profiles/missions/juno.profile
```diff
@@ -48,7 +48,7 @@
   ],
   "meta": {
     "author": "OpenSpace Team",
-    "description": "work in progress scene for juno",
+    "description": "A work in progress profile for juno.",
```
alexanderbock, line 51:
```suggestion
    "description": "A work in progress profile for Juno.",
```

data/profiles/missions/messenger.profile
```diff
@@ -42,7 +42,7 @@
   ],
   "meta": {
     "author": "OpenSpace Team",
-    "description": "This scene contains model and trajectory of the NASA MESSENGER spacecraft with craft pointing data from 2011-03 to 2011-06. In addition, a rendering of Mercury's magnetosphere based on data recorded by MESSENGER can be enabled and viewed around the planet. Along with the mission data, additional maps were added to Mercury showing mineral abundances on the surface and a multi-color mosaic from the MDIS instrument",
+    "description": "This scene contains model and trajectory of the NASA MESSENGER spacecraft with craft pointing data from 2011-03 to 2011-06. In addition, a rendering of Mercury's magnetosphere based on data recorded by MESSENGER can be enabled and viewed around the planet. Along with the mission data, additional maps were added to Mercury showing mineral abundances on the surface and a multi-color mosaic from the MDIS instrument.",
```
alexanderbock, line 45:
```suggestion
    "description": "This scene contains model and trajectory of the NASA MESSENGER spacecraft with pointing data from 2011-03 to 2011-06. In addition, a rendering of Mercury's magnetosphere based on data recorded by MESSENGER can be enabled and viewed around the planet. Along with the mission data, additional maps were added to Mercury showing mineral abundances on the surface and a multi-color mosaic from the MDIS instrument.",
```

modules/server/src/topics/profiletopic.cpp
```diff
@@ -36,13 +37,25 @@ bool ProfileTopic::isDone() const {
...
-
-    const nlohmann::json data = {
+    // @TODO (2025-04-30, emmbr) If we expose the json converters from profile.cpp, we
+    // could use those here instead and minimize the risk of getting the serialization of
+    // the data out of sync
+    nlohmann::json data = {
         { "uiPanelVisibility", global::profile->uiPanelVisibility },
         { "markNodes", global::profile->markNodes },
-        { "name", name }
+        { "filePath", global::configuration->profile}
```
alexanderbock, line 46:
```suggestion
        { "filePath", global::configuration->profile }
```

modules/server/src/topics/profiletopic.cpp
```diff
@@ -36,13 +37,25 @@ bool ProfileTopic::isDone() const {
...
     };
+
+    if (global::profile->meta.has_value()) {
+        data["name"] = global::profile->meta->name.value_or("");
```
alexanderbock, line 50: Does it make it easier on the Typescript side if the value always exist but might be empty?
Compared to:

```cpp
if (global::profile->meta->name.has_value()) {
    data["name"] = *global::profile->meta->name;
}
```

etc

alexanderbock (approved):

modules/server/src/topics/profiletopic.cpp
WeirdRubberDuck: For the typescript side, it doesn't really matter. But since these are very "string-y" values, I personally think it could be nice if they are initialized with empty string, to avoid having to keep track of whether a license was set or not. 

However, what I would like us to do is to move all the JSON serialization code to one place, so that we could use the same serialization for the web UI as for the saving of the profile, for example. Hence the TODO-cmment at the top of this function. Now we have JSON serialization of the same thing in a few places, with a slightly different formatting. 

But that felt like too much work for this PR

WeirdRubberDuck: The drawback of that would be that if we change the format of the profile, the UI code would also have to change, however.

PR #3634, opened 2025-05-01
Issue/3596 saturn ring shadow

### LINKED ISSUE #3596: Saturn rings: shadow issue for points in time
For some points in time there are artifacts for the shadows on the rings of Saturn.
May be related to the angle between the rings and the light source.

Issue:
Image

Working:
Image

benpm:
Fixes #3596 by discarding fragments in the ring shadow map shader when normals are perpendicular to the light source. This removes the artifacts while maintaining the shadow cast by the rings onto the surface of the Saturn globe geometry.

### DIFF
--- a/modules/globebrowsing/shaders/rings_geom_fs.glsl
+++ b/modules/globebrowsing/shaders/rings_geom_fs.glsl
@@ -41,7 +41,7 @@ Fragment getFragment() {
   float radius = length(st);
 
   // Discard if normal is perpendicular to the camera direction
-  if (abs(dot(vs_normal, vec3(0.0, 0.0, 1.0))) <= 1.0e-3) {
+  if (abs(dot(normalize(vs_normal), vec3(0.0, 0.0, 1.0))) <= 1.0e-3) {
     discard;
   }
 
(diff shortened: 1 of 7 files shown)


modules/globebrowsing/shaders/rings_geom_fs.glsl
```diff
@@ -39,6 +40,11 @@ Fragment getFragment() {
   // The length of the texture coordinates vector is our distance from the center
   float radius = length(st);
 
+  // Discard if normal is perpendicular to the camera direction
+  if (abs(dot(normalize(vs_normal), vec3(0.0, 0.0, 1.0))) <= 1.0e-3) {
```
alexanderbock, line 44:
```suggestion
  if (abs(dot(normalize(vs_normal), vec3(0.0, 0.0, 1.0))) <= 0.01) {
```

Changing it to 1e-2 to prevent more issues (for example 2025 APR-20 22:00:00
image


Changing it to `0.01` to make it easier to read

benpm:
seems to not build on CI because the branch doesn't exist in the upstream. im just gonna go ahead and merge it though.

PR #3638, opened 2025-05-05
Cleaning up the code for the setProperty value functions, add better documentation, and add the ability to combine tags

### LINKED ISSUE #1755: Add option to combine multiple tags in `setPropertyValue`
`setPropertyValue` only supports one tag. It would be nice to be able to give it multiple tags and get all object that has all of the tags.

Example: Currently we only have tags for either all minor moons' trails (`moonTrail_minor`) or the trails of all the moons of a planet (e.g. `moonTrail_jupiter`). Combining these (through intersection) would give us all the trails that are both Jupiter moons and minor moons. 

A benefit with this is that it would allow us to re-use tags and don't have to create an abundance of new tags for all these specific groupings.

### LINKED ISSUE #3260: setPropertyValue and setPropertyValueSingle are missing the parameters in the documentation
These functions are missing their parameters in the signature https://docs.openspaceproject.com/en/releases-v0.20.0/generated/scriptingApi/openspace.html#setpropertyvalue it seems to stem from the way they are exported in the luaLibrary https://github.com/OpenSpace/OpenSpace/blob/5ecdd67a10ed625c20454d54c3d3a011b6571a4d/src/scene/scene.cpp#L794 and the corresponding inline file https://github.com/OpenSpace/OpenSpace/blob/5ecdd67a10ed625c20454d54c3d3a011b6571a4d/src/scene/scene_lua.inl#L336 Should we rewrite this so it follows other lua functions and so that the documentation is properly added?

### LINKED ISSUE #3588: Code cleanup of the setPropertyValue function
Setting properties is happening a lot and the code has gotten more complicated over time and it is time for a rewrite and a cleanup. Also:  Add more unit tests to make sure that this change doesn't break anything

alexanderbock:
Also adds unit tests to ensure the setPropertyValue/setPropertyValueSingle/getPropertyValue/property functions still work.

Closes #3588. Closes #3260. Closes #1755

### DIFF
--- a/src/scene/scene.cpp
+++ b/src/scene/scene.cpp
@@ -110,7 +110,7 @@ namespace {
             catch (...) {
                 return false;
             }
-            };
+        };
 
         if (value == "true" || value == "false") {
             return openspace::PropertyValueType::Boolean;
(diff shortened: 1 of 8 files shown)


src/scene/scene_lua.inl
```diff
@@ -166,35 +260,31 @@ std::vector<openspace::properties::Property*> findMatchesInAllProperties(
...
-                    // Check that the nodeName fully matches the node in id
-                    else if (nodePos != 0) {
+                const size_t nodePos = uri.find(nodeName);
+                if (nodePos == std::string::npos) {
+                    return;
+                }
+
+                // Check tag
+                if (isGroupMode) {
+                    if (!ownerMatchesGroupTag(prop->owner(), groupName)) {
```
WeirdRubberDuck, line 278: Nitpicky, but this makes it so that this comment and the one below are on the same "reading level", if you know what I mean
```suggestion
                if (isGroupMode) {
                    // Check tag
                    if (!ownerMatchesGroupTag(prop->owner(), groupName)) {
```

src/scene/scene_lua.inl
```diff
@@ -166,35 +260,31 @@ std::vector<openspace::properties::Property*> findMatchesInAllProperties(
...
                         return;
                     }
                 }
-                else {
+                else if (nodePos != 0) {
+                    // Check that the nodeName fully matches the node in id
```
WeirdRubberDuck, line 283:
```suggestion
                    // Check that the node identifier fully matches the node in the uri
```
?

src/scene/scene_lua.inl
```diff
@@ -736,94 +761,53 @@ namespace {
...
-                )
-            );
-        }
-    }
-    // Literal or tag
-    else {
-        propertyName = name;
-        isLiteral = true;
-    }
+    auto [nodeName, propertyName, isLiteral] = parseRegex(name);
```
WeirdRubberDuck, line 764: Same here. These are not the names, but the identifiers. Please update the variable names to make this clear. 

Also, please rename the input parameter from `name` to `regex` or something

tests/test_lua_property.cpp
```diff
@@ -0,0 +1,260 @@
...
+                CHECK(r[0] == p1.uri());
+                CHECK(r[1] == p2.uri());
+            }
+        });
+
+        triggerScriptRun();
+    }
+}
+
+TEST_CASE("Property: Multiple / 2", "[property]") {
```
WeirdRubberDuck, line 172: What do you mean by `/ 2` here?  Please change to a more understandable name :) 

Also, Missing examples with tag, and wildcard in another place than at the end of the uri

src/scene/scene.cpp
```diff
@@ -848,69 +847,112 @@ scripting::LuaLibrary Scene::luaLibrary() {
...
+                    { "uri", "String" }
+                },
+                "String | Number | Boolean | Table",
+                "Returns the value the property, identified by the provided URI. This "
+                "function will provide an error message if no property matching the URI "
+                "is found.",
+                { "src/scene/scene_lua.inl", 495 }
             },
             codegen::lua::HasProperty,
             codegen::lua::PropertyDeprecated,
```
WeirdRubberDuck, line 958: Can't access the code, but the tag and wildcard behavior needs to be added to the `property` function as well. 

And maybe removeSceneGraphNodesFromRegex``?  But seems like that one does not support the tags...?

WeirdRubberDuck (changes requested):
Here's some comments for code changes - also plan to do some testing but moving on for now

tests/test_lua_setpropertyvalue.cpp
```diff
@@ -0,0 +1,2377 @@
...
+        CHECK(p2 == 1.f);
+        CHECK(p21 == 1.f);
+        triggerScriptRun();
+        CHECK(p1 == 2.f);
+        CHECK(p2 == 1.f);
+        CHECK(p21 == 2.f);
+    }
+}
+
+TEST_CASE("SetPropertyValue: Wildcard Basic Multiple / 2", "[setpropertyvalue]") {
```
WeirdRubberDuck, line 481: Same here, what does `/ 2` mean??

tests/test_lua_setpropertyvalue.cpp
```diff
@@ -0,0 +1,2377 @@
...
+        const double v = glm::mix(1.0, 2.0, 0.1);
+        CHECK(p1 > 1.f);
+        CHECK_THAT(p1, Catch::Matchers::WithinAbs(v, 0.05));
+        CHECK(p2 == 1.f);
+        CHECK(p21 > 1.f);
+        CHECK_THAT(p21, Catch::Matchers::WithinAbs(v, 0.05));
+    }
+}
+
+TEST_CASE("SetPropertyValue: Wildcard Interpolation Multiple /2", "[setpropertyvalue]") {
```
WeirdRubberDuck, line 629: And here

tests/test_lua_setpropertyvalue.cpp
```diff
@@ -0,0 +1,2377 @@
...
+        const double v = glm::mix(1.0, 2.0, t);
+        CHECK(p1 > 1.f);
+        CHECK_THAT(p1, Catch::Matchers::WithinAbs(v, 0.075));
+        CHECK(p2 == 1.f);
+        CHECK(p21 > 1.f);
+        CHECK_THAT(p21, Catch::Matchers::WithinAbs(v, 0.075));
+    }
+}
+
+TEST_CASE("SetPropertyValue: Wildcard Easing Multiple /2", "[setpropertyvalue]") {
```
WeirdRubberDuck, line 754: :)

tests/test_lua_setpropertyvalue.cpp
```diff
@@ -0,0 +1,2377 @@
...
+        updateInterpolations(std::chrono::milliseconds(150));
+        triggerScriptRun();
+        CHECK(p1 == 2.f);
+        CHECK(p2 == 1.f);
+        CHECK(q1 == 0.75f);
+        CHECK(p21 == 2.f);
+    }
+}
+
+TEST_CASE("SetPropertyValue: Wildcard PostScript Multiple /2", "[setpropertyvalue]") {
```
WeirdRubberDuck, line 893: :)

WeirdRubberDuck (commented):
Also, looking at the test for wildcard `*` and thinking that it would be good to add some where the wildcard is not at the beginning or end of the URI?

src/scene/scene.cpp
```diff
@@ -848,69 +847,112 @@ scripting::LuaLibrary Scene::luaLibrary() {
...
+in which the parameter is interpolated. Has to be one of "Linear", "QuadraticEaseIn",
+"QuadraticEaseOut", "QuadraticEaseInOut", "CubicEaseIn", "CubicEaseOut", "CubicEaseInOut",
+"QuarticEaseIn", "QuarticEaseOut", "QuarticEaseInOut", "QuinticEaseIn", "QuinticEaseOut",
+"QuinticEaseInOut", "SineEaseIn", "SineEaseOut", "SineEaseInOut", "CircularEaseIn",
+"CircularEaseOut", "CircularEaseInOut", "ExponentialEaseIn", "ExponentialEaseOut",
+"ExponentialEaseInOut", "ElasticEaseIn", "ElasticEaseOut", "ElasticEaseInOut",
+"BounceEaseIn", "BounceEaseOut", "BounceEaseInOut"
+\\param postscript This parameter specifies a Lua script that will be executed once the
+change of property value is completed. If a duration larger than 0 was provided, it is
+at the end of the interpolation. If 0 was provided, the script runs immediately.
```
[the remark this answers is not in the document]
alexanderbock: I kept the last sentence to keep make it unambiguous.  A valid interpretation of 0 would also be that the postscript never runs since "it never finished" an interpolation

src/scene/scene_lua.inl
```diff
@@ -63,98 +63,192 @@
...
+
+        std::string_view t1 = tagToMatch.substr(0, i);
+        auto t1It = std::find(tags.begin(), tags.end(), t1);
+        std::string_view t2 = tagToMatch.substr(i + 1);
+        auto t2It = std::find(tags.begin(), tags.end(), t2);
+        if (t1It != tags.end() && t2It != tags.end()) {
+            return true;
+        }
+    }
+    if (size_t i = tagToMatch.find(Negation);  i != std::string_view::npos) {
```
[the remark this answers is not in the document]
alexanderbock: Yea, I think so regarding the overthinking.

If the tag contains an intersection _and_ a negation, the if statement inside the interaction checking for negation will trigger and throw an exception.  So if we make it to the second if about the negation, we know for sure that the intersection is not there (i added the additional checks inside only to make it more robust against reordering the instructions later)

src/scene/scene_lua.inl
alexanderbock: I'll leave it as it is since the `isGroupMode` check already belongs to the "Check tag" heading to me

tests/test_lua_property.cpp
alexanderbock: It's the second type of test for multiple, but no other useful difference that would be expressable with a simple name.

Bringing it into the same naming scheme as used in other places (https://github.com/OpenSpace/codegen/blob/master/tests/parsing_luawrapper/parsing_luawrapper_fail.cpp#L210).

src/scene/scene.cpp
alexanderbock: The `propertyValue` function doesn't support wildcards and will only ever match a single property value.

The `removeSceneGraphNodesFromRegex` is used in the assetvalidation script afaik, so we'll need that

alexanderbock: Gotcha!  I'll add the tags into that function if I get the time over the weekend

WeirdRubberDuck: Not sure it's a big deal if it doesn't! Especially since it's so concretely named to explicitly use regex

PR #3643, opened 2025-05-06
webbrowser: Fix CEF deprecation notice and crash

dawagner:
According to https://github.com/chromiumembedded/cef/issues/3685, using the so-called "Alloy bootstrap" (which seems to be the default) is deprecated. This causes a warning when starting OpenSpace.

In some setups (e.g. an AppImage build on Ubuntu 22.04), the warning is followed by a fatal error and a crash:

   [0505/213642.653624:FATAL:path_service.cc(265)] Failed to get the path for 1001

Following the migration instructions from the aforementioned ticket solve both the warning and the crash.

### DIFF
--- a/modules/webbrowser/src/cefhost.cpp
+++ b/modules/webbrowser/src/cefhost.cpp
@@ -57,6 +57,9 @@ CefHost::CefHost([[maybe_unused]] const std::string& helperLocation) {
     settings.windowless_rendering_enabled = true;
     attachDebugSettings(settings);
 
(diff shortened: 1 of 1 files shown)


dawagner:
cc @hn-88 , @alexanderbock

ylvaselling (approved):
This is something I've looked at as well, and that is needed for the UI rewrite also. Thanks!

alexanderbock:
Thanks @dawagner !

PR #3651, opened 2025-05-10
a couple of patches for compiling on Mac

hn-88:
1. On Mac, ALSA will not be declared, leading to errors. So, we need to choose coreaudio and not ALSA for APPLE
2. For newer CMake versions, targets with spaces are not allowed, but we can set the OUTPUT_NAME with spaces.


alexanderbock:
Neat! I didn't know that it actually worked that way

PR #3695, opened 2025-06-02
some patches towards building on MacOS

hn-88:
Unfortunately these changes are not sufficient. Apparently <syncstream> is not supported on Mac's CLang. https://github.com/hn-88/OpenSpace-AppleSiliconMac/issues/9

Would you like me to work towards implementing syncstream functionality within an `ifdef` block, @alexanderbock ? If yes, I can work on that, using the links mentioned in the issue link above.

### DIFF
--- a/modules/webbrowser/cmake/patch/cmake/cef_variables.cmake
+++ b/modules/webbrowser/cmake/patch/cmake/cef_variables.cmake
@@ -24,7 +24,9 @@ endif()
 
 # Determine the project architecture.
 if(NOT DEFINED PROJECT_ARCH)
-  if(CMAKE_SIZEOF_VOID_P MATCHES 8)
+  if(OS_MACOSX AND CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
+    set(PROJECT_ARCH "arm64")
+  elseif(CMAKE_SIZEOF_VOID_P MATCHES 8)
     set(PROJECT_ARCH "x86_64")
   else()
     set(PROJECT_ARCH "x86")
@@ -183,6 +185,14 @@ if(OS_LINUX)
     list(APPEND CEF_LINKER_FLAGS
       -m32
       )
+  elseif(PROJECT_ARCH STREQUAL "arm64")
+    # Apple Silicon (ARM64) architecture.
+    list(APPEND CEF_COMPILER_FLAGS
+      -arch arm64
+      )
+    list(APPEND CEF_LINKER_FLAGS
+      -arch arm64
+     )
   endif()
 
   # Standard libraries.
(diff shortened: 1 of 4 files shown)


modules/webbrowser/cmake/patch/cmake/cef_variables.cmake
```diff
@@ -253,15 +263,15 @@ if(OS_MACOSX)
...
+    -std=c17                        # Use the C99 language standard
     )
   list(APPEND CEF_CXX_COMPILER_FLAGS
     #-fno-exceptions                 # Disable exceptions
     #-fno-rtti                       # Disable real-time type information
     -fno-threadsafe-statics         # Don't generate thread-safe statics
     -fobjc-call-cxx-cdtors          # Call the constructor/destructor of C++ instance variables in ObjC objects
     -fvisibility-inlines-hidden     # Give hidden visibility to inlined class member functions
-    -std=gnu++14                    # Use the C++14 language standard including GNU extensions
+    -std=gnu++2a                    # Use the C++14 language standard including GNU extensions
```
alexanderbock:
```suggestion
    -std=gnu++20                    # Use the C++20 language standard including GNU extensions
```

it was called `2a` in GCC 9 and below, which we don't support anyway

modules/webbrowser/cmake/patch/cmake/cef_variables.cmake
```diff
@@ -329,6 +339,9 @@ if(OS_MACOSX)
       set(CEF_BINARY_DIR "${_CEF_ROOT}/$<CONFIG>")
   endif()
 
+  message("CMAKE_OSX_DEPLOYMENT_TARGET is set to ")
+  message(${CMAKE_OSX_DEPLOYMENT_TARGET})
```
alexanderbock:
```suggestion
  message("CMAKE_OSX_DEPLOYMENT_TARGET is set to ${CMAKE_OSX_DEPLOYMENT_TARGET}")
```

This will work as well and only print a single line, which makes it easier to read in the CMake log

alexanderbock:
I pushed to the `master` branch with a fix that is better than reimplementing `syncstream`.  We only used it as a shorthand for not declaring a mutex explicitly, so I went back to implementing the lock guarding by hand

modules/webbrowser/cmake/patch/cmake/cef_variables.cmake
hn-88: Thanks @alexanderbock - I'd read at https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html about "decay" behaviour if a compiler doesn't support a particular flag, hence I used an earlier flag, because XCode typically lags behind in feature support. But it looks like XCode does support -std=gnu++20 from version 14 or so, https://developer.apple.com/documentation/xcode/build-settings-reference

alexanderbock: :+1:  Lets use the `gnu++20` then as it will probably cause less confusion for the next person to look at it

PR #3704, opened 2025-06-09
Issue/3681 missing attached node apollo

### LINKED ISSUE #3681: Apollo profile - GlobeRotation error "Could not find attached node"
Steps to reproduce: 
- Start the Apollo Profile
- Search for and trigger the action "Setup Apollo 11 landing site"
- Observe the continuous flood of errors from `GlobeRotation`

Image

WeirdRubberDuck:
Fixes #3681 and prevents similar issues from happening in the future by refactoring initialization code for transforms. All transforms should now have their `update` and `initialize` functions called correctly, even in timeline- and multi-transforms

The changes include: 
- Fix misnamed error message category for `GlobeTranslation`
- The attached node for the `GlobeTranslation` (and `GlobeRotation`) is now set in the initialize instead of update, and the timeline and multi-transforms correctly initialize all their member transforms. A benefit of this is that we only get the error message once if the attached node is not found
- Correctly call the `update` and `initialize` functions for renderables and transforms that have other transforms as members
- Make `initialize` function for transforms return void instead of bool
- `ScreenSpaceRenderableRenderable` not calls the intilize functions of its transforms and renderable in `initialize` instead of constructor
- Make code style for `GlobeTranslation` and `GlobeRotation` more consistent
- Some other minor code cleanup

### DIFF
--- a/modules/base/rendering/renderabletrail.cpp
+++ b/modules/base/rendering/renderabletrail.cpp
@@ -244,7 +244,6 @@ RenderableTrail::RenderableTrail(const ghoul::Dictionary& dictionary)
     }
 
     addPropertySubOwner(_appearance);
-
 }
 
 void RenderableTrail::initialize() {
@@ -299,7 +298,10 @@ glm::dvec3 RenderableTrail::translationPosition(Time time) const {
     // Use an empty modelTransform to get the position in local coordinates. Also, the
     // previous frame time does not matter so we just use time 0
     const UpdateData data = { {}, time, Time(0.0) };
-    _translation->update(data);
+    // @TODO (2025-06-09, emmbr) Do we need to also call the update for the translation?
+    // As of now, this borks up the performance of the trail completely, and I'm not sure
+    // the update is needed since we are not using the cached version of the position.
+    //_translation->update(data);
     return _translation->position(data);
 }
 
--- a/modules/base/rendering/renderabletrail.cpp
+++ b/modules/base/rendering/renderabletrail.cpp
@@ -247,6 +247,7 @@ RenderableTrail::RenderableTrail(const ghoul::Dictionary& dictionary)
 }
 
 void RenderableTrail::initialize() {
+    Renderable::initialize();
     _translation->initialize();
 }
 
--- a/modules/base/rendering/renderabletrail.cpp
+++ b/modules/base/rendering/renderabletrail.cpp
@@ -295,6 +295,14 @@ bool RenderableTrail::isReady() const {
     return _programObject != nullptr;
 }
 
+glm::dvec3 RenderableTrail::translationPosition(Time time) const {
+    // Use an empty modelTransform to get the position in local coordinates. Also, the
+    // previous frame time does not matter so we just use time 0
+    const UpdateData data = { {}, time, Time(0.0) };
+    _translation->update(data);
+    return _translation->position(data);
+}
+
 void RenderableTrail::internalRender(bool renderLines, bool renderPoints,
                                      const RenderData& data,
                                      const glm::dmat4& modelTransform,
--- a/modules/base/rendering/renderabletrail.cpp
+++ b/modules/base/rendering/renderabletrail.cpp
@@ -247,6 +247,10 @@ RenderableTrail::RenderableTrail(const ghoul::Dictionary& dictionary)
 
 }
 
+void RenderableTrail::initialize() {
+    _translation->initialize();
+}
+
 void RenderableTrail::initializeGL() {
     ZoneScoped;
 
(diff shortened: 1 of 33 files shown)


modules/base/rendering/renderabletrail.cpp
```diff
@@ -291,6 +295,17 @@ bool RenderableTrail::isReady() const {
...
+glm::dvec3 RenderableTrail::translationPosition(Time time) const {
+    // Use an empty modelTransform to get the position in local coordinates. Also, the
+    // previous frame time does not matter so we just use time 0
+    const UpdateData data = { {}, time, Time(0.0) };
+    // @TODO (2025-06-09, emmbr) Do we need to also call the update for the translation?
+    // As of now, this borks up the performance of the trail completely, and I'm not sure
+    // the update is needed since we are not using the cached version of the position.
+    //_translation->update(data);
+    return _translation->position(data);
+}
```
WeirdRubberDuck: @alexanderbock @Arohdin I ended up not calling the update function here, since it's no longer needed when the initialization is being done correctly. 

Would be good with a second pair of eyes on this, but calling the update here completely borks up the performance for the trails. And, if I understand the update structure and caching of the transform positions correctly, we shouldn't actually need to call the update as long as we are getting the position from the non-cached version of the position function (which is the one used here, that takes the `UpdateData` struct). Do you agree with this?

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -185,11 +185,8 @@ void RenderableTrailOrbit::update(const UpdateData& data) {
 
     // 2
     // Write the current location into the floating position
-    const glm::vec3 p = _translation->position({
-        {},
-        data.time,
-        Time(0.0)
-    });
+    const glm::vec3 p = translationPosition(data.time);
+
```
WeirdRubberDuck:
```suggestion
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -203,10 +203,10 @@ RenderableTrailTrajectory::RenderableTrailTrajectory(const ghoul::Dictionary& di
     _timeStampSubsamplingFactor.onChange([this] { _subsamplingIsDirty = true; });
     addProperty(_timeStampSubsamplingFactor);
 
-    _numberOfReplacementPoints = p.accurateTrailPositions.value_or(
-        _numberOfReplacementPoints
+    _nReplacementPoints = p.accurateTrailPositions.value_or(
+        _nReplacementPoints
     );
```
WeirdRubberDuck:
```suggestion
    _nReplacementPoints = p.accurateTrailPositions.value_or(_nReplacementPoints);
```

modules/base/rendering/renderabletrail.cpp
alexanderbock: Yea, the `update` should store the cached version for the `postition()` function.  Ran into the same issue on my version of the branch and also left out the `update` call.

Having a comment here would be good though so that we notice where issues are when we introduce a Translation subclass that actually needs the update called

modules/base/rendering/renderabletrail.h
```diff
@@ -115,6 +116,15 @@ class RenderableTrail : public Renderable {
 protected:
     explicit RenderableTrail(const ghoul::Dictionary& dictionary);
 
+    /**
+     * Get the trail position for a given time from the translation object. The position
```
alexanderbock:
```suggestion
     * Get the trail position for a given time from the Translation object. The position
```

modules/base/rendering/renderabletrailtrajectory.cpp
```diff
@@ -258,10 +256,8 @@ void RenderableTrailTrajectory::reset() {
 
 void RenderableTrailTrajectory::update(const UpdateData& data) {
     if (_needsFullSweep) {
-
         if (_sweepIteration == 0) {
-            // Max number of vertices
-            constexpr unsigned int maxNumberOfVertices = 1000000;
+            constexpr unsigned int maxVertices = 1000000;
```
alexanderbock:
```suggestion
            constexpr unsigned int MaxVertices = 1000000;
```

While we're already in here

modules/base/rendering/renderabletrail.cpp
WeirdRubberDuck: 👍 Is the current comment ok or should it be rephrased? Feel free to come with a suggestion

alexanderbock: Yea, I think the comment is great. Just didn't want my message to read as if the comment should disappear

WeirdRubberDuck: I guess it could be rephrased to not sound as much like a question, but more "if we add a translation type that needs updating before the position is computed, we need to call the update here" :)

Arohdin (approved):

PR #3708, opened 2025-06-10
A few random fixes I stumbled upon

k2d222:
It's a collection small patches I applied on my branch and thought deserved to be merged.

Commits should be self contained, I'm open to splitting the PR or dropping commits.

Below a couple explanations.

## Updated .gitignore

* compile_commands.json (Clang compilation database) and the .cache/ folder are used by the clangd LSP in several IDEs.
* GPUCache/ folder is generated on certain platforms by Chromium (cef webgui)

## Added fov input for fisheye projections in launcher

I use a spherical display with a 310 degrees fisheye. I added a fov input for fisheye projections. (which I believe existed at some point, though I couldn't bisect the removal)

## ~~Removed static_assert(false, "unimplemented") in non-specialized template functions~~

~~The `static_assert` was triggered by my compiler (visual studio 17 2022?). I suspect it might be a compiler bug, because the template should not be instanciated and cppreference says "[when the static_assert] is evaluated in the context of a template definition and the template is uninstantiated, this declaration has no effect". Removing the default implementation fixes it, however the error message if if failed would not be nice.~~ (Reverted)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ cmake-build-*
 cmake_install.cmake
 install_manifest.txt
 Makefile
+compile_commands.json
+.cache/
+GPUCache/
 
 # OS specific things
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.dir
 *.idea/
 .build-vs/
+.cache/
 .cproject
 .project
 .vs/
@@ -13,11 +14,9 @@ CMakeFiles
 CMakeLists.txt.user
 cmake-build-*
 cmake_install.cmake
+compile_commands.json
 install_manifest.txt
 Makefile
-compile_commands.json
-.cache/
-GPUCache/
 
 # OS specific things
 .DS_Store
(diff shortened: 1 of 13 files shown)


k2d222:
Update on the `static_assert` commit: I troubleshooted the issue and found that updating VSCommunity fixed the bug. Before that I tried many compilation variants and none passed the assert -- **it was likely a MSVC bug that is now fixed**.


Old toolchain:
```
1> [CMake] -- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.26100.
1> [CMake] -- The C compiler identification is MSVC 19.38.33143.0
1> [CMake] -- The CXX compiler identification is MSVC 19.38.33143.0
```

New, working toolchain:
```
1> [CMake] -- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.26100.
1> [CMake] -- The C compiler identification is MSVC 19.44.35209.0
1> [CMake] -- The CXX compiler identification is MSVC 19.44.35209.0
```

alexanderbock:
I undid the delayedvariable change for now, but lets have a coffee to talk about it!

PR #3715, opened 2025-06-13
Feature/WSAdocs

ElonOlsson:
More / different documentation for the WSA project.
* Compact description in mission panel. ( It's been chatGPTed and I kept the original text, so I could extend it if not enough as is)
* Documentation above parameter struct in RenderableTimeVaryingFitsSphere and RenderableFieldlinesSequence, focused on what parameters are required when.
* A bug fix where colorMinMaxRange was used instead of the selected color range

Descriptions that were in the mission panel before, will be moved into the docs repository for more detailed documentation and description

### DIFF
--- a/data/assets/scene/solarsystem/heliosphere/todayssun/mission.asset
+++ b/data/assets/scene/solarsystem/heliosphere/todayssun/mission.asset
@@ -4,35 +4,15 @@ local Mission = {
   Name = "The WSA Simulation Model",
   TimeRange = { Start = "2022 DEC 09 16:14:00", End = timeNow },
   Description = [[
-    This profile shows the Sun and its magnetic polarities, using data from the WSA
-    (Wang-Sheeley-Arge) simulation model version 5.4. The solar coronal portion of WSA is
-    comprised of two potential field type models. The inner model is Potential Field
-    Source Surface (PFSS) which specifies the coronal field from the inner, photospheric
-    boundary at 1 solar radii (Rs) to its outer boundary or source surface at 2.5Rs. The
-    outer model is the Schatten Current Sheet (SCS) model. The radial magnetic field
-    components of the PFSS magnetic field solution at 2.5Rs are used as the inner boundary
-    condition to the SCS model. All data is downloaded dynamically, bringing you the
-    latest data from the model. You can find these assets in the Scene menu under
-    "Solar System/Heliosphere". Under the "WSA Coronal Model" submenu, you will find extra
-    options for the solar field lines and surface features. The four different sets of
-    field lines are all following the magnetic fields but traced from different locations.
-    They can be colored by their different parameters in the data, for example polarity or
-    open or closed field lines. The two elements named "Solar Surface" show the magnetic
-    polarities, but have other options in the settings under "Texture Layer Options". The
-    same goes for the two named "Velocity at Outer Boundary". However, the two named
-    "Magnetic Field at 5Rs" do not have other options. Each asset includes settings for
-    appearance. If field lines are looking too bright against a bright background, try
-    changing the setting for additive blending. Additionally there is an option to save
-    the downloaded data for future use (by default, the data is not cached). Each data
-    file remains active until the next one in the sequence becomes available. To see which
-    file is currently being used, check the "Texture Source" field for each asset in the
-    GUI. In the visualization, the white line shooting out from the Sun, along with the
-    white arc, always points in the direction of Earth. The red arc is the Carrington
-    prime meridian which is longitude 0 of the Sun, and the green grid shows lines with a
-    10-degree separation. If no data is showing, most likely the selected time is outside
-    of the data range (e.g., too far in the future or the past for the dynamic data to be
-    available). Alternatively, the files failed to download. For more information on the
-    simulation model, see https://ccmc.gsfc.nasa.gov/models/WSA~5.4.
+  This visualization shows the Sun's magnetic field using data from the WSA
+  (Wang-Sheeley-Arge) simulation model. The field lines trace the Sun's magnetic field.
+  Some loop back into the Sun (closed, shown in yellow), while others stream into space
+  (open), colored by their position relative to the heliospheric current sheet. On the
+  Sun's surface, white and black indicate areas of positive and negative magnetic
+  polarity, respectively. Understanding these fields is critical: solar activity like
+  Coronal Mass Ejections (CMEs) can disrupt satellites, power grids, and pose risks to
+  astronauts and spacecraft throughout the solar system. All data is fetched live, giving
+  you the latest view of our dynamic Sun.
   ]],
   Milestones = {
     {
--- a/modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
+++ b/modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
@@ -387,7 +387,7 @@ namespace {
         std::optional<std::vector<std::string>> extraVariables;
 
         // Which variable in CDF file that represents which vector field in the data to
-        // trace fieldlines in.
+        // trace field lines in.
         std::optional<std::string> tracingVariable;
 
         // Decides whether or not to cache the downloaded data between runs. By default,
--- a/modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
+++ b/modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
@@ -72,9 +72,10 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo ColorMinMaxInfo = {
-        "ColorQuantityMinMax",
-        "ColorTable Min Value",
-        "Value to map to the lowest and highest end of the color table.",
+        "ColorMinMaxRange",
+        "Color Min Max Range",
+        "A min-max range for what the lowest and highest values of the color table can "
+        "be set to.",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
@@ -150,7 +151,7 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FlowReversedInfo = {
-        "Reversed",
+        "ReversedFlow",
         "Reversed Flow",
         "Toggle to make the flow move in the opposite direction.",
         openspace::properties::Property::Visibility::User
@@ -171,7 +172,7 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FlowSpeedInfo = {
-        "Speed",
+        "FlowSpeed",
         "Speed",
         "Speed of the flow.",
         openspace::properties::Property::Visibility::User
@@ -223,6 +224,26 @@ namespace {
         openspace::properties::Property::Visibility::User
     };
 
+    // This `Renderable` visualizes field lines, mainly a sequence of time steps but work
+    // with only one time step too. A sequence is a data source consisting of multiple
+    // data files that each correspond to a specific time, and is therefor time varying
+    // like the name of the renderable suggests.
+    //
+    // `LoadingType` can be specified in two ways;
+    //
+    // `StaticLoading`: In this case all data is loaded when starting OpenSpace. A
+    // `SourceFolder` is then mandatory. The data format is also required to be set with
+    // `InputFileType`.
+    //
+    // `DynamicDownloading`: This case downloads the data during run time. For this a few
+    // parameters are required: `InfoURL`, `DataURL` and `DataID` are used to specify the
+    // data source to use.
+    //
+    // When using CDF data `SeedPointDirectory` is required. Some prior knowledge of the
+    // data is needed to use it in this way. `TracingVariable` needs to be set and
+    // `ExtraVariables` will have to match what parameters are in the CDF data. Using CDF
+    // directly in this `Renderable` is not recommended. Rather use
+    // `KameleonVolumeToFieldlinesTask` first, to save the data to .osfls or .json
     struct [[codegen::Dictionary(RenderableFieldlinesSequence)]] Parameters {
         enum class [[codegen::map(openspace::RenderableFieldlinesSequence::ColorMethod)]]
         ColorMethod {
@@ -247,7 +268,7 @@ namespace {
         // colorized by. Should be entered as min value, max value
         std::optional<std::vector<glm::vec2>> colorTableRanges;
 
-        // Specifies the total data range to where color map will be applied
+        // [[codegen::verbatim(ColorMinMaxInfo.description)]]
         std::optional<glm::vec2> colorMinMaxRange;
 
         // [[codegen::verbatim(FlowEnabledInfo.description)]]
@@ -275,7 +296,7 @@ namespace {
         std::optional<int> maskingQuantity;
 
         // Ranges for which their corresponding quantity parameter value will be
-        // masked by. Should be entered as a min value, max value pair.
+        // masked by. Should be entered as a list of min value, max value pairs.
         std::optional<std::vector<glm::vec2>> maskingRanges;
 
         // Ranges for which their corresponding parameters values will be
@@ -291,8 +312,9 @@ namespace {
         // [[codegen::verbatim(ColorUseABlendingInfo.description)]]
         std::optional<bool> alphaBlendingEnabled;
 
-        // Set if first/last file should render forever.
-        bool showAtAllTimes;
+        // Set if first/last file should render when time is outside of the sequence
+        // interval. This can be used regardless of LoadingType.
+        std::optional<bool> showAtAllTimes;
 
         // If data sets parameter start_time differ from start of run,
         // elapsed_time_in_seconds might be in relation to start of run.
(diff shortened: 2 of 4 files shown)


modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -150,7 +151,7 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FlowReversedInfo = {
-        "Reversed",
+        "ReversedFlow",
```
alexanderbock, line 154: This is a breaking change that we need to make sure that none of our assets are using the old version and it needs to be in the breaking changes list of the internal wiki

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -171,7 +172,7 @@ namespace {
     };
 
     constexpr openspace::properties::Property::PropertyInfo FlowSpeedInfo = {
-        "Speed",
+        "FlowSpeed",
```
alexanderbock, line 175: This is a breaking change that we need to make sure that none of our assets are using the old version and it needs to be in the breaking changes list of the internal wiki

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
         openspace::properties::Property::Visibility::User
     };
 
+    // This `Renderable` visualizes field lines, mainly a sequence of time steps but work
```
alexanderbock:
```suggestion
    // This `Renderable` visualizes field lines, mainly a sequence of time steps but works
```

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
...
+    // with only one time step too. A sequence is a data source consisting of multiple
```
alexanderbock:
```suggestion
    // with only one time step, too. A sequence is a data source consisting of multiple
```

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
...
+    // data files that each correspond to a specific time, and is therefor time varying
```
alexanderbock:
```suggestion
    // data files that each correspond to a specific time and is therefore time varying
```

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
...
+    // like the name of the renderable suggests.
+    //
+    // `LoadingType` can be specified in two ways;
+    //
+    // `StaticLoading`: In this case all data is loaded when starting OpenSpace. A
+    // `SourceFolder` is then mandatory. The data format is also required to be set with
+    // `InputFileType`.
+    //
+    // `DynamicDownloading`: This case downloads the data during run time. For this a few
```
alexanderbock:
```suggestion
    // `DynamicDownloading`: This case downloads the data during runtime. For this, a few
```

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
...
+    // parameters are required: `InfoURL`, `DataURL` and `DataID` are used to specify the
+    // data source to use.
+    //
+    // When using CDF data `SeedPointDirectory` is required. Some prior knowledge of the
+    // data is needed to use it in this way. `TracingVariable` needs to be set and
+    // `ExtraVariables` will have to match what parameters are in the CDF data. Using CDF
+    // directly in this `Renderable` is not recommended. Rather use
```
alexanderbock:
```suggestion
    // directly in this `Renderable` is not recommended. Rather use the
```

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
...
+    // `KameleonVolumeToFieldlinesTask` first, to save the data to .osfls or .json
```
alexanderbock:
```suggestion
    // `KameleonVolumeToFieldlinesTask` task first, to save the data to .osfls or .json
```

alexanderbock: I know its double, but technically `KameleonVolumeToFieldlinesTask` is the name of the task

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -291,8 +312,9 @@ namespace {
         // [[codegen::verbatim(ColorUseABlendingInfo.description)]]
         std::optional<bool> alphaBlendingEnabled;
 
-        // Set if first/last file should render forever.
-        bool showAtAllTimes;
+        // Set if first/last file should render when time is outside of the sequence
+        // interval. This can be used regardless of LoadingType.
```
alexanderbock:
```suggestion
        // interval. This can be used regardless of LoadingType. If this value is not
        // specified, the field lines are shown at all times.
```

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -573,6 +596,7 @@ RenderableFieldlinesSequence::RenderableFieldlinesSequence(
 
     if (p.colorTableRanges.has_value()) {
         _colorTableRanges = *p.colorTableRanges;
+        _selectedColorRange = _colorTableRanges[_colorQuantityTemp];
```
alexanderbock: Is `_colorQuantityTemp` guaranteed to be >0 and < _colorTableRanges.size() here?  If not, it will lead to a hard crash

modules/fitsfilereader/src/renderabletimevaryingfitssphere.cpp
```diff
@@ -125,6 +127,30 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    // This `Renderable` reads a data sequence from specifically FITS files and makes
+    // textures from them and wraps them onto a sphere. A sequence is a data source
+    // consisting of multiple data files that each correspond to a specific
+    // time, and is therefor time varying like the name of the renderable suggests.
```
alexanderbock:
```suggestion
    // time and is therefore time varying like the name of the renderable suggests.
```

modules/fitsfilereader/src/renderabletimevaryingfitssphere.cpp
```diff
@@ -125,6 +127,30 @@ namespace {
...
+    //
+    // `StaticLoading`: In this case, parameter `TextureSource` is expected to be provided
+    // to specify where the data is. Using this, all data will be loaded on start up.
+    // Other properties required to be provided is at least one entry in the list of
+    // `LayerNames` that assigns a name to the index corresponding to which layer in the
+    // FITS data that will be used. Additionally, a data range for each of these entries
+    // is required to be assigned in `LayerMinMaxCapValues` in a similar manner as
+    // `LayerNames`.
+    //
+    // `DynamicDownloading`: This case downloads the data during run time. In addition to
```
alexanderbock:
```suggestion
    // `DynamicDownloading`: This case downloads the data during runtime. In addition to
```

modules/fitsfilereader/src/renderabletimevaryingfitssphere.cpp
```diff
@@ -166,11 +192,13 @@ namespace {
...
         // This is set to false by default and will delete all the downloaded content when
-        // OpenSpace is shut down. Set to true to save all the downloaded files.
+        // OpenSpace is shut down, if using dynamic downloading. Set to true to save all
+        // the downloaded files.
         std::optional<bool> cacheData;
 
-        // Set if first/last file should render outside of the sequence interval.
-        std::optional<bool> showPastFirstAndLastFile;
+        // Set if first/last file should render when time is outside of the sequence
+        // interval. This can be used regardless of LoadingType.
```
alexanderbock: Same change as above

alexanderbock (approved):

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
ElonOlsson: I checked all our fieldline assets and profiles, and it seems okay.
This stayed the same:
```
        // [[codegen::verbatim(FlowReversedInfo.description)]]
        std::optional<bool> reversedFlow;
```
With this identifier name change in the PropertyInfo i put in the breaking change list something like this:

> `RenderableFieldlinesSequence`s PropertyInfo identifier `Reversed` renamed to `ReversedFlow` to match input parameter name.

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
ElonOlsson: This was used in Juice profile for GanymedeMagnetosphere together with DomainEnabled and FlowEnabled.
Name change + that DomainEnabled and FlowEnabled is noted in release notes, and edited in profile.

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
ElonOlsson: Good catch. This has never worked super well. We have talked about it needing a better overhaul. Right now it will work if only one specified or if a range for all available parameter variables in the data are specified, at least up to the selected  index given with ColorQuantity. Fixed so that it at least checks for these now. Like you suspected, this crashed the static Earth magnetosphere asset

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
alexanderbock: Would say:

The `RenderableFieldlinesSequence`s `Reversed` property was renamed to `ReversedFlow` to match the input parameter name

data/assets/scene/solarsystem/heliosphere/todayssun/mission.asset
```diff
@@ -4,35 +4,15 @@ local Mission = {
...
-    file remains active until the next one in the sequence becomes available. To see which
-    file is currently being used, check the "Texture Source" field for each asset in the
-    GUI. In the visualization, the white line shooting out from the Sun, along with the
-    white arc, always points in the direction of Earth. The red arc is the Carrington
-    prime meridian which is longitude 0 of the Sun, and the green grid shows lines with a
-    10-degree separation. If no data is showing, most likely the selected time is outside
-    of the data range (e.g., too far in the future or the past for the dynamic data to be
-    available). Alternatively, the files failed to download. For more information on the
-    simulation model, see https://ccmc.gsfc.nasa.gov/models/WSA~5.4.
+  This visualization shows the Sun's magnetic field using data from the WSA
```
WeirdRubberDuck: Add one level of indentation

data/assets/scene/solarsystem/heliosphere/todayssun/mission.asset
```diff
@@ -4,35 +4,15 @@ local Mission = {
...
+  (Wang-Sheeley-Arge) simulation model. The field lines trace the Sun's magnetic field.
+  Some loop back into the Sun (closed, shown in yellow), while others stream into space
+  (open), colored by their position relative to the heliospheric current sheet. On the
```
WeirdRubberDuck: "colored by their position relative to the heliospheric current sheet" is a bit difficult to understand here. 

I managed to Google and find out what the sheet is, but what does red VS blue mean?

modules/fieldlinessequence/rendering/renderablefieldlinessequence.cpp
```diff
@@ -223,6 +224,26 @@ namespace {
...
+    //
+    // `LoadingType` can be specified in two ways;
+    //
+    // `StaticLoading`: In this case all data is loaded when starting OpenSpace. A
+    // `SourceFolder` is then mandatory. The data format is also required to be set with
+    // `InputFileType`.
+    //
+    // `DynamicDownloading`: This case downloads the data during runtime. For this, a few
+    // parameters are required: `InfoURL`, `DataURL` and `DataID` are used to specify the
+    // data source to use.
```
WeirdRubberDuck:
```suggestion
    // `LoadingType` can be specified in two ways:
    // 1. `StaticLoading`: In this case, all data is loaded when starting OpenSpace. A
    // `SourceFolder` is then mandatory. The data format is also required to be set with
    // `InputFileType`.
    //
    // 2. `DynamicDownloading`: This case downloads the data during runtime. For this, a few
    // parameters to specify the data source are required: `InfoURL`, `DataURL`, and `DataID`.
```
This will make it a numbered list on the docs page, which is nice for structure. Tried running it just to be sure, and it looks like this:
image

And same below :)

modules/fitsfilereader/src/renderabletimevaryingfitssphere.cpp
```diff
@@ -125,6 +127,30 @@ namespace {
...
+    // `LayerNames` that assigns a name to the index corresponding to which layer in the
+    // FITS data that will be used. Additionally, a data range for each of these entries
+    // is required to be assigned in `LayerMinMaxCapValues` in a similar manner as
+    // `LayerNames`.
+    //
+    // `DynamicDownloading`: This case downloads the data during runtime. In addition to
+    // `LayerNames` and `LayerMinMaxCapValues` more parameters are required to be
+    // specified. `InfoURL`: that will be a first step to give data to 'DataURL` which
+    // is expected to return the list with data files. The `DataID` accompanies the two
+    // URLs to specify which data source to use.
```
WeirdRubberDuck:
```suggestion
    // `LoadingType` can be specified in two ways;
    // 1. `StaticLoading`: In this case, all data will be loaded on startup, from the location 
    // specified in the `TextureSource` parameter. Other required parameters are at least 
    // one entry in the `LayerNames` and `LayerMinMaxCapValues` tables, respectively. 
    // `LayerNames` assigns a name to an index that corresponds to a layer in the
    // FITS data to use. For each entry in `LayerNames`, there should also be a matching 
    // entry in the `LayerMinMaxCapValues` that specifies the min and max range for the data.
    //
    // 2. `DynamicDownloading`: This case downloads the data during runtime. In addition to
    // `LayerNames` and `LayerMinMaxCapValues`, a few more parameters are required in this case: 
    // `InfoURL`: that will be a first step to give data to 'DataURL` which
    // is expected to return the list with data files. The `DataID` accompanies the two
    // URLs to specify which data source to use.
```

? Make it a list, and some text edits.  Found the initial paragraphs a bit hard to read. Note that I have not made sure that the line width is correct


Note that I didn't know how to rewrite this part, but I find it very difficult to understand. You should maybe give it a second look. 
```
    // `InfoURL`: that will be a first step to give data to 'DataURL` which
    // is expected to return the list with data files. The `DataID` accompanies the two
    // URLs to specify which data source to use.
```

modules/fitsfilereader/src/renderabletimevaryingfitssphere.cpp
```diff
@@ -125,6 +127,30 @@ namespace {
...
+    //
+    // In addition, but not required `FitsLayer` can be specified that correspond to one
+    // of the layers in `LayerNames` and `LayerMinMaxCapValues`. `CacheData` and
+    // `ShowPastFirstAndLastFile` are two additional options.
```
WeirdRubberDuck:
```suggestion
    // In addition, but not required, the `FitsLayer` parameter can be specified to use a specific 
    // data layer for the sphere texture. The index should match one of the layers in `LayerNames` and 
    // `LayerMinMaxCapValues`. `CacheData` and  `ShowAtAllTimes` are two other optional parameters.
```

WeirdRubberDuck (approved):
Great!  The mission asset description is huge imrpovement and looks great :)

data/assets/scene/solarsystem/heliosphere/todayssun/mission.asset
```diff
@@ -4,35 +4,19 @@ local Mission = {
...
-    simulation model, see https://ccmc.gsfc.nasa.gov/models/WSA~5.4.
+    This visualization shows the Sun's magnetic field using data from the WSA
+    (Wang-Sheeley-Arge) simulation model. The field lines trace the Sun's magnetic field.
+    Some loop back into the Sun (closed, shown in yellow), while others stream into space
+    (open). The simulation data throughout this visualization is colored red for positive
+    values and blue for negative. Red field lines indicate that the direction of the field
+    is pointing away from the sun and blue that it is pointing inward toward the Sun. The
+    direction of the field is opposite on either side of whats called the heliospheric
+    current sheet. On the Sun's surface, white and black indicate areas of positive and
+    negative magnetic polarity, respectively. Understanding these fields is critical:
```
WeirdRubberDuck, line 15:
```suggestion
    negative magnetic polarity, respectively. Understanding these fields is critical, as
```

WeirdRubberDuck: oops, I was too late!

PR #3721, opened 2025-06-17
Fix `toggleFade` for multiple property owners, and add `openspace.propertyOwner` API function

WeirdRubberDuck:
Summary of changes: 
-  Add a Lua API function to get the identifiers of all property owners matching an expression with tags/wildcards: `openspace.propertyOwner`
- `toggleFade` now works with wildcards and/or tags (as specified in its documentation), utilizing the new Lua API function
- Apply `toggleFade` in actions where relevant. Fixes issues where some actions would not have worked if the object was faded out

- Fix error in `openspace.property` function documentation
- Small cleanup in core_scripts `fadeIn` and `fadeOut`

Should not change existing logic (apart from that some action now fades instead of immediately toggling)

### DIFF
--- a/data/assets/actions/trails/toggle_all_trails.asset
+++ b/data/assets/actions/trails/toggle_all_trails.asset
@@ -38,8 +38,8 @@ local ToggleTrailsInstant = {
   Identifier = "os.ToggleTrailsInstant",
   Name = "Toggle all trails instantly",
   Command = [[
-    openspace.toggleFade("Scene.*Trail.Renderable", 0)
-    openspace.toggleFade("Scene.*trail.Renderable", 0)
+    openspace.toggleFade("Scene.*Trail.Renderable", 0.0)
+    openspace.toggleFade("Scene.*trail.Renderable", 0.0)
   ]],
   Documentation = "Toggle fade instantly for all trails in the Scene",
   GuiPath = "/Trails",


WeirdRubberDuck:
One potential issue (that was there before this as well) is that the `toggleFade` does not apply the same fade state to all matched objects - just inverts the state of each individual match.  Is this something we want to address?

For example, toggling the trails in the default profile leads to the Pluto Keplerian Trail being enabled, which seems a bit weird to me. But maybe we should rather go through the actions and use tags to make sure that they do what we expect them to

https://github.com/user-attachments/assets/3e18b737-4e95-4db0-9deb-5c455196a89b

data/assets/actions/trails/toggle_all_trails.asset
```diff
@@ -72,29 +38,8 @@ local ToggleTrailsInstant = {
...
-      end
-      if (#capList > 0) then
-        openspace.setPropertyValue("Scene.*Trail.Renderable.Fade", newFade)
-      end
-      if (#list > 0) then
-        openspace.setPropertyValue("Scene.*trail.Renderable.Fade", newFade)
-      end
-    end
+    openspace.toggleFade("Scene.*Trail.Renderable", 0)
+    openspace.toggleFade("Scene.*trail.Renderable", 0)
```
alexanderbock:
```suggestion
    openspace.toggleFade("Scene.*Trail.Renderable", 0.0)
    openspace.toggleFade("Scene.*trail.Renderable", 0.0)
```

data/assets/actions/trails/toggle_all_trails_planets_moons_instant.asset
```diff
@@ -2,15 +2,8 @@ local ToggleTrailsInstant = {
...
-    for _,v in pairs(list) do
-      openspace.setPropertyValueSingle(v, not openspace.propertyValue(v))
-    end
-
-    local moonlist = openspace.property("{moonTrail_solarSystem}.Renderable.Enabled")
-    for _,v in pairs(moonlist) do
-      openspace.setPropertyValueSingle(v, not openspace.propertyValue(v))
-    end
+    openspace.toggleFade("{planetTrail_solarSystem}.Renderable", 0)
+    openspace.toggleFade("{moonTrail_solarSystem}.Renderable", 0)
```
alexanderbock:
```suggestion
    openspace.toggleFade("{planetTrail_solarSystem}.Renderable", 0.0)
    openspace.toggleFade("{moonTrail_solarSystem}.Renderable", 0.0)
```

data/assets/scene/solarsystem/telescopes/jwst/targets/hudf.asset
```diff
@@ -77,12 +77,10 @@ asset.onInitialize(function()
   openspace.addScreenSpaceRenderable(HUDFImage)
 
   openspace.addSceneGraphNode(HUDFPosition)
-  openspace.addSceneGraphNode(HUDFImage)
```
alexanderbock, line 80: Great catch!

alexanderbock (approved):

data/assets/actions/trails/toggle_all_trails.asset
alexanderbock: ?

data/assets/actions/trails/toggle_all_trails_planets_moons_instant.asset
alexanderbock: ?

PR #3730, opened 2025-06-26
Adds property confirmation modal hint to properties

### LINKED ISSUE #3006: "Needs confirmation" view option for properties
Some property changes may lead to dire consequences - such as the "Disable all mouse interaction". 

We could add a view option for properties that makes an "Are you sure?" modal appear in the WebGUI before the value is actually changed, i.e. ask the user for confirmation.

engbergandreas:
Adds option to have a modal popup when changing a property, similar to userLevel it only acts as a hint to the GUI on whether it should be shown or not.  

To be tested with  https://github.com/OpenSpace/OpenSpace-WebGui/pull/178

fixes #3006 

Obs. as described by Alex comment 
> Showing the dialog also for numerical properties when the value entered by hand is outside the range for Novice users and users

Is not possible with the current implementation on the GUI side. Is this something we still want? Considering that we have the yellow text and warning when the input is outside the range.


include/openspace/engine/openspaceengine.h
```diff
@@ -88,6 +88,12 @@ class OpenSpaceEngine : public properties::PropertyOwner {
         CameraPath
     };
 
+    enum class PropertyConfirmation {
+        Never = 0,
+        Default = 1,
+        Always = 2,
```
alexanderbock:
```suggestion
        Default,
        Always
```

No need to explicitly number the others

include/openspace/properties/property.h
```diff
@@ -84,6 +84,21 @@ class Property {
...
+     * The Confirmation classes for Property%s. Always ignore GUI settings, Yes and No
+     * should adhere to GUI settings 
+     */
+    enum class Confirmation {
+        /// Do not confirm with respect to GUI settings
+        No = 0,
+        /// Confirm with respect to GUI settings
+        Yes = 1,
+        /// Always confirm
+        Always = 2,
```
alexanderbock:
```suggestion
        Yes,
        /// Always confirm
        Always,
```

include/openspace/properties/property.h
```diff
@@ -84,6 +84,21 @@ class Property {
...
+        /// The default confirmation for properties
+        Default = No
```
alexanderbock: Do we need the `Default` here?  I'd rather say that different parts of the code might want to declare their own default values and we might not want to bake it into the enum itself.

Also I think we talked about having a `Never` setting as well?

src/engine/configuration.cpp
```diff
@@ -76,6 +76,19 @@ namespace {
         // OpenSpace. If it is not provided, it defaults to 'User'
         std::optional<Visibility> propertyVisibility;
 
+        enum class [[codegen::map(openspace::OpenSpaceEngine::PropertyConfirmation)]]
+        PropertyConfirmation
+        {
+            Never,
+            Default,
+            Always
```
alexanderbock: Same here

src/engine/openspaceengine.cpp
```diff
@@ -159,7 +159,22 @@ namespace {
...
+        ShowPropertyConfirmationModalInfo =
+    {
+        "ShowPropertyConfirmationModals",
+        "Show Property Confirmation Modals",
+        "Controls when confirmation modals are shown when making changes to properties. "
+        "'Never': Disables confirmation prompts for all properties, unless a specific "
+        "property explicitly requires confirmation. 'Default': Follows the individual "
+        "settings for each property. Properties marked as 'Yes' or 'Always' will require "
+        "confirmation, while others will not. 'Always': Enables confirmation prompts for "
+        "all changes, regardelss of specific property settings.",
```
alexanderbock:
```suggestion
        "ShowPropertyConfirmation",
        "Show Property Confirmation",
        "Controls when confirmation dialogs are shown when making changes to properties. "
        "'Never': Disables confirmation prompts for all properties, unless a specific "
        "property explicitly requires confirmation. 'Default': Follows the individual "
        "settings for each property. Properties marked as 'Yes' or 'Always' will require "
        "confirmation, while others will not. 'Always': Enables confirmation prompts for "
        "all changes, regardless of specific property settings.",
```

Removing the "modal" as most people wouldn't know what that means

include/openspace/properties/property.h
engbergandreas: I wasnt sure if this was needed or not when looking at the `Visibility` setting as reference. 

Regarding `Never`, I discussed this with @WeirdRubberDuck and we couldn't come up with a situation where it would be necessary to have never (?) But am open to add it again.

src/engine/configuration.cpp
engbergandreas: In this case Default would be that the individual properties determine if they should show the modal or not, but maybe needs another name?

include/openspace/properties/property.h
alexanderbock: I'd think purely for symmetry's sake it might be fine to add it, even if we don't have a specific use-case at the moment.

Regarding the `Default`, I'd probably say that we can determine at each use-case site what the default is, rather than adding a `Default` label to the enum.  Specifically since we won't be able to use this to detect if something is not the default anyway

src/engine/configuration.cpp
alexanderbock: Ah!  If we remove the `Default` label from the other enum, then we only use it here and i think there is no more confusion

include/openspace/properties/property.h
```diff
@@ -84,6 +84,21 @@ class Property {
         Default = Always
     };
 
+    /**
+     * The Confirmation classes for Property%s. Always and Never ignore GUI settings, Yes
```
WeirdRubberDuck:
```suggestion
     * The Confirmation classes for Properties. Always and Never ignore GUI settings, Yes
```
?

src/engine/configuration.cpp
```diff
@@ -76,6 +76,19 @@ namespace {
...
+        };
+        // Determines when the property confirmation modal should be shown when starting
+        // up OpenSpace. If it is not provided, it defaults to 'Default', i.e., each
+        // individual property determines whether it requires a confirmation or not
+        std::optional<PropertyConfirmation> propertyConfirmation;
```
WeirdRubberDuck: The first sentence sounds a bit weird here. How about something like: 
```suggestion
        // Determines the setting for when property confirmation modals should be shown
        // on property value changes. If not provided, it defaults to 'Default', i.e.,
        // each individual property determines whether it requires a confirmation or not.
        std::optional<PropertyConfirmation> propertyConfirmation;
```

Not that we have decided that these descriptions should end with punctuation. It's just not applied to the other parameters in this file yet.

WeirdRubberDuck: Also , below you call them "dialogs" instead of modals. Do we want to do the same here?

src/engine/openspaceengine.cpp
```diff
@@ -159,7 +159,22 @@ namespace {
         "Disable All Mouse Inputs",
         "Disables all mouse inputs. Useful when using touch interaction, to prevent "
         "double inputs on touch (from both touch input and inserted mouse inputs).",
-        openspace::properties::Property::Visibility::User
+        openspace::properties::Property::Visibility::User,
+        openspace::properties::Property::Confirmation::Always
```
WeirdRubberDuck: Should we rename the enum to `NeedsConfirmation` ?  Would read a bit easier

Just a thought

WeirdRubberDuck (commented):
Looks good to me.  Looking at the GUI PR separately

include/openspace/properties/property.h
ylvaselling: Can we call "Gui settings" something else, like Engine property? I thought it was something in the ui

ylvaselling:
I'm not sure I follow what the use case is for properties to have confirmation "always" and never". Isn't it enough to have "yes" and "no"? Especially since we can set "always", "never" and "default" in the ui. 

If the property is set to "always" and the ui is set to "never" the modal shows up, but that seems a bit strange to me. Wouldn't it be confusing for the user when they set a property to never show confirmations, and then they show up anyway? The only way to find where "needs confirmation" of the property is set is in the source code. Personally i think it is enough and less confusing with only "yes" and "no"

PR #3734, opened 2025-07-01
Add a Test Wizard and start recreating new tests

alexanderbock:


src/scene/profile_lua.inl
```diff
@@ -26,6 +26,32 @@
 
 namespace {
 
+/**
+ * Returns the name of the profile with which OpenSpace was started.
+ */
+[[codegen::luawrap]] std::string profile() {
```
Roxeena: This name is a bit confusing, I thought this function would return the full contents of the profile as a dictionary/table. Maybe `profileName` is a more descriptive name for this function?

src/scene/profile_lua.inl
```diff
@@ -26,6 +26,32 @@
...
+    else {
+        return p;
+    }
+}
+
+/**
+* Returns the full path of the profile with which OpenSpace was started.
+*/
+[[codegen::luawrap]] std::filesystem::path profilePath() {
+    return openspace::global::configuration->profile;
```
Roxeena, line 52: Well I think the same here for `openspace::global::configuration->profile`, which only returns the path by the looks of it. But that is not part of this PR I suppose.

support/assetvalidation/main.py
```diff
@@ -1,8 +1,8 @@
 ##########################################################################################
 #                                                                                        #
-# OpenSpace                                                                              #
+# OpenSpace Asset Validation                                                             #
 #                                                                                        #
-# Copyright (c) 2014-2025                                                                #
+# Copyright (c) 2024-2025                                                                #
```
Roxeena, line 5: Why?

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
+##########################################################################################
+#                                                                                        #
+# OpenSpace Visual Test Wizard                                                           #
+#                                                                                        #
+# Copyright (c) 2025                                                                     #
```
Roxeena: Same here, why is the copyright date different than the normal?

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+  test["commands"].append({ "type": "screenshot" })
+
+  # Save the test to disk
+  with open(f"{name}.ostest", "w") as fp:
+    json.dump(test, fp, indent=2)
+    fp.write("\n")
+
+
+async def mainLoop():
+  api = Api("localhost", 4681)
```
Roxeena, line 321: Any particular reason why this port is used?

visualtests/documentation/model_vertex_colors.ostest
```diff
@@ -1,25 +0,0 @@
-{
```
Roxeena, line 1: Why was this test removed?

visualtests/profiles/default/hirise.ostest
```diff
@@ -1,74 +0,0 @@
-{
```
Roxeena, line 1: Same with these, why were they removed?

visualtests/profiles/eclipse/earth1966.ostest
```diff
@@ -1,16 +0,0 @@
-{
```
Roxeena, line 1: And here

visualtests/profiles/default/moon-surface.ostest
```diff
@@ -1,18 +0,0 @@
-{
```
Roxeena, line 1: And here

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+    print("Finalize the test by selecting the empty value.")
+    selection = input()
+    if selection == "":
+      break
+
+    if not selection.isnumeric():
+      print("Invalid index")
+      continue
+
+    idx = int(selection) - 1
```
Roxeena, line 296: Could it make sense to also be able to give a numeric value that is this index directly? It could potentially save some time if you do not need to write and correctly spell the scripts/commands you want to add.

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+  if include == "y" and ("Timestamp" in navstate):
+    res["value"]["timestamp"] = navstate["Timestamp"]
+
+  return res
+
+
+async def createCommandAsset(openspace):
+  assets = await openspace.asset.rootAssets()
+  folder = await openspace.absPath("${ASSETS}")
+  assets = [asset[len(folder)+1:asset.find(".")] for index,asset in assets.items()]
```
Roxeena, line 66:
```suggestion
  assets = [asset[len(folder)+1:asset.find(".")]
  
  print("List of assets:")
  for index,asset in assets.items()]
```
To make the code more readable and consistent maybe?

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+
+  print("List of assets:")
+  for index,asset in enumerate(assets):
+    print(  f"  ({index+1}): {asset}")
+
+  asset = ""
+  while not asset.isnumeric():
+    asset = input("Select which asset to load: ")
```
Roxeena, line 74: So here the user inputs an integer index? The help text could be a little more clearer.

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+    value = input("Enter value of the property: ")
+
+    # Perform some type conversions if necessary
+    if value.lower() == "true":
+      value = True
+    elif value.lower() == "false":
+      value = False
+
+    if value.isnumeric():
+      value = int(value)
```
Roxeena: How about floats and doubles? Or vectors such as colors?

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+    "type": "property",
+    "value": {
+      "property": uri,
+      "value": value
+    }
+  }
+
+
+async def createCommandWait(openspace):
+  wait = input("Number of seconds to wait: ")
```
Roxeena:
```suggestion
  seconds = input("Number of seconds to wait: ")
```
To be more descriptive

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+  if wait == "":
```
Roxeena:
```suggestion
  if seconds == "":
```

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+    return None
+
+  return {
+    "type": "wait",
+    "value": wait
```
Roxeena:
```suggestion
    "value": seconds
```

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+
+  p = ""
+  if pause == "y":
+    p = "true"
+  else:
+    p = "false"
+
+  return {
+    "type": "pause",
+    "value": p
```
Roxeena, line 175: Is this supposed to be written as a string and not a boolean?

support/testwizard/main.py
```diff
@@ -0,0 +1,331 @@
...
+    print(  f"  ({index+1}): {action["Name"]} ({action["Identifier"]})")
+
+  action = ""
+  while not action.isnumeric():
+    action = input("Select which action to execute: ")
+
+    if action == "":
+      return None
+
+    idx = int(action) - 1
```
Roxeena, line 203: Same here, maybe it could be clearer that the user is supposed to enter an index here.

visualtests/README.md
```diff
@@ -1,89 +0,0 @@
-# Visual Test Specification
```
Roxeena, line 1: Why is this file removed? Has it been moved somewhere else?

visualtests/profiles/default/solarsystem.osrec
Roxeena, line 1: Why? Is it not needed anymore or has it been moved?

visualtests/profiles/missions/apollo/11/landing.ostest
```diff
@@ -1,20 +0,0 @@
-{
```
Roxeena, line 1: What about all of these files?

visualtests/profiles/nightsky/constellationlines.ostest
Roxeena, line 1: And this one

visualtests/renderable/renderablemodel/renderablemodel.ostest
Roxeena, line 1: And this one

src/scene/profile_lua.inl
alexanderbock: Agree.  Changed it to `profileName`

src/scene/profile_lua.inl
alexanderbock: The entries in the `Configuration` struct are deliberately named to match the keys in the `openspace.cfg`, so this would be tricky to change even if we wanted to

support/assetvalidation/main.py
alexanderbock: I get the confusion;  the different copyright dates would be if we consider assetvalidation to be a different tool.  I changed this and the testwizards to say "OpenSpace" at the top and the same copyright as the rest to reduce confusion

support/testwizard/main.py
alexanderbock: It's the standard websocket port that we use in OpenSpace

visualtests/documentation/model_vertex_colors.ostest
alexanderbock: Are we using it in the documentation page somewhere already?  If so we should recreate the test in the correct subfolder

visualtests/profiles/default/hirise.ostest
alexanderbock: Same answer for the next few as well.  I recreated a few tests for the profiles without recreating all of them.  The idea is to have a clean sheet of tests and get rid of the old ones. If we need a new HiRISE test (which we do), we can create that one from scratch again very easily now

support/testwizard/main.py
alexanderbock: I don't understand the question.   You _do_ provide the index to which command you want to add.  It's not possible to type the command out in full; see the `idx` variable being used in line 296

support/testwizard/main.py
alexanderbock: I don't see how this code would work?

The new suggestion doesn't use the list comprehension feature to extract the individual assets to extract the base path. The new line 65 would have an undefined variable `asset`.

support/testwizard/main.py
alexanderbock: Leaving this at it is. I think from context it is clear what to enter:

image

support/testwizard/main.py
alexanderbock: That's right... changed `int` to `float`.

Vectors we can't enter by hand at the moment.  We can add it to the wizard when we need it

support/testwizard/main.py
alexanderbock: Good catch

support/testwizard/main.py
alexanderbock: Same as above, I think from context its clear enough

visualtests/README.md
alexanderbock: Yep, all of the different descriptions are now here: https://docs.openspaceproject.com/latest/contribute/imagetesting/index.html

ylvaselling:
Nice! 

image
"Leaving the prompt empty finalizes the test" is a bit hard to understand. make some kind of action to "finish" making the test? Like (10) Save and quit

image
On ending ^ close connection properly?

Make a note in the description in the console that the test get stored in the same folder? Since you have to refresh the file explorer on windows, it is not super obvious

alexanderbock:
We can now save the test by providing `s` or `save` at the prompt.  The path to the test is printed to the console.

The error message is a bug in our Python API that has been fixed in that repository

support/testwizard/main.py
Roxeena: :+1: then I must have misunderstood the code

support/testwizard/main.py
Roxeena: I meant as you do similarly for actions in line 186. To first get all the assets and then at a separate line, loop over them. Right now this, to me, does two things and is not consistent with how it is done in the actions command. If the code I suggested is not correct python code, then that is just because I don't have enough experience with python.

visualtests/documentation/model_vertex_colors.ostest
Roxeena: We created it to test the model loading functionality, and specifically for that color format. Not sure if the documentation folder was the right place for it in but I still think it is valuable to keep the test to enable us to be notified if this feature ever breaks.

visualtests/profiles/default/hirise.ostest
Roxeena: So we want to recreate all of the tests that we already have? I thought this was supposed to make it easier to add new ones but still keep the old ones.

support/testwizard/main.py
Roxeena: That could be good to have documented somewhere or a warning to make that clearer.

PR #3736, opened 2025-07-01
To compile on mac os

hn-88:
With these changes, OpenSpace successfully compiles on MacOS. The fixes to Ghoul in this branch - https://github.com/hn-88/Ghoul/tree/MacOSfixes/ are also needed. I had opened a pull request for Ghoul earlier, that was incomplete. If appropriate, I can open one more pull request for Ghoul and delete the earlier one. 

Notes:
1. Many `path` objects in std::format needed the `.string()` suffix, since XCode still does not support std::formatter overloads.
2. `emplace_back()` in certain situations had problems due to stricter checking for constructors on AppleCLang - in some cases I have written constructors, in others I have manually created temporary structs and used `push_back()`
3. Building on Intel Macs should be possible, if they're running MacOS 14 or above - with MacOS 13, XCode 15.4 is not available, only up to 15.2, which has less support for C++ 20 etc. https://developer.apple.com/xcode/cpp/ - anyway, I'm trying it, at https://github.com/hn-88/OpenSpace-IntelMac/actions


alexanderbock:
Closing the PR due to us dropping MacOS support on the OpenGL side also removing the need to support the Apple Clang fork

PR #3749, opened 2025-07-11
Higher quality rings and ring shadows

### LINKED ISSUE #3656: Improve Saturn Ring Shadow rendering
The current method for displaying Saturn's ring shadow projected onto the Saturn globe is leaving a bit to be desired when compared to a reference:

 _OpenSpace:_
Image

_[Reference:](https://www.esa.int/Science_Exploration/Space_Science/In_the_shadows_of_Saturn_s_rings)_
Image

Increasing the ring shadow map resolution is probably a good start.

---

- [x] Directly sample the ring texture to render shadows from ring onto globe surface
- [x] Apply additional factors to accurately reproduce the darkening of the shadow at low light angle
- [x] Improve code, remove unused uniforms and textures etc.
- [ ] Check that everything works in other SGCT configs

benpm:
This is a minor rendering feature which enables high definition ring shadows on Saturn (and any other globe that we might want rings for in the future...?). It also fixes some visual issues with Saturn's rings. The result:

https://github.com/user-attachments/assets/4f69f640-1d2c-4c92-aebe-ad37edaafa95

Summary of changes:
- Projects the ring texture directly on the globe surface in order to produce shadows, instead of using a depth map. (The globe casting shadow onto the rings still uses the depth map, however)
- The transparency map was inverted by accident previously, it's now fixed and the results look more accurate. I will put up a side-by-side tomorrow
- As the light angle becomes smaller, the ring shadow becomes somewhat darker due to light traveling through more material to reach the surface. Let me know if this is not actually the case but it seems to match reference images reasonably well

This PR will resolve #3656


alexanderbock:
Getting a bunch of uniform location warnings:
```
(W) ProgramObject(..ch') Failed to locate uniform location for 'modelTransform'
(W) ProgramObject(..ch') Failed to locate uniform location for 'shadowMatrix'
(W) ProgramObject(..ch') Failed to locate uniform location for 'shadowMapTexture'
(W) ProgramObject(..ch') Failed to locate uniform location for 'zFightingPercentage'
(W) ProgramObject(..ch') Failed to locate uniform location for 'ringTextureFwrd'
(W) ProgramObject(..ch') Failed to locate uniform location for 'ringTextureBckwrd'
(W) ProgramObject(..ch') Failed to locate uniform location for 'ringTextureUnlit'
(W) ProgramObject(..ch') Failed to locate uniform location for 'sunPositionObj'
(W) ProgramObject(..ch') Failed to locate uniform location for 'camPositionObj'
(W) ProgramObject(..ch') Failed to locate uniform location for 'sunPositionWorld'
(W) ProgramObject(..ch') Failed to locate uniform location for 'camPositionWorld'
```

If possible, we should only set the uniforms if they are actually needed and `if` them out if we know that they are not needed

modules/globebrowsing/shaders/advanced_rings_fs.glsl
```diff
@@ -126,6 +126,7 @@ Fragment getFragment() {
 
   frag.color = diffuse * shadow;
   frag.color.a *= opacity;
+  frag.color.a += (1.0 - shadow) * 0.5;
```
alexanderbock:
```suggestion
  frag.color.a = frag.color.a * opacity + (1.0 - shadow) * 0.5;
```

? More chance for the GLSL compiler to detect the fused multiply-add and optimize

modules/globebrowsing/shaders/localrenderer_vs.glsl
```diff
@@ -37,16 +37,16 @@ out vec4 fs_position;
...
+out vec3 posObjSpace;
+out vec3 normalObjSpace;
 
 #if USE_ACCURATE_NORMALS
   out vec3 ellipsoidTangentThetaCameraSpace;
   out vec3 ellipsoidTangentPhiCameraSpace;
 #endif // USE_ACCURATE_NORMALS
 
-#if USE_ECLIPSE_SHADOWS
-out vec3 positionWorldSpace;
```
alexanderbock, line 47: I think the `positionWorldSpace` should still be inside the `USE_ECLIPSE_SHADOWS` ifdef since it is inside it in the `main` function

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -52,22 +52,29 @@ uniform vec2 vertexResolution;
...
+uniform sampler1D ringTextureColor;
+// Transparency of the rings
+uniform sampler1D ringTextureTransparency;
+uniform vec2 textureOffset;
+uniform vec3 sunPositionWorld; // NEW: world coordinates
+uniform vec3 camPositionWorld; // NEW: world coordinates
+#endif // USE_RING_SHADOWS
 #endif // SHADOW_MAPPING_ENABLED
 
+in vec3 posObjSpace; // Fragment position in object space
```
alexanderbock: For the GLSL coding style, we have the `in/out` variables at the top before the `uniform`s

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -137,19 +144,35 @@ vec4 calcShadow(const ShadowRenderingStruct shadowInfoArray[NSEclipseShadows],
 }
 #endif
 
+float rayPlaneIntersection(vec3 rayOrigin, vec3 rayDirection, vec3 planePoint, vec3 planeNormal) {
```
alexanderbock:
```suggestion
float rayPlaneIntersection(vec3 rayOrigin, vec3 rayDirection, vec3 planePoint,
                           vec3 planeNormal)
{
```

For line length

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
-      sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2( NSSamples - #{i}, -NSSamples + #{i}));
-      sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2( NSSamples - #{i},  0));
-      sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2( NSSamples - #{i},  NSSamples - #{i}));
-    #endfor
-    sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2(0, 0));
-    shadow = sum / (8.0 * NSSamples + 1.f);
-  }
-  frag.color.xyz *= shadow < 0.99 ? clamp(shadow + 0.3, 0.0, 1.0) : shadow;
-#endif
+  // Light thru rings is colored, default full white
```
alexanderbock:
```suggestion
  // Light through rings is colored, default full white
```

Even though the spelling of that word is bonkers

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
+  
+  #if USE_RING_SHADOWS
+    // Calculate ring shadow by projecting ring texture directly onto surface
+    // Assume ring lies in the XZ plane (Y=0) in object space
+    vec3 surfaceToSun = -normalize(lightDirectionObjSpace); // Use world coordinates
+    vec3 p = posObjSpace;
+    vec3 ringPlaneNormal = vec3(0.0, 0.0, 1.0);
+    
+    if (abs(surfaceToSun.y) > 1e-8 && dot(normalObjSpace, lightDirectionObjSpace) < 0.0) {
+      float t = rayPlaneIntersection(p, surfaceToSun, vec3(0.0, 0.0, 0.0), ringPlaneNormal);
```
alexanderbock:
```suggestion
      float t = rayPlaneIntersection(p, surfaceToSun, vec3(0.0), ringPlaneNormal);
```

For making it fit in a single line

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
+      // Calculate distance from ring center
+      float tx = length(ringIntersection.xy) / ringSize;
+      // See advanced_rings_fs.glsl for explanation of textureOffset
+      float texCoord = (tx - textureOffset.x) / (textureOffset.y - textureOffset.x);
+
+      if (texCoord >= 0.0 && texCoord <= 1.0) {
+        // Sample ring transparency texture
+        float ringOpacity = texture(ringTextureTransparency, texCoord).r;
+        
+        // Increase the shadow darkness factor with low angle to simulate the light having to pass thru more material
```
alexanderbock:
```suggestion
        // Increase the shadow darkness factor with low angle to simulate the light having
        // to pass through more material
```

modules/globebrowsing/shaders/advanced_rings_fs.glsl
```diff
@@ -73,16 +73,16 @@ Fragment getFragment() {
   vec4 colorBckwrd = texture(ringTextureBckwrd, texCoord);
   vec4 colorFwrd = texture(ringTextureFwrd, texCoord);
   vec4 colorMult = texture(ringTextureColor, texCoord);
-  vec4 transparency = texture(ringTextureTransparency, texCoord);
+  float transparency = 1.0 - texture(ringTextureTransparency, texCoord).r;
```
alexanderbock:
```suggestion
  float transparency = 1.0 - texture(ringTextureTransparency, texCoord).a;
```

We usually use the alpha channel to signal that we meant to access a 1-dimensional texture and not the red channel out of an RG, RGB, or RGBA texture

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
-      sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2( NSSamples - #{i},  0));
-      sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2( NSSamples - #{i},  NSSamples - #{i}));
-    #endfor
-    sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2(0, 0));
-    shadow = sum / (8.0 * NSSamples + 1.f);
-  }
-  frag.color.xyz *= shadow < 0.99 ? clamp(shadow + 0.3, 0.0, 1.0) : shadow;
-#endif
+  // Light thru rings is colored, default full white
+  vec3 lightColor = vec3(1.0, 1.0, 1.0);
```
alexanderbock:
```suggestion
  vec3 lightColor = vec3(1.0);
```

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
+      vec3 ringIntersection = p + t * surfaceToSun;
+        
+      // Calculate distance from ring center
+      float tx = length(ringIntersection.xy) / ringSize;
+      // See advanced_rings_fs.glsl for explanation of textureOffset
+      float texCoord = (tx - textureOffset.x) / (textureOffset.y - textureOffset.x);
+
+      if (texCoord >= 0.0 && texCoord <= 1.0) {
+        // Sample ring transparency texture
+        float ringOpacity = texture(ringTextureTransparency, texCoord).r;
```
alexanderbock, line 320:
```suggestion
        float ringOpacity = texture(ringTextureTransparency, texCoord).a;
```

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
+        
+        // Increase the shadow darkness factor with low angle to simulate the light having to pass thru more material
+        float angleFactor = clamp(abs(-dot(ringPlaneNormal, surfaceToSun)) / 2.0, 0.0, 0.3);
```
alexanderbock, line 324:
```suggestion
        float angleFactor = clamp(abs(dot(ringPlaneNormal, surfaceToSun)) / 2.0, 0.0, 0.3);
```

If we do the `abs` directly anyway, we don't need the `-`

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
+        
+        // Calculate shadow factor based on ring opacity
+        shadow = ringOpacity;
+        shadow = clamp(shadow + angleFactor, 0.05, 1.0);
```
alexanderbock:
```suggestion
        shadow = clamp(ringOpacity + angleFactor, 0.05, 1.0);
```

modules/globebrowsing/shaders/renderer_fs.glsl
```diff
@@ -268,29 +292,48 @@ Fragment getFragment() {
...
+        lightColor = texture(ringTextureColor, texCoord).rgb;
+      }
+    }
+
+  #endif // USE_RING_SHADOWS
+
+  // Blend the light color passing thru the rings with the pre-shaded color
```
alexanderbock:
```suggestion
  // Blend the light color passing through the rings with the pre-shaded color
```

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -774,6 +774,21 @@ RenderableGlobe::RenderableGlobe(const ghoul::Dictionary& dictionary)
         _ringsComponent->setParentFadeable(this);
         _ringsComponent->initialize();
         addPropertySubOwner(_ringsComponent.get());
+        
+        // Set up notification for shader recompilation when rings state changes
```
alexanderbock: Since we are creating the RingComponent in here, I think it might be better to pass the `onChange` lambda into the constructor instead and let the RingsComponent set the `onChange` on the enabled property itself. That way we can avoid a bunch of ugly code and assuming that there is an Enabled property in this class

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -774,6 +774,21 @@ RenderableGlobe::RenderableGlobe(const ghoul::Dictionary& dictionary)
...
+            _ringsComponent->property("Enabled")
+        );
+        if (enabledProperty) {
+            enabledProperty->onChange([this]() {
+                _shadersNeedRecompilation = true;
+            });
+        }
+        
+        // Set up notification for shader recompilation when rings readiness changes
+        _ringsComponent->onReadinessChange([this]() {
```
alexanderbock, line 791: What do we need the readiness change check for?  It smells to me like that it is only useful if we start OpenSpace with a shader that does not compile?  If that is the case, I think we should get rid of the code as it makes the class more complicated for a use-case that doesn't happen very often for us and never happens for a user

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1242,9 +1257,9 @@ void RenderableGlobe::renderChunks(const RenderData& data, RendererTasks&,
     }
 
     using namespace layers;
-    const bool nightLayersActive =
+    bool nightLayersActive =
```
alexanderbock: If we delete the second assignment further below (which doesn't do anything), we can keep these variables `const`

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1242,9 +1257,9 @@ void RenderableGlobe::renderChunks(const RenderData& data, RendererTasks&,
...
         !_layerManager.layerGroup(Group::ID::NightLayers).activeLayers().empty();
-    const bool waterLayersActive =
+    bool waterLayersActive =
```
alexanderbock: Whats the reason for switching the order of the local and global renderer here?  It makes reading the diff more difficult than it need to be and would say that we should keep the old order

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1449,9 +1477,7 @@ void RenderableGlobe::renderChunkGlobally(const Chunk& chunk, const RenderData&
 
     if (_eclipseShadowsEnabled && !_ellipsoid.shadowConfigurationArray().empty()) {
         calculateEclipseShadows(program, data, ShadowCompType::GLOBAL_SHADOW);
-    }
-
-    // Shadow Mapping
+    }    // Shadow Mapping
```
alexanderbock: I think having the empty lines here improved the readability

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1589,9 +1659,7 @@ void RenderableGlobe::renderChunkLocally(const Chunk& chunk, const RenderData& d
 
     if (_eclipseShadowsEnabled && !_ellipsoid.shadowConfigurationArray().empty()) {
         calculateEclipseShadows(program, data, ShadowCompType::LOCAL_SHADOW);
-    }
-
-    // Shadow Mapping
+    }    // Shadow Mapping
```
alexanderbock: Same here

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1253,44 +1268,57 @@ void RenderableGlobe::renderChunks(const RenderData& data, RendererTasks&,
 
         const glm::vec3 directionToSunCameraSpace = glm::vec3(viewTransform *
             glm::dvec4(directionToSunWorldSpace, 0));
-        // @TODO (abock, 2020-04-14); This is just a bandaid for issue #1136.  The better
-        // way is to figure out with the uniform is optimized away. I assume that it is
-        // because the shader doesn't get recompiled when the last layer of the night
-        // or water is disabled;  so the shader thinks it has to do the calculation, but
-        // there are actually no layers left
+        const glm::vec3 directionToSunObjSpace = glm::vec3(_cachedInverseModelTransform *
+            glm::dvec4(directionToSunWorldSpace, 0));
```
alexanderbock:
```suggestion
            glm::dvec4(directionToSunWorldSpace, 0.0));
```

To signal immediately that it is meant to be a `double`

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1253,44 +1268,57 @@ void RenderableGlobe::renderChunks(const RenderData& data, RendererTasks&,
...
```
alexanderbock:
```suggestion
            glm::dvec4(directionToSunWorldSpace, 0.0));
```

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1253,44 +1268,57 @@ void RenderableGlobe::renderChunks(const RenderData& data, RendererTasks&,
...
+        !_layerManager.layerGroup(Group::ID::NightLayers).activeLayers().empty();
+    waterLayersActive =
+        !_layerManager.layerGroup(Group::ID::WaterMasks).activeLayers().empty();
 
     if (nightLayersActive || waterLayersActive || _performShading) {
         const glm::dvec3 directionToSunWorldSpace =
             directionToLightSource(data.modelTransform.translation, _lightSourceNode);
 
         const glm::vec3 directionToSunCameraSpace = glm::vec3(viewTransform *
             glm::dvec4(directionToSunWorldSpace, 0));
```
alexanderbock:
```suggestion
            glm::dvec4(directionToSunWorldSpace, 0.0));
```

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1253,44 +1268,57 @@ void RenderableGlobe::renderChunks(const RenderData& data, RendererTasks&,
...
```
alexanderbock:
```suggestion
            glm::dvec4(directionToSunWorldSpace, 0.0));
```

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1771,7 +1895,7 @@ void RenderableGlobe::recompileShaders() {
     }
 
     std::vector<std::pair<std::string, std::string>>& pairs =
-        preprocessingData.keyValuePairs;
+                preprocessingData.keyValuePairs;
```
alexanderbock:
```suggestion
        preprocessingData.keyValuePairs;
```

modules/globebrowsing/src/renderableglobe.cpp
```diff
@@ -1787,11 +1911,18 @@ void RenderableGlobe::recompileShaders() {
...
+        "useRingShadows", 
+        std::to_string(_shadowMappingProperties.shadowMapping && _ringsComponent)
+    );
     pairs.emplace_back("showChunkEdges", std::to_string(_debugProperties.showChunkEdges));
     pairs.emplace_back("showHeightResolution", "0");
     pairs.emplace_back("showHeightIntensities", "0");
     pairs.emplace_back("defaultHeight", std::to_string(DefaultHeight));
 
+    // log if ring shadows are enabled
+    LINFO(std::format("Ring shadows enabled: {}", _shadowMappingProperties.shadowMapping && _ringsComponent));
```
alexanderbock:
```suggestion
    LINFO(std::format(
        "Ring shadows enabled: {}",
        _shadowMappingProperties.shadowMapping && _ringsComponent
    ));
```

Columnwidth fixing

modules/globebrowsing/src/ringscomponent.cpp
```diff
@@ -867,4 +876,50 @@ double RingsComponent::size() const {
...
+    return _textureUnlit.get();
+}
+
+ghoul::opengl::Texture* RingsComponent::ringTextureColor() const {
+    return _textureColor.get();
+}
+
+ghoul::opengl::Texture* RingsComponent::ringTextureTransparency() const {
+    return _textureTransparency.get();
+}
```
alexanderbock: I'd probably say to rename these two:  `forwardTexture`, `backwardTexture`, `unlitTexture`, `colorTexture`, and `transparencyTexture`.  The "ring" is selfexplanatory from the call side, since it will be _ringComponent->forwardTexture()` and another "ring" would be double.

At the same time, we can also rename `_textureFwrdPath` to `_textureForwardPath` etc

modules/globebrowsing/src/ringscomponent.h
```diff
@@ -69,15 +72,27 @@ class RingsComponent : public properties::PropertyOwner, public Fadeable {
     );
     void update(const UpdateData& data);
 
-    static documentation::Documentation Documentation();
-
-    bool isEnabled() const;
+    static documentation::Documentation Documentation();    bool isEnabled() const;
```
alexanderbock:
```suggestion
    bool isEnabled() const;

    static documentation::Documentation Documentation();
```

modules/globebrowsing/shaders/localrenderer_vs.glsl
benpm: Wouldn't this cause shader compilation errors if USE_ECLIPSE_SHADOWS is false and USE_RING_SHADOWS is true?

modules/globebrowsing/shaders/advanced_rings_fs.glsl
benpm: Oh okay, so you can use R or A channels for a single-channel tex? Neat

modules/globebrowsing/shaders/localrenderer_vs.glsl
alexanderbock: I only see if being used here?
```
#if USE_ECLIPSE_SHADOWS
  positionWorldSpace = vec3(inverseViewTransform * dvec4(p, 1.0));
  positionWorldSpace = vec3(modelTransform * dvec4(p, 1.0));
#endif // USE_ECLIPSE_SHADOWS
```

modules/globebrowsing/shaders/advanced_rings_fs.glsl
alexanderbock: Yea, or even G or B,  in a single dimensional, they are all the same

modules/globebrowsing/src/renderableglobe.cpp
benpm: is this not useful for when editing shaders on the fly? i do that a lot with my workflow.

modules/globebrowsing/src/renderableglobe.cpp
benpm: Not sure how this happened. My mistake.

modules/globebrowsing/shaders/advanced_rings_fs.glsl
benpm: Just double checked, and using the alpha channel doesn't work. This is actually the reason the rings were not being rendered correctly before, because the original code use the alpha channel, which is all 1.0. It's because the texture is not actually a single channel, it uses RGB channels (which are all the same value for the texture) and it does not have an alpha channel. Do you think it'd be possible to change the texture format to be able to use the alpha channel as intended? Or would the texture image file need to be modified?

modules/globebrowsing/shaders/advanced_rings_fs.glsl
```diff
@@ -70,19 +70,19 @@ Fragment getFragment() {
...
   }
 
-  vec4 colorBckwrd = texture(ringTextureBckwrd, texCoord);
-  vec4 colorFwrd = texture(ringTextureFwrd, texCoord);
-  vec4 colorMult = texture(ringTextureColor, texCoord);
-  vec4 transparency = texture(ringTextureTransparency, texCoord);
+  vec4 colorBckwrd = texture(textureBackwards, texCoord);
+  vec4 colorFwrd = texture(textureForwards, texCoord);
+  vec4 colorMult = texture(textureColor, texCoord);
+  float transparency = 1.0 - texture(textureTransparency, texCoord).r;
```
alexanderbock, line 76:
```suggestion
  float transparency = 1.0 - texture(textureTransparency, texCoord).a;
```

Using the alpha channel for the single-channel textures to be in line with the other shaders

modules/globebrowsing/shaders/advanced_rings_fs.glsl
```diff
@@ -106,7 +106,7 @@ Fragment getFragment() {
       sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2( NSSamples - #{i},  NSSamples - #{i}));
```
alexanderbock:
```suggestion
      sum += textureProjOffset(shadowMapTexture, normalizedShadowCoords, ivec2(NSSamples - #{i},  NSSamples - #{i}));
```

alexanderbock (approved):

PR #3757, opened 2025-07-16
Show full trail for renderableTrailOrbit

### LINKED ISSUE #1791: RenderableTrailOrbit - option to show full orbit
Sometimes you want to be able to see the full orbit. We have the "Show full trail" option for RenderableTrailTrajectory, would be nice to add to the RenderableTrailOrbit too. Depending on the source of the data, the period can be used to show the trail further back in time, but not forward.

Arohdin:
## Information
Adds toggle `Force Full Orbit Trail`, `Start Time` and `End Time` for renderableTrailOrbit. When start and end times are provided, the new toggle can be used to force a trail of one period of length. When enabled, the behavior is as follows:
- Before `start`: Trail is between start and one period forward in time
- Between `start` and `start + period`: Trail is between start and one period forward in time
- Between `start + period` and `end`: Trail is between ingame time and one period backwards in time
- After `end`: Trail is between end and one period backwards in time
If `Start Time` or `End Time` is not provided, the toggle does nothing

## Changes
- **[New]** - Added new toggle `Force Full Orbit Trail` which enables new feature
- **[New]** - Added new optional `StartTime` and `EndTime` for asset type `renderableTrailOrbit`
- **[Breaking]** - Corrected Gaia trail number of samples from 24244 to 13140 as comment indicated sample interval of 40 minutes but previous number of samples did not reflect that

Closes #1791

### DIFF
--- a/data/assets/scene/solarsystem/telescopes/gaia/trail.asset
+++ b/data/assets/scene/solarsystem/telescopes/gaia/trail.asset
@@ -30,8 +30,10 @@ local GaiaTrail = {
     },
     Color = { 0.0, 0.8, 0.7 },
     PointSize = 5,
+    StartTime = "2013 Dec 19 09:55:27",
+    EndTime = "2026 Sep 14 00:00:00",
     Period = 365, -- 1 year orbit time
-    Resolution = 24244 -- Sameple rate of 40 minutes
+    Resolution = 13140 -- Sameple rate of 40 minutes
   },
   GUI = {
     Name = "Gaia Trail",
@@ -57,6 +59,8 @@ local GaiaTrailEclip = {
     },
     Color = { 1.0, 0.0, 0.0 },
     PointSize = 5,
+    StartTime = "2013 Dec 20 00:00:00",
+    EndTime = "2026 Sep 14 00:00:00",
     Period = 365, -- 1 year orbit time
     Resolution = 365 -- Sameple rate of 1 day
   },
--- a/modules/base/rendering/renderabletrailorbit.cpp
+++ b/modules/base/rendering/renderabletrailorbit.cpp
@@ -87,6 +87,33 @@
 // items in memory as was shown to be much slower than the current system.   ---abock
 
 namespace {
+    constexpr openspace::properties::Property::PropertyInfo ForceFullOrbitTrailInfo = {
+        "ForceFullOrbitTrail",
+        "Force Full Orbit Trail",
+        "Forces the trail to always have a visible length of one orbit. If time between "
+        "the start date of the trail and the current time in OpenSpace is less than one "
+        "period, then the trail will show part of the trail that extends into the "
+        "future compared to the current OpenSpace time. The extent of the trail will "
+        "then be from the start date of the trail and one period of time forward.",
+        openspace::properties::Property::Visibility::User
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo StartTimeInfo = {
+        "StartTime",
+        "Start Time",
+        "The start time for the range of this trajectory. The date must be in ISO 8601 "
+        "format: YYYY MM DD HH:mm:ss.xxx.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo EndTimeInfo = {
+        "EndTime",
+        "End Time",
+        "The end time for the range of this trajectory. The date must be in ISO 8601 "
+        "format: YYYY MM DD HH:mm:ss.xxx.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
     constexpr openspace::properties::Property::PropertyInfo PeriodInfo = {
         "Period",
         "Period (in days)",
@@ -108,6 +135,17 @@ namespace {
     };
 
     struct [[codegen::Dictionary(RenderableTrailOrbit)]] Parameters {
+        // [[codegen::verbatim(ForceFullOrbitTrailInfo.description)]]
+        std::optional<bool> forceFullOrbitTrail;
+
+        // [[codegen::verbatim(StartTimeInfo.description)]]
+        std::optional<std::string> startTime
+            [[codegen::annotation("A valid date in ISO 8601 format")]];
+
+        // [[codegen::verbatim(EndTimeInfo.description)]]
+        std::optional<std::string> endTime
+            [[codegen::annotation("A valid date in ISO 8601 format")]];
+
         // [[codegen::verbatim(PeriodInfo.description)]]
         double period;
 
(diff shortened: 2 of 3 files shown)


modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -128,13 +164,34 @@ documentation::Documentation RenderableTrailOrbit::Documentation() {
...
+    , _endTime(EndTimeInfo)
     , _period(PeriodInfo, 0.0, 0.0, 250.0 * 365.25) // 250 years should be enough I guess
     , _resolution(ResolutionInfo, 10000, 1, 1000000)
 {
     const Parameters p = codegen::bake<Parameters>(dictionary);
 
     _translation->onParameterChange([this]() { _needsFullSweep = true; });
 
+    _forceFullOrbitTrail = p.forceFullOrbitTrail.value_or(_forceFullOrbitTrail);
+    _forceFullOrbitTrail.onChange([&] {
```
alexanderbock:
```suggestion
    _forceFullOrbitTrail.onChange([&]() {
```

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -128,13 +164,34 @@ documentation::Documentation RenderableTrailOrbit::Documentation() {
...
+        _needsFullSweep = true;
+        _indexBufferDirty = true;
+        _forceFlag = true;
+    });
+    addProperty(_forceFullOrbitTrail);
+
+    // Start time in ISO 8601 format
+    _startTime = p.startTime.value_or(_startTime);
+    _startTime.onChange([&] {_needsFullSweep = true; _indexBufferDirty = true; });
```
alexanderbock:
```suggestion
    _startTime.onChange([&]() {
        _needsFullSweep = true;
        _indexBufferDirty = true;
    });
```

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -128,13 +164,34 @@ documentation::Documentation RenderableTrailOrbit::Documentation() {
...
+    addProperty(_startTime);
+
+    // End time in ISO 8601 format
+    _endTime = p.endTime.value_or(_endTime);
+    _endTime.onChange([&] {_needsFullSweep = true; _indexBufferDirty = true; });
```
alexanderbock:
```suggestion
    _endTime.onChange([&]() {
        _needsFullSweep = true;
        _indexBufferDirty = true;
    });
```

modules/base/rendering/renderabletrailorbit.h
```diff
@@ -103,6 +124,9 @@ class RenderableTrailOrbit : public RenderableTrail {
     /// A dirty flag to determine whether the index buffer needs to be regenerated and
     /// then reuploaded
     bool _indexBufferDirty = true;
+    /// Flag to help determine if we should apply full sweep or not when
+    /// _forceFullOrbitTrail is enabled
+    bool _forceFlag;
```
alexanderbock:
```suggestion
    bool _forceFlag = false;
```

It might be good for this variable to have a better name.  `forceFlag` is not super descriptive what the flag is used for

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -307,115 +370,151 @@ void RenderableTrailOrbit::update(const UpdateData& data) {
...
-            if (_primaryRenderInformation.first == _primaryRenderInformation.count - 1) {
-                // If it is at the beginning, set it to the end first
-                _primaryRenderInformation.first = 0;
-            }
-            else {
-                // Move the current pointer fowards one step  to be used as the new
-                // floating
-                ++_primaryRenderInformation.first;
+                // if we are on the upper bounds of the array, we start at 0
+                if (_primaryRenderInformation.first == _primaryRenderInformation.count - 1) {
```
alexanderbock: Line length exceeded

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -307,115 +370,151 @@ void RenderableTrailOrbit::update(const UpdateData& data) {
...
+        return Phase::End;
+    }
+    else {
+        return Phase::Normal;
+    }
+}
+
 RenderableTrailOrbit::UpdateReport RenderableTrailOrbit::updateTrails(
-                                                                   const UpdateData& data)
+                                                                const UpdateData& data)
```
alexanderbock, line 442:
```suggestion
                                                                   const UpdateData& data)
```

Arohdin: Adding one more TAB will push the `const` over the the line length limit.

Arohdin:
@micahnyc Currently when `force full orbit trail` is enabled, the trail will always show at least one period length of trail.
This is true before `start` and after `end` as well - is this the wanted behavior?
Or do we want it to only force a "full trail" within the bounds of `start` and `end`?

data/assets/scene/solarsystem/telescopes/gaia/trail.asset
```diff
@@ -30,8 +30,10 @@ local GaiaTrail = {
     },
     Color = { 0.0, 0.8, 0.7 },
     PointSize = 5,
+    StartTime = "2013 Dec 19 09:55:27",
+    EndTime = "2026 Sep 14 00:00:00",
     Period = 365, -- 1 year orbit time
-    Resolution = 24244 -- Sameple rate of 40 minutes
+    Resolution = 13140 -- Sameple rate of 40 minutes
```
Roxeena, line 36: The comment here seems to be incorrect? 40 minutes in seconds would be 2400? Or am I missremembering the unit for this property?

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -87,6 +87,31 @@
 // items in memory as was shown to be much slower than the current system.   ---abock
 
 namespace {
+    constexpr openspace::properties::Property::PropertyInfo ForceFullOrbitTrailInfo = {
+        "ForceFullOrbitTrail",
+        "Force Full Orbit Trail",
```
Roxeena: I think for a user, this could be a bit difficult to understand. I am wondering if "Force One Full Rotation/Orbit" or something like that could be a better name alternative? Right now, it feels very similar to the "Show Full Trail" property and I think specifying that this is just on full rotation could be helpful.

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -87,6 +87,31 @@
...
+        "Forces the trail to always have a visible length of one orbit. If time between "
+        "the start date of the trail and the current time in OpenSpace is less than one "
+        "period, then the trail will show part of the trail that extends into the "
+        "future compared to the current OpenSpace time. The extent of the trail will "
+        "then be from the start date of the trail and one period of time forward.",
+        openspace::properties::Property::Visibility::User
```
Roxeena:
```suggestion
        "Forces the trail to always have a visible length of one orbit. If the time from"
        "the start date of the trail and the current time in OpenSpace is less than one "
        "period (full rotation), then the trail will extends into the future compared to the "
        "current OpenSpace time (if possible).",
        openspace::properties::Property::Visibility::User
```
Just a suggestion but as the previous comment said, I feel like this could be a bit unclear to a regular user, so I tried to make it a bit simpler.

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -307,115 +382,153 @@ void RenderableTrailOrbit::update(const UpdateData& data) {
...
 
+RenderableTrailOrbit::Phase RenderableTrailOrbit::trailPhase(double time) {
+    if (!_forceFullOrbitTrail) {
+        return Phase::Normal;
+    }
+
+    using namespace std::chrono;
+    const double startTime = Time::convertTime(_startTime);
+    const double endTime = Time::convertTime(_endTime);
+    const double periodInSeconds = _period * duration_cast<seconds>(hours(24)).count();
```
Roxeena: There is a global constant with the number of seconds in 1 day here:  https://github.com/OpenSpace/OpenSpace/blob/7604b3dae946aacf9bb316df9a4026e43f8499f3/include/openspace/util/timeconstants.h#L33

Then you would not need the std::chrono library?

modules/base/rendering/renderabletrailorbit.cpp
```diff
@@ -433,13 +546,20 @@ void RenderableTrailOrbit::fullSweep(double time) {
         std::iota(_indexArray.begin() + _resolution, _indexArray.end(), 0);
     }
 
-    _lastPointTime = time;
-
     using namespace std::chrono;
     const double periodSeconds = _period * duration_cast<seconds>(hours(24)).count();
```
Roxeena, line 624: This seems like a value that is used a lot, do we want to save it somewhere so we do not need to recalculate it so often? It would only change if the period would change, which is not that often.

Roxeena (commented):
I get some unexpected behavior when I tried this out. The first picture is with the property disabled, and the second one is with it enabled. 

OpenSpace_000000

OpenSpace_000001

See how the trail of the Moon have moved further into its orbit when it is enabled. Since the Moon has data both behind and ahead in time for more then one full orbit, I would expect that this property left the trail unchanged since there is nothing to "fill in" to make it a complete orbit. Or have I missunderstood this feature?

data/assets/scene/solarsystem/telescopes/gaia/trail.asset
Arohdin: The sampling is a bit backwards...
You're not defining sample interval in seconds, instead you define how many sampling points the trail should have.
So 13140 points equals once every 40 minutes for the period of 365 days.
(365 * 24 * 60 * 60) / 13140 = 2400 

🙃

Arohdin:
@micahnyc Can you test and see if it is OK compared to #1791 ?
An alternative would be that you can specify a time range to show for the trail instead of whatever time it is in OpenSpace at that moment.

PR #3761, opened 2025-07-22
Support relative paths in GeoJSON SpriteTexture

### LINKED ISSUE #3139: SpriteTexture in GeoJSON file should take relative paths
Currently, it only works with absolute paths, which is cumbersome when moving files around

And with relative paths, I mean a path relative to the geojson file itself

alexanderbock:
Set the current working directory temporarily to the folder where thegeojson is located and convert image path via absPath immediately. Add new test for the relative path.

Closes #3139

### DIFF
--- a/data/assets/examples/geojson/geojson_points-relative.asset
+++ b/data/assets/examples/geojson/geojson_points-relative.asset
@@ -1,0 +1,46 @@
+local earth = asset.require("scene/solarsystem/planets/earth/earth")
+
+
+
+local data = asset.resource({
+  Name = "GeoJSON Example Outfacing",
+  Type = "UrlSynchronization",
+  Identifier = "geojson_example_points",
+  Url = {
+    "http://liu-se.cdn.openspaceproject.com/files/examples/geojson/points-relative.geojson",
+    "http://liu-se.cdn.openspaceproject.com/files/misc/target.png",
+  }
+})
+
+
+local ExamplePoints = {
+  Identifier = "Points-Example-Relative",
+  File = data .. "points-relative.geojson",
+  HeightOffset = 20000.0,
+  DefaultProperties = {
+    PointSize = 10.0
+  },
+  Name = "Example Points (Relative Path)"
+}
+
+
+asset.onInitialize(function()
+  openspace.globebrowsing.addGeoJson(earth.Earth.Identifier, ExamplePoints)
+end)
+
+asset.onDeinitialize(function()
+  openspace.globebrowsing.deleteGeoJson(earth.Earth.Identifier, ExamplePoints)
+end)
+
+asset.export(ExamplePoints)
+
+
+
+asset.meta = {
+  Name = "GeoJson Example - Points (Relative Path)",
+  Description = [[GeoJson example asset with points that are facing the camera
+    (default)]],
+  Author = "OpenSpace Team",
+  URL = "http://openspaceproject.com",
+  License = "MIT license"
+}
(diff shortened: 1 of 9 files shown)


data/assets/examples/geojson/geojson_points-relative.asset
```diff
@@ -0,0 +1,46 @@
...
+
+local data = asset.resource({
+  Name = "GeoJSON Example Outfacing",
+  Type = "UrlSynchronization",
+  Identifier = "geojson_example_points",
+  Url = {
+    "http://liu-se.cdn.openspaceproject.com/files/examples/geojson/points-relative.geojson",
+    "http://liu-se.cdn.openspaceproject.com/files/misc/target.png",
+  }
+})
```
WeirdRubberDuck: This was confusing to me, so I think a user would find it even more confusing. It's difficult to understand that the files will end up next to each other

Should we consider putting the files in the repo instead? Or would this be a bad idea? It makes sense to me, especially for this example, as the important part for the example is within the actual geojson file

Otherwise, can we at least put the files next to each other on the cdn server?

data/assets/examples/geojson/geojson_points-relative.asset
```diff
@@ -0,0 +1,46 @@
...
+end)
+
+asset.export(ExamplePoints)
+
+
+
+asset.meta = {
+  Name = "GeoJson Example - Points (Relative Path)",
+  Description = [[GeoJson example asset with points that are facing the camera
+    (default)]],
```
WeirdRubberDuck: Update this description to explain what the example is about

And do we want to format these examples the same way we do for the examples on the docs page?  Was looking at the description at the top first... :)

data/assets/examples/geojson/geojson_points-relative.asset
```diff
@@ -0,0 +1,46 @@
...
+
+
+local ExamplePoints = {
+  Identifier = "Points-Example-Relative",
+  File = data .. "points-relative.geojson",
+  HeightOffset = 20000.0,
+  DefaultProperties = {
+    PointSize = 10.0
+  },
+  Name = "Example Points (Relative Path)"
```
WeirdRubberDuck:
```suggestion
  Name = "Example Points (Relative Textuure Path)"
```
? Otherwise it could seem seems like we mean the path to the file. 

Dot he same change in the identifier and asset name

data/assets/examples/geojson/geojson_points-relative.asset
```diff
@@ -0,0 +1,46 @@
+local earth = asset.require("scene/solarsystem/planets/earth/earth")
+
+
+
+local data = asset.resource({
+  Name = "GeoJSON Example Outfacing",
```
WeirdRubberDuck: Fix name to match this example. This is copied from another one ;)

data/assets/examples/geojson/geojson_points-relative.asset
```diff
@@ -0,0 +1,46 @@
+local earth = asset.require("scene/solarsystem/planets/earth/earth")
```
WeirdRubberDuck: The name of this example file should be changed to use `_` instead of `-` for consistent naming with the other files

And maybe also add "texture" to the name (see comment below). One suggestion: `geojson_points_relative_texture_path`

WeirdRubberDuck (approved):
Implementation looks and works great but please address the comments in the example file

Approved given that the issues there are fixed

data/assets/examples/geojson/geojson_points-relative.asset
alexanderbock: I put them next to each other and also added a comment explaining it

data/assets/examples/geojson/geojson_points-relative.asset
alexanderbock: Yea, ultimately we do want it.  I think we can do that once we move the example files into a folder where they get automatically added to the documentation page;  almost like an indicator whether they are "old" examples or "new" examples.

Issue #3764, opened 2025-08-02
support for Win/Linux arm64 architectures


hn-88:
Currently, `modules/webgui/cmake/nodejs_support.cmake` supports only x86_64 for Linux and Windows. Adding arm64 support is likely to avoid issues like this - https://github.com/hn-88/OpenSpace-AppImage/issues/29

PR #3765, opened 2025-08-02
fix for #3764 - adding arm64 architecture support

hn-88:
adding arm64 architecture support


hn-88:
ARM64 Linux appimage issue https://github.com/hn-88/OpenSpace-AppImage/issues/29 seems to be fixed with this patch - https://github.com/hn-88/OpenSpace-AppImage/actions/runs/16700725308

alexanderbock:
Sweet!  Thanks!!

PR #3772, opened 2025-08-12
Issue/1188 improve console

engbergandreas:
This PR includes some improvements to the console. It now supports nested tab completion for functions and also tab completion for paths. Also changed it so that we will have line breaks instead of horizontal scrolling if there are too many characters to fit on one line. Furthermore, added some new shortcuts: 

Shift + Left/Right arrow key moves to the nearest special character
Shift + Backspace removes whole words until the previous special character
Shift + Delete removes whole words until the next special character 
Enter applies the suggested path or function if any exists
Shift + Tab and Tab now correctly cycle between suggestions 

Tab completion for tabs will cycle through all the folders and files in a given path, or if a partial path is given, it will filter based on what you've written so far.

Path completion: 
`openspace.asset.remove("C:/Users/anden88/Desktop/projects/ab <tab>` -- list all folder & files starting with `ab`
`openspace.asset.remove("C:/Users/anden88/Desktop/projects <tab>` -- list all folders & files in `/../projects`
`openspace.asset.remove("C:/Users/anden88/Desktop/projects/ <tab>` -- same as above, no need to end with the `/`

It can also autocomplete within functions and strings, but it requires the last `" ] or '` character to match correctly. For example 
`openspace.printInfo(openspace.walkDirectory('C:/Users/anden88/Desktop/projects/<tab>'))` works but requires the `'` before the last closing paranthesis. 

Function completion: 
`openspace.a <tab>` -- existing behavior, list all functions starting with `a`
`openspace.printInfo(openspace.statemachine <tab>` -- list all functions that is part of the `statemachine` library

caveat: for nested function tab completion to work, it must at least contain `openspace.` The following will unfortunately not autocomplete `openspace.printInfo(ope <tab>`

I tried to squash any new bugs introduced and fix any unexpected behaviour. I'm sure I've missed something, so please try it out. Also, let me know if there are any other shortcut combinations that you feel are missing. 

Here is what it looks like 
https://github.com/user-attachments/assets/22d2e851-9b5a-490f-b6e4-2a4b20e242f7

### DIFF
--- a/include/openspace/rendering/luaconsole.h
+++ b/include/openspace/rendering/luaconsole.h
@@ -67,6 +67,15 @@ private:
     void registerKeyHandlers();
     void registerKeyHandler(Key key, KeyModifier modifier, std::function<void()> callback);
 
+    // Helper functions for tab autocomplete
+    void autoCompleteCommand();
+    void resetAutoCompleteState();
+    size_t detectContext(std::string_view command);
+    bool gatherPathSuggestions(size_t contextStart);
+    void gatherFunctionSuggestions(size_t contextStart);
+    void filterSuggestions();
+    void cycleSuggestion();
+
     properties::BoolProperty _isVisible;
     properties::BoolProperty _shouldBeSynchronized;
     properties::BoolProperty _shouldSendToRemote;
@@ -85,19 +94,6 @@ private:
     // Map of registered keybinds and their corresponding callbacks
     std::map<KeyWithModifier, std::function<void()>> _keyHandlers;
 
-    struct {
-        int lastIndex;
-        bool hasInitialValue;
-        std::string initialValue;
-    } _autoCompleteInfo;
-
-    struct {
-        int index;
-        std::string suggestion;
-        bool isReverse;
-    } _suggestionInfo;
-
-
     enum class Context {
         None = 0,
         Function,
@@ -107,11 +103,11 @@ private:
     struct {
         Context context;
         bool isDataDirty;
-        std::string baseInput; // What the user typed
-        std::string typedFragment; // The partial word being completed (if any)
+        std::string input; // Part of the command that we're intrested in
         std::vector<std::string> suggestions;
         int currentIndex;
         std::string suggestion;
+        bool cycleReverse = false;
 
     } _autoCompleteState;
 
--- a/include/openspace/rendering/luaconsole.h
+++ b/include/openspace/rendering/luaconsole.h
@@ -34,6 +34,7 @@
 #include <openspace/util/keys.h>
 #include <ghoul/opengl/ghoul_gl.h>
 #include <ghoul/opengl/uniformcache.h>
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
@@ -63,6 +64,8 @@ public:
 private:
     void parallelConnectionChanged(const ParallelConnection::Status& status);
     void addToCommand(const std::string& c);
+    void registerKeyHandlers();
+    void registerKeyHandler(Key key, KeyModifier modifier, std::function<void()> callback);
 
     properties::BoolProperty _isVisible;
     properties::BoolProperty _shouldBeSynchronized;
@@ -79,6 +82,8 @@ private:
     std::vector<std::string> _commandsHistory;
     size_t _activeCommand = 0;
     std::vector<std::string> _commands;
+    // Map of registered keybinds and their corresponding callbacks
+    std::map<KeyWithModifier, std::function<void()>> _keyHandlers;
 
     struct {
         int lastIndex;
--- a/include/openspace/rendering/luaconsole.h
+++ b/include/openspace/rendering/luaconsole.h
@@ -98,6 +98,24 @@ private:
     } _suggestionInfo;
 
 
+    enum class Context {
+        None = 0,
+        Function,
+        Path
+    };
+
+    struct {
+        Context context;
+        bool isDataDirty;
+        std::string baseInput; // What the user typed
+        std::string typedFragment; // The partial word being completed (if any)
+        std::vector<std::string> suggestions;
+        int currentIndex;
+        std::string suggestion;
+
+    } _autoCompleteState;
+
+
     float _currentHeight = 0.f;
     float _targetHeight = 0.f;
     float _fullHeight = 0.f;
--- a/include/openspace/rendering/luaconsole.h
+++ b/include/openspace/rendering/luaconsole.h
@@ -75,6 +75,7 @@ private:
     void gatherFunctionSuggestions(size_t contextStart);
     void filterSuggestions();
     void cycleSuggestion();
+    void applySuggestion();
 
     properties::BoolProperty _isVisible;
     properties::BoolProperty _shouldBeSynchronized;
@@ -101,13 +102,14 @@ private:
     };
 
     struct {
-        Context context;
-        bool isDataDirty;
-        std::string input; // Part of the command that we're intrested in
-        std::vector<std::string> suggestions;
-        int currentIndex;
-        std::string suggestion;
+        Context context = Context::None;
+        bool isDataDirty = true;
+        std::string input = ""; // Part of the command that we're intrested in
+        std::vector<std::string> suggestions = {};
+        int currentIndex = -1;
+        std::string suggestion = "";
         bool cycleReverse = false;
+        size_t insertPosition = 0;
 
     } _autoCompleteState;
 
--- a/include/openspace/rendering/luaconsole.h
+++ b/include/openspace/rendering/luaconsole.h
@@ -91,6 +91,13 @@ private:
         std::string initialValue;
     } _autoCompleteInfo;
 
+    struct {
+        int index;
+        std::string suggestion;
+        bool isReverse;
+    } _suggestionInfo;
+
+
     float _currentHeight = 0.f;
     float _targetHeight = 0.f;
     float _fullHeight = 0.f;
--- a/src/rendering/luaconsole.cpp
+++ b/src/rendering/luaconsole.cpp
@@ -46,6 +46,7 @@
 namespace {
     constexpr std::string_view HistoryFile = "ConsoleHistory";
     constexpr std::string_view JumpCharacters = ".,'/()";
+    constexpr std::string_view PathIdentifier = "\"'[";
 
     constexpr int NoAutoComplete = -1;
 
@@ -183,9 +184,6 @@ LuaConsole::LuaConsole()
         glm::vec4(1.f)
     )
     , _historyLength(HistoryLengthInfo, 13, 0, 100)
-    , _autoCompleteInfo({NoAutoComplete, false, ""})
-    , _suggestionInfo({ NoAutoComplete, "" })
-    , _autoCompleteState({ Context::None, true, "", "", {}, NoAutoComplete })
 {
     addProperty(_isVisible);
     addProperty(_shouldBeSynchronized);
@@ -200,6 +198,8 @@ LuaConsole::LuaConsole()
 
     _historyTextColor.setViewOption(properties::Property::ViewOptions::Color);
     addProperty(_historyTextColor);
+
+    resetAutoCompleteState();
 }
 
 void LuaConsole::initialize() {
@@ -366,26 +366,7 @@ bool LuaConsole::keyboardCallback(Key key, KeyModifier modifier, KeyAction actio
     // If any other key is pressed, we want to remove our previous findings
     // The special case for Shift is necessary as we want to allow Shift+TAB
     if (key != Key::Tab && !modifierShift) {
-        _autoCompleteInfo = {
-                .lastIndex = NoAutoComplete,
-                .hasInitialValue = false,
-                .initialValue = ""
-        };
-        _suggestionInfo = {
-            .index = NoAutoComplete,
-            .suggestion = ""
-        };
-
-        _autoCompleteState = {
-            .context = Context::None,
-            .isDataDirty = true,
-            .baseInput = "",
-            .typedFragment = "",
-            .suggestions = {},
-            .currentIndex = 0,
-            .suggestion = ""
-        };
-
+        resetAutoCompleteState();
     }
         
     // Do not consume modifier keys
@@ -907,16 +888,12 @@ void LuaConsole::registerKeyHandlers() {
                 }
             }
 
+            // TODO fix where insertion should happen 
+            // openspace.asset.remove("C:/Users/anden88/Desktop/projects\Splitter2020")
+            // tries to insert /bin after ") characters
+
             // Clear suggestion
-            _autoCompleteState = {
-                .context = Context::None,
-                .isDataDirty = true,
-                .baseInput = "",
-                .typedFragment = "",
-                .suggestions = {},
-                .currentIndex = 0,
-                .suggestion = ""
-            };
+            resetAutoCompleteState();
         };
 
         if (_autoCompleteState.suggestion != "") {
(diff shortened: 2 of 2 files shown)


include/openspace/rendering/luaconsole.h
```diff
@@ -79,12 +92,27 @@ class LuaConsole : public properties::PropertyOwner {
...
+    };
 
     struct {
-        int lastIndex;
-        bool hasInitialValue;
-        std::string initialValue;
-    } _autoCompleteInfo;
+        Context context = Context::None;
+        bool isDataDirty = true;
+        std::string input = ""; // Part of the command that we're intrested in
```
alexanderbock:
```suggestion
        std::string input; // Part of the command that we're intrested in
```

include/openspace/rendering/luaconsole.h
```diff
@@ -79,12 +92,27 @@ class LuaConsole : public properties::PropertyOwner {
...
+        std::vector<std::string> suggestions = {};
```
alexanderbock:
```suggestion
        std::vector<std::string> suggestions;
```

include/openspace/rendering/luaconsole.h
```diff
@@ -79,12 +92,27 @@ class LuaConsole : public properties::PropertyOwner {
...
+        int currentIndex = -1;
+        std::string suggestion = "";
```
alexanderbock:
```suggestion
        std::string suggestion;
```

include/openspace/rendering/luaconsole.h
```diff
@@ -79,12 +92,27 @@ class LuaConsole : public properties::PropertyOwner {
...
+        bool cycleReverse = false;
+        size_t insertPosition = 0;
+
```
alexanderbock:
```suggestion
```

src/rendering/luaconsole.cpp
```diff
@@ -348,255 +350,23 @@ bool LuaConsole::keyboardCallback(Key key, KeyModifier modifier, KeyAction actio
...
-    }
 
-    // Go to the previous character
-    if (key == Key::Left || (modifierControl && key == Key::B)) {
-        if (_inputPosition > 0) {
-            --_inputPosition;
-        }
-        return true;
-    }
+    KeyWithModifier keyCombination{ key, modifier };
```
alexanderbock:
```suggestion
    KeyWithModifier keyCombination = KeyWithModifier(key, modifier);
```

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +487,95 @@ void LuaConsole::render() {
...
-
-        // Compute the overflow in pixels
-        const float overflow = currentWidth - res.x * 0.995f;
-        if (overflow <= 0.f) {
-            break;
+    // Render the command and suggestions with line breaks where required
+    const size_t totalCommandSize = 2 + currentCommand.size() +
+        _autoCompleteState.suggestion.size();
+    const size_t nCharactersPerRow = static_cast<size_t>(res.x * 0.95 /
+        static_cast<double>(_font->glyph('m')->width));
```
alexanderbock:
```suggestion
    const size_t nCharactersPerRow = static_cast<size_t>(res.x * 0.95f /
        static_cast<float>(_font->glyph('m')->width));
```

Since we don't need to do the calculation in `double precision?

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +487,95 @@ void LuaConsole::render() {
...
+    inputLocation.y += markerStartRow * EntryFontSize * 1.25f * dpiScaling.y;
     // Just offset the ^ marker slightly for a nicer look
     inputLocation.y += 3 * dpiScaling.y;
 
     // Render the ^ marker below the text to show where the current entry point is
     RenderFont(
         *_font,
         inputLocation,
-        (std::string(_inputPosition - nChoppedCharsBeginning + 2, ' ') + "^"),
+        (std::string((_inputPosition + 2)  % nCharactersPerRow, ' ') + "^"),
```
alexanderbock:
```suggestion
        (std::string((_inputPosition + 2) % nCharactersPerRow, ' ') + "^"),
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
     _shouldSendToRemote = (status == ParallelConnection::Status::Host);
 }
 
+void LuaConsole::registerKeyHandler(Key key, KeyModifier modifier,
+    std::function<void()> callback)
```
alexanderbock:
```suggestion
                                    std::function<void()> callback)
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+{
+    _keyHandlers[{ key, modifier }] = std::move(callback);
+}
+
+void LuaConsole::registerKeyHandlers() {
+
```
alexanderbock:
```suggestion
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    registerKeyHandler(Key::Escape, KeyModifier::None, [this]() {
+        _isVisible = false;
+    });
```
alexanderbock:
```suggestion
    registerKeyHandler(
        Key::Escape,
        KeyModifier::None,
        [this]() { _isVisible = false; }
    );
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    registerKeyHandler(Key::C, KeyModifier::Control, [this]() {
+        ghoul::setClipboardText(_commands.at(_activeCommand));
+    });
+
+    // Cut to clipboard
+    registerKeyHandler(Key::X, KeyModifier::Control, [this]() {
+        ghoul::setClipboardText(_commands.at(_activeCommand));
+        _commands.at(_activeCommand).clear();
+        _inputPosition = 0;
+    });
```
alexanderbock: Same here

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+
+    // Cut part after cursor to clipboard ("Kill")
+    registerKeyHandler(Key::K, KeyModifier::Control, [this]() {
+        auto here = _commands.at(_activeCommand).begin() + _inputPosition;
+        auto end = _commands.at(_activeCommand).end();
+        ghoul::setClipboardText(std::string(here, end));
+        _commands.at(_activeCommand).erase(here, end);
+    });
```
alexanderbock:
```suggestion
    registerKeyHandler(
        Key::K,
        KeyModifier::Control,
        [this]() {
            auto here = _commands.at(_activeCommand).begin() + _inputPosition;
            auto end = _commands.at(_activeCommand).end();
            ghoul::setClipboardText(std::string(here, end));
            _commands.at(_activeCommand).erase(here, end);
        }
    );
```

alexanderbock: Same for the rest below

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+
+    // Go to end of command string
+    registerKeyHandler(Key::E, KeyModifier::Control, [this]() {
+        _inputPosition = _commands.at(_activeCommand).size();
+    });
+
+
+    auto executeCommand = [this]() {
+        if (_autoCompleteState.suggestion != "") {
```
alexanderbock:
```suggestion
        if (!_autoCompleteState.suggestion.empty()) {
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+        }
+
+        const std::string cmd = _commands.at(_activeCommand);
+        if (!cmd.empty()) {
+            using Script = scripting::ScriptEngine::Script;
+            global::scriptEngine->queueScript({
+                .code = cmd,
+                .synchronized = Script::ShouldBeSynchronized(_shouldBeSynchronized),
+                .sendToRemote = Script::ShouldSendToRemote(_shouldSendToRemote)
+                });
```
alexanderbock:
```suggestion
            });
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    // everything is set back to normal
+
+    const std::string currentCommand = _commands[_activeCommand];
+    // Determine if are currently in a function or path context
+    if (_autoCompleteState.isDataDirty) {
+        const size_t contextStart = detectContext(currentCommand);
+
+        switch (_autoCompleteState.context) {
+        case Context::Path:
+            if (!gatherPathSuggestions(contextStart)) {
```
alexanderbock:
```suggestion
            const bool hasSuggestions = gatherPathSuggestions(contextStart);
            if (!hasSuggestions) {
```

From the name of the function it is not super clear what `if !gatherPathSuggestions` means

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    // implements the 'hop-over' behavior. As soon as another key is pressed,
+    // everything is set back to normal
+
+    const std::string currentCommand = _commands[_activeCommand];
+    // Determine if are currently in a function or path context
+    if (_autoCompleteState.isDataDirty) {
+        const size_t contextStart = detectContext(currentCommand);
+
+        switch (_autoCompleteState.context) {
+        case Context::Path:
```
alexanderbock: All the case labels need to be indented one more level

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+            if (!gatherPathSuggestions(contextStart)) {
+                return;
+            }
+            break;
+
```
alexanderbock:
```suggestion
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+void LuaConsole::resetAutoCompleteState() {
+    _autoCompleteState = {
+                .context = Context::None,
+                .isDataDirty = true,
+                .input = "",
+                .suggestions = {},
+                .currentIndex = -1,
+                .suggestion = "",
+                .cycleReverse = false,
+                .insertPosition = 0,
```
alexanderbock:
```suggestion
        .context = Context::None,
        .isDataDirty = true,
        .input = "",
        .suggestions = {},
        .currentIndex = -1,
        .suggestion = "",
        .cycleReverse = false,
        .insertPosition = 0
```

Might be worth defining the `_autoCompleteState` struct with a proper name, like `AutoCompleteState` and then this function could disappear and be replaced with `_autoCompleteState = AutoCompleteState();`

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    std::vector<std::filesystem::path> suggestions =
+        ghoul::filesystem::walkDirectory(
+            dirToSearch,
+            ghoul::filesystem::Recursive::No,
+            ghoul::filesystem::Sorted::Yes
+        );
+
+    auto containsNonAscii = [](const std::filesystem::path& p) {
+        auto s = p.generic_u8string();
+        for (auto it = s.rbegin(); it != s.rend(); it++) {
```
alexanderbock, line 1134: Why does this have to be a reverse search?  A forward search will probably be faster

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+            if (static_cast<unsigned char>(*it) > 0x7F) {
+                return true;
+            }
+        }
+        return false;
+    };
+
+    std::vector<std::string> entries;
+    for (const auto& entry : suggestions) {
```
alexanderbock:
```suggestion
    for (const std::filesystem::path& entry : suggestions) {
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+        _autoCompleteState.insertPosition = contextStart + functionEnd;
+    }
+    else {
+        // There is nothing after the path so we render normally at the end
+        _autoCompleteState.insertPosition = currentCommand.size();
+    }
+}
+
+void LuaConsole::filterSuggestions() {
+    auto normalize = [this](const std::string& s) -> std::string {
```
alexanderbock:
```suggestion
    auto normalize = [this](const std::string& s) {
```

If it is not necesasry, we normally don't explicitly specify the return value for lambdas

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +487,95 @@ void LuaConsole::render() {
...
+    // positions, even if linebreaks are added    
+    // Pad suggestion before and after with ' '
+    const std::string suggestion = std::string(firstPart.size() + 2, ' ') +
+        _autoCompleteState.suggestion + std::string(secondPart.size(), ' ');
+    // Pad first part at the end with ' ' 
+    firstPart.insert(firstPart.end(), totalCommandSize - firstPart.size(), ' ');
+    // Pad second part in the beginning with ' ' 
+    secondPart.insert(secondPart.begin(), totalCommandSize - secondPart.size(), ' ');
+
+    auto linebreakCommand = [nCharactersPerRow](const std::string& s) -> std::string {
```
alexanderbock:
```suggestion
    auto linebreakCommand = [nCharactersPerRow](const std::string& s) {
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+        if (candidate.starts_with(input)) {
+            results.insert(result);
+        }
+    }
+
+    _autoCompleteState.suggestions.clear();
+    for (const std::string& candidate : results) {
+        _autoCompleteState.suggestions.push_back(candidate);
+    }
+    //_autoCompleteState.suggestion.assign(results.begin(), results.end());
```
alexanderbock:
```suggestion
    _autoCompleteState.suggestions = std::vector<std::string>(results.begin(), results.end());
```

?

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +674,517 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+        _autoCompleteState.suggestion.size();
+
+    if (_autoCompleteState.context == Context::Function) {
+        if (!_autoCompleteState.suggestion.ends_with('.')) {
+            // We're in a leaf function add paranthesis 
+            currentCommand.insert(_inputPosition, "()");
+            // Set the cursor position to be between the brackets
+            _inputPosition++;
+        }
+    }
```
alexanderbock:
```suggestion
```suggestion
    if (_autoCompleteState.context == Context::Function &&
        !_autoCompleteState.suggestion.ends_with('.'))
    {
        // We're in a leaf function add paranthesis 
        currentCommand.insert(_inputPosition, "()");
        // Set the cursor position to be between the brackets
        _inputPosition++;
    }
```

src/rendering/luaconsole.cpp
alexanderbock: Also, how about 0.925f instead?  With 0.95f on my system the last one or two characters were outside the view

src/rendering/luaconsole.cpp
engbergandreas: Huh, for some reason it keeps tabbing them back to the same level whenever I close a bracket :(

src/rendering/luaconsole.cpp
engbergandreas: I was thinking that it's more probable that a filename or a deeply nested folder might contain non-ASCII characters than the beginning. Based on absolutely nothing 🤷

src/rendering/luaconsole.cpp
engbergandreas: I wish, but that gives E0349 no matching = operator 😢 same reason the `.assign` didn't work

src/rendering/luaconsole.cpp
alexanderbock: Yea, the way it is at the moment is the default for Visual Studio.

Tools -> Options -> Text Editor -> C/C++ -> Code Style -> Formatting -> Indentation -> "Indent case labels"  should be checked.

src/rendering/luaconsole.cpp
```diff
@@ -348,255 +349,23 @@ bool LuaConsole::keyboardCallback(Key key, KeyModifier modifier, KeyAction actio
...
-    if (key >= Key::F1 && key <= Key::F25) {
-        return false;
+    // If any other key is pressed, we want to remove our previous findings
+    // The special case for Shift is necessary as we want to allow Shift+TAB
+    const bool isShiftModifierOnly = (key == Key::LeftShift || key == Key::RightShift);
+    if (key != Key::Tab && !isShiftModifierOnly) {
+        _autoCompleteState = AutoCompleteState();
     }
-
+
```
WeirdRubberDuck: Remove the extra whitespace on this line plz. 
I recommend using an extension that helps you trim trailing whitespaces automatically. Personally, I like this one: 
image
```suggestion

```

src/rendering/luaconsole.cpp
```diff
@@ -348,255 +349,23 @@ bool LuaConsole::keyboardCallback(Key key, KeyModifier modifier, KeyAction actio
...
-                .lastIndex = NoAutoComplete,
-                .hasInitialValue = false,
-                .initialValue = ""
-            };
-        }
+    // Call the registered function for the key combination pressed
+    auto it = _keyHandlers.find(keyCombination);
+    if (it != _keyHandlers.end()) {
+        it->second(); 
+        //return true; // TODO do we want return here or not?
```
WeirdRubberDuck: This should not be here now that you're done? 

Figure out if it should return or not :)

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +486,95 @@ void LuaConsole::render() {
...
-        // Compute the current width of the string and console prefix.
-        const float currentWidth =
-            _font->boundingBox("> " + currentCommand).x + inputLocation.x;
 
-        // Compute the overflow in pixels
-        const float overflow = currentWidth - res.x * 0.995f;
-        if (overflow <= 0.f) {
-            break;
+    // Render the command and suggestions with line breaks where required
+    const size_t totalCommandSize = 2 + currentCommand.size() +
```
WeirdRubberDuck, line 503: What's the magic 2 here and below?  Would be good to have that explained somehow

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +486,95 @@ void LuaConsole::render() {
...
+        _autoCompleteState.suggestion.size();
+    const size_t nCharactersPerRow = static_cast<size_t>(res.x * 0.925f /
```
WeirdRubberDuck: Here's another magic number that is not motivated. Why 0.925?

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +486,95 @@ void LuaConsole::render() {
...
+        static_cast<float>(_font->glyph('m')->width));
```
WeirdRubberDuck: Also, how about updating the line breaking here: 
```suggestion
    const size_t nCharactersPerRow = static_cast<size_t>(
        res.x * 0.925f / static_cast<float>(_font->glyph('m')->width)
    );
```
much easier to read and aligns better with coding guidelines

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +486,95 @@ void LuaConsole::render() {
...
+    // The command is split in 3 parts to render the suggestion in a different color
+    std::string firstPart = _autoCompleteState.insertPosition != 0 ?
+        currentCommand.substr(0, _autoCompleteState.insertPosition) : currentCommand;
+    std::string secondPart = _autoCompleteState.insertPosition != 0 ?
+        currentCommand.substr(_autoCompleteState.insertPosition) : "";
+
+    // We pad the strings with empty spaces so that each part is rendered in their correct
+    // positions, even if linebreaks are added    
+    // Pad suggestion before and after with ' '
+    const std::string suggestion = std::string(firstPart.size() + 2, ' ') +
```
WeirdRubberDuck: Same here, the motivation for the 2 is not obvious to me from reading this

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +486,95 @@ void LuaConsole::render() {
...
-        nChoppedCharsBeginning += nCharsToChopBeginning;
-        nChoppedCharsEnd += nCharsToChopEnd;
+        return result;
+    };
 
-        const size_t displayLength =
-            _commands.at(_activeCommand).size() -
-            nChoppedCharsBeginning - nChoppedCharsEnd;
+    // Offset the command depending on how many rows we require for a nicer look
+    inputLocation.y += nCommandRows * EntryFontSize * 1.25f * dpiScaling.y;
```
WeirdRubberDuck, line 576: More magic numbers :) why is this needed?

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
     _shouldSendToRemote = (status == ParallelConnection::Status::Host);
 }
 
+void LuaConsole::registerKeyHandler(Key key, KeyModifier modifier,
+                                                           std::function<void()> callback)
```
WeirdRubberDuck: Fix parameter alignment: 
```suggestion
void LuaConsole::registerKeyHandler(Key key, KeyModifier modifier,
                                    std::function<void()> callback)
```

https://docs.openspaceproject.com/latest/contribute/development/coding-style.html#handling-line-breaks

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+        }
+    );
+
+    // Go to the next JumpCharacter character
+    registerKeyHandler(
+        Key::Right,
+        KeyModifier::Control,
+        [this]() {
+            auto it = _commands.at(_activeCommand).find_first_of(JumpCharacters,
+                _inputPosition + 1);
```
WeirdRubberDuck: Again, line breaking issues. This does not adhere to code staandard. but not sur ehow to best fix. Maybe: 

```suggestion
            auto it = _commands.at(_activeCommand).find_first_of(
                JumpCharacters,
                _inputPosition + 1
            );
```
but there is still some nesting that goes against the "one statement per line" rule that makes htings easier to debug, so maybe rather split it up like you did above.: 

```suggestion
            std::string current = _commands.at(_activeCommand);
            auto it = current.find_first_of(JumpCharacters, _inputPosition + 1);
```
this is best (easiest to read) IMO

src/rendering/luaconsole.cpp
WeirdRubberDuck: I think that maybe we do want to return here?  If the command trigger any of the handlers, it should be consumed, right?

Not 100% sure, just a feeling

include/openspace/rendering/luaconsole.h
```diff
@@ -79,12 +91,29 @@ class LuaConsole : public properties::PropertyOwner {
...
+        AutoCompleteState();
+
+        Context context;
+        bool isDataDirty;
+        std::string input; // Part of the command that we're intrested in
+        std::vector<std::string> suggestions;
+        int currentIndex;
+        std::string suggestion;
+        bool cycleReverse;
+        size_t insertPosition;
```
WeirdRubberDuck: Maybe these are obvious to future us, but how about adding a one-sentence description of each member variable?  I have no idea what "cyclereverse" means from looking at this, for example

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    registerKeyHandler(
+        Key::Left,
+        KeyModifier::Control,
+        [this]() {
+            std::string current = _commands.at(_activeCommand);
+            std::reverse(current.begin(), current.end());
+            const size_t start = current.size() - (_inputPosition - 1);
+            auto it = current.find_first_of(JumpCharacters, start);
+            if (it != std::string::npos) {
+                _inputPosition = current.size() - it;
```
WeirdRubberDuck: Maybe this is nitpicky, but the `find_first_of`` method for string does not return an iterator, but a `size_t`. So I think this variable name is a bit misleading, and the `auto` is not motivated. 

Something like this might be better? Or even using a more elaborate name like `jumpCharPos` :)  
```suggestion
            size_t pos = current.find_first_of(JumpCharacters, start);
            if (pos != std::string::npos) {
                _inputPosition = current.size() - pos;
```
the same comment goes for all the other places where you have used `string.find_first_pos`

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    // Find the path starting point which can start with either " ' or [ character
+    size_t pathStartIndex = 0;
+    for (size_t i = _inputPosition; i > 0; --i) {
+        if (PathStartIdentifier.find(command[i - 1]) != std::string::npos) {
+            pathStartIndex = i;
+            break;
+        }
+    }
+
+    // TODO anden88 2025-08-08: Detect functions in a smarter way that allows nested
```
WeirdRubberDuck: If this TODO comment should stay around, propertly format it: 
```suggestion
    // @TODO (anden88, 2025-08-08): Detect functions in a smarter way that allows nested
```

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    // function calls. The following example currently does not work.
+    // If the user typed e.g., "openspace.printInfo(open"
+    // We will not be able to autocomplete the last openspace. since the first
```
WeirdRubberDuck:
```suggestion
    // If the user typed e.g., "openspace.printInfo(open",
    // we will not be able to autocomplete the last openspace. since the first
```
?

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+    // Get the entries in directory
+    std::vector<std::filesystem::path> suggestions =
+        ghoul::filesystem::walkDirectory(
+            dirToSearch,
+            ghoul::filesystem::Recursive::No,
+            ghoul::filesystem::Sorted::Yes
+        );
+
+    auto containsNonAscii = [](const std::filesystem::path& p) {
+        auto s = p.generic_u8string();
```
WeirdRubberDuck:
```suggestion
        std::u8string s = p.generic_u8string();
```
Avoid auto
https://docs.openspaceproject.com/latest/contribute/development/coding-style.html#use-the-auto-keyword-sparingly

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+
+    // Get the entries in directory
+    std::vector<std::filesystem::path> suggestions =
+        ghoul::filesystem::walkDirectory(
+            dirToSearch,
+            ghoul::filesystem::Recursive::No,
+            ghoul::filesystem::Sorted::Yes
+        );
+
+    auto containsNonAscii = [](const std::filesystem::path& p) {
```
WeirdRubberDuck, line 1132: This seems like a useful function. Should it be moved to ghoul/misc/stringhelper.h?

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+LuaConsole::AutoCompleteState::AutoCompleteState()
+    : context{ Context::None }
+    , isDataDirty{ true }
+    , input{ "" }
+    , suggestions{ }
+    , currentIndex{ NoAutoComplete }
+    , suggestion{ "" }
+    , cycleReverse{ false }
+    , insertPosition{ 0 }
+{ }
```
WeirdRubberDuck: Move this constructor definition to the top of the file, before the LuaConsole constructor.

that's how we usuelly do it to have all the construction code in one place

src/rendering/luaconsole.cpp
```diff
@@ -893,4 +673,584 @@ void LuaConsole::parallelConnectionChanged(const ParallelConnection::Status& sta
...
+void LuaConsole::filterSuggestions() {
+    auto normalize = [this](const std::string& s) {
+        std::string out = s;
+
+        if (_autoCompleteState.context == Context::Function) {
+            out = ghoul::toLowerCase(out);
+        }
+
+#ifdef WIN32
+        // Treat lowercase paths the same in Windows since for the most part
```
WeirdRubberDuck: Clarify: "the same as..." what?

src/rendering/luaconsole.cpp
```diff
@@ -717,86 +486,95 @@ void LuaConsole::render() {
...
 
-        currentCommand = _commands.at(_activeCommand).substr(
-            nChoppedCharsBeginning,
-            displayLength
-        );
-    }
+    RenderFont(
+        *_font,
+        inputLocation,
+        linebreakCommand("> " + firstPart),
```
WeirdRubberDuck: I guess these characters are the magic 2?  Still would be nice with some explanation, especially if we happen ot ever change this :)

src/rendering/luaconsole.cpp
WeirdRubberDuck: Noted that there are trailing whitespaces in a few other places in this file as well - please fix, preferably using an extension as mentioned below

These are the changes I get just from _saving_ this file (not even making any changes) :) . Quite annoying for git changelog reasons
image

src/rendering/luaconsole.cpp
WeirdRubberDuck: also, see comment above about that it does not return an iterator, but a size_t

WeirdRubberDuck (approved):
Very nice!  Did some testing as well and like how it works

src/rendering/luaconsole.cpp
engbergandreas: If so, it should be thoroughly checked that it is valid on both Windows and Linux systems. This was a combination of lots of googling and GPT use. I'm not sure I trust it. @alexanderbock thoughts?

engbergandreas: I just remember from some uni courses that casting and or comparing a string of chars it to a char/unsigned char can in some cases be dangerous and lead to unintended behavior. Again very uncertain so bring your 🧂

WeirdRubberDuck:
Approved but after some more testing I had some thoughts about the behavior.  We concluded that it might be useful with some more eyes on this and some team-level discussion

I realized that I find it a bit annoying to have to press enter to accept a suggestion now. It leads to three extra keypresses for typing e.g. a Lua function `openspace.asset.add`

But might also be that I'm so used to how it worked before

src/rendering/luaconsole.cpp
alexanderbock: I think this should probably be fine since `generic_u8string` returns a `std::string<char8_t>` that encodes UTF-8 and in that, the first 128 characters are mapping identically to ASCII

alexanderbock:
There was some check to guard against a 1x1 render target that happens for one frame when minimizing the window in here:  https://github.com/OpenSpace/OpenSpace/commit/f6099b14a08c237e86dd40f7a4089124cd1b56ee

src/rendering/luaconsole.cpp
```diff
@@ -716,87 +495,100 @@ void LuaConsole::render() {
...
+    // Render the command and suggestions with line breaks where required, + 2 accounts
+    // for the '> ' characters in the beginning of the command
+    const size_t totalCommandSize = 2 + currentCommand.size() +
+        _autoCompleteState.suggestion.size();
+    // Scalefactor 0.925f chosen arbitraily to fit characters on screen with some margin
+    const size_t nCharactersPerRow = std::max(
+        static_cast < size_t>(1),
+        static_cast<size_t>(res.x * 0.925f / static_cast<float>(_font->glyph('m')->width))
+    );
+    const size_t nCommandRows = totalCommandSize / nCharactersPerRow;
```
Roxeena: Don't we want to use a `std::ceil` here instead of cutting the decimals? If it turns out that this division is 2.3 then we would want 2 full rows of text and one 30% (-ish) full row, right? Meaning 3 rows where the last one is not fully complete.  

I do get some strange behavior when I test it to be needing 2 line breaks 
Skärmbild 2025-08-26 135919

src/rendering/luaconsole.cpp
```diff
@@ -716,87 +495,100 @@ void LuaConsole::render() {
...
+
+    // The command is split in 3 parts to render the suggestion in a different color
+    std::string firstPart = _autoCompleteState.insertPosition != 0 ?
```
Roxeena: Personally, for readability I would make this a regular if else statement, since the same check is performed twice. But this might just be my preference.

src/rendering/luaconsole.cpp
```diff
@@ -716,87 +495,100 @@ void LuaConsole::render() {
...
+        currentCommand.substr(0, _autoCompleteState.insertPosition) : currentCommand;
+    std::string secondPart = _autoCompleteState.insertPosition != 0 ?
+        currentCommand.substr(_autoCompleteState.insertPosition) : "";
+
+    // We pad the strings with empty spaces so that each part is rendered in their correct
+    // positions, even if linebreaks are added
+    // Pad suggestion before and after with ' '
+    const std::string suggestion = std::string(firstPart.size() + 2, ' ') +
+        _autoCompleteState.suggestion + std::string(secondPart.size(), ' ');
```
Roxeena: Im confused about the `+ std::string(secondPart.size(), ' ')` what is the second part? Isn't that just the rest of the string after the point suggestions should be inserted in the original command? But this would be empty? If I write: `openspace.a` then the `insertPosition` would be after the `.a`, right? And there is nothing after that, it is the end of the string. Or is the `currentCommand` including a suggestion like `openspace.absPath` or `openspace.action`?

src/rendering/luaconsole.cpp
```diff
@@ -716,87 +495,100 @@ void LuaConsole::render() {
...
+    // Pad first part at the end with ' '
+    firstPart.insert(firstPart.end(), totalCommandSize - firstPart.size(), ' ');
+    // Pad second part in the beginning with ' '
+    secondPart.insert(secondPart.begin(), totalCommandSize - secondPart.size(), ' ');
```
Roxeena: Shouldn't this be the same as the `const std::string suggestion` in line 520 above?

src/rendering/luaconsole.cpp
```diff
@@ -716,87 +495,100 @@ void LuaConsole::render() {
...
+        size_t count = 0;
+        for (char c : s) {
+            result += c;
+            count++;
 
-        // Compute the current width of the string and console prefix.
-        const float currentWidth =
-            _font->boundingBox("> " + currentCommand).x + inputLocation.x;
+            if (c == '\n') {
+                count = 0;
```
Roxeena, line 562:
```suggestion
                count = 0;
                continue;
```
Makes it clearer that there is no intention to continue this iteration.

src/rendering/luaconsole.cpp
```diff
@@ -716,87 +495,100 @@ void LuaConsole::render() {
...
         *_font,
         inputLocation,
-        "> " + currentCommand,
+        linebreakCommand(secondPart),
         _entryTextColor,
         ghoul::fontrendering::CrDirection::Down
     );
 
+    // Move the marker to the correct row if there are multiple
+    const size_t markerStartRow = nCommandRows - (_inputPosition + 2) / nCharactersPerRow;
```
Roxeena, line 602: From my testing, there seems to be an issue here if the line perfectly fits. (I made the window really slim to test this out)

Skärmbild 2025-08-26 135059

src/rendering/luaconsole.cpp
```diff
@@ -811,8 +603,10 @@ void LuaConsole::render() {
         commandSubset = _commandsHistory;
     }
     else {
+        // Historic lines are reduced by the number of rows the command is occupying
         commandSubset = std::vector<std::string>(
-            _commandsHistory.end() - _historyLength,
+            _commandsHistory.end() - _historyLength + std::min(nCommandRows,
```
Roxeena: Strange line breaking, maybe make the calculation in a separate row and used the stored value? Do we have any style rules for these type of cases?

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+
+    registerKeyHandler(Key::Tab, KeyModifier::Shift, [this]() {
+        _autoCompleteState.cycleReverse = true;
+        autoCompleteCommand();
+    });
+}
+
+void LuaConsole::autoCompleteCommand() {
+    // We get a list of all the available commands or paths and initially find the
+    // first match that starts with how much we typed sofar. We store the index so
```
Roxeena:
```suggestion
    // first match that starts with how much we typed so far. We store the index so
```

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+    // that in subsequent "tab" presses, we will discard previous matches. This
+    // implements the 'hop-over' behavior. As soon as another key is pressed,
+    // everything is set back to normal
+
+    const std::string currentCommand = _commands[_activeCommand];
+    // Determine if are currently in a function or path context
```
Roxeena:
```suggestion
    // Determine if we are currently in a function or path context
```

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+    if (_autoCompleteState.isDataDirty) {
+        const size_t contextStart = detectContext(currentCommand);
+
+        switch (_autoCompleteState.context) {
+        case Context::Path: {
```
Roxeena:
```suggestion
            case Context::Path: {
```

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+                const bool hasSugestions = gatherPathSuggestions(contextStart);
+                if (!hasSugestions) {
+                    return;
+                }
+                break;
+            }
+            case Context::Function:
```
Roxeena, line 1055: Why do the path suggestions return when it is empty and the function suggestions does not? Why not handle them the same? The behavior shown to the user would be same in both cases, right?

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+        }
+
+        filterSuggestions();
+        _autoCompleteState.isDataDirty = false;
+    }
+    cycleSuggestion();
+}
+
+size_t LuaConsole::detectContext(std::string_view command) {
+    // Find the path starting point which can start with either " ' or [ character
```
Roxeena:
```suggestion
    // Find the path starting point which can start with either " ' or [ 
```

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+    size_t pathStartIndex = 0;
+    for (size_t i = _inputPosition; i > 0; --i) {
```
Roxeena: Is the loop really needed here? Won't the `find` command search the entire (sub)string anyway?

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+                throw ghoul::RuntimeError("Unhandled context");
+        }
+
+        filterSuggestions();
+        _autoCompleteState.isDataDirty = false;
+    }
+    cycleSuggestion();
+}
+
+size_t LuaConsole::detectContext(std::string_view command) {
```
Roxeena, line 1068: I would assume that this function returns either `Context::Path` or `Context::Function`, but it returns the beginning of the context but not what kind of context it is? I'm confused, maybe rename the function? `findStartOfContext` or something similar could be a suggestion.

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+    }
+
+    // @TODO (anden88, 2025-08-08): Detect functions in a smarter way that allows nested
+    // function calls. The following example currently does not work.
+    // If the user typed e.g., "openspace.printInfo(open", we will not be able to
+    // autocomplete the last openspace. since the first instance we find is at the
+    // beginning, resulting in the fragment being wrongly assumed as "printInfo(open"
+    size_t functionStartIndex = command.rfind("openspace.");
+
+    if (functionStartIndex == std::string::npos) {
```
Roxeena, line 1085: But this would never identify any other function? It would just find the last `openspace.` command? I'm confused as to how that is useful?

src/rendering/luaconsole.cpp
```diff
@@ -885,12 +679,584 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+            // We're only interested in autocomplete until the next separator "."
+            const size_t offset = input.size();
+            const size_t pos = suggestionNormalized.find('.', offset);
+
+            if (pos != std::string::npos) {
+                result = result.substr(0, pos + 1); // include the "."
+            }
+        }
+
+        if (suggestionNormalized.starts_with(input)) {
```
Roxeena, line 1231: Isn't it better to have an early out? So Before you start looking for dots, you check if the suggestion starts with the input string, if not you continue to the next suggestion. This feel like it will go faster since many suggestions will be skipped before even trying to find any dots.

Roxeena (commented):
Tested it out a bit and it is really nice. Personally I am more used to "applying the suggestion" by pressing the Right key than the Enter key. But this could be confusing if you are trying to step to the Right and accidentally apply the suggestion that you did not want.

engbergandreas:
Add right key to apply suggestion the way Enter key does now. Also fix crashes

WeirdRubberDuck (changes requested):
Fix bugs mentioned in the meeting plz. We want no crashes :)

src/rendering/luaconsole.cpp
engbergandreas: The second part was not needed here, good catch! For clarification thought we could have something like this `openspace.action.action(openspace.<TAB>)` then the first part would be  `openspace.action.action(openspace.` and the second part would be `)`

engbergandreas: I just remember why it looked like this. Had a discussion on this with Alex when I wrote it and we decided to have them all be the same size for symmetry's sake.

src/rendering/luaconsole.cpp
engbergandreas: We're looking backwards in the command and check each character if it exists in the `PathStartIdentifier`

engbergandreas: We can't use the std::string::find_last_of here https://en.cppreference.com/w/cpp/string/basic_string/find_last_of since that would break if we have something like `openspace.absPath("C:/")` It would put the cursor after the last `"`

src/rendering/luaconsole.cpp
engbergandreas: Yeah I agree but it also has the side effect of setting `_autoCompleteState.context` so I'm not sure what is best tbh

src/rendering/luaconsole.cpp
engbergandreas: Since every openspace lua function contains `openspace.` as its first lua library. But yes if you're thinking of things like `math.acos` then no it wont find that. And I think its fine to not support it as well

engbergandreas: it allows us to do something like `openspace.printInfo(openspace.action.<TAB>` which wasn't possible before

src/rendering/luaconsole.cpp
engbergandreas: I'm not sure I follow, this handles both Functions and Paths

src/rendering/luaconsole.cpp
```diff
@@ -716,88 +496,106 @@ void LuaConsole::render() {
...
+    );
+    // We're intrested in the zero based index when computing the input position
+    // If the characters fit on one line we should not add any extra rows
+    nCommandRows = nCommandRows > 1 ? nCommandRows - 1 : 0;
+
+    // The command is split in 3 parts to render the suggestion in a different color
+    std::string firstPart = _autoCompleteState.insertPosition != 0 ?
+        currentCommand.substr(0, _autoCompleteState.insertPosition) : currentCommand;
+    std::string secondPart = _autoCompleteState.insertPosition != 0 ?
+        currentCommand.substr(_autoCompleteState.insertPosition) : "";
```
WeirdRubberDuck: Agreeing with Malin's comment about splitting this up into an if-statement for readability. How about: 

```suggestion
    // The command is split in 3 parts to render the suggestion in a different color
    std::string firstPart = currentCommand;
    std::string secondPart = "";
    if (_autoCompleteState.insertPosition != 0) {
        firstPart = currentCommand.substr(0, _autoCompleteState.insertPosition);
        secondPart = currentCommand.substr(_autoCompleteState.insertPosition);
    }
```
? 

Also, the comment need some explanation of what you mean with three parts - it is quite confusing that you now have variables for part 1 and 2 but not for part three. 

If I understand correctly, the first part is the text before the suggestion, the second part is after the suggestion, and then the final part is the suggestion itself?  Clarify that in the comment, and maybe rename the `secondPart` variable to `lastPart` ? to clarify that it's at the end and not in the center?

WeirdRubberDuck: Updated suggestion after response to questions above ;) Please rename the variables to reflect that the first part is before the suggestion and the second part is after the suggestion, and add clarification to the comment:
```suggestion
    // The command is split into 3 parts to render the suggestion in a different color: 
    // the part before the suggestion, the suggestion, and the part after the suggestion
    std::string beforeSuggestion = currentCommand;
    std::string afterSuggestion = "";
    if (_autoCompleteState.insertPosition != 0) {
        beforeSuggestion = currentCommand.substr(0, _autoCompleteState.insertPosition);
        afterSuggestion = currentCommand.substr(_autoCompleteState.insertPosition);
    }
```

src/rendering/luaconsole.cpp
```diff
@@ -716,88 +496,106 @@ void LuaConsole::render() {
...
+    // positions, even if linebreaks are added
+    // Pad suggestion before and after with ' '
+    const std::string suggestion = std::string(firstPart.size() + 2, ' ') +
+        _autoCompleteState.suggestion + std::string(secondPart.size(), ' ');
+    // Pad first part at the end with ' '
+    firstPart.insert(firstPart.end(), totalCommandSize - firstPart.size(), ' ');
+    // Pad second part in the beginning with ' '
+    secondPart.insert(secondPart.begin(), totalCommandSize - secondPart.size(), ' ');
+
+    auto linebreakCommand = [nCharactersPerRow](const std::string& s) {
```
WeirdRubberDuck, line 546: This lambda could use an explanatory comment about what it does. It seems to be called for all rendered and is hence important enough to be explained :)

src/rendering/luaconsole.cpp
WeirdRubberDuck: So, conclusion is still to move it to stringhelper.h ? Then please do this @engbergandreas  :)

src/rendering/luaconsole.cpp
WeirdRubberDuck: And change this into a forward search?  Especially if moving it to the stringhelper.h file, as it does not make much sense to do a backwards search when moving this file out of its context

src/rendering/luaconsole.cpp
```diff
@@ -886,12 +688,589 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+        // Normalize paths to lowercase so they are treated equivalently
+        if (_autoCompleteState.context == Context::Path) {
+            out = ghoul::toLowerCase(sanitizeInput(out));
+        }
+#endif // WIN32
+
+        return out;
+    };
+
+    // We're only interested in unique matches
```
WeirdRubberDuck:
```suggestion
    // We're only interested in unique matches, and want them sorted alphabetically, 
    // so use a set for the results
```

src/rendering/luaconsole.cpp
```diff
@@ -886,12 +688,589 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+    std::set<std::string> results;
+
+    std::string input = _autoCompleteState.input;
+
+    input = normalize(input);
```
WeirdRubberDuck:
```suggestion
    std::string input = _autoCompleteState.input;
    input = normalize(input);
```

src/rendering/luaconsole.cpp
```diff
@@ -886,12 +688,589 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+        _autoCompleteState.suggestion
+    );
+    // Set cursor to the end of the command
+    _inputPosition = _autoCompleteState.insertPosition +
+        _autoCompleteState.suggestion.size();
+
+    if (_autoCompleteState.context == Context::Function &&
+        !_autoCompleteState.suggestion.ends_with('.'))
+    {
+        // We're in a leaf function add paranthesis
```
WeirdRubberDuck:
```suggestion
        // We're in a leaf function => add parentheses
```

src/rendering/luaconsole.cpp
WeirdRubberDuck: However, there is no need to sort or remove duplicates until the very end, and since you create a vector from this set in the end I would recommed just using a std::vector right away. See other suggestion

That would make the code more explicit and hence easier to read/understand, and probably also be a bit faster :)

src/rendering/luaconsole.cpp
```diff
@@ -886,12 +688,589 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+            }
+        }
+
+        if (suggestionNormalized.starts_with(input)) {
+            results.insert(result);
+        }
+    }
+
+    _autoCompleteState.suggestions =
+        std::vector<std::string>(results.begin(), results.end());
```
WeirdRubberDuck:
```suggestion
    std::vector<std::string> results;
    results.reserve(_autoCompleteState.suggestions.size());

    std::string input = _autoCompleteState.input;
    input = normalize(input);

    for (const std::string& suggestion : _autoCompleteState.suggestions) {
        std::string suggestionNormalized = normalize(suggestion);
        std::string result = sanitizeInput(suggestion);

        if (_autoCompleteState.context == Context::Function) {
            // We're only interested in autocomplete until the next separator "."
            const size_t offset = input.size();
            const size_t pos = suggestionNormalized.find('.', offset);

            if (pos != std::string::npos) {
                result = result.substr(0, pos + 1); // include the "."
            }
        }

        if (suggestionNormalized.starts_with(input)) {
            results.push_back(result);
        }
    }
    results.shrink_to_fit();

    // We're only interested in unique suggestions, and we want them sorted alphabetically
    std::sort(results.begin(), results.end());
    results.erase(std::unique(results.begin(), results.end()), results.end());

    _autoCompleteState.suggestions = std::move(results);
```

WeirdRubberDuck (approved):
Nice!  Using the right arrow key feels much more natural to me :) 

Had some code-related comments, as well as some old comments that I still think needs adressing. But otherwise it looks good

src/rendering/luaconsole.cpp
```diff
@@ -886,12 +688,589 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+            // behavior results in abc.de -> abc. -> abc -> 'empty string'
+            if (JumpCharacters.find(command[_inputPosition - 1]) != std::string::npos) {
+                command.erase(_inputPosition - 1, 1);
+                --_inputPosition;
+                return;
+            }
+
+            // Find the position of the last JumpCharacter before _inputPosition
+            size_t start = 0;
+            for (size_t i = _inputPosition; i > 0; --i) {
```
WeirdRubberDuck: To match Alex's changes of converting all for loops to postfix increments, go through all for loops in your added code and change to post instead of prefix :) 

```suggestion
            for (size_t i = _inputPosition; i > 0; i--) {
```

WeirdRubberDuck: So yeah, do the same for all the for loops you have in this code

src/rendering/luaconsole.cpp
```diff
@@ -886,12 +688,589 @@ void LuaConsole::setCommandInputButton(Key key) {
...
+    );
+
+    // Remove character before _inputPosition
+    registerKeyHandler(
+        Key::BackSpace,
+        KeyModifier::None,
+        [this]() {
+            if (_inputPosition > 0) {
+                _commands[_activeCommand].erase(_inputPosition - 1, 1);
+                --_inputPosition;
```
WeirdRubberDuck: And cases like this as well: 
```suggestion
                _inputPosition--;
```

src/rendering/luaconsole.cpp
engbergandreas: After discussion decided not to move it after all since it takes std::path as input argument

PR #3781, opened 2025-08-18
Add rendering for mouse interaction

### LINKED ISSUE #1399: Navigation handler draw velocity vector
When using the orbital navigator, the velocity added is based on the mouse's position after mouse_down, the vector of the mouse could be drawn to give the navigator an indication of how much velocity is being added to the camera.

alexanderbock:
Adds a new BoolProperty to the NavigationHandler that, if enabled, shows the interaction of the mouse while pressing any button.

image
(in the image, the OS mouse cursor is at the end of the white line, but is not captured in the screenshot)

Closes #1399

### DIFF
--- a/include/openspace/navigation/navigationhandler.h
+++ b/include/openspace/navigation/navigationhandler.h
@@ -99,6 +99,8 @@ public:
     void mousePositionCallback(double x, double y);
     void mouseScrollWheelCallback(double pos);
 
+    void renderOverlay() const;
+
     std::vector<std::string> listAllJoysticks() const;
     void setJoystickAxisMapping(std::string joystickName,
         int axis, JoystickCameraStates::AxisType mapping,
@@ -208,6 +210,14 @@ private:
     properties::BoolProperty _disableJoystickInputs;
     properties::BoolProperty _useKeyFrameInteraction;
     properties::FloatProperty _jumpToFadeDuration;
+
+    struct {
+        properties::BoolProperty enable;
+        bool isMouseFirstPress = false;
+        bool isMousePressed = false;
+        glm::vec2 clickPosition;
+        glm::vec2 currentPosition;
+    } _mouseVisualizer;
 };
 
 } // namespace openspace::interaction
--- a/include/openspace/navigation/navigationhandler.h
+++ b/include/openspace/navigation/navigationhandler.h
@@ -37,6 +37,7 @@
 #include <openspace/navigation/pathnavigator.h>
 #include <openspace/properties/scalar/boolproperty.h>
 #include <openspace/properties/scalar/floatproperty.h>
+#include <openspace/properties/vector/vec4property.h>
 #include <openspace/util/mouse.h>
 #include <openspace/util/keys.h>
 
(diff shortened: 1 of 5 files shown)


include/openspace/navigation/navigationhandler.h
```diff
@@ -208,6 +210,14 @@ class NavigationHandler : public properties::PropertyOwner {
...
     properties::BoolProperty _useKeyFrameInteraction;
     properties::FloatProperty _jumpToFadeDuration;
+
+    struct {
+        properties::BoolProperty enable;
+        bool isMouseFirstPress = false;
+        bool isMousePressed = false;
+        glm::vec2 clickPosition;
+        glm::vec2 currentPosition;
+    } _mouseVisualizer;
```
Roxeena, line 224: Should this be a property owner class? So the user could choose the color and size of the rendered line using properties.

src/rendering/helper.cpp
```diff
@@ -42,6 +42,12 @@
 
 namespace {
 
+struct VertexXYUVRGBA {
+    std::array<GLfloat, 2> xy;
+    std::array<GLfloat, 2> uv;
```
Roxeena, line 47: Do we need the uv components? Will we ever apply a texture to a line?

include/openspace/navigation/navigationhandler.h
alexanderbock: I added a PropertyOwner to the anonymous struct since we skip a bit of code that way

src/rendering/helper.cpp
alexanderbock: The line rendering is using a shader that needs xyuvrgba values for the vertices.  So we don't need it for the line rendering, but to reuse the same shader, we need it

Roxeena: I noticed, but I am wondering if this generic shader would ever need the uv component? Since it will only render lines anyway and I don't really see a need for a texture when rendering a 2D line. But this could become a new issue instead of being part of this PR.

alexanderbock: That xyuvrgba shader is also used by the function to render planes, which is using the UV coordinates.   So I think setting the uv coordinates but reusing the existing shader is preferable

src/rendering/helper.cpp
```diff
@@ -501,6 +562,47 @@ void renderBox(const glm::vec2& position, const glm::vec2& size, const glm::vec4
...
 }
 
+void renderLine(const glm::vec2& startPosition, const glm::vec2& endPosition,
+                const glm::vec2& size, const glm::vec4& startColor,
+                const glm::vec4& endColor)
+{
+    auto& shdr = shaders.xyuvrgba;
+    shdr.program->activate();
+    shdr.program->setUniform(shdr.cache.proj, ortho(glm::vec2(0.f), glm::vec2(1.f)));
+    shdr.program->setUniform(shdr.cache.hasTexture, 0);
```
engbergandreas, line 572: Did we run out of vowels? 😅

engbergandreas (approved):
Looks great. Should we add some anti-aliasing for this? The edges are kind of sharp at times. Not a dealbreaker though.

image

PR #3782, opened 2025-08-21
Create Deb with CPACK

hn-88:
These changes to the packaging.cmake allows the creation of deb files by calling `cpack` after the build step. @alexanderbock - please use the "squash and merge" functionality to squash all these trial-and-error commits to a single commit, to not clutter up the codebase.

Successful build of deb on Ubuntu 24.04 is at https://github.com/hn-88/OpenSpace-AppImage/actions/runs/17118519369

Test of this deb is at https://github.com/hn-88/OpenSpace-AppImage/actions/runs/17119724311


hn-88:
@alexanderbock - in case you're trying to test cpack (deb or zip) with jenkins using Docker, I have found that the docker image runs out of disk space and the build does not complete. With docker, we would need to mount separate larger volumes instead of the default 20 GB, or maybe increase the default container size - 

https://www.baeldung.com/ops/docker-mounting-multiple-volumes

https://forums.docker.com/t/cannot-change-default-20g-size-limit-on-windows-docker-host/31879/2

alexanderbock:
Closing it in favor of #4286

PR #3784, opened 2025-08-25
Fixed incorrect rendering of planet-onto-ring shadows in fisheye projection mode

### LINKED ISSUE #3776: Saturn shadowing onto rings is incorrectly rendered in fisheye projection
The shadows cast by Saturn onto the rings in fisheye projection mode is incorrect because the depth map used for shadow rendering is part of the normal render pass, so it is rendered 5 times and in different directions. This causes parts of the shadow to be missing. 

_The shadow is only visible on part of the rings:_
Image

_One of the five depth maps, which is cut off, causing the artifact_
Image

_A visualization of how the 5 views in fisheye mode are composed_
Image

The pass that renders the scene from the point of view of the light should either be separated or removed, in favor of a different method of shadowing the rings.

benpm:
The shadows cast by Saturn onto its rings in fisheye projection mode are rendered incorrectly because the depth map used for shadow rendering is rendered during the normal render pass, so it is rendered 5 times and in different directions. This causes parts of the shadow to be missing. (See #3776 for more details)

Screenshot 2025-08-25 120237

I've simply replaced shadow-mapping with a ray/ellipsoid intersection test. However, there are other use cases for shadow-mapping, like in [project/moon-shot](https://github.com/OpenSpace/OpenSpace/tree/project/moon-shot), so I've left most of the shadow-map rendering logic intact so that the project can be eventually updated and merged.

fixes #3776

### DIFF
--- a/data/assets/scene/solarsystem/planets/saturn/saturn.asset
+++ b/data/assets/scene/solarsystem/planets/saturn/saturn.asset
@@ -62,7 +62,7 @@ local Saturn = {
       }
     },
     Shadows = {
-      Enabled = true,
+      Enabled = false,
       DistanceFraction = 40.0
     }
   },
--- a/modules/globebrowsing/shaders/advanced_rings_fs.glsl
+++ b/modules/globebrowsing/shaders/advanced_rings_fs.glsl
@@ -24,6 +24,7 @@
 
 #include "PowerScaling/powerScaling_fs.hglsl"
 #include "fragment.glsl"
+#include "ellipsoid.glsl"
 
 #define NSSamplesMinusOne #{nShadowSamples}
 #define NSSamples (NSSamplesMinusOne + 1)
@@ -32,8 +33,9 @@ in vec2 vs_st;
 in float vs_screenSpaceDepth;
 in vec4 shadowCoords;
 in vec3 vs_normal;
+// Fragment position in object space
+in vec3 posObj;
 
-uniform sampler2DShadow shadowMapTexture;
 uniform sampler1D textureForwards;
 uniform sampler1D textureBackwards;
 uniform sampler1D textureUnlit;
@@ -47,6 +49,7 @@ uniform vec3 camPositionObj;
 uniform float nightFactor;
 uniform float zFightingPercentage;
 uniform float opacity;
+uniform vec3 ellipsoidRadii;
 
 
 Fragment getFragment() {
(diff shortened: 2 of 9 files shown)


data/assets/scene/solarsystem/planets/saturn/saturn.asset
```diff
@@ -62,7 +62,7 @@ local Saturn = {
       }
     },
     Shadows = {
-      Enabled = true,
+      Enabled = false,
```
Roxeena, line 65: Why do we want to turn this off by default now?

modules/globebrowsing/shaders/ellipsoid.glsl
```diff
@@ -0,0 +1,35 @@
...
+        return true;
+    }
+    
+    // If both intersections exist and C > 0, check if the smaller root t1 >= 0
+    // t1 = (-b - sqrt(discriminant)) / a
+    // Since we need t1 >= 0: -b - sqrt(discriminant) >= 0
+    // This means: -b >= sqrt(discriminant), so b <= -sqrt(discriminant)
+    // Since sqrt(discriminant) >= 0, this means b <= 0
+    return b <= 0.0;
+}
```
Roxeena: Missing new line

modules/globebrowsing/shaders/advanced_rings_fs.glsl
```diff
@@ -86,28 +89,12 @@ Fragment getFragment() {
     discard;
   }
 
+  // Check if ray from fragment to sun intersects the ellipsoid (globe)
+  // This creates more accurate shadowing for rings
+  bool intersectsGlobe = rayIntersectsEllipsoid(posObj, sunPositionObj, vec3(0.0), ellipsoidRadii);
```
Roxeena: Why the `vec(0.0)`? Shouldn't this be the center of Saturn? In which coordinate system?

modules/globebrowsing/shaders/rings_fs.glsl
```diff
@@ -75,28 +78,12 @@ Fragment getFragment() {
     }
 }
 
+  // Check if ray from fragment to sun intersects the ellipsoid (globe)
+  // This creates more accurate shadowing for rings
+  bool intersectsGlobe = rayIntersectsEllipsoid(posObj, sunPositionObj, vec3(0.0), ellipsoidRadii);
```
Roxeena: Same question here

Roxeena (approved):
Tried it out and it seems to work well in both fisheye and regular configurations

alexanderbock:
Starting the default profile I get two warnings:
```
(W) ProgramObject(..am') Failed to locate uniform location for 'shadowMatrix'
(W) ProgramObject(..am') Failed to locate uniform location for 'zFightingPercentage'
```

alexanderbock:
In Debug mode on Windows when approaching Saturn, getting a related asset:
```
(D) SessionRecordin.opic Starting new SessionRecording state subscription

    File:       A:\openspace\OpenSpace\ext\ghoul\src\opengl\programobject.cpp, line 1686
    Function:   ghoul::opengl::ProgramObject::setUniform
    Assertion:  location != -1    Location must not be -1
(I)gnore / Ignore (P)ermanently / (A)ssertion / (S)tacktrace / (E)xit:
```

alexanderbock:
I pushed some changes for the coding style straight away.  All about exceeding line lengths and the copyright header for the ellipse.glsl was missing.  Feel free to check though (https://github.com/OpenSpace/OpenSpace/pull/3784/commits/eaa9e6fb39e099699cebceff233b52bda787c437)

modules/globebrowsing/shaders/advanced_rings_fs.glsl
alexanderbock: As far as I understand it, it is in the local coordinate system, meaning that `vec3(0.0)` would be the center?

data/assets/scene/solarsystem/planets/saturn/saturn.asset
benpm: This is the legacy shadows, I'm turning it off by default because it's executing additional drawing unnecessarily (there is currently nothing in OpenSpace that will actually render anything differently if this setting is disabled). I'd like to keep the legacy shadow stuff around though because I'll still need them around.

PR #3794, opened 2025-09-14
New ScreenSpaceText objects

alexanderbock:
This pull request adds two new ScreenSpaceRenderables:
  - `ScreenSpaceText`:  Can be used to print a fixed text onto the screen
  - `ScreenSpaceDate`:  Can print the current in-game simulation time to the screen.  Uses the same parameters and properties as `DashboardItemDate`

Other changes:
To reduce confusion, all of the ScreenSpaceRenderable classes that cannot be directly instantiated are now prefixed with `ScreenSpaceRenderable`, for example `ScreenSpaceRenderableFramebuffer` instead of `ScreenSpaceFramebuffer`.   The `ScreenSpace` prefix is used for types that can be instantiated by the user.

There is a new abstract class `ScreenSpaceRenderableText` that groups functionality for ScreenSpace types that want to print text to the screen.

The ScreenSpaceRenderableFramebuffer files have been moved from the base module into the core.

The PR also adds examples and image tests


src/rendering/screenspacerenderableframebuffer.cpp
```diff
@@ -68,35 +67,37 @@ ScreenSpaceFramebuffer::ScreenSpaceFramebuffer(const ghoul::Dictionary& dictiona
...
 
         if (iIdentifier == 0) {
-            setIdentifier("ScreenSpaceFramebuffer");
+            setIdentifier("ScreenSpaceRenderableFramebuffer");
         }
         else {
-            setIdentifier("ScreenSpaceFramebuffer" + std::to_string(iIdentifier));
+            setIdentifier(
+                "ScreenSpaceRenderableFramebuffer" + std::to_string(iIdentifier)
+            );
```
WeirdRubberDuck: std::format?

src/rendering/screenspacerenderableframebuffer.cpp
```diff
@@ -68,35 +67,37 @@ ScreenSpaceFramebuffer::ScreenSpaceFramebuffer(const ghoul::Dictionary& dictiona
         iIdentifier = id();
```
WeirdRubberDuck: Nitpicky, but this variable name was confusing to me, and the variable could be initialized in the if-statement instead of outside :)  Maybe:

```suggestion
        int framebufferIdCounter= id();
```
?

modules/base/rendering/screenspacedate.cpp
```diff
@@ -0,0 +1,108 @@
...
+        global::timeManager->time().j2000Seconds(),
+        _timeFormat.value().c_str()
+    );
+
+    try {
+        // @CPP26(abock): This can be replaced with std::runtime_format
+        _buffer = std::vformat(_formatString.value(), std::make_format_args(time));
+    }
+    catch (const std::format_error&) {
+        LERRORC("DashboardItemDate", "Illegal format string");
```
WeirdRubberDuck: Wrong category
```suggestion
        LERRORC("ScreenSpaceDate", "Illegal format string");
```

or maybe add a `_loggerCat`? Easier to keep in sync when more log messages are added

data/assets/examples/screenspacerenderable/screenspacedate/date.asset
```diff
@@ -0,0 +1,15 @@
+-- Basic
+-- This example creates a screen space renderable that shows the current simulation time.
+
+local Item = {
+  Type = "ScreenSpaceDate",
+  Identifier = "ScreenSpaceDate_Example"
```
WeirdRubberDuck: Add real names to all the examples, following our naming scheme pattern

```suggestion
  Identifier = "ScreenSpaceDate_Example",
  Name = "ScreenSpaceDate Example - Basic"
```

data/assets/examples/screenspacerenderable/screenspacedate/date.asset
```diff
@@ -0,0 +1,15 @@
+-- Basic
+-- This example creates a screen space renderable that shows the current simulation time.
+
+local Item = {
```
WeirdRubberDuck: In the existing examples, this variable is named "Object". We should be consistent

I personally like "Item" a bit better though

modules/base/rendering/screenspacedate.cpp
alexanderbock: Since it is just one, I'll change it to the correct category

data/assets/examples/screenspacerenderable/screenspacedate/date.asset
alexanderbock: I'll make it `Object` now and then have a separate commit to rename them to Item

PR #3739, opened 2025-07-03
Faster rendering for renderableOrbitalKepler

Arohdin:
## Info ##
This PR contains the following changes to `renderableOrbitalKepler`:
- Precomputes relevant vertices on CPU rather than filtering all vertices on the GPU.
- Removes geometry shader for trails rendering
- Changed GUI name of RenderingMode "Points+Trails" to "Points and Trails"

The changes are not breaking changes and should not affect the visuals either.

## Performance ##

The performance improvements varies based on the datasets as performance is limited based on things such as number of orbits, number of vertices per orbit, how long the trail is and how much overlap/overdraw there is for the trails/points.

Below are some before and after numbers based on asteroid assets.

### Trails ###
- Inner main, Main, Outer main:
  - Before: 80 fps
  - After: 90 fps
- Transneptunian:
  - Before: 29 fps
  - After 300+ fps
- Amor, Apollo, Outer main, Main, Inner main, Jupiter trojan, Transneptunian:
  - Before: 14 fps
  - After: 65 fps

### Points ###
- Inner main, Main, Outer main:
  - Before: 80 fps
  - After: 88 fps
- Transneptunian:
  - Before: 30 fps
  - After 300+ fps
- Amor, Apollo, Outer main, Main, Inner main, Jupiter trojan, Transneptunian:
  - Before: 14 fps
  - After: 65 fps
 
### Trails and Points ###
- Inner main, Main, Outer main:
  - Before: 42 fps
  - After: 50 fps
- Transneptunian:
  - Before: 15 fps
  - After 300+ fps
- Amor, Apollo, Outer main, Main, Inner main, Jupiter trojan, Transneptunian:
  - Before: 7 fps
  - After: 34 fps

PR #3799, opened 2025-09-18
feature/multithreaded-asteroids

Arohdin:
## This PR adds the following
- Multithreading and a bettery system for updating which segments to render for each orbit
- Multithreading for loading of data
- Support for `Camera View Direction` in addition to existing `Camera Position Normal`

## Faster loading of data
Test with `.../scene/solarsystem/sssb/mpc.asset` (1.4 million objects). Loading times changed from `3 minutes` to `12 seconds`.

## Faster and smarter `update` function
Update function is now multithreaded and can also determine when new segments needs to be calculated, thus mostly doing minimal work for each frame (unless very high delta time).

### Compared to previous performance improvements in #3739:
**Datasets: Amor, Apollo, Outer main, Main, Inner main, Jupiter trojan, Transneptunian**
(ingame time paused)
- **Mode: Trails**
  - Before: 65 fps
  - After: 300 fps

- **Mode: Points**
  - Before: 65 fps
  - After: 350 fps

- **Mode: Trails + Points**
  - Before: 34 fps
  - After: 280 fps

### It also makes it possible to have better framerate at higher delta times:
**Datasets: Amor, Apollo, Outer main, Main, Inner main, Jupiter trojan, Transneptunian**
**Mode: Trails**
- **Before**
  - 14 days/sec: 32 fps
  - 30 days/sec: 25 fps

- **After**
  - 14 days/sec: 180 fps
  - 30 days/sec: 120 fps

### DIFF
--- a/modules/space/rendering/renderableorbitalkepler.cpp
+++ b/modules/space/rendering/renderableorbitalkepler.cpp
@@ -40,22 +40,40 @@
 #include <ghoul/logging/logmanager.h>
 #include <chrono>
 #include <cmath>
+#include <execution>
 #include <fstream>
 #include <random>
 #include <vector>
 
 namespace {
     // The possible values for the _renderingModes property
-    enum RenderingMode {
+    enum RenderMode {
         RenderingModeTrail = 0,
         RenderingModePoint,
         RenderingModePointTrail
+    };
+
+    enum PointRenderingMode {
+        ViewDirection = 0,
+        PositionNormal
     };
 
     constexpr openspace::properties::Property::PropertyInfo PathInfo = {
         "Path",
         "Path",
         "The file path to the data file to read.",
+        openspace::properties::Property::Visibility::AdvancedUser
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo PointRenderingModeInfo =
+    {
+        "PointRenderingMode",
+        "Point Rendering Mode",
+        "Controls how the points will be oriented. \"Camera View "
+        "Direction\" rotates the points so that they are orthogonal to the viewing "
+        "direction of the camera (useful for planar displays), and \"Camera Position "
+        "Normal\" rotates the points towards the position of the camera (useful for "
+        "spherical displays, like dome theaters).",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
@@ -193,6 +211,13 @@ namespace {
         // The file format that is contained in the file.
         Format format;
 
+        enum class [[codegen::map(PointRenderingMode)]] PointRenderingMode {
+            ViewDirection [[codegen::key("Camera View Direction")]],
+            PositionNormal [[codegen::key("Camera Position Normal")]]
+        };
+        // [[codegen::verbatim(PointRenderingModeInfo.description)]]
+        std::optional<PointRenderingMode> pointRenderingMode;
+
         // [[codegen::verbatim(SegmentQualityInfo.description)]]
         int segmentQuality;
 
@@ -261,22 +286,24 @@ RenderableOrbitalKepler::Appearance::Appearance()
     , enableMaxSize(EnableMaxSizeInfo, true)
     , maxSize(MaxSizeInfo, 5.f, 0.f, 45.f)
     , renderingModes(RenderingModeInfo)
+    , pointRenderOption(PointRenderingModeInfo)
     , trailFade(TrailFadeInfo, 20.f, 0.f, 30.f)
     , enableOutline(EnableOutlineInfo, true)
     , outlineColor(OutlineColorInfo, glm::vec3(0.f), glm::vec3(0.f), glm::vec3(1.f))
     , outlineWidth(OutlineWidthInfo, 0.2f, 0.f, 1.f)
 {
     renderingModes.addOptions({
-        { RenderingModeTrail, "Trails" },
-        { RenderingModePoint, "Points" },
-        { RenderingModePointTrail , "Points and Trails" }
+        { RenderMode::RenderingModeTrail, "Trails" },
+        { RenderMode::RenderingModePoint, "Points" },
+        { RenderMode::RenderingModePointTrail , "Points and Trails" }
     });
+    renderingModes.onChange([this]() { changedRenderType = true; });
     addProperty(renderingModes);
-
     color.setViewOption(properties::Property::ViewOptions::Color);
     addProperty(color);
     addProperty(trailWidth);
     addProperty(trailFade);
+    addProperty(pointRenderOption);
     addProperty(pointSizeExponent);
     addProperty(enableMaxSize);
     addProperty(maxSize);
@@ -288,6 +315,7 @@ RenderableOrbitalKepler::Appearance::Appearance()
 
 RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
     : Renderable(dict)
+    , _nThreads(static_cast<int>(std::ceil(std::thread::hardware_concurrency() / 2.0)))
     , _segmentQuality(SegmentQualityInfo, 2, 1, 10)
     , _startRenderIdx(StartRenderIdxInfo, 0, 0, 1)
     , _sizeRender(RenderSizeInfo, 1, 1, 2)
@@ -299,11 +327,12 @@ RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
     addProperty(Fadeable::_opacity);
 
     _segmentQuality = static_cast<unsigned int>(p.segmentQuality);
-    _segmentQuality.onChange([this]() { updateBuffers(); });
+    _segmentQuality.onChange([this]() { _updateDataBuffersAtNextRender = true; });
     addProperty(_segmentQuality);
 
     _appearance.color = p.color;
     _appearance.trailFade = p.trailFade.value_or(_appearance.trailFade);
+    _appearance.trailFade.onChange([this]() { _forceUpdate = true; });
     _appearance.trailWidth = p.trailWidth.value_or(_appearance.trailWidth);
     _appearance.enableMaxSize = p.enableMaxSize.value_or(_appearance.enableMaxSize);
     _appearance.maxSize = p.maxSize.value_or(_appearance.maxSize);
@@ -313,21 +342,43 @@ RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
     _appearance.pointSizeExponent =
         p.pointSizeExponent.value_or(_appearance.pointSizeExponent);
 
+    _appearance.pointRenderOption.addOption(
+        PointRenderingMode::ViewDirection,
+        "Camera View Direction"
+    );
+    _appearance.pointRenderOption.addOption(
+        PointRenderingMode::PositionNormal,
+        "Camera Position Normal"
+    );
+    if (p.pointRenderingMode.has_value()) {
+        switch (*p.pointRenderingMode) {
+        case Parameters::PointRenderingMode::ViewDirection:
+            _appearance.pointRenderOption = PointRenderingMode::ViewDirection;
+            break;
+        case Parameters::PointRenderingMode::PositionNormal:
+            _appearance.pointRenderOption = PointRenderingMode::PositionNormal;
+            break;
+        }
+    }
+    else {
+        _appearance.pointRenderOption = PointRenderingMode::ViewDirection;
+    }
+
     if (p.renderingMode.has_value()) {
         switch (*p.renderingMode) {
             case Parameters::RenderingMode::Trail:
-                _appearance.renderingModes = RenderingModeTrail;
+                _appearance.renderingModes = RenderMode::RenderingModeTrail;
                 break;
             case Parameters::RenderingMode::Point:
-                _appearance.renderingModes = RenderingModePoint;
+                _appearance.renderingModes = RenderMode::RenderingModePoint;
                 break;
             case Parameters::RenderingMode::PointsTrails:
-                _appearance.renderingModes = RenderingModePointTrail;
+                _appearance.renderingModes = RenderMode::RenderingModePointTrail;
                 break;
         }
     }
     else {
-        _appearance.renderingModes = RenderingModeTrail;
+        _appearance.renderingModes = RenderMode::RenderingModeTrail;
     }
     addPropertySubOwner(_appearance);
 
@@ -336,8 +387,8 @@ RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
     _startRenderIdx = p.startRenderIdx.value_or(0);
     _startRenderIdx.onChange([this]() {
         if (_contiguousMode) {
-            if ((_numObjects - _startRenderIdx) < _sizeRender) {
-                _sizeRender = static_cast<unsigned int>(_numObjects - _startRenderIdx);
+            if ((_nOrbits - _startRenderIdx) < _sizeRender) {
+                _sizeRender = static_cast<unsigned int>(_nOrbits - _startRenderIdx);
             }
             _updateDataBuffersAtNextRender = true;
         }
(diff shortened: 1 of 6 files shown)


alexanderbock:
Removing or adapting the `RenderSize` from the main_belt_asteroid.asset as well.  Maybe it needs some adjustment of the trail length as well?

Arohdin:
> Removing or adapting the `RenderSize` from the main_belt_asteroid.asset as well. Maybe it needs some adjustment of the trail length as well?

The whole dataset is over 1 million orbits. I think that it's smart to keep the `RenderSize`, but perhaps we can increase it from 50k to a couple of 100k?

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -313,21 +342,43 @@ RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
...
+    if (p.pointRenderingMode.has_value()) {
+        switch (*p.pointRenderingMode) {
+        case Parameters::PointRenderingMode::ViewDirection:
+            _appearance.pointRenderOption = PointRenderingMode::ViewDirection;
+            break;
+        case Parameters::PointRenderingMode::PositionNormal:
+            _appearance.pointRenderOption = PointRenderingMode::PositionNormal;
+            break;
+        }
+    }
```
engbergandreas, line 363:
```suggestion
        switch (*p.pointRenderingMode) {
          case Parameters::PointRenderingMode::ViewDirection:
              _appearance.pointRenderOption = PointRenderingMode::ViewDirection;
              break;
          case Parameters::PointRenderingMode::PositionNormal:
              _appearance.pointRenderOption = PointRenderingMode::PositionNormal;
              break;
       }
    }
```

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -562,32 +642,31 @@ void RenderableOrbitalKepler::render(const RenderData& data, RendererTasks&) {
...
-            "Start index {} out of range [0, {}]", _startRenderIdx.value(), _numObjects
+            "Start index {} out of range [0, {}]", _startRenderIdx.value(), _nOrbits
         ));
     }
 
     long long endElement = _startRenderIdx + _sizeRender - 1;
-    endElement = (endElement >= _numObjects) ? _numObjects - 1 : endElement;
-    if (endElement < 0 || endElement >= _numObjects) {
+    endElement = (endElement >= _nOrbits) ? _nOrbits - 1 : endElement;
+    if (endElement < 0 || endElement >= _nOrbits) {
```
engbergandreas, line 665: Won't `endElemnt >= _nOrbits` always be false due to the above check? Can be removed?

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -684,120 +814,141 @@ void RenderableOrbitalKepler::updateBuffers() {
...
+                        p0Start = startVertexIndex;
+                        p1Start = correctVertexIndex;
+                        if (lastVertexIndex - correctVertexIndex == 1) {
+                            p1Length = 0;
+                            p0Length = correctTrailLength - 1;
+                        }
+                        else {
+                            p1Length = lastVertexIndex - correctVertexIndex;
+                            p0Length = correctTrailLength - p1Length;
+                        }
```
engbergandreas, line 940: Does it make sense to write a comment describing the different scenarios here as well like you've done for the rest of the if-statements

modules/space/rendering/renderableorbitalkepler.cpp
Arohdin: No, the check `_startRenderIdx >= _nOrbits` will not force `endElement >= _nOrbits` to always be false.
The first check just makes sure that the starting index is not more than the number of orbits that we have.
The second check makes sure that the starting index + the size we want to render is not more than the number of orbits.

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -40,25 +40,43 @@
...
 namespace {
     // The possible values for the _renderingModes property
-    enum RenderingMode {
+    enum RenderMode {
         RenderingModeTrail = 0,
         RenderingModePoint,
         RenderingModePointTrail
     };
 
+    enum PointRenderingMode {
```
alexanderbock:
```suggestion
    enum class RenderMode {
RenderingModeTrail = 0,
RenderingModePoint,
RenderingModePointTrail
};

    enum class PointRenderingMode {
```

Even if it requires some `static_cast<std::underlying_type_t<RenderMode>>`s to make it compile.  It's going to be a bit more expressive when we are trying to make an int fit into the enum class

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -40,25 +40,43 @@
...
+
     constexpr openspace::properties::Property::PropertyInfo PathInfo = {
         "Path",
         "Path",
         "The file path to the data file to read.",
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
+    constexpr openspace::properties::Property::PropertyInfo PointRenderingModeInfo =
+    {
```
alexanderbock:
```suggestion
    constexpr openspace::properties::Property::PropertyInfo PointRenderingModeInfo = {
```

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -40,25 +40,43 @@
...
+        "PointRenderingMode",
+        "Point Rendering Mode",
+        "Controls how the points will be oriented. \"Camera View "
+        "Direction\" rotates the points so that they are orthogonal to the viewing "
+        "direction of the camera (useful for planar displays), and \"Camera Position "
+        "Normal\" rotates the points towards the position of the camera (useful for "
+        "spherical displays, like dome theaters).",
```
alexanderbock:
```suggestion
        "Controls how the points will be oriented. \"Camera View Direction\" rotates the "
        "points so that they are orthogonal to the viewing direction of the camera "
        "(useful for planar displays), and \"Camera Position Normal\" rotates the points "
        "towards the position of the camera (useful for spherical displays, like dome "
        "theaters).",
```

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -288,6 +315,7 @@ RenderableOrbitalKepler::Appearance::Appearance()
 
 RenderableOrbitalKepler::RenderableOrbitalKepler(const ghoul::Dictionary& dict)
     : Renderable(dict)
+    , _nThreads(static_cast<int>(std::ceil(std::thread::hardware_concurrency() / 2.0)))
```
alexanderbock:
```suggestion
    , _nThreads(std::max(1, std::thread::hardware_concurrency() / 2))
```

?

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -427,30 +478,49 @@ bool RenderableOrbitalKepler::isReady() const {
...
 }
 
-void RenderableOrbitalKepler::update(const UpdateData&) {
+void RenderableOrbitalKepler::update(const UpdateData& data) {
     if (_updateDataBuffersAtNextRender) {
-        _updateDataBuffersAtNextRender = false;
         updateBuffers();
     }
+
+    if(_appearance.changedRenderType) {
```
alexanderbock:
```suggestion
    if (_appearance.changedRenderType) {
```

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -427,30 +478,49 @@ bool RenderableOrbitalKepler::isReady() const {
...
+        _forceUpdate = true;
+        _appearance.changedRenderType = false;
+    }
+
+    std::for_each(
+        std::execution::par_unseq,
+        _threadIds.cbegin(),
+        _threadIds.cend(),
```
alexanderbock:
```suggestion
        _threadIds.begin(),
        _threadIds.end(),
```

I did some testing a while back and there is actually no difference between `begin` and `end`.  So i'd opt for the simpler version

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -613,64 +692,115 @@ void RenderableOrbitalKepler::updateBuffers() {
...
-            _vertexBufferData[vertexBufIdx + j].period = orbit.period;
+    std::vector<int> orbitIdHolder;
+    orbitIdHolder.resize(_sizeRender);
+    std::iota(orbitIdHolder.begin(), orbitIdHolder.end(), 0);
+
+    std::for_each(
+        std::execution::par_unseq,
+        orbitIdHolder.begin(),
+        orbitIdHolder.end(),
+        [&](int& index) {
```
alexanderbock:
```suggestion
        [&](int index) {
```

Since we don't modify it. And we shouldn't since it is unsequenced anyway

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -613,64 +692,115 @@ void RenderableOrbitalKepler::updateBuffers() {
...
+            const kepler::Parameters& orbit = _parameters[index];
+
+            ghoul::Dictionary d;
```
alexanderbock: Not to open another can of worms, but how about having a:
```cpp
std::vector<ghoul:Dictionary> dicts;
dicts.resize(_sizeRender);
for (ghoul::Dictionary& dict : dicts) {
    dict.setValue("Type", std::string("KeplerTranslation"));
}
```

before the `for_each`
and then use:
```cpp
            ghoul::Dictionary& d = dicts[index];
            d.setValue("Eccentricity", orbit.eccentricity);
            d.setValue("SemiMajorAxis", orbit.semiMajorAxis);
            d.setValue("Inclination", orbit.inclination);
            d.setValue("AscendingNode", orbit.ascendingNode);
            d.setValue("ArgumentOfPeriapsis", orbit.argumentOfPeriapsis);
            d.setValue("MeanAnomaly", orbit.meanAnomaly);
            d.setValue("Period", orbit.period);
            d.setValue("Epoch", orbit.epoch);
```

inside?

We'd skip all of the allocations of the dictionaries

alexanderbock: Alternative:   Extract the working bits out of the `KeplerTranslation` into a standalone function that gets all of the parameters to calculate the position.  That way we could skip the Dictionary and the KeplerTranslation's constructor entirely

modules/space/rendering/renderableorbitalkepler.cpp
Arohdin: Will keep as is for now as the whole section needs to be reworked anyways.

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -684,120 +814,145 @@ void RenderableOrbitalKepler::updateBuffers() {
...
-        const int nSegments = _segmentSizeRaw[i] - 1;
-        const int offset = static_cast<int>(std::floor(frac * nSegments));
-
-        // Set start vertex ID in buffer
-        _startIndexPoints[i] = startVertexIndex + offset;
-        startVertexIndex += _segmentSizeRaw[i];
-    }
-}
+void RenderableOrbitalKepler::threadedSegmentCalculations(const int threadId,
+                                                                const UpdateData& data)
```
alexanderbock:
```suggestion
void RenderableOrbitalKepler::threadedSegmentCalculations(int threadId,
                                                          const UpdateData& data)
```

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -684,120 +814,145 @@ void RenderableOrbitalKepler::updateBuffers() {
...
+            const int pointHead = static_cast<int>(std::floor(frac * nSegments));
+
+            // We can always do this since it has no cost
+            _startIndexPoints[i] = startVertexIndex + pointHead;
+
+            // There is a lot of what seems to be "magic numbers" in this section.
+            // They will most likely disappear when we change our method of determining
+            // the trail fade amount is changed.
+            if (_renderTrails) {
+
```
alexanderbock:
```suggestion
```

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -684,120 +814,145 @@ void RenderableOrbitalKepler::updateBuffers() {
...
+                // When rendering a trail we don't know if the trail will pass over
+                // the starting point of the orbit or not. If the trail passes over the
+                // starting point of the orbit, then we can't draw the entire trail as
+                // line strip. Instead we need to divide the line strip into two parts,
+                // where p0 and p1 denotes the respctive line strips (parts).
+                int p0Start;
+                int p0Length;
+                int p1Start;
+                int p1Length;
```
alexanderbock:
```suggestion
                int p0Start = -1;
                int p0Length = -1;
                int p1Start = -1;
                int p1Length = -1;
```

Just to prevent uninitialized variable warnings

modules/space/rendering/renderableorbitalkepler.h
```diff
@@ -76,58 +78,74 @@ class RenderableOrbitalKepler : public Renderable {
         properties::Vec3Property outlineColor;
         /// Specifies how much if the point should be covered by the outline
         properties::FloatProperty outlineWidth;
+
+        bool changedRenderType = false;
```
alexanderbock:
```suggestion
        bool isRenderTypeDirty = false;
```

To make it clearer that it is a dirty flag?

modules/space/rendering/renderableorbitalkepler.cpp
```diff
@@ -684,120 +814,145 @@ void RenderableOrbitalKepler::updateBuffers() {
...
-                p1Start = correctVertexIndex - 1;
-                p1Length = lastVertexIndex - correctVertexIndex + 1;
-                p0Start = startVertexIndex;
-                p0Length = correctTrailLength - p1Length + 1;
-            }
-            else {
-                // If the entire trail is within the bounds of the orbit
-                p0Start = correctVertexIndex;
-                p0Length = correctTrailLength;
+    int offset = std::accumulate(_orbitsPerThread.begin(),
```
alexanderbock:
```suggestion
    int offset = std::accumulate(
        _orbitsPerThread.begin(),
```

modules/space/rendering/renderableorbitalkepler.h
```diff
@@ -76,58 +78,74 @@ class RenderableOrbitalKepler : public Renderable {
...
     std::vector<GLint> _startIndexTrails;
     std::vector<GLint> _segmentSizeTrails;
     std::vector<kepler::Parameters> _parameters;
 
+    /// Extra data for more efficient updating of vectors
+    struct updateInfo {
+        double timestamp = std::numeric_limits<double>::min();
+        double timePerStep = 0.0;
+    };
+    std::vector<updateInfo> _updateHelper;
```
alexanderbock:
```suggestion
    struct UpdateInfo {
        double timestamp = std::numeric_limits<double>::min();
        double timePerStep = 0.0;
    };
    std::vector<UpdateInfo> _updateHelper;
```

modules/space/rendering/renderableorbitalkepler.h
```diff
@@ -76,58 +78,74 @@ class RenderableOrbitalKepler : public Renderable {
...
 
     // Point cache
     UniformCache(modelTransform, viewTransform, projectionTransform,
-        cameraPositionWorld, cameraUpWorld,  inGameTime, color,
-        pointSizeExponent, enableMaxSize, maxSize, enableOutline,
-        outlineColor, outlineWeight, opacity)
+        renderOption, cameraViewDirectionUp, cameraViewDirectionRight,
+        cameraPositionWorld, cameraUpWorld, inGameTime, color, pointSizeExponent,
+        enableMaxSize, maxSize, enableOutline, outlineColor, outlineWeight, opacity)
         _uniformPointCache;
```
alexanderbock:
```suggestion
    // Line cache
    UniformCache(modelViewTransform, projectionTransform, trailFadeExponent,
        colorFadeCutoffValue, inGameTime, color, opacity) _uniformTrailCache;

    // Point cache
    UniformCache(modelTransform, viewTransform, projectionTransform, renderOption,
        cameraViewDirectionUp, cameraViewDirectionRight, cameraPositionWorld,
        cameraUpWorld, inGameTime, color, pointSizeExponent, enableMaxSize, maxSize,
        enableOutline, outlineColor, outlineWeight, opacity) _uniformPointCache;
```

alexanderbock (changes requested):
Running it through Tracy I saw some troubling behavior (see the VRAM graph)

The many blocks are all `threadedSegmentCalculations` that don't go away when pausing the time

Image

alexanderbock:
The threaded calculation also runs when the time is paused

Arohdin:
> Running it through Tracy I saw some troubling behavior (see the VRAM graph)
> 
> The many blocks are all `threadedSegmentCalculations` that don't go away when pausing the time
> 
> Image

When checking Tracy it looks like issue is CEF related.
Issue persists under `Empty` profile as well.
If GPU accelerated CEF rendering is disabled in `openspace.cfg` the issue stops.

PR #3813, opened 2025-10-06
Make TUIO port configurable

k2d222:
Make the TUIO port (touch) configurable at load-time.

### DIFF
--- a/modules/touch/include/tuioear.h
+++ b/modules/touch/include/tuioear.h
@@ -52,7 +52,7 @@ namespace openspace {
 
 class TuioEar : public TUIO::TuioListener {
 public:
-    TuioEar(int port = 3333);
+    explicit TuioEar(int port = 3333);
     ~TuioEar();
 
     /**
--- a/modules/touch/touchmodule.cpp
+++ b/modules/touch/touchmodule.cpp
@@ -76,7 +76,7 @@ namespace {
 
     struct [[codegen::Dictionary(TouchModule)]] Parameters {
         // [[codegen::verbatim(TuioPortInfo.description)]]
-        std::optional<int> tuioPort;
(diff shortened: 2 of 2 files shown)


k2d222:
to configure, set in `openspace.cfg`:
```lua
ModuleConfigurations = {
    Touch = {
        TuioPort = 3334
    },
    // ...
}
```

k2d222:
I don't have permission to see the CI failure. The PR compiles for me.

alexanderbock:
I think it's because it is on your repo and we don't execute Jenkinsfiles from unknown origins.  I don't see any issues that Linux would complain about either

modules/touch/include/tuioear.h
```diff
@@ -52,7 +52,7 @@ namespace openspace {
 
 class TuioEar : public TUIO::TuioListener {
 public:
-    TuioEar();
+    TuioEar(int port = 3333);
```
alexanderbock:
```suggestion
    explicit TuioEar(int port = 3333);
```

modules/touch/touchmodule.cpp
```diff
@@ -66,18 +73,28 @@ namespace {
         "relatively spherical objects.",
         openspace::properties::Property::Visibility::AdvancedUser
     };
+
+    struct [[codegen::Dictionary(TouchModule)]] Parameters {
+        // [[codegen::verbatim(TuioPortInfo.description)]]
+        std::optional<int> tuioPort;
```
alexanderbock:
```suggestion
        std::optional<int> tuioPort [[codegen::inrange(1, 65535)]];
```

alexanderbock: Don't think `0` is a valid port number?

k2d222: It exists, but it is reserved. I applied the suggestion.

PR #3835, opened 2025-10-28
Video examples cleanup and documentation

WeirdRubberDuck:
Add documentation for the classes using the Video Player, and convert examples into our new format

This also includes changing the `RenderablePlane`'s size parameter to optional, to make the examples as minimal as possible.

### DIFF
--- a/modules/video/src/videoplayer.cpp
+++ b/modules/video/src/videoplayer.cpp
@@ -135,7 +135,7 @@ namespace {
 
         // The mode of how the video is played back. The Default is `RealTimeLoop`,
         // which means that the video is played in realtime using the `Play` command
-        // in the user interface under `Video Player`.
+        // in the user interface.
         std::optional<PlaybackMode> playbackMode;
     };
 #include "videoplayer_codegen.cpp"


WeirdRubberDuck:
In case someone is interested in how the documentation for the video classes looks on the docs: 
image

modules/video/src/renderablevideoplane.cpp
```diff
@@ -27,12 +27,24 @@
 #include <openspace/documentation/documentation.h>
 #include <openspace/documentation/verifier.h>
 
+namespace {
+    // This `Renderable` creates a textured 3D plane where the texture is a video.
+    //
+    // The video can either be played back based on a given simulation time
```
Roxeena, line 33: Might be nitpicky but is this clear enough that the user needs to trigger the video themselves when the video is not mapped to the simulation time? Do we want to emphasis this more to avoid further confusion?

WeirdRubberDuck: Hm, this is what I tried to emphasize in this description 🤔 Any suggestion for how to make it clearer? 

Maybe adding another sentence after this paragraph, saying something like this: 
"Note that, unless playback is mapped to simulation time, the video must be started manually via the user interface settings."
?

WeirdRubberDuck: It would also be great to create a better UI component for video playback, to make it easier to play videos. I'll create an issue for this :)

data/assets/examples/renderable/renderablevideoplane/videoplane.asset
```diff
@@ -0,0 +1,37 @@
+-- Basic
+-- Creates a video plane located in 3D space that plays a video file. The dimensions of
+-- the plane are automatically set to match the aspect ratio of the video.
+
+-- The video file is here downloaded from a URL to a location on our servers. This
```
Roxeena: (Comment that disappeared) I think this could potentially be confusing. It sounds like we are downloading the content form any URL onto our servers and then send a link to the downloaded object there. But this could just be me misunderstanding this too.

modules/video/src/renderablevideoplane.cpp
Roxeena: I agree that a better UI would be great, it is a little hidden right now inside many steps in the current UI. And I like the suggested addition to make it clearer.

WeirdRubberDuck:
Good catch with the video layer on Earth @Roxeena. Fixing this seems a bit out of scope for this PR, however. And I wonder why the mapping is so off 🤔

WeirdRubberDuck:
> Good catch with the video layer on Earth @Roxeena. Fixing this seems a bit out of scope for this PR, however. And I wonder why the mapping is so off 🤔

Nvm it's actually quite apparent - the video itself is a bit off. Here's the video (left) vs the blue marble texture. So the solution is probably to update the example video. 

image

Roxeena (approved):
:+1: I think it is fine to leave the video as is in this PR. Maybe it could be nice with some documentation or comment somewhere about it that we know the example is off aligned. Maybe we can create a new issue for it?

WeirdRubberDuck:
> 👍 I think it is fine to leave the video as is in this PR. Maybe it could be nice with some documentation or comment somewhere about it that we know the example is off aligned. Maybe we can create a new issue for it?

I'll make an issue :)

PR #3837, opened 2025-10-28
Add angular fisheye texture mapping for spheres

### LINKED ISSUE #3720: Render fisheye images or videos
Add the ability to render a fisheye image or a video onto a RenderableSphere by using only the top half of the sphere.

WeirdRubberDuck:
Add Angular Fisheye texture mapping for spheres, allowing us to show fisheye images or video onto a `RenderableSphere` (covering half of the sphere). 

Includes examples for `RenderableSphereImageLocal`, `RenderableSphereImageOnline`, and `RenderableVideoSphere`. 

Note that it duplicates the added documentation for `RenderableVideoSphere` added in PR #3835 and adds the details of the texture mapping. That PR should be merged first to keep this in sync.

Also fixes an issue with `RenderableVideoSphere` not updating correctly when changing sphere properties during runtime. 

Closes #3720


modules/base/shaders/sphere_vs.glsl
```diff
@@ -36,11 +36,51 @@ uniform mat4 modelViewProjection;
...
+
+// Remap equirectangular texture coordinates into angular fisheye
+vec2 equiToAngularFisheye(vec2 textureCoords) {
+  vec2 pos2 = textureCoords * 2.0 - 1.0; // Map [0,1] tex coords to [-1,1]
+
+  // 2D equi to 3D vector
+  float lat = pos2.y * 0.5 * M_PI;
+  float lon = pos2.x * M_PI;
+
+  // Map to 3D position, with Z begin the north pole
```
WeirdRubberDuck:
```suggestion
  // Map to 3D position, with Z being the north pole
```

data/assets/examples/renderable/renderablesphereimagelocal/sphereimagelocal_fisheye.asset
```diff
@@ -0,0 +1,28 @@
...
+-- will cover half of the sphere.
+
+local Node = {
+  Identifier = "RenderableSphereImageLocal_Example_FisheyeMapping",
+  Renderable = {
+    Type = "RenderableSphereImageLocal",
+    Texture = openspace.absPath("${DATA}/test3.jpg"),
+    TextureProjection = "Angular Fisheye",
+    -- Set orientation to also render the inside of the sphere (which is the correct view
+    -- for a fisheye/fulldome video)
```
Roxeena:
```suggestion
    -- for a fisheye/fulldome image)
```

data/assets/examples/renderable/renderablesphereimageonline/sphereimageonline_fisheye.asset
```diff
@@ -0,0 +1,27 @@
...
+-- online URL and mapped to the sphere using Angular Fisheye projection. The image will
+-- cover half of the sphere.
+
+local Node = {
+  Identifier = "RenderableSphereImageOnline_Example_FisheyeMapping",
+  Renderable = {
+    Type = "RenderableSphereImageOnline",
+    URL = "http://data.openspaceproject.com/examples/renderableplaneimageonline.jpg",
+    TextureProjection = "Angular Fisheye",
+    -- Set orientation to also render the inside of the sphere
```
Roxeena: Should we also add here that it is the correct view for a fisheye/fulldome image?

data/assets/examples/renderable/renderablevideosphere/videosphere_fisheye.asset
```diff
@@ -0,0 +1,42 @@
+-- Fulldome Fisheye Video
+-- Creates a 3D sphere with an angular fisheye video (fulldome) mapped onto its surface.
+-- The video will be shown on half of the sphere.
+
+-- The video file is here downloaded from a URL to a location on our servers. This
```
Roxeena: This is an URL sync so (if I remember and understand them correct) that means we can give any URL and it will be downloaded to the local machine, right? So that does not require our servers at all? Or maybe I am remembering it wrong. But either way I still think this could be confusing as I mentioned in the other PR too.

data/assets/examples/renderable/renderablevideosphere/videosphere_fisheye.asset
```diff
@@ -0,0 +1,42 @@
...
+local Node = {
+  Identifier = "RenderableVideoSphere_Example",
+  Renderable = {
+    Type = "RenderableVideoSphere",
+    Video = video,
+    TextureProjection = "Angular Fisheye",
+    -- Set orientation to also render the inside of the sphere (which is the correct view
+    -- for a fisheye/fulldome video)
+    Orientation = "Both",
+    -- Increasing the segments makes the sphere smoother and reduces distortion at the
```
Roxeena:
```suggestion
    -- Increasing the number of segments makes the sphere smoother and reduces distortion at the
```
Just to be a bit more clearer?

modules/base/rendering/renderablesphere.cpp
```diff
@@ -88,6 +88,20 @@ namespace {
...
 
+    enum class TextureProjection {
+        Equirectangular,
+        AngularFisheye
+    };
+
+    constexpr openspace::properties::Property::PropertyInfo TextureProjectionInfo = {
+        "TextureProjection",
+        "Texture Projection",
+        "Specifies the projection mapping to use any texture loaded onto the sphere "
```
Roxeena:
```suggestion
        "Specifies the projection mapping to use for any texture loaded onto the sphere "
```

modules/base/rendering/renderablesphere.cpp
```diff
@@ -145,10 +159,11 @@ namespace {
...
     };
 
-    // This `Renderable` represents a simple sphere with an image. The image that is shown
-    // should be in an equirectangular projection/spherical panoramic image or else
-    // distortions will be introduced. The `Orientation` parameter determines whether the
-    // provided image is shown on the inside, outside, or both sides of the sphere.
+    // This `Renderable` represents a simple sphere with an image. Per default, the
+    // sphere uses an equirectangular projection for the image mapping.
+    //
+    // The `Orientation` parameter determines whether the provided image is shown on
```
Roxeena, line 165: Should this be here? I actually kind of like it but it will be duplicated since it is a property in itself and will have its documentation further down the page already.  Although it looks like we used to have it here before as well 🤷‍♀️

modules/base/rendering/renderablesphereimagelocal.cpp
```diff
@@ -44,6 +44,10 @@ namespace {
     // This `Renderable` shows a sphere with an image provided by a local file on disk. To
     // show a sphere with an image from an online source, see
     // [RenderableSphereImageOnline](#base_screenspace_image_online).
+    //
+    // Per default, the sphere uses an equirectangular projection for the image mapping
+    // and hence expects an equirectangular image. However, it can also be used show
```
Roxeena:
```suggestion
    // and hence expects an equirectangular image. However, it can also be used to show
```

modules/base/rendering/renderablesphereimageonline.cpp
```diff
@@ -68,6 +68,10 @@ namespace {
     // will be downloaded when the `Renderable` is added to a scene graph node. To show a
     // sphere with an image from a local file, see
     // [RenderableSphereImageLocal](#base_screenspace_image_local).
+    //
+    // Per default, the sphere uses an equirectangular projection for the image mapping
+    // and hence expects an equirectangular image. However, it can also be used show
```
Roxeena:
```suggestion
    // and hence expects an equirectangular image. However, it can also be used to show
```

modules/video/src/renderablevideosphere.cpp
```diff
@@ -28,12 +28,27 @@
 #include <openspace/documentation/verifier.h>
 #include <openspace/util/sphere.h>
 
+namespace {
+    // This `Renderable` creates a textured 3D sphere where the texture is a video. Per
+    // default, the sphere uses an equirectangular projection for the image mapping
+    // and hence expects a video in equirectangular format. However, it can also be used
+    // to play fisheye videos by changing the `TextureProjection`.
```
Roxeena, line 35: Should we mention here again that it only covers half the sphere?

Roxeena (commented):
:+1: Tested the examples too.

modules/base/rendering/renderablesphere.cpp
WeirdRubberDuck: Yeah we used to have it here before, so I just kept it. Also in general I think it's nice to highlight some important properties/arguments/members in the main documentation string, so having it here does not hurt :) We do this for e.g. [RenderablePotinCloud](https://docs.openspaceproject.com/latest/reference/asset-components/Renderable/RenderablePointCloud.html) as well

However, the docs for `RenderableSphere` is not actually visible anywhere on docs.openspaceproject.com, so it doesn't matter much what is in this documentation at the moment 😅 It might in the future though, I guess

data/assets/examples/renderable/renderablesphereimageonline/sphereimageonline_fisheye.asset
WeirdRubberDuck: Yep, it should match the other example. Good catch :)

data/assets/examples/renderable/renderablevideosphere/videosphere_fisheye.asset
WeirdRubberDuck: I get your point. So

```suggestion
-- The video file is downloaded from a URL. This
```
?

WeirdRubberDuck: But in general, the URL sync is a bit confusing (e.g. it returns the path to the folder where the fgile is located and not the file itself). So some type of comment is necessary here, I think 🤔

modules/base/rendering/renderablesphereimageonline.cpp
WeirdRubberDuck: So many typos... Thank you :)

modules/video/src/renderablevideosphere.cpp
WeirdRubberDuck: I think that in this case it's fine to just check the documentation for the `TextureProjection` argument?

data/assets/examples/renderable/renderablevideosphere/videosphere_fisheye.asset
WeirdRubberDuck: Also, could not find the comment about this in the other PR 🤔 Are you certain that you published it?

Roxeena: Strange that comment seem to just have disappeared. I'll add it again in the other PR.

modules/base/rendering/renderablesphere.cpp
Roxeena: I agree that it is nice that it is here, I just wondered what we thought about the possible duplication ☺️

modules/base/shaders/sphere_vs.glsl
```diff
@@ -36,11 +36,51 @@ uniform mat4 modelViewProjection;
...
+
+  // 2D equi to 3D vector
+  float lat = pos2.y * 0.5 * M_PI;
+  float lon = pos2.x * M_PI;
+
+  // Map to 3D position, with Z being the north pole
+  vec3 pos3 = vec3(
+    cos(lat)*cos(lon),
+    sin(lat),
+    cos(lat)*sin(lon)
```
engbergandreas:
```suggestion
    cos(lat) * cos(lon),
    sin(lat),
    cos(lat) * sin(lon)
```

modules/base/rendering/renderablesphere.cpp
```diff
@@ -67,8 +67,8 @@ namespace {
         openspace::properties::Property::Visibility::AdvancedUser
     };
 
-    enum class Orientation : int {
-        Outside = 0,
+    enum class Orientation {
+        Outside ,
```
alexanderbock:
```suggestion
        Outside,
```

modules/base/rendering/renderablesphere.cpp
```diff
@@ -243,6 +267,15 @@ RenderableSphere::RenderableSphere(const ghoul::Dictionary& dictionary)
     _mirrorTexture = p.mirrorTexture.value_or(_mirrorTexture);
     addProperty(_mirrorTexture);
 
+    _textureProjection.addOptions({
+        { static_cast<int>(TextureProjection::Equirectangular), "Equirectangular" },
+        { static_cast<int>(TextureProjection::AngularFisheye), "Angular Fisheye" },
```
alexanderbock:
```suggestion
        { static_cast<int>(TextureProjection::AngularFisheye), "Angular Fisheye" }
```

modules/base/shaders/sphere_vs.glsl
```diff
@@ -36,11 +36,51 @@ uniform mat4 modelViewProjection;
...
+
+  float coverAngle = M_PI; // 180 degrees
+
+  // 3D vector to normalized 2D fisheye [-1,1]
+  float r = 2.0 / coverAngle * atan(sqrt(dot(pos3.xz, pos3.xz)), pos3.y);
+  float theta = atan(pos3.z, pos3.x);
+  vec2 fisheye2D = vec2(r * cos(theta), r * sin(theta));
+
+  if (r > 1.0) {
+    return vec2(-1, -1); // Indicate invalid coordinates
```
alexanderbock:
```suggestion
    return vec2(-1.0, -1.0); // Indicate invalid coordinates
```

alexanderbock (approved):

WeirdRubberDuck:
> In the VideoSphere, there seems to be an issue with a black border on the last pixel of the image?
> 

Good catch @alexanderbock ! There's also a stretch on the other side of the hemisphere (visible in your image). I noticed this at some point, but apparently forgot to fix it. 

The issue seems to be related to the interpolation of the texture coordinates - moving the coordinate mapping code to the fragment shader fixes it. Doing that should not be an issue, right?

image

There is still some aliasing going on at the edges, but this is at least much less of an issue.

alexanderbock:
Doing that in the fragment shader shouldn't be an issue I think 👍   And yea, the aliasing isn't really noticable unless you are really close

PR #3841, opened 2025-11-15
these changes fix cpack for creating zip files

hn-88:
Tested cpack -G ZIP works at workflow https://github.com/hn-88/OpenSpace-AppImage/blob/testing/.github/workflows/windows-latest.yml

Tested the contents of the zip file by comparing with the release zip file created with deploy.bat, comparison workflow is at https://github.com/hn-88/OpenSpace-AppImage/blob/main/.github/workflows/Compare%20OpenSpace%20Windows%20ZIP%20Contents.yml

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,7 @@ if (UNIX AND NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   endif ()
 endif ()
 
-add_subdirectory(ext)
+add_subdirectory(ext EXCLUDE_FROM_ALL)
 add_subdirectory(src)
 
 add_subdirectory(support/coding/codegen)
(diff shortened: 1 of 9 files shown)


alexanderbock:
I think there might be something wrong with the script?  Getting this error when trying to run the `package` target:

```
CMake Error at A:/openspace/OpenSpace/build/modules/globebrowsing/cmake_install.cmake:36 (file):
9>    file INSTALL cannot make directory
9>    "A:/openspace/OpenSpace/build/_CPack_Packages/win64/ZIP/OpenSpace
9>    ../modules/globebrowsing/gdal_data": No such file or directory.
9>  Call Stack (most recent call first):
9>    A:/openspace/OpenSpace/build/modules/cmake_install.cmake:97 (include)
9>    A:/openspace/OpenSpace/build/cmake_install.cmake:47 (include)
```
and the `globebrowsing/gdal_data` folder does exist, so I'm wondering if it is an issue with the current working directory?


Also, is the code looking for the DirectX compiler necessary?  I thought  that CEF was already including those DLLs automatically?

hn-88:
@alexanderbock , did you run the packaging with `cpack -G ZIP`? Because that is what [I have tested](https://github.com/hn-88/OpenSpace-AppImage/actions/runs/19386935800/job/55475159611). I have not tested the NSIS bundling. I tested the zip file against the release zip file, to make sure the same files were bundled. You could also take a look at my workflow linked in the above comment, to see if there is some env var or some other difference in the way you have called cmake (eg. cmake variables) vs what I had done.

DirectX compiler DLL - I did not know that CEF bundles those. If CEF bundles the correct dll, then we can copy from the CEF path instead, since we already download the correct architecture's CEF.

alexanderbock:
Closing it in favor of #4286

PR #3845, opened 2025-11-19
Added color mapping options to exoplanet asset

micahnyc:
In order to color map the circles correctly, I had to change the circle to grayscale so it could be colorized. Then it has a static color to bring it back to the blue color we had before.

[2.zip](https://github.com/user-attachments/files/23617921/2.zip)

### DIFF
--- a/data/assets/scene/digitaluniverse/exoplanets.asset
+++ b/data/assets/scene/digitaluniverse/exoplanets.asset
@@ -63,11 +63,10 @@ local Object = {
       ring centered on each host star. The ring is not intended to signify an orbit, but
       serve only as a marker. The labels list the host star name, and if there is more
       than one planet, will list the number of planets in parentheses.
-      Colors: When enabling color mapping, matplotlib 'sequential-cool' is used, by
-      default mapping the number of planets found, with options available for discovery
-      year and method. For number of planets, the colors are start at teal for one planet,
-      and progress to magenta for the maximum number of planets found in a system (which is
-      currently 8). Census: 5,589 planets in 4,139 systems.]]
+      Colors: When enabling color mapping, the colormap matplotlib 'sequential-cool'
+      is used, mapping lower values to the color teal and increasing to magenta. By
+      default the number of found planets is mapped to the colormap, with options
+      available for discovery year and method. Census: 5,589 planets in 4,139 systems.]]
   }
 }
 


alexanderbock:
Added to the data server

alexanderbock:
There is still some quite big difference compared to 0.21.2 though:

Master:
master

0.21.2:

data/assets/scene/digitaluniverse/exoplanets.asset
```diff
@@ -28,7 +30,20 @@ local Object = {
     },
     Opacity = 1.0,
     Texture = {
-      File = textures .. "target-blue.png",
+      File = textures .. "target-white.png",
+    },
+    Coloring = {
+      ColorMapping = {
+        File = colormaps.Sequential2.Cool,
```
Roxeena, line 37: When I tried this with the colormap, I thought that it was hard to distinguish between the color for like 1 and 2  planets, their color was too similar. I was thinking, if we know that the range of values are 1-8 then could we maybe use a qualitative color map instead? Then every value would have its own color and it would be easier to distinguish between them. But this should probably be discussed with the rest of the team.

data/assets/scene/digitaluniverse/exoplanets.asset
```diff
@@ -47,8 +62,12 @@ local Object = {
...
       ring centered on each host star. The ring is not intended to signify an orbit, but
       serve only as a marker. The labels list the host star name, and if there is more
-      than one planet, will list the number of planets in parentheses. Census: 5,589
-      planets in 4,139 systems.]]
+      than one planet, will list the number of planets in parentheses.
+      Colors: When enabling color mapping, matplotlib 'sequential-cool' is used, by
+      default mapping the number of planets found, with options available for discovery
+      year and method. For number of planets, the colors are start at teal for one planet,
+      and progress to magenta for the maximum number of planets found in a system (which is
+      currently 8). Census: 5,589 planets in 4,139 systems.]]
```
Roxeena:
```suggestion
      than one planet, will list the number of planets in parentheses.
      Colors: When enabling color mapping, the colormap matplotlib 'sequential-cool'
      is used, mapping lower values to the color teal and increasing to magenta. By
      default the number of found planets is mapped to the colormap, with options
      available for discovery year and method. Census: 5,589 planets in 4,139 systems.]]
```
?

alexanderbock: "going up" -> "increasing"
?

alexanderbock:
Updated the target-white.png image server-side again

PR #3850, opened 2025-11-21
Added mouse4 and mouse5 buttons mappings to (ctrl and ctrl+shift) in order to allow landing on planets with mouse only

micahnyc:


src/interaction/mousecamerastates.cpp
```diff
@@ -90,8 +92,8 @@ void MouseCameraStates::updateStateFromInput(const MouseInputState& mouseState,
         _currentSensitivityRamp * SensitivityAdjustmentIncrease;
 
     // Update the mouse states
-    if (primaryPressed && !keyShiftPressed && !keyAltPressed) {
-        if (keyCtrlPressed) {
+    if ( (primaryPressed || button4Pressed) && !keyShiftPressed && !keyAltPressed) {
```
alexanderbock:
```suggestion
    if ((primaryPressed || button4Pressed) && !keyShiftPressed && !keyAltPressed) {
```

PR #3857, opened 2025-12-03
Make the profile editor (launcher GUI) and thus the Qt6 dependency optional

braunms:
The reason for the requested change is that the dependency on newer versions of Qt6 (>=6.8.0) prevents OpenSpace from running on the Windows Server 2016 operating system [1] [2].

Running the binary file on Windows Server 2016 causes OpenSpace to crash immediately with the following error message: 
"The procedure entry point SetThreadDescription could not be loaded in the dynamic link library Qt6Core.dll". 
Since the Qt libraries are included in the main.cpp file, no crash or log files are created, and using the --bypassLauncher command line option does not cause the Qt6 DLLs to be skipped.

The solution proposed here is to add an additional option in CMAKE to make the profile editor (launcher GUI), which is the only project that uses Qt, optional. However, if the profile launcher is skipped, you must either change the desired profile and window options in openspace.cfg or set it using the command line parameters.

[1] https://qt-project.atlassian.net/browse/QTBUG-134075
[2] https://lists.qt-project.org/pipermail/development/2025-March/046207.html

### DIFF
--- a/apps/OpenSpace/CMakeLists.txt
+++ b/apps/OpenSpace/CMakeLists.txt
@@ -143,14 +143,13 @@ add_custom_command(TARGET OpenSpace POST_BUILD
 
 end_header("Dependency: SGCT")
 
-# Profile Editor / Launcher GUI 
-option(OPENSPACE_PROFILE_EDITOR "Build profile editor (launcher GUI)" ON)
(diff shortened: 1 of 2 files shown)


alexanderbock:
Awesome!  I just made some adaptations to bring it into coding style.   Thanks a bunch!

PR #3878, opened 2025-12-16
Fix travel speed indicator computation and clean up code a bit

### LINKED ISSUE #3225: `RenderableTravelSpeed` indicator rotation issue
There is an orientation problem for the travel speed indicator, so it won't reach the correct object due to being affected by the transformation of the parent node.

**To Reproduce**
Create a SGN with a `RenderableTravelSpeed`, setting the Earth as parent and the Moon as target, for example. 

Observe how the speed indicator will not travel in the correct direction

WeirdRubberDuck:
It incorrectly included the parent node's rotation

closes #3225

### DIFF
--- a/modules/space/rendering/renderabletravelspeed.cpp
+++ b/modules/space/rendering/renderabletravelspeed.cpp
@@ -285,7 +285,7 @@ void RenderableTravelSpeed::update(const UpdateData& data) {
         glm::inverse(sourceNode->modelTransform()) * glm::dvec4(_targetNode->worldPosition(), 1.0)
     );
 
-    // Assumes source position is at origin, i.e. a the parent node position
+    // Assumes source position is at origin, i.e. the parent node position
     _travelTime = glm::length(targetPosition) / _travelSpeed;
     _directionVector = glm::normalize(targetPosition);
 


modules/space/rendering/renderabletravelspeed.cpp
```diff
@@ -273,18 +277,22 @@ void RenderableTravelSpeed::update(const UpdateData& data) {
...
+    SceneGraphNode* sourceNode = parent();
+    ghoul_assert(sourceNode, "Renderable have to be owned by scene graph node");
+
+    // Target position, in the reference frame of the source node (to correctly inherit parent transform)
+    glm::dvec3 targetPosition = glm::dvec3(
+        glm::inverse(sourceNode->modelTransform()) * glm::dvec4(_targetNode->worldPosition(), 1.0)
+    );
 
-    _travelTime = glm::distance(_targetPosition, _sourcePosition) / _travelSpeed;
+    // Assumes source position is at origin, i.e. a the parent node position
```
alexanderbock:
```suggestion
    // Assumes source position is at origin, i.e. the parent node position
```

Issue #2913, opened 2023-10-06
New York Server map layers not working


micahnyc:
Currently a few layers on the new york map server are not working

Moon Wac
Mars Themis
Europa

ylvaselling:
![image](https://github.com/OpenSpace/OpenSpace/assets/24265984/4e017e2b-92f9-4338-956a-4d09a78dc62e)
Getting issues with MOC WA Color New York today as well

Arohdin:
Still issue on 0.20.0 RC1 for MOC WA Color New York. 
Works for Sweden and Utah versions.
![image](https://github.com/OpenSpace/OpenSpace/assets/9076357/87f93031-64ef-45aa-b8dd-46c5d6dbcd17)

Also got similar, but not the same, results on 0.19.2

alexanderbock:
Is this one fixed now? @micahnyc

micahnyc:
In prep for Utah server shutting down, I now found all the broken layers on AMNH server and replaced the files with working ones. The few listed above and 8 others.

PR #3890, opened 2026-01-22
Fix New York map layers

micahnyc:
Fixed wms for AMNH mola and made it the default for New york mars instead of utah. 

Closes https://github.com/OpenSpace/OpenSpace/issues/2913


micahnyc:
This layer was the only one that need asset updates, all the other broken layers were fixed by replacing the .ppg/.pjg files on the server

data/assets/scene/solarsystem/planets/mars/layers/newyork_layers.asset
```diff
@@ -8,7 +8,7 @@ asset.require("./colorlayers/themis_ir_night_newyork", false)
 asset.require("./colorlayers/ctx_mosaic_newyork", false)
 
 -- The New York server does not have this map, use the Utah server for now
```
alexanderbock:
```suggestion
```

PR #3896, opened 2026-01-30
Feature/topics

### LINKED ISSUE #3910: UI issues in missions/rosetta profile
For some reason, the UI panels do not render their content when running the Rosetta profile. Only the missions panel (which is shown at startup) renders its content

Image

Tested a couple of other mission profiles (JWST, New Horizons) and did not observe the same issue

ylvaselling:
* Creates a new topic where you can subscribe to _all property updates_
* Moves the topics into core
* Adds a global access to passing messages to a topic

As this is kind of a big rewrite, we might want to have some discussion about if there is some nice way we can make it easy to add new topics (I don't think this particular way of setting the topic, with a string, should be the final form)

To be used with [this PR](https://github.com/OpenSpace/OpenSpace-WebGui/pull/195)

Closes #3910


src/engine/configuration.cpp
```diff
@@ -243,6 +246,9 @@ namespace {
         // used if this is left out
         std::optional<HttpProxy> httpProxy;
 
+        //ghoul::Dictionary server [[codegen::reference("core_topicmanager")]];
```
WeirdRubberDuck: This will currently be interpreted as the documentation for this dictionary. Remove this comment and add a meaningful description of these settings. 

And, add the `codegen::reference` to the correct place? :)

src/properties/property.cpp
```diff
@@ -252,6 +253,14 @@ void Property::notifyChangeListeners() {
...
         p.second();
     }
+    nlohmann::json payload;
+    payload["event"] = "property_changed";
+    payload["payload"] = {
+        { "property", uri() },
+        { "value", nlohmann::json::parse(jsonValue()) }
+    };
+    global::topicManager->passDataToTopic("propertyTree", payload);
+
```
WeirdRubberDuck:
```suggestion
```

src/topic/topics/downloadeventtopic.cpp
```diff
@@ -22,12 +22,13 @@
...
 
-#include <modules/server/include/topics/downloadeventtopic.h>
+#include <openspace/topic/topics/downloadeventtopic.h>
 
 #include <openspace/engine/globals.h>
 #include <openspace/util/downloadeventengine.h>
 #include <ghoul/format.h>
-#include <string_view>
+#include <ghoul/logging/logmanager.h>
+#include <openspace/topic/connection.h>
```
WeirdRubberDuck:
```suggestion
#include <openspace/engine/globals.h>
#include <openspace/topic/connection.h>
#include <openspace/util/downloadeventengine.h>
#include <ghoul/format.h>
#include <ghoul/logging/logmanager.h>
```

src/topic/topics/enginemodetopic.cpp
```diff
@@ -22,12 +22,12 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#include "modules/server/include/topics/enginemodetopic.h"
+#include "openspace/topic/topics/enginemodetopic.h"
 
-#include <modules/server/include/connection.h>
+#include <openspace/topic/connection.h>
 #include <openspace/engine/globals.h>
+#include <openspace/query/query.h>
```
WeirdRubberDuck:
```suggestion
#include <openspace/engine/globals.h>
#include <openspace/query/query.h>
#include <openspace/topic/connection.h>
```

src/topic/topics/errorlogtopic.cpp
```diff
@@ -22,14 +22,11 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#include <modules/server/include/topics/errorlogtopic.h>
+#include <openspace/topic/topics/errorlogtopic.h>
 
-#include <modules/server/include/logging/notificationlog.h>
-#include <ghoul/logging/log.h>
+#include <openspace/topic/notificationlog.h>
+#include <openspace/topic/connection.h>
```
WeirdRubberDuck: Please go through all your include changes and make sure they are correctly sorted. I will stop commenting on this now

src/topic/topics/propertytreetopic.cpp
```diff
@@ -0,0 +1,102 @@
...
+        _isSubscribedTo = true;
+        _requestedResourceIsSubscribable = true;
+    }
+    if (event == StopSubscription) {
+        _isSubscribedTo = false;
+        resetCallbacks();
+    }
+    if (event == PropertyChanged) {
+        const nlohmann::json& payload = json.at("payload").get<nlohmann::json>();
+
```
WeirdRubberDuck:
```suggestion
```
nitpicky

src/topic/connection.cpp
```diff
@@ -22,51 +22,48 @@
...
+#include <openspace/topic/topics/actionkeybindtopic.h>
+#include <openspace/topic/topics/skybrowsertopic.h>
+#include <openspace/topic/topics/propertytreetopic.h>
+#include <openspace/topic/topics/subscriptiontopic.h>
+#include <openspace/topic/topics/timetopic.h>
+#include <openspace/topic/topics/topic.h>
+#include <openspace/topic/topics/triggerpropertytopic.h>
+#include <openspace/topic/topics/versiontopic.h>
+#include <openspace/engine/configuration.h>
+#include <openspace/engine/globals.h>
```
WeirdRubberDuck, line 28: These should be above the topic ones

src/topic/connection.cpp
```diff
@@ -250,4 +248,20 @@ void Connection::setAuthorized(bool status) {
     _isAuthorized = status;
 }
 
+Topic* Connection::findTopicByType(const std::string& type) {
+    const auto it = std::find_if(
+        _topics.begin(),
+        _topics.end(),
+        [&type](const auto& pair) {
+            //LINFO(pair.second->type());
```
WeirdRubberDuck:
```suggestion
```
?

src/topic/topicmanager.cpp
```diff
@@ -253,4 +256,13 @@ void ServerModule::removePreSyncCallback(CallbackHandle handle) {
     _preSyncCallbacks.erase(it);
 }
 
+void TopicManager::passDataToTopic(const std::string& topicType,
+                                   const nlohmann::json& jsonData) {
+    for (const ConnectionData& connectionData : _connections) {
```
WeirdRubberDuck:
```suggestion
                                   const nlohmann::json& jsonData)
{
    for (const ConnectionData& connectionData : _connections) {
```

src/topic/topicmanager.cpp
```diff
@@ -22,40 +22,43 @@
...
-#include <algorithm>
-#include <optional>
-#include <thread>
+#include <ghoul/misc/templatefactory.h>
 
 namespace {
     struct [[codegen::Dictionary(ServerModule)]] Parameters {
         std::optional<ghoul::Dictionary> interfaces;
         std::optional<std::vector<std::string>> allowAddresses;
         std::optional<int> skyBrowserUpdateTime;
```
WeirdRubberDuck: If you feel like it, please add some documentation to these parameters while you're at it :) 

This is especially important since this is seemingly now added as a required (?) parameter in the Configuration 
(and hence will show at the top of this list: https://docs.openspaceproject.com/latest/reference/asset-components/Other/Configuration.html)

src/topic/topicmanager.cpp
```diff
@@ -22,40 +22,43 @@
...
+#include <ghoul/io/socket/websocketserver.h>
+#include <ghoul/logging/logmanager.h>
 #include <ghoul/misc/profiling.h>
-#include <algorithm>
-#include <optional>
-#include <thread>
+#include <ghoul/misc/templatefactory.h>
 
 namespace {
     struct [[codegen::Dictionary(ServerModule)]] Parameters {
```
WeirdRubberDuck:
```suggestion
    struct [[codegen::Dictionary(TopicManager)]] Parameters {
```

include/openspace/topic/topics/topic.h
```diff
@@ -43,6 +41,7 @@ class Topic {
         int code = 500);
     virtual void handleJson(const nlohmann::json& json) = 0;
     virtual bool isDone() const = 0;
+    virtual std::string type() const;
```
WeirdRubberDuck, line 46: Should the Topic subclasses overload this one?  Doesn't seem like they do?

include/openspace/topic/topics/propertytreetopic.h
```diff
@@ -22,12 +22,37 @@
...
+    void resetCallbacks();
+
+    const int UnsetCallbackHandle = -1;
+
+    bool _requestedResourceIsSubscribable = false;
+    bool _isSubscribedTo = false;
+    int _onChangeHandle = UnsetCallbackHandle;
+    int _onDeleteHandle = UnsetCallbackHandle;
+    int _onMetaDataChangeHandle = UnsetCallbackHandle;
+    properties::Property* _prop = nullptr;
```
WeirdRubberDuck: Is there a reason for this variable name to be shortened?  If not, please write it out. 

According to coding style: https://docs.openspaceproject.com/latest/contribute/development/coding-style.html#abbreviations-in-names-must-be-avoided

src/engine/configuration.cpp
WeirdRubberDuck: Also, should this be optional? (like most the other parameters)

include/openspace/topic/topics/propertytreetopic.h
```diff
@@ -2,7 +2,7 @@
  *                                                                                       *
  * OpenSpace                                                                             *
  *                                                                                       *
- * Copyright (c) 2014-2026                                                               *
+ * Copyright (c) 2014-2025                                                               *
```
WeirdRubberDuck:
```suggestion
 * Copyright (c) 2014-2026                                                               *
```

src/topic/topics/propertytreetopic.cpp
```diff
@@ -0,0 +1,102 @@
+/*****************************************************************************************
+ *                                                                                       *
+ * OpenSpace                                                                             *
+ *                                                                                       *
+ * Copyright (c) 2014-2025                                                               *
```
WeirdRubberDuck:
```suggestion
 * Copyright (c) 2014-2026                                                               *
```

engbergandreas:
Ran into runtime issues with the cameratopic saying that it was compiled without globebrowsing support. Turns out that the `OPENSPACE_MODULE_SPACE_ENABLED` was not set even though it is enabled in CMake. 

Alex comment: 

> The _ENABLED compile definition is only set for projects that actually depend on the Space module in this case.  I think she moved the cameratopic into the core which doesn't depend on the Space module so it doesn't get the flag defined

engbergandreas:
Getting a crash when exiting OpenSpace in the `topicmanager.cpp` destructor 
image
Call stack: 
image

alexanderbock:
> Ran into runtime issues with the cameratopic saying that it was compiled without globebrowsing support. Turns out that the `OPENSPACE_MODULE_SPACE_ENABLED` was not set even though it is enabled in CMake.
> 
> Alex comment:
> 
> > The _ENABLED compile definition is only set for projects that actually depend on the Space module in this case.  I think she moved the cameratopic into the core which doesn't depend on the Space module so it doesn't get the flag defined

Fixed this since the functions that were previously gated by the Space module now all live in the core as well.

alexanderbock:
Is it just me or has the update rate of the UI slowed down?

There also seems to be some issue with the Deltatime not updating correctly:
https://github.com/user-attachments/assets/6c435568-bf6b-49cd-8035-c57a9c7288ab

alexanderbock:
> Getting a crash when exiting OpenSpace in the `topicmanager.cpp` destructor image Call stack: image

Fixed the crash

src/engine/configuration.cpp
ylvaselling: @alexanderbock Not really sure what to do about this, any thoughts?

include/openspace/topic/topics/topic.h
ylvaselling: For the current implementation, the topics don't need to overload this function -  since no one uses the global "send to topic" function but the propertyTree, all other types are "" which works. One option is to create overloaded functions when there is a need to be able to send data through simple calls to a global function.

The idea is that from any module you can pass json to a topic, just specify which topic and then the topicmanager will pass it to the topic that has that as its type, if it has been registered. This is brittle now because it uses a string to compare it to the types. This is what i referred to in the description of the PR.

But this is an architectural problem that might require some discussion and input from @alexanderbock (and maybe should be solved in a different PR)

ylvaselling:
> Is it just me or has the update rate of the UI slowed down?
> 
> There also seems to be some issue with the Deltatime not updating correctly: https://github.com/user-attachments/assets/6c435568-bf6b-49cd-8035-c57a9c7288ab

Fixed the time issue.

You probably need to use the built version to be able to compare properly. 

Run `npm run build` and then `npm preview` to be able to get a fair comparison between localhost:4680 and the dev gui.

If the problem persists, that is not good 😓

alexanderbock:
Should check if this fixes https://github.com/OpenSpace/OpenSpace/issues/3910 ?

ylvaselling:
> Should check if this fixes #3910 ?

For me it fixes it

include/openspace/topic/topics/topic.h
alexanderbock: I think we can leave the `type` parameter now and introduce it to the other types one by one?

src/engine/configuration.cpp
alexanderbock: A good documentation would be great.  Also agree that it should be `std::optional<ghoul::Dictionary> server [[codegen::reference("core_topicmanager")]];`  including a good documentation above

src/engine/globals.cpp
```diff
@@ -82,13 +85,13 @@ namespace {
...
         sizeof(MissionManager) +
         sizeof(ModuleEngine) +
-        sizeof(OpenSpaceEngine) +
         sizeof(ParallelPeer) +
         sizeof(RaycasterManager) +
         sizeof(RenderEngine) +
         sizeof(std::vector<std::unique_ptr<ScreenSpaceRenderable>>) +
         sizeof(SyncEngine) +
         sizeof(TimeManager) +
+        sizeof(TopicManager) +
```
alexanderbock:
```suggestion
```

There already is a `sizeof(TopicManager)` above

src/topic/topics/skybrowsertopic.cpp
```diff
@@ -22,10 +22,11 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#include "modules/server/include/topics/skybrowsertopic.h"
+#include "openspace/topic/topics/skybrowsertopic.h"
```
alexanderbock: This topic will have to live in the SkybrowserModule though?  Since it depends on there header files of it

src/topic/server.cpp
```diff
@@ -22,11 +22,16 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
  ****************************************************************************************/
 
-#include <modules/server/servermodule.h>
+#include <openspace/topic/topicmanager.h>
 
-#include <modules/server/include/connection.h>
-#include <openspace/documentation/documentation.h>
+#include <modules/globebrowsing/globebrowsingmodule.h>
```
alexanderbock, line 27: We can't depend on the globebrowsing module from the core

src/topic/topicmanager.cpp
```diff
@@ -123,8 +128,8 @@ void ServerModule::internalInitialize(const ghoul::Dictionary& configuration) {
     _skyBrowserUpdateTime = p.skyBrowserUpdateTime.value_or(_skyBrowserUpdateTime);
 }
 
-void ServerModule::preSync() {
-    // Set up new connections
+void TopicManager::preSync() {
+    // Set up new connections.
```
alexanderbock:
```suggestion
    // Set up new connections
```

src/engine/globals.cpp
```diff
@@ -132,6 +135,14 @@ void create() {
     memoryManager = new MemoryManager;
 #endif // WIN32
 
+#ifdef WIN32
+    topicManager = new (currentPos) TopicManager;
+    ghoul_assert(topicManager, "No topicManager");
+    currentPos += sizeof(TopicManager);
+#else // ^^^ WIN32 / !WIN32 vvv
```
WeirdRubberDuck:
```suggestion
#else // ^^^^ WIN32 / !WIN32 vvvv
```
to match changes in master

src/engine/globals.cpp
```diff
@@ -654,6 +667,13 @@ void destroy() {
     delete openSpaceEngine;
 #endif // WIN32
 
+    LDEBUGC("Globals", "Destroying 'TopicManager'");
+#ifdef WIN32
+    topicManager->~TopicManager();
+#else // ^^^ WIN32 / !WIN32 vvv
```
WeirdRubberDuck:
```suggestion
#else // ^^^^ WIN32 / !WIN32 vvvv
```

src/engine/openspaceengine.cpp
```diff
@@ -1156,6 +1160,8 @@ void OpenSpaceEngine::preSynchronization() {
         global::interactionMonitor->updateActivityState();
     }
 
+    global::topicManager->preSync();
+
```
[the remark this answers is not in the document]
ylvaselling: @alexanderbock

ylvaselling:
Moved the skybrowser topic to the skybrowser module.

Changes:
1) Topic factory now lives in `FactoryManager`
2) Topics are added to the factory in `TopicManager`
3) `AuthorizationTopic` is now not constructed with the password; instead this password lives in `Connection`, and is exposed to the AuthorizationTopic via a `get` function
4) Skybrowser update time now lives in SkyBrowserModule (I did not add it as a property bc it felt overengineered to me - it is only an in argument for the skybrowser module)
5) Added documentation for the Server (TopicManager) as well as linked the ServerInterface to the TopicManager in args, plus added ServerInterface to the DocsEngine so it gets added to the docs.

Questions:
1) Do these documentations make sense? I wasn't 100% sure what they actually are doing or how they work
2) Should we rename TopicManager to Server? it looks funky in the Configuration docs now
image
3) Is it ok to expose the password in a get function for the authorization topic?

@alexanderbock

src/topic/connection.cpp
engbergandreas, line 71:
```suggestion
    try {
        const nlohmann::json j = nlohmann::json::parse(message.c_str());
        handleJson(j);
    }
    catch (const std::domain_error& e) {
        LERROR(std::format("JSON handling error from: {}. {}", message, e.what()));
    }
    catch (const std::out_of_range& e) {
        LERROR(std::format("JSON handling error from: {}. {}", message, e.what()));
    }
```

The nested try catch seems a bit weird

engbergandreas: Double-check the indentation on this one as it got messed up when writing the review

include/openspace/topic/topics/authorizationtopic.h
```diff
@@ -22,27 +22,26 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
```
engbergandreas, line 22: To discuss the purpose of the `autorizationtopic`. To me its not clear how a user would use this, and why we would need it when we have black/white list already

include/openspace/topic/topics/bouncetopic.h
```diff
@@ -22,10 +22,10 @@
  * OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.                                         *
```
engbergandreas, line 22: Rename this topic to `pingtopic`?

src/documentation/core_registration.cpp
```diff
@@ -64,6 +64,8 @@
 #include <openspace/util/task.h>
 #include <openspace/util/time.h>
 #include <openspace/util/timerange.h>
+#include <openspace/topic/serverinterface.h>
+#include <openspace/topic/topicmanager.h>
```
engbergandreas:
```suggestion
#include <openspace/topic/serverinterface.h>
#include <openspace/topic/topicmanager.h>
#include <openspace/util/spicemanager.h>
#include <openspace/util/task.h>
#include <openspace/util/time.h>
#include <openspace/util/timerange.h>
```

src/properties/property.cpp
```diff
@@ -250,6 +251,13 @@ void Property::notifyChangeListeners() {
     for (const std::pair<OnChangeHandle, std::function<void()>>& p : _onChangeCallbacks) {
         p.second();
     }
+    nlohmann::json payload;
+    payload["event"] = "property_changed";
```
engbergandreas, line 259: I think it would be good to unify these keys and similar ones like `StartSubscription` in some util file. As we discussed, it should probably not be part of this PR, but we have similar keys sprinkled all over the place. It can be hard sometimes to know from the frontend and other places, and especially for our users, what those keys are

src/topic/topics/propertytreetopic.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+PropertyTreeTopic::~PropertyTreeTopic() {}
+
+bool PropertyTreeTopic::isDone() const {
+    return !_isSubscribedTo;
+}
+
+std::string PropertyTreeTopic::type() const {
+    return "propertyTree";
+}
```
engbergandreas, line 56: Having these as enums or something more robust in the future

src/topic/server.cpp
```diff
@@ -252,4 +306,23 @@ void ServerModule::removePreSyncCallback(CallbackHandle handle) {
...
 }
 
+void TopicManager::passDataToTopic(const std::string& topicType,
+                                   const nlohmann::json& jsonData)
+{
+    for (const ConnectionData& connectionData : _connections) {
+        if (auto topic = connectionData.connection->findTopicByType(topicType)) {
+            topic->handleJson(jsonData);
+        }
+    }
```
engbergandreas, line 316:
```suggestion
    for (const ConnectionData& connectionData : _connections) {
        Topic* topic = connectionData.connection->findTopicByType(topicType)
        if (topic) {
            topic->handleJson(jsonData);
        }
    }
```
Or at the very least

```suggestion
    for (const ConnectionData& connectionData : _connections) {
        if (Topic* topic = connectionData.connection->findTopicByType(topicType)) {
            topic->handleJson(jsonData);
        }
    }
```
idk what our stance is on defining variables inside the if-statement

src/properties/property.cpp
ylvaselling: Made an issue for this

include/openspace/topic/topics/authorizationtopic.h
ylvaselling: Made an issue for this

include/openspace/topic/topics/bouncetopic.h
ylvaselling: Made an issue for this

src/documentation/core_registration.cpp
ylvaselling: ? s is before t?

src/topic/topics/propertytreetopic.cpp
ylvaselling: Made an issue

src/topic/server.cpp
ylvaselling: @alexanderbock

alexanderbock:
> Moved the skybrowser topic to the skybrowser module.
> 
> Changes:
> 
>     1. Topic factory now lives in `FactoryManager`
> 
>     2. Topics are added to the factory in `TopicManager`
> 
>     3. `AuthorizationTopic` is now not constructed with the password; instead this password lives in `Connection`, and is exposed to the AuthorizationTopic via a `get` function
> 
>     4. Skybrowser update time now lives in SkyBrowserModule (I did not add it as a property bc it felt overengineered to me - it is only an in argument for the skybrowser module)
> 
>     5. Added documentation for the Server (TopicManager) as well as linked the ServerInterface to the TopicManager in args, plus added ServerInterface to the DocsEngine so it gets added to the docs.
> 
> 
> Questions:
> 
>     1. Do these documentations make sense? I wasn't 100% sure what they actually are doing or how they work
> 
>     2. Should we rename TopicManager to Server? it looks funky in the Configuration docs now
> 
> 
> image 3) Is it ok to expose the password in a get function for the authorization topic?
> 
> @alexanderbock

Yea, I think that shouldn't be a problem and makes more sense there anyway.  Don't see a reason why we'd want multiple auth topics with different passwords anyway

src/topic/server.cpp
alexanderbock: Yep, that's fine 👍 

I avoid doing it if it causes the line length to be too long since:

```cpp
if (int i = verylongfunctionname();
    i < 0)
{
```

looks a bit ugly to me, but it's within the guidelines

Issue #2594, opened 2023-04-11
Expose the idle behavior in a better way


alexanderbock:
A lot of users don't know about the idle behavior and we should make it more obvious.  One suggestion is to place it in the top right corner of the rendering window together with the frictions, but in a different color.

alexanderbock:
Either:
1. adding space between the friction buttons and the idle button
2. gray for off and active-blue color
3. slowly strobing a color

alexanderbock:
Should also have an action/keybinding

PR #3906, opened 2026-02-05
Add keybind for toggling idle behavior

WeirdRubberDuck:
Let me know if you have input on the choice of keybind or the description. Went with "R", for "Rotate"

Mentioned in issue: #2594

### DIFF
--- a/data/assets/actions/default_actions.asset
+++ b/data/assets/actions/default_actions.asset
@@ -42,6 +42,17 @@ local ToggleRollFriction = {
 }
 
 
+-- Idle behavior actions
+local ToggleIdleBehavior = {
+  Identifier = "os.ToggleIdleBehavior",
+  Name = "Toggle idle behavior",
+  Command = [[openspace.invertBooleanProperty("NavigationHandler.OrbitalNavigator.IdleBehavior.ApplyIdleBehavior")]],
+  Documentation = [[Toggles the idle behavior of the orbital navigator. If enabled, the camera will automatically rotate around the focus until the next navigation happens]],
+  GuiPath = "/Navigation",
+  IsLocal = true
+}
+
+
 -- UI actions
 
 local ToggleMainGui = {
@@ -224,6 +235,9 @@ asset.onInitialize(function()
   openspace.action.registerAction(ToggleZoomFriction)
   openspace.action.registerAction(ToggleRollFriction)
 
+  -- Idle behavior
+  openspace.action.registerAction(ToggleIdleBehavior)
+
   -- UI
   openspace.action.registerAction(ToggleMainGui)
   openspace.action.registerAction(ToggleNativeUi)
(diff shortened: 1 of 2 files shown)


micahnyc:
This keybinding would conflict with the space weather profiles:  

https://github.com/OpenSpace/OpenSpace/blob/19b76c39f62d3fe04f5f3df40b8bc287cee743f3/data/profiles/spaceweather/solarstorm2012.profile#L41

https://github.com/OpenSpace/OpenSpace/blob/19b76c39f62d3fe04f5f3df40b8bc287cee743f3/data/profiles/spaceweather/bastilleday2000.profile#L51

WeirdRubberDuck:
> This keybinding would conflict with the space weather profiles:
> 
> https://github.com/OpenSpace/OpenSpace/blob/19b76c39f62d3fe04f5f3df40b8bc287cee743f3/data/profiles/spaceweather/solarstorm2012.profile#L41
> 
> https://github.com/OpenSpace/OpenSpace/blob/19b76c39f62d3fe04f5f3df40b8bc287cee743f3/data/profiles/spaceweather/bastilleday2000.profile#L51

Good catch!  Did a search through the assets but apparently missed the profiles... Oops

There aren't that many letter keys that remain completely unused, it seems. Maybe `W`? Or some other "special" key?

Would appreciate suggestions. Or maybe the keybind isn't even needed once it can be toggled from the top bar?(https://github.com/OpenSpace/OpenSpace-WebGui/pull/199)

alexanderbock:
Would be nice to have a keybind for all of the top bar items.  @micahnyc   How about in the profiles move "D" -> "V"  (_V_olume makes more sense than _D_ensity volume as a mnemonic) and then "R" -> "D"  and then use R for the IdleBehavior?

micahnyc:
> Would be nice to have a keybind for all of the top bar items. @micahnyc How about in the profiles move "D" -> "V" (_V_olume makes more sense than _D_ensity volume as a mnemonic) and then "R" -> "D" and then use R for the IdleBehavior?

yeah i think this makes sense and feels right that our default keybinds get precedence over the profiles

WeirdRubberDuck:
> > Would be nice to have a keybind for all of the top bar items. @micahnyc How about in the profiles move "D" -> "V" (_V_olume makes more sense than _D_ensity volume as a mnemonic) and then "R" -> "D" and then use R for the IdleBehavior?
> 
> yeah i think this makes sense and feels right that our default keybinds get precedence over the profiles


So, should I update the other keybinds as part of this PR, or would it be preferable to have a separate commit message for those?   (I guess I could also just not squash the commits in this PR)

alexanderbock:
I'd make a separate commit in `master` moving the other keybinds and then directly squash-merge this PR afterwards.  That way they are close to each other in time, but separate commits so they won't get forgotten in the changelog

WeirdRubberDuck:
Ok, making the  "D" -> "V" and  "R" -> "D" remapping of keys makes sense in the bastilleday profile. 

However, in the solarstorm2012 profile, the keys are mapped to very different actions, having nothing to do with volumes. Does it still make sense to do the same change here? 
image

alexanderbock:
Agreeing totally with Micah that a good engine keybind is >> good profile keybind,  so I'd say that we move the keybinds for the solarstorm as well.  You can move them to some other key that makes sense and I put an item on the discussion list to go through all of the profile keybinds and rearrange and see what we do with them for the 0.22 release

PR #3909, opened 2026-02-11
updated config files to be more user friendly 

micahnyc:
I updated their names, and put them into folders. I also did a small bit of cleanup on some.

I did wind up deleting the spherical_mirror.json since it created a Rendering+GUI but the GUI was warped making it unusable. 

Open to any change or other ideas for folders etc but this feels better to me than what we had before.  

image


alexanderbock:
I think we need to find a better way to solve this (for example by providing a path in the meta data of the JSON, for example?   To maintain pain-free(ish) compatibility, we don't use capital letters in file names and adding spaces in filenames also makes life pretty difficult.

Definitely in favor of making the drop down menu more user-friendly, but maybe loading each JSON, extracting a path out of it and using that might easier to maintain. We have to load them anyway to extract the description text.

micahnyc:
I think your idea of loading the name from the json and showing it in the menu is definitely better. Closing this.

PR #3913, opened 2026-02-12
Fix linear path rotation issue when using duration

WeirdRubberDuck:
The position interpolation now occurs over the same duration as the rotation, and rotation duration limitation is included in path duration computation. 

This makes the behavior of the linear paths more intuitive. Now, it's just a linear interpolation from the start view to the end view, over a given duration if provided. 

Previously, the position interpolation was done separately from the rotation, and the rotation continued after reaching the new position. This led to weird behavior when flying between two close positions with different orientations, especially when specifying a custom duration. 

That is, this fixes an issue where linear paths behaved very weirdly when flying short distances and specifying a duration. Reported by a user from the Support slack, for a use case of using a linear path to interpolate between two navigation states with close positions .

Here's an example that led to weird behavior before this fix, and more intuitive behavior after: 
```lua
-- Settings before flying
openspace.time.setTime("2025-03-29T10:40:00")
openspace.setPropertyValueSingle("NavigationHandler.PathNavigator.IncludeRoll", true)

-- First trigger this flight
openspace.pathnavigation.createPath({
    TargetType = "NavigationState",
    Duration = 10.0,
    PathType = "Linear",
    NavigationState = {
        Anchor = "Earth",
        Pitch = 0.9816455871291493,
        Position = {
            -2981801.1859588623,
            6451204.055419922,
            -317543.15113830566
        },
        ReferenceFrame = "Root",
        Timestamp = "2025 MAR 29 10:40:00",
        Up = {
            0.03664153104209272,
            -0.032227641974919896,
            -0.9988086790249815
        },
        Version = 1,
        Yaw = 0.03705242248990501
    }
})


-- Then trigger the second flight, after the first one is finished
openspace.pathnavigation.createPath({
    TargetType = "NavigationState",
    Duration = 20.0,
    PathType = "Linear",
    NavigationState = {
        Anchor = "Earth",
        Pitch = 3.117073918091984,
        Position = {
            -2971245.494392395,
            6271231.989303589,
            -767341.3227539062
        },
        ReferenceFrame = "Root",
        Timestamp = "2025 MAR 29 10:40:00",
        Up = {
            0.06783458938283171,
            0.15277667432105824,
            0.9859298941946427
        },
        Version = 1,
        Yaw = -0.18006276296578352
    }
})
```

### DIFF
--- a/src/navigation/path.cpp
+++ b/src/navigation/path.cpp
@@ -183,8 +183,8 @@ Path::Path(Waypoint start, Waypoint end, Type type, std::optional<float> duratio
 
         // Seconds per pi angles. Per default, it takes 5 seconds to turn 90 degrees
         constexpr float BaseFactor = 5.f / glm::half_pi<float>();
-        float factor =
-            BaseFactor * global::navigationHandler->pathNavigator().linearRotationSpeedFactor();
+        float factor = BaseFactor *
+            global::navigationHandler->pathNavigator().linearRotationSpeedFactor();
 
         // Always at least 1 second
         const float minTurnDuration = std::max(angle * factor, 1.f);
@@ -297,7 +297,7 @@ void Path::resetPlaybackVariables() {
     _shouldQuit = false;
 }
 
-CameraPose Path::linearInterpolatedPose(double distance, double displacement
+CameraPose Path::linearInterpolatedPose(double distance, double displacement,
                                         double speedScale)
 {
     ghoul_assert(_type == Type::Linear, "Path type must be linear");


include/openspace/navigation/path.h
```diff
@@ -116,7 +116,7 @@ class Path {
      * path. Note that the linear path is a special case, to avoid risks of precision
      * problems for long paths.
      */
-    CameraPose linearInterpolatedPose(double distance, double displacement);
+    CameraPose linearInterpolatedPose(double distance, double displacement, double speedScale);
```
alexanderbock:
```suggestion
    CameraPose linearInterpolatedPose(double distance, double displacement,
        double speedScale);
```

src/navigation/path.cpp
```diff
@@ -172,6 +172,26 @@ Path::Path(Waypoint start, Waypoint end, Type type, std::optional<float> duratio
...
+        const glm::dvec3 a = ghoul::viewDirection(_start.rotation());
+        const glm::dvec3 b = ghoul::viewDirection(_end.rotation());
+        const float angle = static_cast<float>(std::acos(glm::dot(a, b)));
+
+        // Seconds per pi angles. Per default, it takes 5 seconds to turn 90 degrees
+        float factor = 5.f / glm::half_pi<float>();
+        factor *= global::navigationHandler->pathNavigator().linearRotationSpeedFactor();
+
+        const float minTurnDuration = std::max(angle * factor, 1.f); // Always at least 1 second
+        const float turnDuration = std::max(minTurnDuration, estimatedDuration);
```
alexanderbock:
```suggestion
        constexpr float BaseFactor = 5.f / glm::half_pi<float>();
        float factor =
            BaseFactor * global::navigationHandler->pathNavigator().linearRotationSpeedFactor();

        // Always at least 1 second
        const float minTurnDuration = std::max(angle * factor, 1.f);
        const float turnDuration = std::max(minTurnDuration, estimatedDuration);
```

?  A bit easier to see where the constant is

src/navigation/path.cpp
```diff
@@ -275,7 +295,7 @@ void Path::resetPlaybackVariables() {
     _shouldQuit = false;
 }
 
-CameraPose Path::linearInterpolatedPose(double distance, double displacement) {
+CameraPose Path::linearInterpolatedPose(double distance, double displacement, double speedScale) {
```
alexanderbock:
```suggestion
CameraPose Path::linearInterpolatedPose(double distance, double displacement
                                        double speedScale)
{
```

PR #3947, opened 2026-03-08
Add Profile add-ons

### LINKED ISSUE #3580: Add support for profile "variants"
For a few profiles, we want to give people the option to optionally include a fixed selection of other asset files.  Examples are the minor planets for Jupiter and Saturn for the default profile, other L2-orbiting satellites for Euclid, JWST, etc; but there are other examples as well.

This solution would ask the user in the launcher to pick variants and pass them to the engine to include in the initial profile load

alexanderbock:
Adds the ability for profiles to have add-ons, meaning optional elements that some users might want to opt-in for specific profiles. Note that this feature specifically only adds the ability to _have_ variants and adds add-ons to our provided assets. It does not include a user interface in the launcher to actually include add-ons in user-generated profiles.
The profile add-ons can be set from commandline arguments, the `openspace.cfg`, and the get stored when the settings for the last started profile are stored.

This change also adds some add-ons to existing profiles and removes the "asteroids" profile whose content has become an add-on for the "default" profile.

UI for launcher for add-ons variants looks like this:
image (20)

Closes #3580

### DIFF
--- a/apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
+++ b/apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
@@ -30,6 +30,7 @@
 #include "notificationwindow.h"
 #include "settingsdialog.h"
 #include "splitcombobox.h"
+#include "usericon.h"
 #include <openspace/engine/configuration.h>
 #include <openspace/engine/settings.h>
 #include <openspace/openspace.h>
@@ -83,12 +84,7 @@ namespace {
         constexpr QRect EditProfileButton(
             LeftRuler, TopRuler + 180, SmallItemWidth, SmallItemHeight
         );
-        constexpr QRect AddonBox(
-            LeftRuler,
-            TopRuler + 210,
-            SmallItemWidth,
-            SmallItemHeight
-        );
+        constexpr QRect AddonBox(LeftRuler, TopRuler + 210, ItemWidth, SmallItemHeight);
         constexpr QRect OptionsLabel(LeftRuler + 10, TopRuler + 255, 151, 24);
         constexpr QRect WindowConfigBox(LeftRuler, TopRuler + 285, ItemWidth, ItemHeight);
         constexpr QRect NewWindowButton(
@@ -303,8 +299,17 @@ LauncherWindow::LauncherWindow(bool profileEnabled, const Configuration& globalC
     _addonBox.combobox = new QComboBox(centralWidget);
     _addonBox.combobox->setObjectName("config");
     _addonBox.combobox->setGeometry(geometry::AddonBox);
+    _addonBox.combobox->setPlaceholderText("Addons");
     _addonBox.combobox->setAccessibleName("Select Addons");
     _addonBox.model = new QStandardItemModel;
+    connect(
+        _addonBox.model,
+        &QStandardItemModel::itemChanged,
+        this,
+        &LauncherWindow::updatePlaceholderText
+    );
+
+
     _addonBox.combobox->setModel(_addonBox.model);
     updateAddonsBox(globalConfig.profile.profile + ".profile");
     for (const std::string& addon : globalConfig.profile.addons) {
@@ -730,6 +735,26 @@ void LauncherWindow::updateStartButton() const {
     _startButton->setEnabled(!profilePath.empty() && !configPath.empty());
 }
 
+void LauncherWindow::updatePlaceholderText() {
+    int count = 0;
+    for (int i = 0; i < _addonBox.model->rowCount(); i++) {
+        QStandardItem* item = _addonBox.model->item(i);
+        const int state = item->data(Qt::CheckStateRole).toInt();
+        const bool checked = state == Qt::Checked;
+        if (checked) {
+            count++;
+        }
+    }
+
+    if (count > 0) {
+        std::string text = std::format("Addons ({} selected)", count);
+        _addonBox.combobox->setPlaceholderText(QString::fromStdString(text));
+    }
+    else {
+        _addonBox.combobox->setPlaceholderText("Addons");
+    }
+}
+
 void LauncherWindow::updateAddonsBox(const std::string& profile) {
     // Get a list of all of the potential variants
     std::vector<std::filesystem::path> addonsCore = ghoul::filesystem::walkDirectory(
@@ -747,7 +772,6 @@ void LauncherWindow::updateAddonsBox(const std::string& profile) {
 
     // First clear the model of the previous results
     _addonBox.model->clear();
-    _addonBox.combobox->setPlaceholderText("Addons");
 
     // Then recreate the new ones
     const std::filesystem::path coreCandidate = _profilePath / profile;
@@ -763,8 +787,17 @@ void LauncherWindow::updateAddonsBox(const std::string& profile) {
         std::filesystem::exists(userCandidate) ? userCandidate : coreCandidate
     );
 
-    auto toItem = [](const Addon& addon) {
-        QStandardItem* item = new QStandardItem(QString::fromStdString(addon.name));
+    QIcon icon = userIcon();
+
+    auto toItem = [&icon](const Addon& addon, bool isUser) {
+        std::string name = isUser ? std::format(" {}", addon.name) : addon.name;
+        QStandardItem* item = new QStandardItem(QString::fromStdString(name));
+        if (isUser) {
+            item->setIcon(icon);
+        }
+
+        item->setToolTip(QString::fromStdString(addon.description));
+
         item->setData(QString::fromStdString(addon.identifier));
         item->setFlags(Qt::ItemIsUserCheckable | Qt::ItemIsEnabled);
         item->setData(Qt::Unchecked, Qt::CheckStateRole);
@@ -789,14 +822,16 @@ void LauncherWindow::updateAddonsBox(const std::string& profile) {
         for (const Addon& addon : p.addons.custom) {
             profileAddons.insert(addon.identifier);
 
-            QStandardItem* i = toItem(addon);
+            constexpr bool IsUserFolder = false;
+            QStandardItem* i = toItem(addon, IsUserFolder);
             _addonBox.model->appendRow(i);
         }
         // Then the recommended
         for (const Addon& addon : p.addons.recommended) {
             profileAddons.insert(addon.identifier);
 
-            QStandardItem* i = toItem(addon);
+            constexpr bool IsUserFolder = false;
+            QStandardItem* i = toItem(addon, IsUserFolder);
             _addonBox.model->appendRow(i);
         }
     }
@@ -814,7 +849,8 @@ void LauncherWindow::updateAddonsBox(const std::string& profile) {
                 continue;
             }
 
-            QStandardItem* i = toItem(addon);
+            constexpr bool IsUserFolder = true;
+            QStandardItem* i = toItem(addon, IsUserFolder);
             _addonBox.model->appendRow(i);
         }
         // Then add all of the addons found in the core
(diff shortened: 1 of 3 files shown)


alexanderbock:
Adding a label about the checkboxes

alexanderbock:
Variant -> Add-ons

alexanderbock:
Extracting the add-on into a separate file that can be reused by different profiles

WeirdRubberDuck:
Or maybe it should rather be a combobox multiselect in this type of style, with a list of the current selections: 
image

alexanderbock:
Intermediate step in the design:
image


It's left to automatically add all addon files under not recommended

alexanderbock:
Now all addon files not explicitly mentioned are in the list as well;  ready for another pull request

alexanderbock:
image

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -78,16 +83,22 @@ namespace {
...
             LeftRuler, TopRuler + 180, SmallItemWidth, SmallItemHeight
         );
-        constexpr QRect OptionsLabel(LeftRuler + 10, TopRuler + 230, 151, 24);
-        constexpr QRect WindowConfigBox(LeftRuler, TopRuler + 260, ItemWidth, ItemHeight);
+        constexpr QRect AddonBox(
+            LeftRuler + 10,
+            TopRuler + 215,
+            SmallItemWidth - 10,
+            SmallItemHeight
+        );
```
WeirdRubberDuck: How about not adding this extra spacing and keeping the boxes aligned? Especially since the width control does not quite seem to work as expected. 

```suggestion
        constexpr QRect AddonBox(
            LeftRuler,
            TopRuler + 215,
            SmallItemWidth,
            SmallItemHeight
        );
```

It's clear from the grouping and title that it belongs to the profile selection :) 

Image

I also wonder why the width is not being applied correctly... 🤔 the `SmallItemWidth` should be the same as the New and Edit buttons, I guess

apps/OpenSpace/ext/launcher/resources/qss/launcher.qss
```diff
@@ -61,6 +61,18 @@ LauncherWindow  QComboBox#config:disabled {
...
+  border: 1px solid rgb(225, 225, 225);
+  border-radius: 2px;
+  padding: 1px 18px 1px 3px;
+  min-width: 14em;
+  font-size: 10pt;
+  font-family: Segoe UI;
+  font-weight: bold;
+  color: white;
+}
+
```
WeirdRubberDuck: I looked a bit at the qss and this should probably be removed, as it's jsut a duplicate of the styles for the other dropdowns. Instead, the dropdown could use the same css as the, by specifying the same "object name", i.e. `config`

Duplicating the styles easily leads to mismatched behavior. Currently, this dropdown does not have the focus outline, for example (this should be fixed by making sure it uses the same style as the other dropdowns)

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -293,6 +300,23 @@ LauncherWindow::LauncherWindow(bool profileEnabled, const Configuration& globalC
         newProfileButton->setMenu(menu);
     }
 
+    _addonBox.combobox = new QComboBox(centralWidget);
+    _addonBox.combobox->setObjectName("addons");
```
WeirdRubberDuck: This will lead to the style being the same as for the other two dropdowns
```suggestion
    _addonBox.combobox->setObjectName("config");
```

An alternative would be to clean up the QSS file a bit to make all the QComboBoxes in the launcher window always have the same style, no matter the object name.  Let me know if I can help with this

There seems to be some cleaning needed in general... the `:focus` styles should, for example, be at the bottom of the file rather than the top, as the styles are applied from top to bottom.

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -704,13 +730,132 @@ void LauncherWindow::updateStartButton() const {
     _startButton->setEnabled(!profilePath.empty() && !configPath.empty());
 }
 
+void LauncherWindow::updateAddonsBox(const std::string& profile) {
+    // Get a list of all of the potential variants
+    //const std::filesystem::path& path,
+        //Recursive recursive, Sorted sorted,
+        //std::function<bool(const std::filesystem::path&)> filter
```
WeirdRubberDuck:
```suggestion
```
?

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -704,13 +730,132 @@ void LauncherWindow::updateStartButton() const {
...
+    std::vector<std::filesystem::path> addonsUser = ghoul::filesystem::walkDirectory(
+        _userProfilePath,
+        ghoul::filesystem::Recursive::Yes,
+        ghoul::filesystem::Sorted::Yes,
+        [](const std::filesystem::path& path) { return path.extension() == ".addon"; }
+    );
+
+    // First clear the model of the previous results
+    _addonBox.model->clear();
+    _addonBox.combobox->setPlaceholderText("Addons");
```
WeirdRubberDuck:
```suggestion
    _addonBox.combobox->setPlaceholderText("Add-ons");
```
? 

We seem to alternate a bit between the two, which I guess is needed. But since you use `add-on` in the descriptive text maybe we should do this here too?

data/profiles/missions/euclid.profile
```diff
@@ -6,6 +6,19 @@
...
+      {
+        "version": { "major": 1, "minor": 0 },
+        "identifier": "l2_spacecraft",
+        "name": "Spacecraft in L2",
+        "description": "Includes other spacecraft that are in the Earth's L2 orbit.",
+        "assets": [
+          "scene/solarsystem/telescopes/jwst/jwst"
+        ]
+      }
+    ]
```
WeirdRubberDuck, line 20: This feature is a little weird to me. What's the motivation for being able to add custom add-ons in the profile? 

This feature will require adding a more complex UI to the profile editor (compared to just selecting a list of add-ons, for example), so I'm wondering if it's really needed. 

As a user, I might think: "Ooh, that L2 spacecraft add-on seems neat! I'll go and add it to one of my custom profiles...", but then I won't be able to find it.

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
WeirdRubberDuck: Based on our discussions about simplifying the qss file, I suggest applying this suggestion and removing the changes from the QSS file for this PR :)

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
alexanderbock: I changed the accessible name to Addons as well.  I felt "Add-ons" was grabbing too much visual attention

data/profiles/missions/euclid.profile
alexanderbock: I strongly feel that we will want to be able to add custom add-ons.  The apollo profile is the perfect example.  To me it doesn't make sense to have the Apollo 17 boulders as a standalone addon that we'd want to include in any random profile, so it appearing in the list available to all profiles would be just noise.

I don't think it would require a more complex UI, since ultimately we'd want users to create their own addons so that complex UI would need to be there anyway. But instead of saving the addon to disk, it would be saved directly into the profile instead.

apps/OpenSpace/main.cpp
```diff
@@ -1258,6 +1258,15 @@ int main(int argc, char* argv[]) {
         "Specifies the profile that should be used to start OpenSpace and that overrides "
         "the profile specified in the `openspace.cfg` and the settings."
     ));
+    parser.addCommand(std::make_unique<ghoul::cmdparser::SingleCommand<std::string>>(
+        commandlineArguments.profileAddons,
+        "--addons",
+        "-a",
+        "Specifies the add-ons of the profile that should be loaded. The provided addons "
```
WeirdRubberDuck:
```suggestion
        "Specifies the addons of the profile that should be loaded. The provided addons "
```

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -78,16 +83,22 @@ namespace {
         constexpr QRect EditProfileButton(
             LeftRuler, TopRuler + 180, SmallItemWidth, SmallItemHeight
         );
-        constexpr QRect OptionsLabel(LeftRuler + 10, TopRuler + 230, 151, 24);
-        constexpr QRect WindowConfigBox(LeftRuler, TopRuler + 260, ItemWidth, ItemHeight);
+        constexpr QRect AddonBox(
+            LeftRuler,
+            TopRuler + 210,
+            SmallItemWidth,
```
WeirdRubberDuck:
```suggestion
            ItemWidth,
```
If we want it to be the full width (currently overridden in QSS, but at least it's good to have it be correct here)

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -704,13 +730,128 @@ void LauncherWindow::updateStartButton() const {
...
+    std::vector<std::filesystem::path> addonsUser = ghoul::filesystem::walkDirectory(
+        _userProfilePath,
+        ghoul::filesystem::Recursive::Yes,
+        ghoul::filesystem::Sorted::Yes,
+        [](const std::filesystem::path& path) { return path.extension() == ".addon"; }
+    );
+
+    // First clear the model of the previous results
+    _addonBox.model->clear();
+    _addonBox.combobox->setPlaceholderText("Addons");
```
WeirdRubberDuck: How about adding the number of selected addons in the placeholder?  Like
```
Addons (2 selected)
```
or something?

So that it's visible from the top level that addons are included?

src/scene/profile_lua.inl
```diff
@@ -53,7 +53,14 @@ namespace {
...
 [[codegen::luawrap]] std::filesystem::path profilePath() {
-    return global::configuration->profile;
+    return global::configuration->profile.profile;
+}
+
+/**
+ * Returns the enabled add-ons of the profile that was started.
+ */
+[[codegen::luawrap]] std::vector<std::string> profileAddons() {
+    return global::configuration->profile.addons;
```
WeirdRubberDuck, line 63: Reminder to self that we should use this to list the loaded addons in the UI somewhere, close to the information about the loaded profile :) 

Or, maybe better, it could be part of the profile topic

apps/OpenSpace/ext/launcher/src/launcherwindow.cpp
```diff
@@ -293,6 +300,23 @@ LauncherWindow::LauncherWindow(bool profileEnabled, const Configuration& globalC
...
+    _addonBox.combobox->setModel(_addonBox.model);
+    updateAddonsBox(globalConfig.profile.profile + ".profile");
+    for (const std::string& addon : globalConfig.profile.addons) {
+        for (int i = 0; i < _addonBox.model->rowCount(); i++) {
+            QStandardItem* item = _addonBox.model->item(i);
+            std::string id = item->data().toString().toStdString();
+            if (id == addon) {
+                item->setData(Qt::Checked, Qt::CheckStateRole);
+            }
+        }
```
WeirdRubberDuck, line 322: Maybe out of scope and could be an issue, but do we want to add our little "user folder" icon to the items that come from the user folder?  Maybe overkill, until we have a way for users to create their own addons? 

E.g, here I have one from my user folder, but it's hard to know apart form that it's the first one in the list. And it's a bit confusing why it breaks the alphabetical ordering. 
Image

Adding the icon to the left or right of the addon name would solve it 
Image

PR #3960, opened 2026-03-17
Rename hover circle -> hover indicator

### LINKED ISSUE #2171: Rename "HoverCircle" to "HoverIndicator" or similar
There is actually nothing that forces this indicator in the skybrowser to have a circular shape. It just depends upon the texture used. It could be a crosshair, 🍌, or whatever the user desires

did not implement it as part of the fix for #2153, since renaming the functions and everything could be considered a breaking change, if some user has already changed the texture for the indicator.

WeirdRubberDuck:
Breaking change due to identifier and Lua API change

But both the identifier and function are mostly used by our own user interface. 

closes #2171

### DIFF
--- a/data/assets/modules/skybrowser/hover_indicator.asset
+++ b/data/assets/modules/skybrowser/hover_indicator.asset
@@ -23,7 +23,7 @@ local Indicator = {
   GUI = {
     Name = "Hover Indicator",
     Description = [[A marker that shows the position on the night sky of the object
-      hovered in the sky browser UI. The circle will hide/show up dynamically, depending
+      hovered in the sky browser UI. The marker will hide/show up dynamically, depending
       on the interaction with the items in the UI]],
     Path = "/SkyBrowser",
     Hidden = true


data/assets/modules/skybrowser/hover_indicator.asset
```diff
@@ -21,31 +21,30 @@ local circle = {
...
 
-asset.export(circle)
-
+asset.export(Indicator)
 
 
 asset.meta = {
-  Name = "SkyBrowser Hover Circle",
+  Name = "SkyBrowser Hover Indicator",
   Description = [[Includes a circular marker that shows the position on the night sky
```
ylvaselling, line 48:
```
  Description = [[Includes a circular marker that shows the position on the night sky
```

I don't know if you wanted to remove all references to it being a circle (personally I don't mind it) but here is another reference

data/assets/modules/skybrowser/hover_indicator.asset
```diff
@@ -21,31 +21,30 @@ local circle = {
     Texture = imageFolder .. "ring.png"
   },
   GUI = {
-    Name = "Hover Circle",
-    Description = [[A circular marker that shows the position on the night sky
-      of the object hovered in the sky browser UI. The circle will hide/show up
-      dynamically, depending on the interaction with the items in the UI]],
+    Name = "Hover Indicator",
+    Description = [[A marker that shows the position on the night sky of the object
+      hovered in the sky browser UI. The circle will hide/show up dynamically, depending
```
ylvaselling: Don't know if you wanted to remove all references to a "circle" (personally i dont mind it) but here is "circle"

WeirdRubberDuck: Good catch!  In this asset we could theoretically call it circle since we load a circle texture. But I'd prefer to change this to indicator as well :)

data/assets/modules/skybrowser/hover_indicator.asset
WeirdRubberDuck: Here I think it's fine to mention the circular marker. That is the texture that we load, after all

If the user changes it here, they're responsible for also changing the description, I'd say

PR #3964, opened 2026-03-19
Fix broken rendering of constellation bounds

WeirdRubberDuck:
Made this a PR since it didn't feel like a proper fix, and I could use some input. I just reverted the code back to how it was before the OpenGL 4.6 rewrite, essentially. 

Does anyone know why we need to `pscTransform` call here?  Seems like the only place we use it?

For context, without this fix, the constellation bounds render with faulty depth values:
image

With this fix, they correctly render behind Earth, like they used to: 
image

### DIFF
--- a/modules/space/shaders/constellationbounds_vs.glsl
+++ b/modules/space/shaders/constellationbounds_vs.glsl
@@ -43,5 +43,5 @@ void main() {
   // confusing
   vec4 position = pscTransform(pos, modelTransform);
   out_data.position = pos;
-  gl_Position =  z_normalization(viewProjection * position);
+  gl_Position = z_normalization(viewProjection * position);
 }


modules/space/rendering/renderableconstellationbounds.cpp
```diff
@@ -193,7 +193,7 @@ void RenderableConstellationBounds::initializeGL() {
     const GLint positionAttrib = _program->attributeLocation("in_position");
     glEnableVertexArrayAttrib(_vao, positionAttrib);
     glVertexArrayAttribFormat(_vao, positionAttrib, 3, GL_FLOAT, GL_FALSE, 0);
-    glVertexArrayAttribBinding(_vao, 0, 0);
+    glVertexArrayAttribBinding(_vao, positionAttrib, 0);
```
ylvaselling, line 196: I think this can be 
```
  constexpr GLuint positionAttrib = 0;
  glEnableVertexArrayAttrib(_vao, positionAttrib);
  glVertexArrayAttribFormat(_vao, positionAttrib, 3, GL_FLOAT, GL_FALSE, 0);
  glVertexArrayAttribBinding(_vao, positionAttrib, 0);
```

Because in the shader, `layout(location = 0) in vec3 in_position;` ?

ylvaselling (approved):
Looks good to me

modules/space/shaders/constellationbounds_vs.glsl
```diff
@@ -37,9 +37,11 @@ uniform mat4 modelTransform;
...
   vec4 pos = vec4(in_position, 0.0);
-  gl_Position = z_normalization(
-    viewProjection * pscTransform(pos, modelTransform)
-  );
+  // @TODO (2026-03-19, emmbr) This should be looked over so the pscTransform is not
+  // needed or reimplemented. It currently changes the input pos attribute, which is
+  // confusing
+  vec4 position = pscTransform(pos, modelTransform);
+  out_data.position = pos;
+  gl_Position =  z_normalization(viewProjection * position);
```
ylvaselling:
```suggestion
  gl_Position = z_normalization(viewProjection * position);
```

modules/space/rendering/renderableconstellationbounds.cpp
WeirdRubberDuck: It currently gets that value from the shader, so `positionAttrib` will be `0` already. Just wanted to fix the mismatch of using the number in this line, and the `positionAttrib` variable in the others.

In other places in the code, we alternate between doing it like this (i.e., getting the constant from the shader) or just specifying the number, like this, without even using a variable:

```
 glEnableVertexArrayAttrib(_vao, 0);
 glVertexArrayAttribFormat(_vao, 0, 3, GL_FLOAT, GL_FALSE, 0);
 glVertexArrayAttribBinding(_vao, 0, 0);
``` 

as I'm not sure what we prefer, I'd rather not change it as part of this PR

alexanderbock:
We _really_ need to get rid of the powerscaled coordinate remnants..  😓

modules/space/rendering/renderableconstellationbounds.cpp
alexanderbock: Yea, let's leave it here but come up with a way how to specify this.  I'm swinging back and forth on it as well

PR #3983, opened 2026-04-05
Add support for different blackout colors and images. Invert BlackoutFactor value

### LINKED ISSUE #2924: Blackout factor - add ability to use image or alternate color
Right now the blackout factor is only for black, it would be nice to be able to:

- set a color other than black
- set an image that is used instead of color

(Dev note, we will make a special screenspace renderable that is used by render engine etc)

alexanderbock:
This adds the ability to change the color used for the global blackout fade configurable through the RenderEngine. It also adds the ability to show an image on top of the blackout image.

Two breaking changes:
1. The blackout values have been moved into a separate PropertyOwner, thus changing the URI to the blackout factor
2. The blackout factor value was inverted compared to the previous solution for consistency with other sliders

Closes #2924

### DIFF
--- a/include/openspace/rendering/framebufferrenderer.h
+++ b/include/openspace/rendering/framebufferrenderer.h
@@ -160,7 +160,8 @@ public:
         const glm::ivec4& viewport);
     void performDeferredTasks(const std::vector<DeferredcasterTask>& tasks,
         const glm::ivec4& viewport);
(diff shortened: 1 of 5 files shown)


alexanderbock:
Known issue to be fixed;  transparency is weird when using factor and image factor at the same time

image

PR #3989, opened 2026-04-13
Use new relay server for Celestrak

### LINKED ISSUE #3973: Celestrak reports 403 error when requesting KVN data
When accessing data twice, the second access gets:

https://celestrak.org/NORAD/elements/gp.php?GROUP=starlink&FORMAt=kvn

```
GP data has not updated since your last successful
download of GROUP=starlink at 2026-03-27 08:11:36 UTC.
Data is updated once every 2 hours.
```

This causes the initialization to fail

alexanderbock:
New branch https://github.com/OpenSpace/OpenSpace/tree/issue/3973-celestrak that uses a new relay server set up to (i) ease load on the Celestrak server and (ii) not have the rate limiting.

The source for the relay server is here: https://github.com/openspace/relay.openspaceproject.com

Instead of requesting:
http://www.celestrak.org/NORAD/elements/gp.php?GROUP=x-comm&FORMAT=kvn

we now request:
https://relay.openspaceproject.com/celestrak?GROUP=x-comm&FORMAT=kvn

instead.  The server will only download a new copy if 12h have passed since the last request.

Every other solution was failing for some reason; the final straw was support for dome environments where multiple computers were requesting data from Celestrak from the same public IP, which meant the first computer managed to download the files correctly and the other image generators received the error message

Closes #3973

### DIFF
--- a/data/assets/scene/solarsystem/planets/earth/satellites/communications/amateur.asset
+++ b/data/assets/scene/solarsystem/planets/earth/satellites/communications/amateur.asset
@@ -6,7 +6,7 @@ local omm = asset.resource({
   Name = "Satellite OMM Data (Amateur Radio)",
   Type = "UrlSynchronization",
   Identifier = "satellite_omm_data_amateur_radio",
-  Url = "http://www.celestrak.org/NORAD/elements/gp.php?GROUP=amateur&FORMAT=kvn",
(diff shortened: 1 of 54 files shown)


alexanderbock:
Merging it as the "ok" from TS came in

PR #3996, opened 2026-04-14
Add some Renderable docs 

WeirdRubberDuck:
and update the docs of RenderablePointCloud subclasses to utilize the features like linking and code highlighting. 

Please check that the texts are understandable. 

Edit: I changed the formatting so that the first sentence does not include the word "Renderable", but just directly explains what it does (as suggested in Slack)

### DIFF
--- a/modules/base/rendering/renderablelabel.cpp
+++ b/modules/base/rendering/renderablelabel.cpp
@@ -170,6 +170,13 @@ namespace {
         Property::Visibility::AdvancedUser
     };
 
+    /**
+     * This `Renderable` is used to render a single text label in 3D space. The label
+     * is rendered as a billboard, that is, it is always oriented to face the camera.
+     *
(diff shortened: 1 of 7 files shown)


alexanderbock (approved):

modules/base/rendering/renderablelabel.cpp
```diff
@@ -170,6 +170,13 @@ namespace {
         Property::Visibility::AdvancedUser
     };
 
+    /**
+     * Renders a single text label in 3D space. The label is rendered as a billboard,
+     * that is, it is always oriented to face the camera.
+     *
+     * The label can be rendered with a fade-in effect based on the distance from the
+     * camera.
+     */
```
WeirdRubberDuck: Just realized that the comment style needs updating from block comments to `//`

PR #4005, opened 2026-04-17
Improve touch gestures and interaction

WeirdRubberDuck:
Fixes broken touch gestures, improves gesture detection reliability and stability, and makes touch a bit easier to use and understand (hopefully).

Note that I've focused on making it work, rather than cleaning up the code. There is definitely room for improvement and cleaning of the code, at some point. 

👆 Gestures:
* **Pan** with 2 fingers and swipe with a fixed finger distance (this was broken before!)
* **Zoom** by pinching with 2 fingers. Zooming also looks at finger rotation to allow rolling during the zooming motion
* **Rotate** is still one finger
* **Rolling** either by 1) holding one finger still and rotating the other around it, or 2) touching down 2 fingers and rotating around the center point
* **Triple-click** refocuses on the current focus node (useful now that panning works) 

🦦 Other changes: 
* Single-tap and double-tap are now more reliably detected
* Direct manipulation will only be applied if the touch down points hit the globe/object (that is, it will not interrupt another touch gesture once the touch points hit a target) 
* Zoom, rotate, and "pure" roll will keep moving the camera while the gesture is being held. This is not the case for panning, as it felt like this needed a bit more control

It would be great if someone could try out the interaction and see if they find any issues. The rolling is (to me) the most "finicky" one, but at least it's not worse (maybe better) than how it worked before. 

There are definitely edge cases that I've missed, though, and I expect that we will find more issues in the future once we start using it


include/openspace/interaction/touchinputstate.h
```diff
@@ -75,14 +76,19 @@ class TouchInputState {
     void updateOrAddTouchInput(TouchInput input);
     void removeTouchInput(TouchInput input);
 
+    void clearTapData();
+    void evaluateTap(TouchInput lastRemovedInput);
```
alexanderbock, line 80:
```suggestion
    void evaluateTap(const TouchInput& lastRemovedInput);
```

Looks like we don't store the value in the function

include/openspace/navigation/orbitalnavigator/directmanipulation.h
```diff
@@ -120,6 +126,10 @@ class DirectManipulation : public PropertyOwner {
     std::set<std::string> _sortedDefaultRenderableTypes;
 
     std::vector<SelectedBody> _selectedNodeSurfacePoints;
+    int _lastNSurfacePoints = 0;
+
+    glm::dvec2 _firstMousePressPos = glm::dvec2(0.0);
+    bool _invalidFirstTouch = false;
```
alexanderbock, line 132:
```suggestion
    bool _isInvalidFirstTouch = false;
```

?

src/interaction/interactionhandler.cpp
```diff
@@ -251,22 +251,29 @@ MouseButton InteractionHandler::secondaryMouseButton() const {
...
-        if (action == MouseAction::Press) {
-            _mouseVisualizer.isMouseFirstPress = true;
-            _mouseVisualizer.isMousePressed = true;
-        }
-        else if (action == MouseAction::Release) {
-            _mouseVisualizer.isMousePressed = false;
-            _mouseVisualizer.currentPosition = glm::vec2(0.f);
-            _mouseVisualizer.clickPosition = glm::vec2(0.f);
-        }
+    // The mouse visualizer keeps track of some useful states - let's alway update them
```
alexanderbock, line 268:
```suggestion
    // The mouse visualizer keeps track of some useful states - let's always update them
```

src/interaction/touchinputstate.cpp
```diff
@@ -24,10 +24,21 @@
...
+#include <ghoul/logging/logmanager.h>
+#include <format>
+
+namespace {
+    std::chrono::milliseconds now() {
+        return std::chrono::duration_cast<std::chrono::milliseconds>(
+            std::chrono::high_resolution_clock::now().time_since_epoch()
+        );
+    }
+}
```
alexanderbock, line 36:
```suggestion
} // namespace
```

src/interaction/touchinputstate.cpp
```diff
@@ -135,38 +153,57 @@ void TouchInputState::updateOrAddTouchInput(TouchInput input) {
...
-                else {
-                    _isTap = true;
-                }
-                _time = timestamp;
-            }
+void TouchInputState::clearTapData() {
+    // Reset the timers so we don't keep detecting the combo taps if tapping continuously.
+    // After a triple, we should start on single tap again
+    if (_isDoubleTap || _isTripleTap) {
+        _lastTapTime = std::chrono::milliseconds::zero();
```
alexanderbock, line 167: Is this doing something different than `std::chrono::milliseconds(0)`?   No request for changing; just curious

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -345,12 +345,16 @@ void DirectManipulation::updateCameraFromInput() {
     std::vector<TouchInputHolder> touchInputs =
         global::interactionHandler->touchInputState().touchPoints();
 
+    const SceneGraphNode* anchor = global::navigationHandler->anchorNode();
+    ghoul_assert(anchor != = nullptr, "Must have an anchor");
```
alexanderbock, line 349:
```suggestion
    ghoul_assert(anchor != nullptr, "Must have an anchor");
```

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -364,21 +368,43 @@ void DirectManipulation::updateCameraFromInput() {
...
-        // No fingers, no input
+        // No fingers, no input. Or, the first touch points did not touch the node
         _selectedNodeSurfacePoints.clear();
         _isActive = false;
+        _invalidFirstTouch = false;
         return;
     }
 
+    if (_lastNSurfacePoints == 0 && !_selectedNodeSurfacePoints.empty()) {
+        // This is the first touch intersecting the object! Check if all the initial touch points
```
alexanderbock, line 392:
```suggestion
        // This is the first touch intersecting the object. Check if all the initial touch points
```

No need to shout? :-)

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -364,21 +368,43 @@ void DirectManipulation::updateCameraFromInput() {
...
+        // intersected the surface
+        _invalidFirstTouch = false;
+        for (const TouchPoint& point : touchPoints) {
+            if (!computeSurfacePoint(point.initialPosition, anchor).has_value()) {
```
alexanderbock, line 396:
```suggestion
            std::optional<glm::dvec3> point = computeSurfacePoint(point.initialPosition, anchor);
            if (!point.has_value()) {
```

A bit easier to read I feel

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -440,53 +468,64 @@ void DirectManipulation::updateNodeSurfacePoints(
...
+            continue;
+        }
+
+        // Note that node is saved as the direct input solver was initially
+        // implemented to handle touch contact points on multiple nodes
+        _selectedNodeSurfacePoints.push_back({
+            .id = touchPoint.id,
+            .node = anchor,
+            .coordinates = *surfacePoint
+        });
```
alexanderbock, line 486:
```suggestion
        if (surfacePoint.has_value()) {
            // Note that node is saved as the direct input solver was
            // initially implemented to handle touch contact points on
            // multiple nodes
            _selectedNodeSurfacePoints.push_back({
                .id = touchPoint.id,
                .node = anchor,
                .coordinates = *surfacePoint
            });
        }
```

?

Guesstimated the new line lengths

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -440,53 +468,64 @@ void DirectManipulation::updateNodeSurfacePoints(
...
+                                                         const SceneGraphNode* node) const
+{
+    const Camera* camera = global::navigationHandler->camera();
     const glm::dquat camRotation = camera->rotationQuaternion();
     const glm::dvec3 camPos = camera->position();
 
-    std::vector<SelectedBody> surfacePoints;
+    // Normalized -1 to 1 coordinates on screen
+    const double xCo = 2.0 * (touchPosition.x - 0.5);
+    const double yCo = -2.0 * (touchPosition.y - 0.5);
```
alexanderbock, line 500:
```suggestion
    const glm::vec2 co = 2.0 * (touchPosition - 0.5);
```

?

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -440,53 +468,64 @@ void DirectManipulation::updateNodeSurfacePoints(
...
-        const size_t id = touchPoint.id;
-        // Normalized -1 to 1 coordinates on screen
-        const double xCo = 2.0 * (touchPoint.position.x - 0.5);
-        const double yCo = -2.0 * (touchPoint.position.y - 0.5);
-
-        const glm::dvec3 cursorInWorldSpace = camRotation *
-            glm::dvec3(glm::inverse(camera->projectionMatrix()) *
+    const glm::dvec3 cursorInWorldSpace = camRotation *
+        glm::dvec3(glm::inverse(camera->projectionMatrix()) *
             glm::dvec4(xCo, yCo, -1.0, 1.0));
```
alexanderbock, line 504:
```suggestion
            glm::dvec4(co.x, co.y, -1.0, 1.0));
```

If the previous is okay

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -93,94 +124,79 @@ TouchCameraStates::computeVelocities(const std::vector<TouchInputHolder>& touchP
...
         case InteractionType::Rotation: {
             // Add rotation velocity
-            constexpr glm::dvec2 Scale = glm::dvec2(5.0);
-            updateVelocities.globalRotation = Scale * _sensitivity * -glm::dvec2(
-                currentInput.speedX(),
-                currentInput.speedY()
-            );
+            glm::dvec2 diff = normalizeByAspectRatio(glm::dvec2(
+                endFinger0.pos - startFinger0.pos
+            ));
```
alexanderbock, line 137:
```suggestion
            glm::dvec2 diff = normalizeByAspectRatio(
                glm::dvec2(endFinger0.pos - startFinger0.pos)
            );
```

?

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -93,94 +124,79 @@ TouchCameraStates::computeVelocities(const std::vector<TouchInputHolder>& touchP
...
+
+            const TouchInput& startFinger1 = touchPoints[1].firstInput();
+            const TouchInput& endFinger1 = touchPoints[1].latestInput();
+
+            glm::dvec2 diffStart = normalizeByAspectRatio(glm::dvec2(
+                startFinger0.pos - startFinger1.pos
+            ));
+            glm::dvec2 diffEnd = normalizeByAspectRatio(glm::dvec2(
+                endFinger0.pos - endFinger1.pos
+            ));
```
alexanderbock, line 154:
```suggestion
            glm::dvec2 diffStart = normalizeByAspectRatio(
                glm::dvec2(startFinger0.pos - startFinger1.pos)
            );
            glm::dvec2 diffEnd = normalizeByAspectRatio(
                glm::dvec2(endFinger0.pos - endFinger1.pos)
            );
```

?

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -93,94 +124,79 @@ TouchCameraStates::computeVelocities(const std::vector<TouchInputHolder>& touchP
...
-            // triggered using a three-finger gesture, which is often reserved for other
-            // windows-related interaction, so I could not test it
-            updateVelocities.localRotation = _sensitivity * Scale * glm::dvec2(
+            // Here we don't use the difference between the first and current position,
+            // but instead the speed of the movement, to allow for more responsive panning
+            glm::dvec2 speed = normalizeByAspectRatio(glm::dvec2(
                 currentInput.speedX(),
                 currentInput.speedY()
-            );
+            ));
```
alexanderbock, line 196:
```suggestion
            glm::dvec2 speed = normalizeByAspectRatio(
                glm::dvec2(currentInput.speedX(), currentInput.speedY())
            );
```

?

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -219,35 +233,104 @@ TouchCameraStates::interpretInteraction(const std::vector<TouchInputHolder>& inp
...
+        // Check that both fingers are moving - otherwise, dot products would be nan
+        constexpr float MoveEpsilon = 0.01f;
+        bool isFingerMoving0 = glm::length(dir0) > MoveEpsilon;
+        bool isFingerMoving1 = glm::length(dir1) > MoveEpsilon;
+        bool areBothFingersMoving = isFingerMoving0 && isFingerMoving1;
 
-    // If average distance between 3 fingers are constant we have panning
-    if (avgDistance < InterpretPanDistance && inputs.size() == 3) {
-        return InteractionType::Pan;
+        // Pinch: movement in opposite directions (checked against the centroid).
```
alexanderbock, line 259:
```suggestion
        // Pinch: movement in opposite directions (checked against the centroid)
```

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -219,35 +233,104 @@ TouchCameraStates::interpretInteraction(const std::vector<TouchInputHolder>& inp
...
+
+    // We have roll if one finger is still, or the total roll angles around the centroid
+    // is over a given threshold (all fingers are moving)
+
+    // 1) Find the slowest moving finger - to find potentially still fingers
     double minDiff = 1000.0;
+    for (const TouchInputHolder& inputHolder : inputs) {
+        if (!inputHolder.isMoving()) {
+            minDiff = 0.0;
+            break; // Found a still finger, no need to continue
```
alexanderbock, line 302:
```suggestion
            // Found a still finger, no need to continue
            break;
```

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -219,35 +233,104 @@ TouchCameraStates::interpretInteraction(const std::vector<TouchInputHolder>& inp
...
+    }
+
+    const float InputStillThreshold = 0.0005f;
+    bool hasStillFinger = std::abs(minDiff) < InputStillThreshold;
+    if (hasStillFinger) {
+        return InteractionType::Roll;
+    }
+
+
+    // 2: All fingers are rotating around a still centroid
```
alexanderbock, line 326:
```suggestion
    // 2: All fingers are rotating around a stationary centroid
```

?

alexanderbock (approved):
Don't have a touch screen anymore, so can't test it on my own

src/interaction/touchinputstate.cpp
WeirdRubberDuck: Very good question! I had to look it up 😅   Seems like we could have used either one here, and that the `zero()` represents a zero-length duration  that e.g. allows comparing with hours etc as well

https://en.cppreference.com/cpp/chrono/duration/zero
https://stackoverflow.com/questions/68675472/are-all-of-zero-in-chrono-the-same-thing

Do you have a preference for which to use? (I don't, I think)

src/navigation/orbitalnavigator/directmanipulation.cpp
WeirdRubberDuck: Looks like the direction of the y coordinate is flipped here too. Would something like this still be preferble to you ?
```
    glm::dvec2 co = 2.0 * glm::dvec2(touchPosition - 0.5);
    co.y *= -1.0;
```

src/navigation/orbitalnavigator/directmanipulation.cpp
```diff
@@ -364,21 +368,46 @@ void DirectManipulation::updateCameraFromInput() {
...
+        _isInvalidFirstTouch = false;
+        for (const TouchPoint& point : touchPoints) {
+            std::optional<glm::dvec3> surfacePoint =
+                computeSurfacePoint(point.initialPosition, anchor);
+
+            if (!surfacePoint.has_value()) {
+                _isInvalidFirstTouch = true;
+                break;
+            }
+        }
```
WeirdRubberDuck: OBS! This misses the edge case when zooming in.... Need to update the computation, again...

src/interaction/touchinputstate.cpp
alexanderbock: If they have a special function for it then we should use it as well, i'd say

WeirdRubberDuck:
It would be great if someone could try out the interaction as well - let me know if you need to borrow the touch screen!

src/navigation/orbitalnavigator/directmanipulation.cpp
alexanderbock: Oh yea, I didn't see that!    In that case maybe leave it as it is instead

src/interaction/touchinputstate.cpp
```diff
@@ -135,38 +150,57 @@ void TouchInputState::updateOrAddTouchInput(TouchInput input) {
...
+}
+
+void TouchInputState::evaluateTap(const TouchInput& lastRemovedInput) {
+    TouchInputHolder& inputHolder = _touchPoints.front();
+
+    if (!inputHolder.holdsInput(lastRemovedInput)) {
+        return;
+    }
+
+    const bool isWithinTapTime = inputHolder.gestureTime() < 0.18;
```
Roxeena, line 180: Should we make this 0.18 configurable? As I understand it, is this the time that is ok between taps to be considered a double/triple tap? For accessibility reasons, it could be nice to make that configurable somewhere.

src/navigation/orbitalnavigator/orbitalnavigator.cpp
```diff
@@ -564,6 +565,12 @@ void OrbitalNavigator::resetVelocities() {
 }
 
 void OrbitalNavigator::updateCamera(double deltaTime) {
+    // Special touch gesture for reargetting focus node
```
Roxeena:
```suggestion
    // Special touch gesture for retargeting focus node
```
Not sure if this is what you meant instead or reargetting

src/navigation/orbitalnavigator/touchcamerastates.cpp
```diff
@@ -93,94 +124,78 @@ TouchCameraStates::computeVelocities(const std::vector<TouchInputHolder>& touchP
...
-                endFinger1.pos.x * aspectRatio,
-                endFinger1.pos.y
+            glm::dvec2 diffEnd = normalizeByAspectRatio(
+                glm::dvec2(endFinger0.pos - endFinger1.pos)
             );
-            double distToCentroidEnd = glm::length(endVec0 - endVec1) / 2.0;
+            double zoomFactor = glm::length(diffEnd) - glm::length(diffStart);
 
-            double zoomFactor = distToCentroidEnd - distToCentroidStart;
+            constexpr double ZoomScale = 5.0;
```
Roxeena, line 157: To make it consistent with the scale for rotation and roll, this should be named just scale, right? 

By the way, do we want to collect the scales and other threshold parameters in one place, like the top of the file, so it is easier to potentially adjust them later? Or even make them adjustable by the user as a setting? But this does not need to stall the merging of this PR of course.

Roxeena (commented):
Unfortuneatly, I do not have access to any touch devices to try it out on, but I took a look at the code.

src/navigation/orbitalnavigator/touchcamerastates.cpp
WeirdRubberDuck: This specific interaction is split up into two parts to allow both zooming and rolling at the same time. Hence, the individual names for the scales here  :) 

Also, I would prefer not to move the constants. Since there are so many different scale constants, I think it's easier to adjust the setting in the place of the code where it has its effect. Otherwise, the coder would first have to go here to see which constant affects which motion and how, and then update the value. 

Regarding adding settings for the individual sensitivities, this is definitely something we can add in the future if we see that there is a need for it. For now, I suggest we leave just the one Sensitivity setting -- having too many settings might make the feature more complex to use.

src/interaction/touchinputstate.cpp
WeirdRubberDuck: Not quite sure if this specific one should be configurable (the time for double-tap is, though). If this value is configurable, it makes it harder to set the double-tap timing since it becomes dependent on this value

I had commented that myself, though, in the last walkthrough of the code 😅 (see line 144 in the removed code). There was a previous comment about this 0.18 being a magic number from tuioear.cpp, but I failed to locate this number anywhere in the tuio code. Based on this, I'm not sure what to do about this constant, TBH

Arohdin: From an accessibility point of view it could make sense to make it configurable as the tap-time might be slower for some individuals (I assume that `0.18` is the max amount of time the finger touches the screen to register a tap).
That being said, a comment is probably sufficient for this PR.

src/interaction/touchinputstate.cpp
```diff
@@ -24,10 +24,18 @@
...
+            std::chrono::high_resolution_clock::now().time_since_epoch()
+        );
+    }
+} // namespace
+
 namespace openspace {
 
 void TouchInputState::initialize() {
-    _time = std::chrono::duration_cast<std::chrono::milliseconds>(
+    _lastTapTime = std::chrono::duration_cast<std::chrono::milliseconds>(
```
Arohdin: Could use `now()` instead

PR #4008, opened 2026-04-20
added exports for commonly exported globes to allow for a version of …

micahnyc:
…deprecation;

Since many files created by users are including the Earth, Moon, Mars or the Sun, I thought it would be nice to export those for them so they could have a version to update their files. 

This doesn't utilize our typical method of deprecation where we have a warning message print for the users, however I think this would still be beneficial to them, instead we can mention it in the slack and in the monthly meetings over the next year.

### DIFF
--- a/data/assets/scene/solarsystem/planets/earth/earth.asset
+++ b/data/assets/scene/solarsystem/planets/earth/earth.asset
@@ -1,4 +1,4 @@
-asset.require("./globe")
+local Earth = asset.require("./globe").Earth
 asset.require("./actions")
 asset.require("./trail")
 asset.require("./atmosphere")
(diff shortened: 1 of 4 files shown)


alexanderbock:
I'd be against adding it as we (i) don't really do this anywhere else, so it would be a surprising behavior and (ii) since there is no deprecation notice, only a small minority would change their assets anyway (`localResource` anyone?), so we would basically trade-off a breaking change now for a breaking change in 0.23.
Additionally, having edited the existing assets its a really simple CTRL+F replacement from "planets/earth/moon/moon" to "planets/earth/moon/globe" etc.   I could see us providing a Lua script the user could execute in OpenSpace that goes through the users folder and suggests those changes?  Something like `openspace.updateAssetsVersion`?

If we _really_ want it, then we should make this pattern for all globes instead and not just a few

PR #4013, opened 2026-04-22
Add Bouncing Interpolation mode to Properties

alexanderbock:
Add as the ability bounce (or pingpong) between property values.  Adds a new 6th optional boolean parameter to the `setPropertyValue` and `setPropertyValueSingle` functions to control whether a parameter change should bounce between the current value and the new value.

Example:
```lua
openspace.setPropertyValueSingle("Scene.Earth.Renderable.Opacity", 1.0);
openspace.setPropertyValueSingle("Scene.Earth.Renderable.Opacity", 0.0, 3, "Linear", "", true);
```

Will cause the opacity of Earth to fade from 1.0 to 0.0 over 3 seconds, then back to 1.0 over 3 seconds, back to 0.0 over seconds etc.
Any property that is interpolatable (anything that derives from `NumericalProperty` can be used for this.

Also adds a new function `openspace.stopPropertyBouncing` that takes the URI of a property and stops the bouncing interpolation. As soon as the call is made, it will interpolate back to the original value.

Note: The implementation is really ugly, but it works and I couldn't think of a way to improve it at the moment

### DIFF
--- a/include/openspace/scene/scene.h
+++ b/include/openspace/scene/scene.h
@@ -177,9 +177,8 @@ public:
      *        finishes
      * \param easingFunction A function that determines who the interpolation occurs
      * \param isBouncing If this value is set to `true`, the property will interpolate to
-     *        the current value, then back to the original value, until manually stopped.
-     *        In this mode, the \p postScript will be called each time the interpolation
-     *        is at the starting value or final value
+     *        the provided new value, then back to the original value, until manually
+     *        stopped.
      *
      * \pre \p prop must not be `nullptr`
      * \pre \p durationSeconds must be positive and not 0
--- a/include/openspace/scene/scene.h
+++ b/include/openspace/scene/scene.h
@@ -176,8 +176,8 @@ public:
      * \param postScript A Lua script that will be executed when the interpolation
      *        finishes
      * \param easingFunction A function that determines who the interpolation occurs
-     * \param isBouncing If this value is set to `true`, the property will interpolate to
-     *        the provided new value, then back to the original value, until manually
+     * \param shouldBounce If this value is set to `true`, the property will interpolate
+     *        to the provided new value, then back to the original value, until manually
      *        stopped.
      *
      * \pre \p prop must not be `nullptr`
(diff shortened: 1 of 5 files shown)


micahnyc:
A nice addition would be to be able to set a 2nd easing function for the opposite direction

micahnyc:
works great!!!! https://www.dropbox.com/scl/fi/dmkngz8oskom0gqduo16c/bouncetesting.mp4?rlkey=6ia6nqotuvv0vk47xm0as6tts&dl=0

alexanderbock:
I'd want to put the second interpolation function for a later update.   Having a plan to improve the non-bouncing interpolation as well, by being able to specify more parameters to configure the curve (different time for ease-in from ease-out, for example)

include/openspace/scene/scene.h
```diff
@@ -176,14 +176,18 @@ class Scene : public PropertyOwner {
      * \param postScript A Lua script that will be executed when the interpolation
      *        finishes
      * \param easingFunction A function that determines who the interpolation occurs
+     * \param isBouncing If this value is set to `true`, the property will interpolate to
```
Roxeena: I would think that `isBouncing` is a flag whether it is currently bouncing, and not a way to trigger it to start bouncing. Maybe `shouldBounce` is more descriptive?

include/openspace/scene/scene.h
```diff
@@ -207,6 +211,18 @@ class Scene : public PropertyOwner {
...
+     * Calling this function will cause the interpolation of the provided property to stop
+     * the next time it is at a defined state. For normal interpolation, calling this
+     * function is a noop, as the interpolation will be automatically stopped at the end
+     * of the interpolation. For bouncing interpolations, it will cause the bouncing
+     * interpolation of the property to be stopped when it reaches the original value the
+     * next time.
+     *
+     * \param prop The property whose interpolation should be stopped
+     */
+    void stopInterpolation(Property* prop);
```
Roxeena: Since the function is used to stop the bouncing behavior and not the interpolation itself, the name is confusing. Maybe `stopBouncing` matches its purpose better? And if it is called for a non-bouncing property, it will still noop as before.

Roxeena (commented):
Tested it out and it looks really cool 👍

PR #4015, opened 2026-04-24
Add support for CSV format from Celestrak and change requesting format

alexanderbock:


data/assets/scene/solarsystem/planets/earth/satellites/weather/terra.asset
```diff
@@ -1,121 +0,0 @@
-local earthAsset = asset.require("scene/solarsystem/planets/earth/transforms")
```
alexanderbock, line 1: Removed this file as it was forgotten in the Asset Split commit

PR #4017, opened 2026-04-24
Update satellite datasets based on Celestrak

### LINKED ISSUE #4016: Update Celestrak catalogue
Some of the Celestrak satellite groups that we use no longer exist and should be removed. Celestrak has added new groups.

Removed:
  - noaa
  - other
  - molniya
  - raduga
  - gorizont
  - iridium

alexanderbock:
Removes satellite groups that are no longer provided by Celestrak:
noaa, other, molniya, raduga, gorizont, iridium

Adds new satellite groups:
Eutelsat, Hulianwang Digui, OneWeb, Qianfan, SatNOGS, Telesat, Geo Protected Zone, Geo Protected Zone Plus, GNSS

Closes #4016


data/assets/scene/solarsystem/planets/earth/satellites/communications/eutelsat.asset
```diff
@@ -3,21 +3,21 @@ local earthAsset = asset.require("scene/solarsystem/planets/earth/transforms")
 
 
 local omm = asset.resource({
-  Name = "Satellite OMM Data (Gorizont)",
+  Name = "Satellite OMM Data (Eutelsat)",
```
Roxeena:
```suggestion
  Name = "Satellite CSV Data (Eutelsat)",
```

Should we indicate that this satellite group uses CSV data format instead of OMM? The same goes for all other CSV-based satellites.

Roxeena (commented):
👍

data/assets/scene/solarsystem/planets/earth/satellites/communications/eutelsat.asset
alexanderbock: Not sure that it would matter to the user?

Roxeena: No, but it would matter to us, right? We did rename them when we switched from TLE to OMM. Another alternative would be to remove the format part from the names completely. I just think that it could be confusing for us developers if it says it is OMM data and then we set the format to CSV (as it was for me when I read this PR). But this doesn't need to be changed with this PR, we can make a new issue for it and fix it later. What do you think?

alexanderbock: Nevermind.. i misread the suggestion.  :+1:

Issue #3967, opened 2026-03-21
Session recording "save frames" with a video texture is not frame accurate


hn-88:
**Is your feature request related to a problem? Please describe.**
A fly-by of a sphere with a video texture, when recorded and played back with "save frames" enabled, results in exported frames which jump frames on the video texture.

**Describe the solution you'd like**
If possible, frame-accurate exports - for eg, if the frame-rate of the export equals the frame rate of the video texture, each frame of the video texture should be exported in one frame of the saved frames.

**Describe alternatives you've considered**
Exporting in Blender would be an alternative.

**Additional context**
Example screenshots from NOAA SOS site's world population movie, https://sos.noaa.gov/catalog/datasets/world-population-movie/

on a sphere texture, adjacent screenshots show exported frame OpenSpace_000041.png shows 1675CE, then 000042 shows 1677CE, then upto frame 000068 1677CE is shown before jumping to 1686CE in frame 000069 shown below,

<img width="1280" height="720" alt="OpenSpace_000068.png" src="https://github.com/user-attachments/assets/1f52986d-dc91-4c8b-a320-1c1fc8fcca88" />

<img width="1280" height="720" alt="OpenSpace_000069.png" src="https://github.com/user-attachments/assets/3c5ec9b4-e5d4-4f5d-ba5a-b656bbf6fa91" />

alexanderbock:
This is an issue with the video player as we use it in an async manner to keep the framerate up, which causes slight variations when a video frame is being shown.

It should definitely be fixed, but if you need a solution now, I'd recommend using `ffmpeg` to convert the video into a pile of PNGs and then use a `TemporalTileProvider` in the `Folder` mode:  https://github.com/OpenSpace/OpenSpace/blob/master/data/assets/scene/solarsystem/planets/earth/noaa-sos/atmosphere/ltg_vaisala.asset#L31-L45

hn-88:
Thanks for the workaround, @alexanderbock - I guess an example for this way of usage is `data/assets/scene/solarsystem/planets/earth/noaa-sos/land/fire.asset`. I'll try using that method. I guess the fps would be set by choosing the filenames and simulation time seconds/second appropriately.

alexanderbock:
Exactly.. you'd rename the files to be a specific timestamp at which you want them to show up and OpenSpace using the in-game time to pick which image to show

PR #4032, opened 2026-05-05
Videoplayer stutter in clustered environments 

### LINKED ISSUE #2900: Video stutter on cluster (0.19.0)
Video on Globe seems to sometimes have stutter issue when rendering in a cluster environment (Version 0.19.0).
Similar issue was raised in Slack by Dylan Salas on their cluster - unclear if they used video plane or globe.

Arohdin:
Changed logic for video player that results in:
- Smoother playback in clustered environment
- Better sync for videos of type `RealTimeLoop` during playback and looping

Closes #2900 

**Note:** Known issue with `VideoPlayer` and seesion recordings have not been addressed (#3967)


modules/video/include/videoplayer.h
```diff
@@ -83,6 +84,20 @@ class VideoPlayer : public PropertyOwner, public Syncable {
     static openspace::Documentation Documentation();
 
 private:
+    // Flags used to syncronize playback
```
alexanderbock:
```suggestion
    // Flags used to synchronize playback
```

modules/video/include/videoplayer.h
```diff
@@ -83,6 +84,20 @@ class VideoPlayer : public PropertyOwner, public Syncable {
...
+    struct syncFlags {
```
alexanderbock:
```suggestion
    struct SyncFlags {
```

modules/video/include/videoplayer.h
```diff
@@ -83,6 +84,20 @@ class VideoPlayer : public PropertyOwner, public Syncable {
...
+        bool goToStart = false;
+        bool play = false;
```
alexanderbock:
```suggestion
        bool shouldGoToStart = false;
        bool shouldPlay = false;
```

?

modules/video/include/videoplayer.h
```diff
@@ -146,6 +157,12 @@ class VideoPlayer : public PropertyOwner, public Syncable {
     BoolProperty _playAudio;
     BoolProperty _loopVideo;
 
+    // Variables used when syncronizing play, pause and looping behavior
+    const bool _isMaster;
```
alexanderbock:
```suggestion
```

Can use `global::windowDelegate->isMaster()` whereever necessary

modules/video/include/videoplayer.h
```diff
@@ -146,6 +157,12 @@ class VideoPlayer : public PropertyOwner, public Syncable {
...
+    syncFlags _syncflags;
```
alexanderbock:
```suggestion
    syncFlags _syncFlags;
```

modules/video/include/videoplayer.h
```diff
@@ -146,6 +157,12 @@ class VideoPlayer : public PropertyOwner, public Syncable {
...
+    PlaybackState _playbackState = PlaybackState::Undefined;
+    std::chrono::duration<long long, std::milli> _goTime;
```
alexanderbock:
```suggestion
    std::chrono::duration<long long, std::milli> _goTime = std::chrono::duration<long long, std::milli>::zero();
```

alexanderbock: Not sure about the variable name, but don't have a better suggestion

modules/video/src/videoplayer.cpp
```diff
@@ -246,6 +238,8 @@ VideoPlayer::VideoPlayer(const ghoul::Dictionary& dictionary)
     , _reload(ReloadInfo)
     , _playAudio(AudioInfo, false)
     , _loopVideo(LoopVideoInfo, true)
+    , _isMaster(global::windowDelegate->isMaster())
+    , _goTime(0)
```
alexanderbock:
```suggestion
```

With the changes in the header file

modules/video/src/videoplayer.cpp
```diff
@@ -310,7 +301,7 @@ VideoPlayer::VideoPlayer(const ghoul::Dictionary& dictionary)
         { MpvKey::IsSeeking, "seeking" },
         { MpvKey::Mute, "mute" },
         { MpvKey::Seek, "seek" },
-        { MpvKey::Loop, "loop-file" }
+        { MpvKey::EndOfFile, "eof-reached"}
```
alexanderbock:
```suggestion
        { MpvKey::EndOfFile, "eof-reached" }
```

modules/video/src/videoplayer.cpp
```diff
@@ -524,13 +509,10 @@ void VideoPlayer::update() {
...
-    // Renders a frame libmpv has been updated
-    if (_wakeup) {
-        const uint64_t result = mpv_render_context_update(_mpvRenderContext);
-        if ((result & MPV_RENDER_UPDATE_FRAME)) {
-            renderFrame();
-            _wakeup = 0;
-        }
+    // Renders a frame
+    const uint64_t result = mpv_render_context_update(_mpvRenderContext);
+    if ((result & MPV_RENDER_UPDATE_FRAME)) {
```
alexanderbock, line 582:
```suggestion
    if (result & MPV_RENDER_UPDATE_FRAME != 0) {
```

?

modules/video/src/videoplayer.cpp
```diff
@@ -770,6 +752,14 @@ void VideoPlayer::handleMpvProperties(mpv_event* event) {
...
             break;
         }
+        case MpvKey::EndOfFile: {
+            if (_isMaster && _loopVideo && _playbackState == PlaybackState::Playing) {
+                if(_isMaster) {
+                    _syncflags.goToStart = true;
+                }
+            }
+            break;
+        }
```
alexanderbock:
```suggestion
        case MpvKey::EndOfFile:
            if (_isMaster && _loopVideo && _playbackState == PlaybackState::Playing) {
                _syncflags.goToStart = true;
            }
            break;
```

Duplicate check and no need for the extra scope

modules/video/src/videoplayer.cpp
```diff
@@ -787,27 +777,68 @@ void VideoPlayer::destroy() {
...
+            const long long dt = _goTime.count() - now.count();
+            if (dt <= 0.0) {
+                _syncflags.play = true;
+            }
+        }
+        else if (_playbackMode == PlaybackMode::MapToSimulationTime) {
+            // MapToSimulationTime
+            _correctPlaybackTime = correctVideoPlaybackTime();
+        }
+    }
```
alexanderbock:
```suggestion
    if (!isMaster) {
        return;
    }

    _correctPlaybackTime = _currentVideoTime;
    if (_playbackMode == PlaybackMode::RealTimeLoop &&
        _playbackState == PlaybackState::Waiting)
    {
        const auto now = std::chrono::duration_cast<std::chrono::milliseconds>(
            std::chrono::high_resolution_clock::now().time_since_epoch()
        );
        const long long dt = _goTime.count() - now.count();
        if (dt <= 0.0) {
            _syncflags.play = true;
        }
    }
    else if (_playbackMode == PlaybackMode::MapToSimulationTime) {
        _correctPlaybackTime = correctVideoPlaybackTime();
    }
```

modules/video/src/videoplayer.cpp
```diff
@@ -787,27 +777,68 @@ void VideoPlayer::destroy() {
...
-    if (_correctPlaybackTime < 0.0) {
-        return;
-    }
-    // Ensure the nodes have the same time as the master node
-    const bool isMappingTime = _playbackMode == PlaybackMode::MapToSimulationTime;
-    if (!isMaster) {
-        if ((_correctPlaybackTime - _currentVideoTime) > glm::epsilon<double>()) {
-            seekToTime(_correctPlaybackTime, PauseAfterSeek(isMappingTime));
+    if (_playbackMode == PlaybackMode::RealTimeLoop) {
+        // Ensure syncronized playback and states between Nodes and Master
```
alexanderbock:
```suggestion
        // Ensure synchronized playback and states between Nodes and Master
```

alexanderbock (approved):

modules/video/include/videoplayer.h
Arohdin: will keep `_goTime` for now

modules/video/src/videoplayer.cpp
Arohdin: I don't think this is needed (right now).

Documentation for `mpv_render_context_update()`
```
* @return a bitset of mpv_render_update_flag values (i.e. multiple flags are
*         combined with bitwise or). Typically, this will tell the API user
*         what should happen next. E.g. if the MPV_RENDER_UPDATE_FRAME flag is
*         set, mpv_render_context_render() should be called. If flags unknown
*         to the API user are set, or if the return value is 0, nothing needs
*         to be done.
```

Documentation for `mpv_render_update_flag`:
```
/**
 * Flags returned by mpv_render_context_update(). Each value represents a bit
 * in the function's return value.
 */
typedef enum mpv_render_update_flag {
    /**
     * A new video frame must be rendered. mpv_render_context_render() must be
     * called.
     */
    MPV_RENDER_UPDATE_FRAME         = 1 << 0,
} mpv_render_context_flag;
```

If I understand correctly `(result & MPV_RENDER_UPDATE_FRAME)` will **only** return `0`or `1` as integer and this will always be `false` or `true`.
However, we could add explicit check if we want to make it safer for any future API changes.

Which do you prefer @alexanderbock ?

alexanderbock: I think having the explicit check would be a good idea to already prepare for an update of the library when they change that behavior

Arohdin:
Note: Requires more testing in the dome, do not merge yet.

PR #4056, opened 2026-05-08
Issue/3702 stereo images

### LINKED ISSUE #3702: Add the ability to render stereoscopic images in the RenderableImagePlane
This would be useful to be able to render stereoscopic images by separating them into a left/right pair and then rendering the correct eye inside OpenSpace.
This would, of course, only work if OpenSpace is also rendering in stereo, otherwise one of the images would be selected.

Roxeena:
Fixes #3702.

This adds the functionality to add a second `RightTexture` (or `RightURL`) to `RenderablePlaneImageLocal` and `RenderablePlaneImageOnline`. The `Texture` (left) and the `RightTexture`/`RightURL` will be routed to the respective eye during stereo rendering.

Notes:
* I added two new examples for the stereo option.
* I also fixed a silent issue we had where the `LazyLoading` for the local version was broken. The images were not unloaded when the plane was disabled. 
* During stereo rendering, only the first (left) image affects the aspect ratio for the `AutoScale` property. 
* The image testing server will probably need to be updated with the new examples.
* If anyone has a proper stereo image pair to test with and to provide as an example, that would be great to add. 

The new stereo example (using the single_sbs_stereo.config):
OpenSpace_000001


modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,447 @@
...
+                break;
+            case WindowDelegate::Frustum::RightEye:
+                if (_useStereo && _rightTexture) {
+                    unit.bind(*_rightTexture);
+                }
+                else if (!_useStereo){
+                    unit.bind(*_texture);
+                }
+                break;
+        }
```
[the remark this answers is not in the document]
Roxeena: Is this really a problem?

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+        "Right Eye Texture",
+        "Sets the path of the texture that is displayed on this plane for the right eye "
+        "during stereo rendering. The path can either be specified as a filepath to a "
+        "local image on disc, or a web URL to an online image.",
+        Property::Visibility::User
+    };
+
+    // Creates a textured 3D plane, where the texture image is loaded from either a
+    // filepath to a local image on disc, or a web URL to an online image. If an online
+    // image is used, the `IsUrl` parameter must be set to `true`."
```
WeirdRubberDuck: Add some information about that this supports stereo rendering and that in that case, the RightTexture should be set

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+                    LDEBUGC(
+                        "RenderablePlaneImage",
+                        std::format("Loaded right texture from '{}'", path)
+                    );
+                    return texture;
+                }
+            );
+        }
+    }
```
WeirdRubberDuck: This huge if-else statement is really difficult to read/overview. How about breaking out the texture loading for each case into individual functions or something?

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+        [](const DownloadManager::MemoryFile&) {
+            LDEBUGC(
+                "RenderablePlaneImage",
+                "Download to memory finished"
+            );
+        },
+        [](const std::string& err) {
+            LDEBUGC(
+                "RenderablePlaneImage",
+                "Download to memory failed: " + err
```
WeirdRubberDuck:
```suggestion
                std::format("Download to memory failed: {}", err)
```

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+        case WindowDelegate::Frustum::LeftEye:
+            if (_texture) {
+                unit.bind(*_texture);
+            }
+            break;
+        case WindowDelegate::Frustum::RightEye:
+            if (_useStereo && _rightTexture) {
+                unit.bind(*_rightTexture);
+            }
+            else if (!_useStereo && _texture){
```
WeirdRubberDuck:
```suggestion
            else if (!_useStereo && _texture) {
```

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
```diff
@@ -0,0 +1,26 @@
...
+-- This example shows how to create a textured plane in 3D space, where the texture is
+-- loaded from two local files on disk to create a stereo image.
+
+local Node = {
+  Identifier = "RenderablePlaneImage_Example_Stereo",
+  Renderable = {
+    Type = "RenderablePlaneImage",
+    Size = 3.0E11,
+    Texture = openspace.absPath("${DATA}/test2.jpg"),
+    RightTexture = openspace.absPath("${DATA}/test.jpg")
```
WeirdRubberDuck: For the future, maybe we should consider having a "real" stereo texture for the example instead of two completely different images :) Make an issue?

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
```diff
@@ -0,0 +1,26 @@
+-- Stereo
+-- This example shows how to create a textured plane in 3D space, where the texture is
+-- loaded from two local files on disk to create a stereo image.
+
```
WeirdRubberDuck, line 6:
```suggestion
-- This example shows how to create a textured plane in 3D space, where the texture is
-- loaded from two local files on disk to create a stereo image. The two different images
-- will be used for the left and right eye, respectively, when OpenSpace is started with 
-- stereo rendering.

```
? Or otherwsie clarify what we mean by stereo image

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+        "base_renderable_planeimage",
+        RenderablePlane::Documentation()
+    );
+}
+
+RenderablePlaneImage::RenderablePlaneImage(const ghoul::Dictionary& dictionary)
+    : RenderablePlane(dictionary,
+        {
+            .automaticallyUpdateRenderBin = false,
+            .shouldUpdateIfDisabled = true
```
WeirdRubberDuck: The other two renderables still use the default settings, right?  Is there a reason for the `shouldUpdateIfDisabled` setting being true here?

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+        })
```
WeirdRubberDuck:
```suggestion
        }
    )
```

modules/base/rendering/renderableplaneimage.cpp
```diff
@@ -0,0 +1,441 @@
...
+                            std::unique_ptr<ghoul::opengl::Texture>
+                        {
+                            std::unique_ptr<ghoul::opengl::Texture> texture =
+                                ghoul::io::texture::loadTexture(
+                                    reinterpret_cast<void*>(file.buffer),
+                                    file.size,
+                                    2,
+                                    ghoul::opengl::Texture::SamplerInit{ .filter =
+                                        ghoul::opengl::Texture::FilterMode::LinearMipMap
+                                    },
```
WeirdRubberDuck: This `.filter` should be on the next line according to our code style. But, with this amount of indentation, it does not fit. 

This is just one of the resons why it would be nice to break out the loading of the different textures into separate functions - it would get rid of a lot of this indentation and make the code easier to read

modules/base/rendering/renderableplaneimage.cpp
WeirdRubberDuck: Or, actually, maybe do one function for local VS online loading, to reduce the duplicated code? The same settings should be used for all the texture loading anyways

modules/base/rendering/renderableplaneimage.h
```diff
@@ -0,0 +1,78 @@
...
+
+    std::future<DownloadManager::MemoryFile> _imageFuture;
+    std::future<DownloadManager::MemoryFile> _rightImageFuture;
+
+    ghoul::opengl::Texture* _texture = nullptr;
+    ghoul::opengl::Texture* _rightTexture = nullptr;
+    glm::vec2 _textureDimensions = glm::vec2(0.f);
+
+    bool _isUrl = false;
+    bool _useStereo = false;
```
WeirdRubberDuck:
```suggestion
    bool _isStereo = false;
```
? More aligned with the "isUrl" setting, and "use" makes it sound like it's a choice controllable by the user, to me. In this case, it isn't. 

Should it be? That is, should there be a property that allows the user to only use one of the textures, if possible? Might be useful for debugging

Feel free to ignore

modules/base/rendering/renderableplaneimageonline.cpp
```diff
@@ -102,6 +102,16 @@ RenderablePlaneImageOnline::RenderablePlaneImageOnline(
     });
 }
 
+void RenderablePlaneImageOnline::initializeGL() {
+    RenderablePlane::initializeGL();
+
+    LWARNINGC(
+        "RenderablePlaneImageOnline",
+        "RenderablePlaneImageOnline is depricated, use RenderablePlaneImage with the "
```
WeirdRubberDuck:
```suggestion
        "RenderablePlaneImageOnline is deprecated, use RenderablePlaneImage with the "
```

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -128,6 +128,11 @@ RenderablePlaneImageLocal::RenderablePlaneImageLocal(const ghoul::Dictionary& di
 void RenderablePlaneImageLocal::initializeGL() {
     RenderablePlane::initializeGL();
 
+    LWARNINGC(
+        "RenderablePlaneImageLocal",
+        "RenderablePlaneImageLocal is depricated, use RenderablePlaneImage instead"
```
WeirdRubberDuck:
```suggestion
        "RenderablePlaneImageLocal is deprecated, use RenderablePlaneImage instead"
```

WeirdRubberDuck (changes requested):
Requesting changes as I think the huge if-else statement really needs fixing, and had some comments on documentation being needed. 

I also think that if we're doing this change of removing the "Online" and "Local" versions here, we should do this to all other similar objects:
* RenderableSphereImage*
* ScreenSpaceImage*
* ScreenSpaceTimeVaryingImage*

But I guess that might be too tight for this release

Have not tested it more than loading the provided example. I would very much like to test it with an actual stereo image example and see how it looks, but I cannot do that in time for the planned code freeze.

modules/base/rendering/renderableplaneimage.h
Roxeena: I'm not sure what you mean by `That is, should there be a property that allows the user to only use one of the textures, if possible?` Currently, if only one texture is set, then both are bound to the left and right eye, which would be the same behaviour?

modules/base/rendering/renderableplaneimage.cpp
Roxeena: Yes, the other types still use the previous settings, and this solves the `LazyLoading` bug I mentioned in the description. Essentially, the unloading of the texture when the user disabled the object is done in the update function, but now that the update function does not run for disabled objects, it was never unloaded. Until it was enabled again, and then it was immediately loaded again. Now it properly unloads the texture when the object is disabled.

modules/base/rendering/renderableplaneimage.h
WeirdRubberDuck: I meant that if there are two textures set, it might be useful to disable the stereo during runtime, without having to "remove" the right texture. Hence the suggestion of a "UseStereo" property :) 

But just a suggestion, could do some testing first ot see whether it would be useful

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
alexanderbock: We could also make a red-green pair using the create single color image function.

modules/base/rendering/renderableplane.cpp
```diff
@@ -249,8 +246,9 @@ RenderablePlane::DistanceScalingSettings::DistanceScalingSettings(
     addProperty(scaleByDistanceMinHeight);
 }
 
-RenderablePlane::RenderablePlane(const ghoul::Dictionary& dictionary)
-    : Renderable(dictionary, { .automaticallyUpdateRenderBin = false })
+RenderablePlane::RenderablePlane(const ghoul::Dictionary& dictionary,
+                                 RenderableSettings settings)
+    : Renderable(dictionary, settings)
```
alexanderbock:
```suggestion
    : Renderable(dictionary, std::move(settings))
```

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -72,19 +88,36 @@ Documentation RenderablePlaneImageLocal::Documentation() {
 }
 
 RenderablePlaneImageLocal::RenderablePlaneImageLocal(const ghoul::Dictionary& dictionary)
-    : RenderablePlane(dictionary)
+    : RenderablePlane(dictionary,
```
alexanderbock:
```suggestion
    : RenderablePlane(
        dictionary,
```

modules/base/rendering/renderableplane.h
```diff
@@ -43,7 +43,12 @@ namespace openspace {
 
 class RenderablePlane : public Renderable {
 public:
-    explicit RenderablePlane(const ghoul::Dictionary& dictionary);
+    explicit RenderablePlane(const ghoul::Dictionary& dictionary,
+        RenderableSettings settings = {
+            .automaticallyUpdateRenderBin = false,
+            .shouldUpdateIfDisabled = false
+        }
+    );
```
alexanderbock:
```suggestion
        RenderableSettings settings = RenderableSettings();
```

Or why do we not want to update the render bin automatically?

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -72,19 +88,36 @@ Documentation RenderablePlaneImageLocal::Documentation() {
...
+        {
+            .automaticallyUpdateRenderBin = false,
```
alexanderbock:
```suggestion
```

Same question here

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -45,14 +47,28 @@ namespace {
...
     // Creates a textured 3D plane, where the texture is provided by a local file on disk.
+    // To use a stereo image set, a second image can be specified with the `RightTexture`
+    // property that will be used for the right eye during stereo rendering.
     struct [[codegen::Dictionary(RenderablePlaneImageLocal)]] Parameters {
         // [[codegen::verbatim(TextureInfo.description)]]
-        std::filesystem::path texture;
+        std::string texture;
+
+        // [[codegen::verbatim(RightTextureInfo.description)]]
+        std::optional<std::string> rightTexture;
```
alexanderbock:
```suggestion
        std::optional<std::filesystem::path> rightTexture;
```

The `texture` is already changes in the master branch

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -72,19 +88,36 @@ Documentation RenderablePlaneImageLocal::Documentation() {
...
+    _texturePath = absPath(p.texture).string();
     _textureFile = std::make_unique<ghoul::filesystem::File>(_texturePath.value());
     _textureFile->setCallback([this]() { _textureIsDirty = true; });
 
     _texturePath.onChange([this]() { loadTexture(); });
     addProperty(_texturePath);
 
+    if (p.rightTexture.has_value()) {
+        _isStereo = true;
+        _rightTexturePath = absPath(*p.rightTexture).string();
```
alexanderbock:
```suggestion
        _rightTexturePath = p.rightTexture->string();
```

With the change to a std::filesystem::path

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -97,52 +130,48 @@ RenderablePlaneImageLocal::RenderablePlaneImageLocal(const ghoul::Dictionary& di
...
     RenderablePlane::initializeGL();
 
     if (!_isLoadingLazily) {
         loadTexture();
-        _textureIsDirty = false;
     }
 }
 
 void RenderablePlaneImageLocal::deinitializeGL() {
-    _textureFile = nullptr;
```
alexanderbock, line 138: Why this removal?

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -153,19 +182,26 @@ void RenderablePlaneImageLocal::update(const UpdateData& data) {
...
+            _rightTexture = nullptr;
+        }
+
         _shouldUnloadTexture = false;
+        _textureIsDirty = false;
     }
 
     if (_textureIsDirty) [[unlikely]] {
         loadTexture();
-        _textureIsDirty = false;
```
alexanderbock, line 161: This would cause the texture to be constantly loaded, no?

alexanderbock (commented):
Didn't run it yet

WeirdRubberDuck:
If this is not going into the release, how about adding the same feature to the RenderableSphereImage and ScreenSpaceImage as part of this PR too? 😇

modules/base/rendering/renderableplaneimage.cpp
WeirdRubberDuck: Gotcha. Then I guess we need to do this for all other image renderables and screenspace renderables that have lazy loading, too?  And would it be worth adding a comment about it, or is it obvious enough to future us?

Roxeena: Created a new issue for it #4073

modules/base/rendering/renderableplane.h
Roxeena: That was the value that was there before. https://github.com/OpenSpace/OpenSpace/blob/7fa609f31163cbbd2dacf34cbadaa53b187a3696/modules/base/rendering/renderableplane.cpp#L253 I didn't want to change the functionality. Or should I?

modules/base/rendering/renderableplaneimagelocal.cpp
Roxeena: It's a unique pointer, so this is not needed anyway, right? But I can put it back to make the deletion clear.

modules/base/rendering/renderableplaneimagelocal.cpp
Roxeena: No, the `loadTexture` function sets the flag. It needs to check the status of both textures, so I thought it was a better location for it. ,

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
Roxeena: But this would not have any stereo correction, right? We might move the two identical (except for color) images depending on the eye, but there won't be any additional depth information in the images themselves since they do not have different perspectives of the motif? Isn't it better to find an actual stereo image pair and use that?

modules/base/rendering/renderableplaneimage.h
Roxeena: I'm not sure which way would be more confusing for the user 🤔 Either they have to have `UseStereo = true` or it will always be stereo when a `RightTexture` is present. Although having the `UseStereo` property opens up a risk that they might be out of sync.

Roxeena:
> If this is not going into the release, how about adding the same feature to the RenderableSphereImage and ScreenSpaceImage as part of this PR too? 😇

Made an issue for it.

modules/base/rendering/renderableplaneimageonline.cpp
```diff
@@ -44,15 +45,30 @@ namespace {
...
+        "during stereo rendering. If this value is changed, the image at the new path "
+        "will automatically be loaded and displayed.",
         Property::Visibility::User
     };
 
     // Creates a textured 3D plane, where the texture image is loaded from the internet
-    // through a web URL.
+    // through a web URL. To use a stereo image set, a second image can be specified with
+    // the `RightURL` property that will be used for the right eye during stereo
+    // rendering.
```
WeirdRubberDuck: The grammar is not quite correct here ("that" refers to the property, not the image). How about:
```suggestion
    // through a web URL. 
    //
    // For stereo rendering, a second image can be specified using the `RightURL`
    // property. This image will be used for the right eye.
```
? 

Maybe having a new paragraph nitpicky, but this seems to be such a secondary feature of the renderable that it could be moved from the "main paragraph.

modules/base/rendering/renderableplaneimagelocal.cpp
```diff
@@ -45,14 +47,28 @@ namespace {
...
+        "RightTexture",
+        "Right Eye Texture",
+        "Sets the path of the texture that is displayed on this plane for the right eye "
+        "during stereo rendering.",
         Property::Visibility::User
     };
 
     // Creates a textured 3D plane, where the texture is provided by a local file on disk.
+    // To use a stereo image set, a second image can be specified with the `RightTexture`
+    // property that will be used for the right eye during stereo rendering.
```
WeirdRubberDuck: Same comment here as below. How about: 
```suggestion
    //
    // For stereo rendering, a second image can be specified using the `RightTexture`
    // property. This image will be used for the right eye.
```

modules/base/rendering/renderableplaneimage.h
WeirdRubberDuck: As I said, feel free to ignore :)  We could also just add a setting alter if we notice that it would be useful

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
WeirdRubberDuck: Yeah, using a red-green pair might be confusing, as users could possibly think that it is stereo, while the only difference would be the color

On the other hand, a real stereo image pair might have very little visual difference, so maybe it's not clear enough that it's two different images?

I think I'd personally lean more towards a real stereo-image pair, but TBH it's not a big deal as long as we can test the example

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
```diff
@@ -0,0 +1,28 @@
...
+-- This example shows how to create a textured plane in 3D space, where the texture is
+-- loaded from two local files on disk to create a stereo image. The two different images
+-- will be used for the left and right eyes, respectively, when OpenSpace is used with
+-- stereo rendering.
+
+local Node = {
+  Identifier = "RenderablePlaneImageLocal_Example_Stereo",
+  Renderable = {
+    Type = "RenderablePlaneImageLocal",
+    Size = 3.0E11,
```
WeirdRubberDuck, line 11:
```suggestion
```
To keep example minimal?

(kitty is huuuuge!)

data/assets/examples/renderable/renderableplaneimageonline/planeimageonline_stereo.asset
```diff
@@ -0,0 +1,28 @@
...
+-- This example shows how to create a textured plane in 3D space, where the texture is
+-- from the internet though two web URLs. to create a stereo image. The two different
+-- images will be used for the left and right eyes, respectively, when OpenSpace is used
+-- with stereo rendering.
+
+local Node = {
+  Identifier = "RenderablePlaneImageOnline_Example_Stereo",
+  Renderable = {
+    Type = "RenderablePlaneImageOnline",
+    Size = 3.0E11,
```
WeirdRubberDuck, line 11:
```suggestion
```

modules/base/rendering/renderableplaneimageonline.cpp
```diff
@@ -44,15 +45,30 @@ namespace {
...
+        "eye.",
+        Property::Visibility::User
+    };
+
+    constexpr Property::PropertyInfo RightTextureInfo = {
+        "RightTexture",
+        "Right Eye Image URL",
+        "Sets the URL of the texture that is displayed on this plane for the right eye "
+        "during stereo rendering. If this value is changed, the image at the new path "
+        "will automatically be loaded and displayed.",
```
WeirdRubberDuck, line 58: Just realized that the stereo will only be correct when looking straight onto the plane. Do we need to add info on this? What do you think?

Maybe the examples should be billboarded for this reason? Or am I overthinking?  Feel free to ignore and do what you think is best :)

WeirdRubberDuck (changes requested):
Tested it in the SBS stereo config and found one issue: 

If I try to set the right texture to an empty string (to disable the stereo during runtime) nothing happens. 

It would be great if we could fix this, either by adding a "UseStereo" so that stereo rendering can easily be disabled during runtime, or by making sure that the left texture is used if no right texture is specified. 

I personally prefer the "UseStereo" setting, since then I can disable it easily without "losing" the texture I had set as the right texture. This is useful if I quickly want to disable stereo and later re-enable it, for example. 

Once this is fixed, from my side this PR is ok to merge (might not be able to go in and approve it for a while)

modules/base/rendering/renderableplaneimage.h
WeirdRubberDuck: Nvm... Duplicating the comment from my review here as it's relevant for this discussion: 

> 
> If I try to set the right texture to an empty string (to disable the stereo during runtime) nothing happens. 
> 
> It would be great if we could fix this, either by adding a "UseStereo" so that stereo rendering can easily be disabled during runtime, or by making sure that the left texture is used if no right texture is specified. 
> 
> I personally prefer the "UseStereo" setting, since then I can disable it easily without "losing" the texture I had set as the right texture. This is useful if I quickly want to disable stereo and later re-enable it, for example.

data/assets/examples/renderable/renderableplaneimagelocal/planeimagelocal_stereo.asset
Roxeena: Since the other RenderablePlane examples are this size I will keep them the same for consistency.

modules/base/rendering/renderableplaneimage.h
Roxeena: After some discussion with Alex, we decided not to have a property for this. However, you can change the right texture at runtime, and the stereo flag will change accordingly, effectively enabling you to turn on/off stereo.

modules/base/rendering/renderableplaneimageonline.cpp
Roxeena: I have made the examples billboarded now.

PR #4074, opened 2026-05-13
Feature/update assetvalidation scripts

engbergandreas:
Update the asset validation to using the updated Python API

### DIFF
--- a/support/assetvalidation/assetvalidation.py
+++ b/support/assetvalidation/assetvalidation.py
@@ -36,6 +36,16 @@ verbose = False
 # Logging object
 logger = logging.getLogger(__name__)
 
+class OpenSpaceCrashException(Exception):
+  """
+  Raised when OpenSpace crashes mid-run, carrying the number of assets that were
+  successfully validated before the crash.
+  """
+  def __init__(self, completedCount: int, cause: Exception):
+    super().__init__(str(cause))
+    self.completedCount = completedCount
+    self.cause = cause
+
 def log(*values: object, logLevel = logging.DEBUG):
   """Custom log function to log messages to file and optionally to console"""
   msg = ' '.join(str(v) for v in values)
@@ -175,9 +185,14 @@ async def waitForAssetState(eventTopic, path: str, expectedState: str) -> str:
         )
         return state
 
-async def internalRun(openspace, assets: list[pathlib.Path], osDir: str, api: Api):
+async def internalRun(
+  openspace, assets: list[pathlib.Path], osDir: str, api: Api,
+  startIndex: int = 0, totalCount: int = 0
+):
   """Logic for running the asset validation tests"""
   assetCount = 0
+  if totalCount == 0:
+    totalCount = len(assets)
 
   logsettings = {
     "timeStamping": False,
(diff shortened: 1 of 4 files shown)


support/assetvalidation/assetvalidation.py
```diff
@@ -193,103 +219,184 @@ async def unloadAssets():
...
+  startIndex = 0
+
+  while remaining:
+    if startOpenSpace:
+      log("Starting OpenSpace...", logLevel = logging.INFO)
+      process = subprocess.Popen(
+        [executable, "--bypassLauncher", "--profile", "empty"],
+        cwd=os.path.dirname(executable),
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.PIPE
```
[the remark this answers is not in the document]
engbergandreas: @alexanderbock issue or we good?

engbergandreas:
TODO: The timeout is an okay-ish idea to solve the issue of forever waiting. However, if there is a large dataset that needs downloading or otherwise takes a really long time to load (for some reason), those assets would be skipped..

support/assetvalidation/assetvalidation.py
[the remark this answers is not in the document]
alexanderbock: I think we're good with that.. don't think that it would actually be a problem

support/assetvalidation/readme.md
```diff
@@ -1,15 +1,27 @@
...
+## Setup
 The project provides a `requirements.txt` with the necessary dependencies. Install using e.g., pip `pip install -r /path/to/requirements.txt`
 
-# Running validation tests
-
+## Running validation tests
 | Parameter | Description |
 | --------- | ----------- |
-| `--dir` | Points to the base folder of the OpenSpace version that is used to execute the validation. There needs to be a compiled version of OpenSpace available in the folder such that `bin/RelWithDebInfo/OpenSpace.exe`(on Windows) or `bin/OpenSpace` (on Linux) exists and is executable.|
+| `--dir` | Points to the base folder of the OpenSpace version that is used to execute the validation. There needs to be a compiled version of OpenSpace available in the folder such that `bin/RelWithDebInfo/OpenSpace.exe`(on Windows) or `bin/OpenSpace` (on Linux) exists and is executable. |
```
alexanderbock: Since the assetvalidation lives in the OpenSpace folder, do we need this?  In the test runner I removed it assumes that we run it from script folder

support/assetvalidation/assetvalidation.py
```diff
@@ -33,9 +33,21 @@
...
+path = ""
 # Logging object
 logger = logging.getLogger(__name__)
 
+class OpenSpaceCrashException(Exception):
+  """
+  Raised when OpenSpace crashes mid-run, carrying the number of assets that were
+  successfully validated before the crash.
+  """
+  def __init__(self, completedCount: int, cause: Exception):
```
alexanderbock, line 64: Should we do proper Python code styling?  Which would be `completed_count` here

support/assetvalidation/readme.md
engbergandreas: Nope, can update to do the same here 👍

support/assetvalidation/assetvalidation.py
engbergandreas: is snake_case something we wan't in all Python scripts? API is currently using camelCase, but it does look a little weird mixing them

alexanderbock: I guess we could change it to snake_case to be Python-consistent?

PR #4101, opened 2026-05-26
Fix crash for saveSettingsToProfile

### LINKED ISSUE #4100: Crash when using saveCurrentSettingsToProfile
OpenSpace crashes when using lua function `openspace.saveCurrentSettingsToProfile()`

Arohdin:
Fix crash due to path problems when using `openspace.saveSettingsToProfile()`.
Created PR just to for someone else to test the code as we are close to release.

Closes #4100 

Note:
Potential problems where some scripts are being saved in a weird format will be fixed in another issue.
Current fix prevents crash and saves the scripts in the profile.

### DIFF
--- a/src/scene/profile_lua.inl
+++ b/src/scene/profile_lua.inl
@@ -76,7 +76,7 @@ namespace {
                                                 bool shouldOverwrite = false)
 {
     const bool receivedSaveFileName = saveFileName.has_value();
-    if (!saveFileName.has_value()) {
+    if (!receivedSaveFileName) {
         std::time_t t = std::time(nullptr);
         std::tm* utcTime = std::gmtime(&t);
         ghoul_assert(utcTime, "Conversion to UTC failed");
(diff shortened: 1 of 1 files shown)


src/scene/profile_lua.inl
```diff
@@ -90,59 +91,51 @@ namespace {
...
-        absFilename = absPath("${USER_PROFILES}/" + *saveFilePath + ".profile").string();
+        absFilename = std::format(
+            "{}/{}.profile", absPath("${USER_PROFILES}"), *saveFileName
+        );
     }
 
-    if (std::filesystem::is_regular_file(absFilename) && !shouldOverwrite) {
+    if (std::filesystem::is_regular_file(absFilename) &&
+        !shouldOverwrite &&
+        receivedSaveFileName)
```
alexanderbock:
```suggestion
    if (std::filesystem::is_regular_file(absFilename) && !shouldOverwrite &&
        receivedSaveFileName)
```

alexanderbock (approved):
We'll leave it to a future fix to make the save function actually work again. It's great that it doesn't crash anymore

PR #4103, opened 2026-05-27
Feature/model node animation

Roxeena:
To be used with https://github.com/OpenSpace/Ghoul/pull/136

Makes it possible to specify a number of transformations to attach to a number of internal model nodes identified by name. There is a simple example file to demonstrate how it would be used. The idea is to use this feature to attach a separate transformation to the solar panels for ISS as an example to make them point toward the Sun. A parameter can also be specified to either replace any existing transformation on that node or to add it on top of it. This can be useful since we do not know how each model is structured with their transformations. Sometimes, there are nodes containing just transformations, and sometimes there can be nodes with both meshes and transformations. 

EDIT:
There is also a new Lua function (and a new library) to print the model tree to help with finding the correct names for nodes to apply a custom transformation to.

### DIFF
--- a/data/assets/examples/renderable/renderablemodel/model_node_transform.asset
+++ b/data/assets/examples/renderable/renderablemodel/model_node_transform.asset
@@ -1,0 +1,52 @@
+-- Node Transform
+-- This example loads a model and applies a custom transformation to one of its internal
+-- model nodes.
+
+-- Load the example model from OpenSpace servers
+-- If you want to use your own model, this block of code can be safely deleted
+local model = asset.resource({
+  Name = "Animated Box",
+  Type = "HttpSynchronization",
+  Identifier = "animated_box",
+  Version = 1
+})
+
+local Node = {
+  Identifier = "RenderableModel_Example_Node_Transform",
+  Renderable = {
+    Type = "RenderableModel",
+    GeometryFile = model .. "BoxAnimated.glb",
+    -- Use the line below insted of the one above if you want to use your own model
+    --GeometryFile = "C:/path/to/model.fbx",
+    CustomTransformNodeName = "nodes[2]",
+    CustomTransform = {
+      Translation = {
+        Type = "StaticTranslation",
+        Position = {10, 10, 10}
+      }
+    }
+  },
+  GUI = {
+    Name = "RenderableModel - Node Transform",
+    Path = "/Examples"
+  }
+}
+
+
+asset.onInitialize(function()
+  openspace.addSceneGraphNode(Node)
+end)
+
+asset.onDeinitialize(function()
+  openspace.removeSceneGraphNode(Node)
+end)
+
+
+-- Model credit
+--[[
+  Author = Cesium, https://cesium.com/
+  URL = https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxAnimated
+  License =
+    Creative Commons Attribution 4.0 International License,
+    https://creativecommons.org/licenses/by/4.0/
+]]
--- a/data/assets/examples/renderable/renderablemodel/model_node_transform.asset
+++ b/data/assets/examples/renderable/renderablemodel/model_node_transform.asset
@@ -18,11 +18,20 @@ local Node = {
     GeometryFile = model .. "BoxAnimated.glb",
     -- Use the line below insted of the one above if you want to use your own model
     --GeometryFile = "C:/path/to/model.fbx",
-    CustomTransformNodeName = "nodes[2]",
-    CustomTransform = {
-      Translation = {
-        Type = "StaticTranslation",
-        Position = {10, 10, 10}
+    CustomTransforms = {
+      {
+        NodeName = "nodes[2]",
+        Translation = {
+          Type = "StaticTranslation",
+          Position = {10, 0, 10}
+        }
+      },
+      {
+        NodeName = "nodes[1]",
+        Translation = {
+          Type = "StaticTranslation",
+          Position = {10, 10, 0}
+        }
       }
     }
   },
--- a/data/assets/examples/renderable/renderablemodel/model_node_transform.asset
+++ b/data/assets/examples/renderable/renderablemodel/model_node_transform.asset
@@ -1,6 +1,6 @@
 -- Node Transform
 -- This example loads a model and applies a custom transformation to one of its internal
--- model nodes.
+-- model nodes identified by name.
 
 -- Load the example model from OpenSpace servers
 -- If you want to use your own model, this block of code can be safely deleted
@@ -18,21 +18,14 @@ local Node = {
     GeometryFile = model .. "BoxAnimated.glb",
     -- Use the line below insted of the one above if you want to use your own model
     --GeometryFile = "C:/path/to/model.fbx",
-    CustomTransformsShouldOverride = false,
     CustomTransforms = {
-      ["nodes[2]"] = {
+      ["nodes[2]"] = { -- Name of the internal node to apply the custom transformation to
         Translation = {
           Type = "StaticTranslation",
           Position = {10, 0, 0}
         }
       }
-    },
-    -- Animation Parameters:
-    EnableAnimation = true,
-    -- Start the animation and play it once at this time
-    AnimationStartTime = "2024 07 09 12:00:00",
-    -- Bounce the animation after the set start time
-    AnimationMode = "BounceFromStart",
+    }
   },
   GUI = {
     Name = "RenderableModel - Node Transform",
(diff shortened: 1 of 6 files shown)


data/assets/examples/renderable/renderablemodel/model_node_transform.asset
```diff
@@ -0,0 +1,53 @@
...
+
+local Node = {
+  Identifier = "RenderableModel_Example_Node_Transform",
+  Renderable = {
+    Type = "RenderableModel",
+    GeometryFile = model .. "BoxAnimated.glb",
+    -- Use the line below instead of the one above if you want to use your own model
+    --GeometryFile = "C:/path/to/model.fbx",
+    CustomTransforms = {
+      ["nodes[2]"] = { -- Name of the internal node to apply the custom transformation to
```
alexanderbock: This seems strange since the `animated_box` does have correct names in Blender:

Image

data/assets/examples/renderable/renderablemodel/model_node_transform.asset
```diff
@@ -0,0 +1,53 @@
...
+        Translation = {
+          Type = "StaticTranslation",
+          Position = {10, 0, 0}
```
alexanderbock:
```suggestion
          Position = { 10.0, 0.0, 0.0 }
```

modules/base/rendering/renderablemodel.cpp
```diff
@@ -608,11 +642,102 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            std::string cleanNodeName = nodeName;
+            std::replace_if(
+                cleanNodeName.begin(),
+                cleanNodeName.end(),
+                [](auto ch) {
+                    std::string toRemove = ". \t\n";
+                    return toRemove.find(ch) != std::string::npos;
+                },
+                '_'
+            );
```
alexanderbock:
```suggestion
            std::string cleanNodeName = makeIdentifier(nodeName);
```

?

modules/base/rendering/renderablemodel.cpp
```diff
@@ -608,11 +642,102 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            if (nodeTransform.translation.has_value()) {
+                _customNodeTransforms[nodeName].translation =
+                    Translation::createFromDictionary(*nodeTransform.translation);
+
+                LDEBUG(std::format(
+                    "Applied custom translation on node '{}' for model '{}'",
+                    nodeName, _file
+                ));
+
+                _customNodeTransforms[nodeName].translation->setIdentifier(
```
alexanderbock: Instead of adding them all individually, it would be cleaner to first create a propertyowner "Custom Transforms" with then "nodes[2]" as a subowner that then has a Translation, Rotation, and Scaling inside.

Otherwise we'll end up with a long and unmanagable propertyowner list and it is mimicking the eay the GeoJSONs and Layers are handled

Image

modules/base/rendering/renderablemodel.cpp
```diff
@@ -608,11 +642,102 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            );
+
+            if (nodeTransform.translation.has_value()) {
+                _customNodeTransforms[nodeName].translation =
+                    Translation::createFromDictionary(*nodeTransform.translation);
+
+                LDEBUG(std::format(
+                    "Applied custom translation on node '{}' for model '{}'",
+                    nodeName, _file
+                ));
```
alexanderbock, line 683:
```suggestion
```

Not sure if the debug messages ardd too much.

modules/base/rendering/renderablemodel.cpp
```diff
@@ -608,11 +642,102 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+
+            if (nodeTransform.rotation.has_value()) {
+                _customNodeTransforms[nodeName].rotation = Rotation::createFromDictionary(
+                    *nodeTransform.rotation
+                );
+
+                LDEBUG(std::format(
+                    "Applied custom rotation on node '{}' for model '{}'",
+                    nodeName, _file
+                ));
```
alexanderbock, line 701:
```suggestion
```

modules/base/rendering/renderablemodel.cpp
```diff
@@ -608,11 +642,102 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
...
+            }
+
+            if (nodeTransform.scale.has_value()) {
+                _customNodeTransforms[nodeName].scale =
+                    Scale::createFromDictionary(*nodeTransform.scale);
+
+                LDEBUG(std::format(
+                    "Applied custom scale on node '{}' for model '{}'",
+                    nodeName, _file
+                ));
```
alexanderbock, line 718:
```suggestion
```

modules/base/rendering/renderablemodel.cpp
```diff
@@ -608,11 +642,102 @@ RenderableModel::RenderableModel(const ghoul::Dictionary& dictionary)
     };
 
     _originalRenderBin = renderBin();
+
+    _customNodeTransformsShouldOverride =
+        p.customTransformsShouldOverride.value_or(_customNodeTransformsShouldOverride);
+
+    if (p.customTransforms.has_value()) {
+        _hasCustomNodeTransforms = true;
```
alexanderbock: It seems all that `_hasCustomNodeTransforms` does is guard against for-looping over the empty map, which is save to do anyway. So I don't think we need it?

data/assets/examples/renderable/renderablemodel/model_node_transform.asset
alexanderbock:
```suggestion
      ["inner_box"] = { -- Name of the internal node to apply the custom transformation to
```

With the extra change in the Ghoul repo

modules/base/rendering/renderablemodel.cpp
Roxeena: I wanted it to kind of mirror the logging behaviour from the scene graph nodes, as here
https://github.com/OpenSpace/OpenSpace/blob/3820a1a9b4faa1735a38873ab2dacdd7a9bcb704/src/scene/scenegraphnode.cpp#L443
but I can remove it if we don't want it.

modules/base/rendering/renderablemodel_lua.inl
```diff
@@ -0,0 +1,69 @@
...
+#include <ghoul/io/model/modelreaderbase.h>
+#include <ghoul/logging/logmanager.h>
+#include <ghoul/misc/assert.h>
+
+namespace {
+
+/**
+ * Prints the model tree of the given model file. Will only work for assimp models.
+ */
+[[codegen::luawrap]] void printModelTree(std::filesystem::path filename) {
```
WeirdRubberDuck:
```suggestion
[[codegen::luawrap]] void printModelTree(std::filesystem::path filepath) {
```
since it's the full name, not just the path?

modules/base/rendering/renderablemodel_lua.inl
```diff
@@ -0,0 +1,69 @@
...
+#include <ghoul/io/model/modelreaderassimp.h>
+#include <ghoul/io/model/modelreaderbase.h>
+#include <ghoul/logging/logmanager.h>
+#include <ghoul/misc/assert.h>
+
+namespace {
+
+/**
+ * Prints the model tree of the given model file. Will only work for assimp models.
+ */
```
WeirdRubberDuck: Add `\param` and `\return` to explain the input and output parameter. Will show in documentation :) 

Example: https://docs.openspaceproject.com/latest/reference/scripting-api/openspace.navigation.html#loadnavigationstatefromfile

Roxeena: I guess the void does not need a `\return`? I added some more information in the Doxygen comments.

modules/base/rendering/renderablemodel_lua.inl
```diff
@@ -30,23 +30,39 @@
...
-    ghoul_assert(!filename.empty(), "Filename must not be empty");
+    /**
+     * Prints the model tree of the \p filepath model file. The node names detected during
+     * reading and the hierarchy of the nodes will be printed. This is useful for debugging
+     * and finding correct node names for applying custom transformations to internal model
+     * nodes.
+     *
+     * \param filepath The model file on disk whose model tree should be printed.
+     *
+     * \throw ModelLoadException If there was an error reading the \p filepath
```
WeirdRubberDuck: I don't know if these `\throw`s will work, or if they will even cause issues when building the documentation. I'd recommend trying to build the domentation reference based on your local OpenSpace version and check

WeirdRubberDuck: Same goes for `\pre`

Roxeena: Yes, the `\throw` and `\pre` didn't work, so I baked it into the description itself.

PR #4107, opened 2026-06-02
Added inversion to size mapping component

micahnyc:


modules/base/rendering/pointcloud/sizemappingcomponent.cpp
```diff
@@ -72,6 +72,13 @@ namespace {
         Property::Visibility::AdvancedUser
     };
 
+    constexpr Property::PropertyInfo InvertScaleInfo = {
+        "InvertScale",
+        "Invert scale",
+        "If true, the scale factor is inverted before being applied to the data values.",
```
WeirdRubberDuck: Maybe nitpicky, but to avoid possible confusion with the "ScaleFactor" property, how about being a bit more explicit in this description: 


```suggestion
        "If true, the points will be scaled based on the inverse of the data value "
        "(1/value), so that larger values result in a smaller size, and vice versa.",
```

WeirdRubberDuck (approved):
Looks good!  Took the liberty of adding an example asset so that we can create a visual test for this more easily. 

Hope you don't mind!

data/assets/examples/renderable/renderablepointcloud/pointcloud_scaledfromdata_inverse.asset
```diff
@@ -0,0 +1,41 @@
...
+    SizeSettings = {
+      SizeMapping = {
+        -- The options for the columns that the points can be scaled by. The first
+        -- alternative in the list is chosen per default
+        ParameterOptions = { "a", "b" },
+        -- Specify which option we want to use for size mapping at start up. Here we
+        -- use the last of the provided options rather than the first one, which is
+        -- otherwise used by default
+        Parameter = "b",
+        -- Invert the size mapping, so that higher values in the column result in smaller points and vice versa
```
alexanderbock:
```suggestion
        -- Invert the size mapping, so that higher values in the column result in
        -- smaller points and vice versa
```

modules/base/rendering/pointcloud/renderablepointcloud.cpp
```diff
@@ -752,6 +752,7 @@ RenderablePointCloud::RenderablePointCloud(const ghoul::Dictionary& dictionary)
             [this]() { _dataIsDirty = true; }
         );
         _sizeSettings.sizeMapping->isRadius.onChange([this]() { _dataIsDirty = true; });
+        _sizeSettings.sizeMapping->invertScale.onChange([this]() { _dataIsDirty = true; });
```
alexanderbock:
```suggestion
        _sizeSettings.sizeMapping->invertScale.onChange([this]() {
            _dataIsDirty = true;
        });
```

modules/base/rendering/pointcloud/sizemappingcomponent.cpp
```diff
@@ -204,6 +217,8 @@ SizeMappingComponent::SizeMappingComponent(const ghoul::Dictionary& dictionary)
     }
 
     isRadius = p.isRadius.value_or(isRadius);
+
```
alexanderbock:
```suggestion
```

alexanderbock (approved):

PR #4112, opened 2026-06-04
Add drag-and-drop support for Models

alexanderbock:
Makes it possible to drag an drop any supported 3D model support. It gets added as a new scene graph node in 50m in front of the camera.

### DIFF
--- a/scripts/drag_drop_handler.lua
+++ b/scripts/drag_drop_handler.lua
@@ -20,7 +20,7 @@ end
 -- Lua doesn't enjoy \ that are used by Windows extensively. So we convert all \ into /
 filename = filename:gsub("\\", "/")
 basename = basename:gsub("\\", "/")
-basename_without_extension = basename:sub(0, #basename - extension:len())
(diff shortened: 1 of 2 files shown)


micahnyc:
Worked well, 

- I was thinking maybe we could use the bounding sphere of the item instead of 50. (would require setting the frame after like we do with the focus).

- With a more complex model, it created the cache and so locked up for a while. (To be expected I guess)

- When dragging in a 2nd model, it winds up parented to the 1st model, which makes sense but was unexpected to me as it also didn't show up in front of the camera. My first thought is to have the navigationhandler keep a property like "LastSelectedGlobe" but my 2nd thought is to have  the engine know the "ClosetGlobe" and then maybe we can use that in other places as well like this one: https://github.com/openspace/openspace/issues/2696


----Large models might want bounding spere calc

OpenSpace_000001

OpenSpace_000000

alexanderbock:
New issue #4119 for the bounding sphere

PR #4130, opened 2026-06-16
Add the ability for users to define their own Lua scripts in the user folder

alexanderbock:
Creates a new user folder `scripts`. Any `.lua` script placed in that folder will be automatically loaded into the LuaState that is used to execute assets, the console, etc.

This allows users to provide their own functions and enable more reuse. Wth the ShowComposer we are seeing more users copy-pasting functions in ScriptComponents that is unwieldy and there is currently no better way to organize code.


src/engine/openspaceengine.cpp
```diff
@@ -1823,9 +1836,7 @@ LuaLibrary OpenSpaceEngine::luaLibrary() {
             codegen::lua::VramInUse,
             codegen::lua::RamInUse
         },
-        {
-            absPath("${SCRIPTS}/core_scripts.lua")
-        }
+        std::move(scripts)
```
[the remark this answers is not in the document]
alexanderbock: Going to be addressed in a separate PR

src/engine/openspaceengine.cpp
```diff
@@ -1800,6 +1800,23 @@ void OpenSpaceEngine::removeModeChangeCallback(CallbackHandle handle) {
...
+    if (FileSys.hasRegisteredToken("${USER_SCRIPTS}")) {
+        std::vector<std::filesystem::path> userFiles = ghoul::filesystem::walkDirectory(
+            absPath("${USER_SCRIPTS}"),
+            ghoul::filesystem::Recursive::Yes,
+            ghoul::filesystem::Sorted::Yes,
+            [](const std::filesystem::path& p) {
+                return p.extension() == ".lua" && std::filesystem::is_directory(p);
+            }
+        );
+        scripts.insert(scripts.end(), userFiles.begin(), userFiles.end());
```
Roxeena: Do we want to log something to let the user know that OpenSpace has successfully added their scripts?

Roxeena (commented):
I am not too sure of how to test this, but it did not work for me. I created a new file (_test.lua_) in a new folder (_\user\scripts_). This file then contained my new function as:
```lua
function helloUniverse()
    print("Hello Universe!")
end
```
And then I tried to call this function (`helloUniverse()`) in the OpenSpace console. I also tried to create a new script component in the ShowComposer that calls this function. Both cases yielded an error that the function was not recognised.

alexanderbock:
This feature goes very much hand-in-hand with https://github.com/OpenSpace/OpenSpace/pull/4136 to make it more obvious how to register functions

src/engine/openspaceengine.cpp
```diff
@@ -1800,6 +1800,23 @@ void OpenSpaceEngine::removeModeChangeCallback(CallbackHandle handle) {
...
 
 LuaLibrary OpenSpaceEngine::luaLibrary() {
+    std::vector<std::filesystem::path> scripts;
+    // Adding the core scripts
+    scripts.push_back(absPath("${SCRIPTS}/core_scripts.lua"));
+
+    // Add user defined scripts
+    if (FileSys.hasRegisteredToken("${USER_SCRIPTS}")) {
+        std::vector<std::filesystem::path> userFiles = ghoul::filesystem::walkDirectory(
+            absPath("${USER_SCRIPTS}"),
```
WeirdRubberDuck: Add this directory if it does not exist, similar to the other folders in the user folder? 

Done here:
https://github.com/OpenSpace/OpenSpace/blob/41955785a126edd54d4ed2fad0508f78c5a9acc7/src/engine/openspaceengine.cpp#L1022

src/engine/openspaceengine.cpp
```diff
@@ -1800,6 +1800,23 @@ void OpenSpaceEngine::removeModeChangeCallback(CallbackHandle handle) {
...
+            ghoul::filesystem::Recursive::Yes,
+            ghoul::filesystem::Sorted::Yes,
+            [](const std::filesystem::path& p) {
+                return p.extension() == ".lua" && std::filesystem::is_directory(p);
```
WeirdRubberDuck: I don't think this `std::filesystem::is_directory(p)` is correct here. The added files will not be directories, so this will not work. 

What did you intend to check?  `is_regular_file`?

WeirdRubberDuck: Tested without the is_directory check, and then everything works as expected

WeirdRubberDuck (changes requested):
Nice, but unfortunately, there is a detail in the code that makes it not work. See comments

src/engine/openspaceengine.cpp
WeirdRubberDuck: Otherwise, users have to know that they need to add this folder to make this work, and the feature becomes much more invisible

src/engine/openspaceengine.cpp
alexanderbock: ☝️   Yea, was missing a `!` before the check

src/engine/openspaceengine.cpp
alexanderbock: Not sure I understand;  the directory is already being created if it doesn't exist?

WeirdRubberDuck: No, it is not - I had to create it manually :) 

The code I linked to seems to be where we create the other folders in the user folder?

WeirdRubberDuck (commented):
👍 A few comments to consider before merging: 

1. I like the change to move the function to the global namespace, but now it's no longer possible to include it in the autocomplete (by tabbing) in the console. It was when the function was registered under the `openspace` library. Is this something we can easily address?
2. The scripts folder in the users folder should be created automatically if it does not exist

src/engine/openspaceengine.cpp
alexanderbock: That is really strange, for me it definitely created it automatically:


https://github.com/user-attachments/assets/d0806441-ad07-40b4-9184-012f93f47716



I did add it into `createUserDirectoriesIfNecessary` just in case anyway

WeirdRubberDuck: Damn, my bad. It showed up at the very end in my user folder, not among the folders. And i have quite a few files in the folder, so I did not see it... 

So sorry for the inconvenience and stubbornness :s

alexanderbock: All good!

PR #4131, opened 2026-06-16
Updated Testwizard to new api version

micahnyc:


alexanderbock:
I must have forgotten to push the change to remove the Python script;   it has been replaced with a web-based wizard at https://tools.openspaceproject.com/

PR #4134, opened 2026-06-17
Fix powerscaling case sensitive includes

hn-88:
Some module shaders still used `powerScaling` instead of `powerscaling` in their includes, which broke shaders on Linux and other case sensitive filesystems.


alexanderbock:
Forgot to set it as well, but for next time, if you say "Closing #XXXX" in the initial message, closing the PR will automatically close the issue as well

PR #4136, opened 2026-06-17
More explicit registering of new global Lua functions

alexanderbock:
Previously, Lua functions were registered implicitly with documentation and implementation separated. This change adds a function that is used to register the function together with the documentation, which now also checks that the function dies not already exist. It was previously also possible for files to insert documentation into the wrong namespace, a feature that has been removed now. The correctness was tested by comparing the generated documentation files before and after.

## Before
```lua
openspace.space.documentation = {
  {
    Name = "tleToSpiceTranslation",
    Arguments = {
      { "tlePath", "String" }
    },
    Return = "( Translation, SpiceKernel )",
    Documentation = [[
      Takes the provided TLE file, converts it into a SPICE kernel and returns a
      SpiceTranslation instance that can be used to access the information in the TLE
      file using SPICE's superior integral solver.

      The second return value is the spice kernel that should be loaded and unloaded by
      whoever called this function.
    ]]
  }
}

openspace.space.tleToSpiceTranslation = function(tlePath)
  -- ...
end
```

## After
```lua
registerFunction(
  {
    Name = "tleToSpiceTranslation",
    Arguments = {
      { "tlePath", "String" }
    },
    Return = "( Translation, SpiceKernel )",
    Documentation = [[
      Takes the provided TLE file, converts it into a SPICE kernel and returns a
      SpiceTranslation instance that can be used to access the information in the TLE
      file using SPICE's superior integral solver.

      The second return value is the spice kernel that should be loaded and unloaded by
      whoever called this function.
    ]],
    Function = function(tlePath)
      -- ...
    end
  }
)
```

As part of this two bugs were found and removed:
  1. `setDefaultDashboard` function was defined in the documentation but never declared
  2. `addExoplanetSystems` was using the wrong variable name


scripts/core_scripts.lua
```diff
@@ -1,64 +1,163 @@
-openspace.documentation = {
+registerFunction(
   {
     Name = "markInterestingNodes",
     Arguments = {{ "sceneGraphNodes", "String[]" }},
     Documentation = [[This function marks the scene graph nodes identified by name as
       interesting, which will provide shortcut access to focus buttons and featured
-      properties.]]
-  },
+      properties.]],
```
WeirdRubberDuck: For consistency and readability, how about always linebreaking before and after the `[[` brackets ?

That is, always doing this: 
```
    Documentation = [[
      This function marks the scene graph nodes identified by name as
      interesting, which will provide shortcut access to focus buttons and featured
      properties.
    ]],
```
Makes both the code itself and and the text in the docs easier to read

WeirdRubberDuck: If so, please make this change for all relevant functions. It seems to only be this file that needs updating

WeirdRubberDuck (approved):
This is a huge improvement - very nice! 

My only thought was that the library the function is added to is no longer visible in the code file. This is a non-issue, however, since it's pretty apparent that the function is registered to the library for the specific module.  Rather, it's nice to remove this redundancy, and it reduces the risk of errors from faultily specifying the library. 

Maybe overkill, but I double-checked that the documentation is still correctly added to the docs page when using the docswriter

WeirdRubberDuck:
Realization after looking at https://github.com/OpenSpace/OpenSpace/pull/4130:  

Do we want to allow the user to specify a library to register the function to, using the `registerFunction` function?  Or is it enough that their custom scripts will always be added to the base `openspace.` library? 

Also, can we expose the information of their loaded functions somehow?  Any documentation that they write will not be visible anywhere (this could be a future issue, to show the documentation of the "local" Lua API somehwhere in the UI)

alexanderbock:
Yea, I also think that we could add that later into an in-game documentation of the Lua functions.  Could see that as an addition to the Lua console where we should the documentation

PR #4138, opened 2026-06-17
Fix return values for geoPositionFromCamera functions

micahnyc:
fixes scale models and nightsky functions models and nightsky functions


visualtests/profiles/nightsky/lookwest.ostest
```diff
@@ -0,0 +1,32 @@
+{
+  "profile": "nightsky",
+  "commands": [
+    {
+      "type": "script",
+      "value": "openspace.navigation.flyToGeo('Earth', 42.4777983094, -74.378376190727, 0, 0)"
+    },
```
alexanderbock, line 21: This one would better be served with a NavigationState command since that also sets the up vector, look-at etc

alexanderbock: Also replaced the `time` command below

visualtests/profiles/nightsky/lookwest.ostest
```diff
@@ -0,0 +1,32 @@
...
+    {
+      "type": "script",
+      "value": "openspace.fadeIn(\"Scene.CardinalDirectionSphere.Renderable\")"
+    },
```
alexanderbock, line 28:
```suggestion
    {
      "type": "property",
      "value": {
        "property": "Scene.CardinalDirectionSphere.Renderable.Enabled",
        "value": true
      }
    },
```

Otherwise, especially since there is no wait command, the screenshot will be taken before the fading is finished

visualtests/profiles/nightsky/lookwest.ostest
```diff
@@ -0,0 +1,32 @@
...
+    {
+      "type": "action",
+      "value": "os.nightsky.LevelHorizonPitch"
+    },
+    {
+      "type": "action",
+      "value": "os.FadeDownTrails"
+    },
+    {
```
alexanderbock, line 45:
```suggestion
    {
      "type": "wait",
      "value": 20
    },
    {
```

To make sure any globe tile loads correctly

alexanderbock (approved):

PR #4179, opened 2026-07-02
Fix globe browsing skirt artefact

### LINKED ISSUE #4178: Globe fading artefact
When fading a globe, there are "skirt" artefacts to the tiles that show up.

https://github.com/user-attachments/assets/c9614600-ffa6-4126-8207-f3d264268d2a

ylvaselling:
Before:

https://github.com/user-attachments/assets/e39ea2c8-d882-4991-82cd-bd9968bcabd6

After:

https://github.com/user-attachments/assets/6993ddae-7c8b-4768-ba88-26c896198951

The fix is that each vertex is given a float (0 or 1) if it is a skirt (pushed in edge) or not. If the opacity is less than 0 (includes fading) then a vertex with skirt value 1 will be discarded.

### DIFF
--- a/modules/globebrowsing/shaders/globalrenderer_vs.glsl
+++ b/modules/globebrowsing/shaders/globalrenderer_vs.glsl
@@ -120,9 +120,10 @@ void main() {
   );
 
   // Get the height value and apply skirts
-  out_data.isSkirt = tileVertexSkirtLength() > 0.0 ? 1.0 : 0.0;
+  float skirtLength = tileVertexSkirtLength();
+  out_data.isSkirt = skirtLength > 0.0 ? 1.0 : 0.0;
   float height =
-    tileHeight(in_texCoords, out_data.levelWeights) - tileVertexSkirtLength();
+    tileHeight(in_texCoords, out_data.levelWeights) - skirtLength;
 
 #if USE_ACCURATE_NORMALS
   // Calculate tangents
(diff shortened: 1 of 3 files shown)


modules/globebrowsing/shaders/globalrenderer_vs.glsl
```diff
@@ -119,6 +120,7 @@ void main() {
   );
 
   // Get the height value and apply skirts
+  out_data.isSkirt = tileVertexSkirtLength() > 0.0 ? 1.0 : 0.0;
```
engbergandreas: Im just curious, could this not be a boolean? :O

alexanderbock: Unfortunately not.. GLSL doesn't allow `bool` as a type to pass between shaders

alexanderbock (approved):
:+1:

It introduces some artifacts when Earth is half-transparent and while being close to the surface where a skirt would be needed, but I think that is acceptable:

Image

modules/globebrowsing/shaders/globalrenderer_vs.glsl
ylvaselling: Not allowed :/ https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.html#output-variables

ylvaselling:
> 👍
> 
> It introduces some artifacts when Earth is half-transparent and while being close to the surface where a skirt would be needed, but I think that is acceptable:
> Image

Yea.. it is not perfect but I don't think it got worse :)

https://github.com/user-attachments/assets/290b0a51-ee02-4376-a130-d4bba7f0519b

PR #4196, opened 2026-07-20
Update joystick mapping and refactor to get rid of joystick_helper.asset

WeirdRubberDuck:
Update joystick input mapping to make it easier to fly around the current focus node, and simplify joystick assets by removing the joystick_helper. 

Summary of mapping changes: 

1. Move roll input to the upper trigger buttons
    - Note that these use the WebSocket commands to apply a roll, as we cannot currently bind a given camera motion to a button, only axes
    - This was the main motivation for the change, as we could not roll nicely when flying to e.g. land on a surface
2. Move Pause to `DPad.UP`, to put all the time control actions in the same place
3. Move "Set time to yesterday" to `X` for Xbox and `Square` for PlayStation
4. Move "Change of focus node" commands to the `Back`/`Start` for Xbox, and `Share`/`Options` for PlayStation
5. Remove roll friction binding (we concluded that this is never needed)

Other changes: 
* Fix a bug that a bound joystick mapping could not be unbound, and unbind mapping on asset deinitialize
* Combine duplicated mappings (wired and wireless version of same controller), to ultimately remove the need of the joystick_helper
* Remove the joystick_helper.asset file (⚠️ **Breaking change** ⚠️ )
* Fix a bug where the joystick input was not reset when a controller is disconnected

Tested controllers:
- [X] Xbox
- [X] Xbox Wireless
- [X] Space Mouse - Compact
- [ ] Space Mouse - Enterprise version
- [x] PS4
- [x] PS5 

FYI: 
* Next, I will look at refactoring the joystick binding code, so we can add it to the UI. With this, the mapping configs as they look in this file will likely be replaced with a format that is easier to edit, like a JSON file.


WeirdRubberDuck:
Future me: Remember to update docs: https://docs.openspaceproject.com/releases-v0.22/using-openspace/navigation/joystick.html

alexanderbock (approved):

Roxeena (approved):
:+1:

PR #4197, opened 2026-07-21
Queue messages to be sent via main thread

engbergandreas:
Fixes an issue where we tried locking the same mutex deep inside ghoul networking files when disconnecting (or connecting) an app. Topic messages are now put in a queue and sent from the main thread on presync instead of synchronous on the same thread

Appeared with the recent ghoul network changes https://github.com/OpenSpace/Ghoul/pull/137

### DIFF
--- a/src/topic/topics/errorlogtopic.cpp
+++ b/src/topic/topics/errorlogtopic.cpp
@@ -25,7 +25,9 @@
 #include <openspace/topic/topics/errorlogtopic.h>
 
 #include <openspace/documentation/schema.h>
+#include <openspace/engine/globals.h>
 #include <openspace/topic/notificationlog.h>
+#include <openspace/topic/server.h>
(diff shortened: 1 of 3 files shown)


alexanderbock (approved):

src/topic/topics/errorlogtopic.cpp
```diff
@@ -126,15 +131,35 @@ void ErrorLogTopic::createLog() {
...
+    std::vector<nlohmann::json> messagesToSend;
+    {
+        const std::unique_lock lock(_queuedMessagesMutex);
+        std::swap(messagesToSend, _queuedMessages);
+    }
+
+    for (const nlohmann::json& message : messagesToSend) {
+        sendData(message);
+    }
+
```
WeirdRubberDuck:
```suggestion
```

WeirdRubberDuck (approved):

PR #4198, opened 2026-07-29
Add Hidden flags to Actions and ScreenSpaceRenderables

alexanderbock:
Only adds the backend data storage, not any UI components

### DIFF
--- a/src/interaction/actionmanager_lua.inl
+++ b/src/interaction/actionmanager_lua.inl
@@ -106,6 +106,10 @@ struct [[codegen::Dictionary(Action)]] Action {
     // Determines whether the provided command will be executed locally or will be sent to
     // connected computers in a cluster or parallel connection environment.
     std::optional<bool> isLocal;
+
+    // Determines whether the provided command will be hidden from the user interface.
+    // This is just a hint that will be forwarded to the user interface.
+    std::optional<bool> isHidden;
 };
 
 /**
@@ -140,7 +144,8 @@ struct [[codegen::Dictionary(Action)]] Action {
         .guiPath = action.guiPath.value_or("/"),
         .color = action.color,
         .textColor = action.textColor,
-        .isLocal = openspace::Action::IsLocal(action.isLocal.value_or(false))
+        .isLocal = openspace::Action::IsLocal(action.isLocal.value_or(false)),
+        .isHidden = openspace::Action::IsHidden(action.isHidden.value_or(false))
     };
     global::actionManager->registerAction(std::move(a));
 }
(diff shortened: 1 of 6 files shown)


src/interaction/actionmanager_lua.inl
WeirdRubberDuck:
```suggestion
* keys: `Identifier`, `Command`, `Name`, `Documentation`, `GuiPath`,
* `IsLocal`, and `IsHidden`.
```

WeirdRubberDuck:
```suggestion
* Returns information about the action as a table with the keys: `Identifier`, `Command`,
* `Name`, `Documentation`, `GuiPath`, `IsLocal`, and `IsHidden`.
```

WeirdRubberDuck (approved):
Looks good to me, but remember to also update the JSON schema for the actions topic: https://github.com/OpenSpace/OpenSpace/blob/master/src/topic/topics/actionkeybindtopic.cpp#L59

And we need to remember to update the generated typescript types for the API and UI. 

If you want, I can go ahead and update the UI code based on this change right away?  Would be good to have it be testable

src/rendering/screenspacerenderable.cpp
```diff
@@ -144,6 +144,14 @@ namespace {
         Property::Visibility::User
     };
 
+    constexpr Property::PropertyInfo IsHiddenInfo = {
+        "Hidden",
+        "Hidden",
```
WeirdRubberDuck: Just realized that for scene graph nodes, this property is called "GuiHidden" rather than "Hidden". Not sure that it matters that they are different, but wanted to check if we're ok with this inconsistency?

alexanderbock: Yea, consistency there is probably better.. Will change

alexanderbock:
@WeirdRubberDuck  Yea, the UI responding to both would be very useful

src/topic/topics/actionkeybindtopic.cpp
```diff
@@ -73,7 +73,8 @@ Schema ActionKeybindTopic::Schema() {
                 "guiPath": { "type": "string" },
                 "identifier": { "type": "string" },
                 "isLocal": { "type": "boolean" },
-                "name": { "type": "string" }
+                "name": { "type": "string" },
+                "isHidden" { "type": "boolean" }
```
WeirdRubberDuck, line 77:
```suggestion
                "isHidden": { "type": "boolean" }
```

WeirdRubberDuck:
Just realized that the `ScreenSpaceSkyBrowser` class also has a "IsHidden" property already. Should we remove this one and instead replace it with the one that now exists for all ScreenSpaceRenderables?

I'm a bit confused about what it's used for. It seems to be always set to true, and we don't use it anywhere in the UI. Do we even use the screenspaceskybrowser for the skybrowser feature nowadays? 🤔

PR #4206, opened 2026-08-11
Feature/rename parallelpeer to astrocast

engbergandreas:
This PR renames the `ParallelPeer` to `Astrocast`

Also does some cleanup on the code


alexanderbock (approved):

data/assets/dashboard/astrocast.asset
```diff
@@ -18,8 +18,8 @@ asset.export(Item)
 
 
 asset.meta = {
-  Name = "Dashboard - ParallelConnection",
-  Description = "This asset provides a ParallelConnection dashboard item",
+  Name = "Dashboard - AstroCast",
+  Description = "This asset provides an Astrocast dashboard item",
```
WeirdRubberDuck: Even if we stick with the name, we need to decide if it is "AstroCast" or "Astrocast" :)

and if it should be written in caps when talking about it in docs  ("an astrocast connection" versus "an Astrocast connection")

data/assets/examples/dashboarditem/dashboarditemparallelconnection/parallelconnection.asset
```diff
@@ -1,9 +1,9 @@
 -- Basic
--- This example adds a dashboard item that shows the status of the parallel connection.
+-- This example adds a dashboard item that shows the status of the Astrocast.
```
WeirdRubberDuck:
```suggestion
-- This example adds a dashboard item that shows the status of the Astrocast connection.
```
?

data/assets/dashboard/astrocast.asset
alexanderbock: I've used Astrocast sofar in my text

src/network/astrocast.cpp
```diff
@@ -538,80 +667,97 @@ void ParallelPeer::requestHostship(std::optional<std::string> hostPassword) {
...
 }
 
-void ParallelPeer::sendScript(std::string script) {
+void Astrocast::sendScript(std::string script) {
     if (!isHost()) {
         return;
     }
 
+    // @TODO (anden88 2026-08-11): This is a temporary solution to avoid sending scripts
+    // that are related to astrocast. Proabbly what we want is to have the JavaScript
```
WeirdRubberDuck:
```suggestion
    // that are related to astrocast. Probably what we want is to have the JavaScript
```

src/network/astrocast.cpp
```diff
@@ -691,127 +834,111 @@ void ParallelPeer::setStatus(ParallelConnection::Status status) {
...
-    kf._followNodeRotation = navHandler.orbitalNavigator().followingAnchorRotation();
-    if (kf._followNodeRotation) {
-        kf._position = glm::inverse(focusNode->worldRotationMatrix()) * kf._position;
-        kf._rotation = navHandler.orbitalNavigator().anchorNodeToCameraRotation();
-    }
-    else {
-        kf._rotation = navHandler.camera()->rotationQuaternion();
-    }
+Astrocast::Message Astrocast::receiveMessage() {
+    // Header consists of...
```
WeirdRubberDuck:
```suggestion
```
? At first look, this comment looks unfinished, but either way it's a little redundant?

src/network/astrocast.cpp
```diff
@@ -38,23 +38,181 @@
...
+    constexpr std::string_view _loggerCat = "Astrocast";
     constexpr size_t MaxLatencyDiffs = 64;
 
+    // Gonna do some UTF-like magic once we reach 255 to introduce a second byte or so
+    static constexpr uint8_t ProtocolVersion = 7;
+
+    class ConnectionLostError final : public ghoul::RuntimeError {
+    public:
+        explicit ConnectionLostError(bool shouldLogError_ = true)
+            : ghoul::RuntimeError("Astrocast connection lost", "Astrocast")
```
WeirdRubberDuck, line 68: I was a bit confused about the input parameter when this error was used to not log error ("what does false mean?)"

How about introducing a `BooleanType` to make it a bit clearer what the value means? 


```suggestion
    BooleanType(ShouldLogError);

    class ConnectionLostError final : public ghoul::RuntimeError {
    public:
        explicit ConnectionLostError(ShouldLogError shouldLogError_ = ShouldLogError::Yes)
            : ghoul::RuntimeError("Astrocast connection lost", "Astrocast")
```

The line
```cpp
throw ConnectionLostError(false);
```
then becomes: 
```cpp
throw ConnectionLostError(ShouldLogError::No);
```

WeirdRubberDuck (approved):
👍 Didn't try running the astrocast featue, but code and name changes looks good to me

src/network/astrocast.cpp
```diff
@@ -691,127 +834,111 @@ void ParallelPeer::setStatus(ParallelConnection::Status status) {
...
+    constexpr size_t HeaderSize =
+        2 * sizeof(char) + // OS
+        sizeof(uint8_t) +  // Protocol version
+        sizeof(uint8_t) +  // Message type
+        sizeof(uint32_t);  // message size
 
-    kf._focusNode = focusNode->identifier();
-    kf._scale = navHandler.camera()->scaling();
-    kf._timestamp = global::windowDelegate->applicationTime();
+    // Create basic buffer for receiving first part of messages
```
WeirdRubberDuck: First part?  Isn't this the entire message?

data/assets/examples/dashboarditem/dashboarditemastrocastconnection/astrocastconnection.asset
```diff
@@ -1,9 +1,9 @@
 -- Basic
```
WeirdRubberDuck, line 1: This file (and folder) needs renaming

src/network/astrocast.cpp
```diff
@@ -159,43 +316,42 @@ ParallelPeer::ParallelPeer()
     addProperty(_cameraKeyframeInterval);
```
WeirdRubberDuck, line 318: Did a quick test using astrocast.openspaceproject.com, and wondering if some of these properties should be made readonly?

Does it make sense to be able to edit them?  There is no way to start a connection using just these proeprties, right?

engbergandreas: I think with the website, it makes less sense to edit them. But yes you can in fact edit these to start a connection but it is going to be much harder than using the website since you have to know the ports etc. 

Doing readonly does not block setPropertyValue to be updating these values right, it would just block updating from the UI?

WeirdRubberDuck: Correct, it's just a hint to the GUI

And gotcha. Does the connection happen automatically when filling in all the properties, then?  I kind of expected a "connect" trigger property or something :)

engbergandreas: In the ImGui there are connect/disconnect buttons https://github.com/OpenSpace/OpenSpace/blob/af202b2f738de3ea5fc81f5d894f86923b870f2f/modules/imgui/src/guiastrocastcomponent.cpp#L65 that triggers the connection

PR #4210, opened 2026-08-14
Fixes to `ScreenSpaceTimeVaryingImageOnline`

WeirdRubberDuck:
* Fix a rendering bug where the plane was rendered with whatever texture was bound when outside valid data time. 
* Add new properties `ShowBefore` and `ShowAfter` to decide whether the image should stay visible before/after the first/last time + add an example
* Add a trigger property to jump to the first time in the data file (there is no way of knowing what this is from the UI otherwise)

I feel like this renderable could use some testing in general, so please try it out and see if there are other imporvement we want to make

### DIFF
--- a/modules/base/rendering/screenspacetimevaryingimageonline.cpp
+++ b/modules/base/rendering/screenspacetimevaryingimageonline.cpp
@@ -146,6 +146,13 @@ void ScreenSpaceTimeVaryingImageOnline::computeSequenceEndTime() {
     _sequenceEndTime = last + avg;
 }
 
+void ScreenSpaceTimeVaryingImageOnline::render(const RenderData& renderData) {
+    if (!_texture) {
+        return;
+    }
+    ScreenSpaceRenderable::render(renderData);
+}
+
 void ScreenSpaceTimeVaryingImageOnline::update() {
     if (_timestamps.empty()) {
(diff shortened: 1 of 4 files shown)


modules/base/rendering/screenspacetimevaryingimageonline.cpp
```diff
@@ -54,6 +54,29 @@ namespace {
...
+        "If enabled, the image with the earliest timestamp is shown when the simulation "
+        "time is before the first timestamp specified in the data file.",
+        Property::Visibility::User
+    };
+
+    constexpr Property::PropertyInfo ShowAfterInfo = {
+        "ShowAfter",
+        "Show image after last time",
+        "If enabled, the image with the latest timestamp is shown when the simulation "
+        "time is after the lasttimestamp specified in the data file.",
```
WeirdRubberDuck:
```suggestion
        "time is after the last timestamp specified in the data file.",
```

WeirdRubberDuck:
Without the rendering fix, going to a time outside the times in the data file could result in something like this... :)

https://github.com/user-attachments/assets/5cb604fd-a1db-4101-9406-d40b8c3ab5f8

alexanderbock (approved):

PR #4254, opened 2026-09-01
Fix dashboarditemangle crash 

WeirdRubberDuck:
Fixes a crash that happened when loading some of our DashboardItemAngle example assets in the empty profile (tested, and it happened in both master and 0.22.0). 

The issue is that we created a "local buffer" of only 128 characters, but the generated string when two nodes were in the same location could be longer than that. Consequently, the old `std::format_to` call led to undefined behavior (the crash)

The suggested solution removes the local buffer entirely rather than increasing its size. I could not quite see why it was needed, and increasing it could still mean crashing if the names of the objects were long enough. Let me know if I missed something that opts for keeping it (ping @alexanderbock)  

I made the same corresponding change in the other dashboarditem class that used a similar approach

### DIFF
--- a/modules/base/dashboard/dashboarditemangle.cpp
+++ b/modules/base/dashboard/dashboarditemangle.cpp
@@ -221,8 +221,6 @@ DashboardItemAngle::DashboardItemAngle(const ghoul::Dictionary& dictionary)
         }
     }
     addProperty(_destination.nodeIdentifier);
-
-    _localBuffer.resize(128);
 }
 
 void DashboardItemAngle::update() {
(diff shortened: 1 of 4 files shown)


alexanderbock:
The local buffer was there to reduce the number of memory allocations per frame, so I'd argue to keep the local buffer. The `std::format` will allocate a new string every frame for the dashboard, which in of itself is not that big of a deal, but having these allocations everywhere is slowly eating up performance.  I think the error is rather that we take that buffer and create a `std::string` out of it to store into `_buffer`.  So moving `_localBuffer` into DashboardItemText would be the better option.

Shouldn't be a problem to create a buffer that is way too big, even if we are paying 1k of RAM per Dashboard, that will be fine and 999 characters should be enough for everyone

alexanderbock (approved):

PR #4272, opened 2026-09-02
Small Assetvalidator Improvements

WeirdRubberDuck:
1. Makes it possible to use deeper paths in the `--filter`. Previously, only one folder, like `"examples/*"`, worked, but now it's also possible to load only the assets in subfolders like `"examples/screenspacerenderable/*"`

2. Also adds a very simple progress bar that shows how many assets will be loaded and the current progress.  Just went with a simple text-based approach that requires no other dependencies. LMK what you think. We could also do something fancier

image

@engbergandreas it would be great if you could check if the progress bar behaves acceptably when the validator fails to run (you probably know more about when this can happen) :)


alexanderbock (approved):

engbergandreas (approved):
Looks good to me 👍

PR #4280, opened 2026-09-03
Toggle Shutdown function takes parameter

### LINKED ISSUE #4256: Toggle shutdown function should take a parameter
Instead of a config parameter, the default value could become a property that stores the default 3s

alexanderbock:
Replace the openspace.cfg setting for the toggle shutdown timer with a property.  The `toggleShutdown` Lua script now takes an optional parameter.

Closes #4256


engbergandreas (approved):
Works 👍

PR #4286, opened 2026-09-08
Conversion to vcpkg

alexanderbock:
Converts most of the dependencies from submodules into vcpkgs

See also:
  - https://github.com/OpenSpace/Ghoul/pull/138
  - https://github.com/sgct/sgct/pull/126
  - https://github.com/OpenSpace/codegen/pull/24

To test (and also needed once the PR is merged):
1. Once per computer
  1. Somewhere permanent (I did to C:/Development/ext):  `git clone https://github.com/microsoft/vcpkg.git`, so that it gets cloned into C:/Development/ext/vcpkg
  2. In that folder `.\bootstrap-vcpkg.bat`
  3. Set environment variables:
    1. `VCPKG_ROOT` -> C:/Development/ext/vcpkg
    2. `VCPKG_DEFAULT_BINARY_CACHE` -> some other permanent folder (I did C:/Development/ext/vcpkg-cache)
2. Check out this PR (during testing preferably _not_ into an existing OpenSpace foldet to truely test everything
3. `cmake --preset windows`
4. `cmake --build --parallel --preset windows`
5. Run and test the applications

As a sideeffect, this change also updates submodules:
- OpenSpace
  - date: v3.0.4 -> v3.0.5
  - curl: who knows which version -> v8.22.0 and using SSPI on Windows


apps/OpenSpace/CMakeLists.txt
```diff
@@ -81,11 +81,17 @@ create_new_application(OpenSpace
...
+# needed here.
+find_path(STB_INCLUDE_DIR "stb_image.h" REQUIRED)
+
 target_include_directories(
   OpenSpace PRIVATE
   ${OPENVR_INCLUDE_DIRS}
   ${SGCT_OPENVR_INCLUDE_DIRECTORY}
   ${SPOUT_INCLUDE_DIRS}
+  ${STB_INCLUDE_DIR}
 )
```
petersteneteg, line 95: In principle we never want to have any calls to target_include_directories, only target_link_libraries, with proper cmake libraries i.e. FOO::BAR. Anything else is a bit of a code smell

apps/OpenSpace/CMakeLists.txt
```diff
@@ -101,25 +107,12 @@ endif ()
...
-add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/ext/sgct)
-target_link_libraries(OpenSpace PRIVATE sgct)
-
-set_target_properties(sgct PROPERTIES FOLDER "External")
-set_target_properties(glfw PROPERTIES FOLDER "External")
-if (TARGET SGCTTest)
-  set_target_properties(SGCTTest PROPERTIES FOLDER "External")
-endif ()
+# SGCT is resolved through the vcpkg overlay port at
+# apps/OpenSpace/ext/sgct/support/vcpkg/ports/sgct, which builds the enclosing submodule
```
petersteneteg, line 111: The sgct overlay port should probably not live inside of sgct, since it requires you to clone that just to consume the dependency, which prevents any use of caching etc. you want to overlay to live inside of OpenSpace.

apps/OpenSpace/CMakeLists.txt
```diff
@@ -144,14 +137,29 @@ if (OPENSPACE_APPLICATION_OPENSPACE-LAUNCHER)
...
-    add_custom_command(
-      TARGET OpenSpace POST_BUILD
-      COMMAND "${WINDEPLOYQT_EXECUTABLE}" --verbose 0 --no-compiler-runtime --no-translations \"$<TARGET_FILE:OpenSpace>\"
-      COMMENT "Deploying Qt libraries"
-    )
+    # Qt comes from vcpkg and is linked dynamically (LGPL compliance). The Qt DLLs
+    # themselves are picked up by the $<TARGET_RUNTIME_DLLS:OpenSpace> copy further down,
+    # but the runtime plugins (platforms/qwindows.dll, styles, imageformats, tls, ...) are
+    # loaded at runtime and not seen as link dependencies, so copy them next to the
+    # executable. The vcpkg qtbase port does not ship windeployqt, so this is done by hand.
```
petersteneteg, line 144: qtbase in vcpkg does come with the deployment tools. vcpkg_installed/<tripplet>/tools/Qt6/bin/

apps/OpenSpace/CMakeLists.txt
alexanderbock: Wouldn't that mean that every consumer of the library would have to write their own overlay port file though?

apps/OpenSpace/CMakeLists.txt
alexanderbock: (edited)

It was moved into a feature with 6.11:  https://github.com/microsoft/vcpkg/discussions/52444

micahnyc:
For me instead of 

cmake --preset windows

i had to do 

cmake --preset windows.msvc

